PR #1383, opened 2020-07-09
Refactor primitives

### LINKED ISSUE #1250: Break out primitives into a v1 module and make generic.
Let's set up a series of types that are defined as "v1" storage. We can do a migration to v2 storage all in one fell swoop. The runtime then doesn't need to worry about upgrades to one storage type at a time and the node-side code can use a primitives-version runtime API to know exactly which types to provide at any time.

The runtime has a hard time with types that refer to the block hash or number type concretely, as it has to use `system::BlockNumber` and `system::Hash` types and can't specialize them.

So we should set up types as generic, typically as 

```rust
struct SomePrimitive<H = Hash, N = BlockNumber> {
    relay_parent: H,
    relay_parent_number: N,
}
```

All primitive types should be rewritten this way.
And all code using primitive types should be retargeted towards `v1` primitives.
Runtime code should be using explicitly instantiated versions with `<T as system::Trait>::BlockNumber` and `<T as system::Trait>::Hash` (usually `T::BlockNumber` and `T::Hash` for short), while node-side code will use the default instantiations.

This is considered substantial only because of the amount of code that needs to be changed. However the default instantiations should cut down on that a lot.

When we move onto v2, most types will be exactly the same. Instead of re-declaring them I figure we can just define stuff like this

```rust
mod v2 {
    pub use v1::{TypeA, TypeB, TypeC};

    struct TypeD<H = Hash, N = BlockNumber> { ... }
}
```

And we can also establish some global primitives like `BlockNumber` and `Hash`, which should never change, but best to leave those as minimal as possible.

### LINKED ISSUE #1276: Change primitives to match guide's description of candidate types
Might have a bit of fallout in `runtime/common`, `validation`, and `network`, but shouldn't be too much.

Some other concrete updates:
  - `InclusionInherent::inclusion` needs to take the candidate commitments alongside the abridged candidate receipt
  - `Inclusion::process_candidates` needs to check that the encoded commitments match the hash in the abridged receipt.
  - `Inclusion` should store commitments in a separate map alongside the `PendingAvailability`, since they may be large. This will make bitfield processing more efficient for candidates that include code upgrades.

rphmeier:
Closes #1250 
Closes #1276 

This also moves all previous "parachain" primitives into a `v0` module. Nothing is left at the crate root of `polkadot-primitives`.

I've also put stuff like `BlockNumber`, `Hash`, and `UncheckedExtrinsic` into `v0` and `v1` which might be somewhat contentious because those things actually can't change once the chain is in-flight. Everything else should be malleable.


coriolinus:
After this is merged, when writing new code, should we generally choose to import v0 primitives, or v1? Where you've updated the imports in this PR, I see both, without any obvious rubric for choosing which.

rphmeier:
You'd want to use the `v1` primitives everywhere. We're going to phase out all code using the `v0` primitives or migrate it over

rphmeier:
Thanks. Will merge after updating candidate backing and related fallout to new primitives.

Issue #1392, opened 2020-07-10
Enable authority-discovery by default


tomaka:
It will play an important roles for parachains to work, and it would be better to test a bit in advance to see if it works rather than enabling it at the last second and crossing fingers.

rphmeier:
@tomaka what labels should apply to this?

PR #1395, opened 2020-07-13
*: Enable authority discovery by default

mxinden:
Instead of having to explicitly enable the authority discovery module on
validator and sentry nodes, this commit enables the module by default.

Today there is no way for non validator or sentry nodes to run the
module. That might change in the future.

Closes https://github.com/paritytech/polkadot/issues/1392

### DIFF
--- a/cli/src/cli.rs
+++ b/cli/src/cli.rs
@@ -59,7 +59,12 @@ pub struct RunCmd {
 	#[structopt(long = "force-westend")]
 	pub force_westend: bool,
 
-	/// Enable the authority discovery module.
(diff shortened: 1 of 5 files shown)


tomaka:
(tests are failing, in case you haven't noticed)

tomaka:
cc @ddorgan 
This PR is a CLI breaking change that removes the `--enable-authority-discovery` option (it is now enabled by default).

ddorgan:
@tomaka thanks for the heads up!

rphmeier:
Thanks

PR #1420, opened 2020-07-16
Companion for #6629 (weight params refactor)

tomusdrw:
Companion for https://github.com/paritytech/substrate/pull/6629


runtime/common/src/claims.rs
```diff
@@ -661,12 +661,14 @@ mod tests {
 	pub struct Test;
 	parameter_types! {
 		pub const BlockHashCount: u32 = 250;
-		pub const MaximumBlockWeight: u32 = 4 * 1024 * 1024;
-		pub const MaximumBlockLength: u32 = 4 * 1024 * 1024;
-		pub const AvailableBlockRatio: Perbill = Perbill::from_percent(75);
+		pub BlockWeights: system::limits::BlockWeights =
```
kianenigma, line 664: here we can probably ignore all of this and pass a `()` or something that indicates no limit at all, since we don't care at all about weight in any test cases except for one or two pallets in substrate.

runtime/common/src/lib.rs
```diff
@@ -79,9 +76,31 @@ parameter_types! {
...
+			weights.reserved = Some(
+				MAXIMUM_BLOCK_WEIGHT - NORMAL_DISPATCH_RATIO * MAXIMUM_BLOCK_WEIGHT,
+			);
+		})
+		.avg_block_initialization(AVERAGE_ON_INITIALIZE_RATIO)
+		.build_or_panic();
 }
 
-const_assert!(AvailableBlockRatio::get().deconstruct() >= AVERAGE_ON_INITIALIZE_WEIGHT.deconstruct());
+const_assert!(NORMAL_DISPATCH_RATIO.deconstruct() >= AVERAGE_ON_INITIALIZE_RATIO.deconstruct());
```
kianenigma, line 103: can (and should imo) put this in the integrity_test of system. Same potentially applies to all other `const_asserts!` in the root runtime files.

runtime/kusama/src/lib.rs
```diff
@@ -927,8 +926,13 @@ impl pallet_proxy::Trait for Runtime {
 pub struct CustomOnRuntimeUpgrade;
 impl frame_support::traits::OnRuntimeUpgrade for CustomOnRuntimeUpgrade {
 	fn on_runtime_upgrade() -> frame_support::weights::Weight {
+<<<<<<< HEAD
```
kianenigma, line 929: oops

runtime/common/src/lib.rs
tomusdrw: What do you mean by `integrity_test`?

kianenigma: https://crates.parity.io/frame_support/traits/trait.IntegrityTest.html I recently added examples of this to staking that you can look at. Just adds compile time checks to each pallet that are invoked upon constructing a runtime.

tomusdrw: Ah, I see. I'm not sure how is this supposed to work in that particular case.
While I was writing a response to this comment I realised that we might be calculating `max_extrinsic` size incorrectly in the original PR and some extra checks might be missing.

It's now updated, and as a consequence this is now validated via `BlockWeights::validate` method. Basically if `AVG_INIT_RATE` is too high, `max_extrinsic` will go down to `0` for some class and it's going to be rejected as invalid configuration.

kianenigma: Great, I must admit I was a bit uncertain about ensuring that the `validate` was correct in the original PR. Glad to see this bug is caught here and now.

thiolliere:
bot merge

PR #1424, opened 2020-07-16
Availability store subsystem guide

montekki:
This is based off #1401

### DIFF
--- a/roadmap/implementers-guide/src/node/utility/availability-store.md
+++ b/roadmap/implementers-guide/src/node/utility/availability-store.md
@@ -123,7 +123,7 @@ On finality event:
 
 ### PoV pruning
 
-We keep a record about every PoV we keep tracking its state and time after which this PoV should be pruned.
+We keep a record about every PoV we keep, tracking its state and the time after which this PoV should be pruned.
 
 As the state of the `Candidate` changes, so does the `Prune At` time according to the rules defined earlier.
 
--- a/roadmap/implementers-guide/src/node/utility/availability-store.md
+++ b/roadmap/implementers-guide/src/node/utility/availability-store.md
@@ -145,7 +145,7 @@ Chunk pruning is organized in a similar schema as PoV pruning.
 
 ### Included blocks caching
 
-In order to process finality events correctly we need to cache the set of parablocks included into each relay block beginning with the last finalized block and up to the most recent heads. We have to cache this data since we are only able to query this info from the state for the `k` last blocks where `k` is a relatively small number (for more info see [`Assumptions`][Assumptions])
+In order to process finality events correctly we need to cache the set of parablocks included into each relay block beginning with the last finalized block and up to the most recent heads. We have to cache this data since we are only able to query this info from the state for the `k` last blocks where `k` is a relatively small number (for more info see `Assumptions`)
 
 These are used to update Chunk pruning and PoV pruning records upon finality:
 When another block finality notification is received:
@@ -195,4 +195,3 @@ Basically we need to test the correctness of data flow through state FSMs descri
 
 [RAM]: ../../types/overseer-protocol.md#runtime-api-message
 [ASM]: ../../types/overseer-protocol.md#availability-store-message
-[Assumptions]: ../../node/index.md#assumptions
--- a/roadmap/implementers-guide/src/node/utility/availability-store.md
+++ b/roadmap/implementers-guide/src/node/utility/availability-store.md
@@ -104,20 +104,14 @@ On finality event:
 
 - For the finalized block and any earlier block (if any) update pruning records of `PoV`s and chunks to keep them for respective periods after finality.
 
-### Note any new candidates backed in the block by `hash`
+### Note any backed, included and timedout candidates in the block by `hash`.
 
 - Create a `(sender, receiver)` pair.
-- Dispatch a [`RuntimeApiMessage`][RAM]`::Request(hash, RuntimeApiRequest::AvailabilityCores(sender)` and listen on the receiver for a response.
-- For every occupied core returned on the previous step
-  * Create a `(sender, receiver)` pair.
-  * Dispatch a [`RuntimeApiMessage`][RAM]`::Request(hash, RuntimeApiRequest::CandidatePendingAvailability(para_id, sender)`.
-  * Wait on the `receiver` for the receipt of the candidate, check if the candidate is a new one and if it is add it to the set of new candidates backed in the block `hash`.
-  * For every newly-backed candidate update pruning records of any blocks that the node stored previously.
-
-### Note any newly-included candidates backed in the block
-
-- > TODO: figure out how to do this using the existing API
-- Update records tracking which parablocks were included in which relay-chain blocks.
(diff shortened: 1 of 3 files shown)


roadmap/implementers-guide/src/types/overseer-protocol.md
```diff
@@ -48,14 +50,17 @@ Messages to and from the availability store.
 
 ```rust
 enum AvailabilityStoreMessage {
-	/// Query the PoV of a candidate by hash.
-	QueryPoV(Hash, ResponseChannel<PoV>),
+	/// Query the AvailableData of a candidate by hash.
+	QueryPoV(Hash, ResponseChannel<Option<AvailableData>>),
```
coriolinus, line 54: `type AvailableData` is not yet defined.

roadmap/implementers-guide/src/types/overseer-protocol.md
```diff
@@ -12,6 +12,8 @@ enum OverseerSignal {
   StartWork(Hash),
   /// Signal to stop (or phase down) work localized to the relay-parent hash.
   StopWork(Hash),
+  /// A relay block was finalized.
+  BlockFinalized(Hash),
```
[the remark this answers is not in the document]
montekki: honestly I would rather not add yet another way to bypass the intended communication interface.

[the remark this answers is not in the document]
coriolinus: @rphmeier What's your take on this?

roadmap/implementers-guide/src/types/overseer-protocol.md
montekki: defined in availability.md

roadmap/implementers-guide/src/node/utility/availability-store.md
```diff
@@ -50,3 +94,104 @@ On `QueryChunk` message:
...
+  - Check that the data is still available.
+  - Send finality notification about the block in question.
+  - Wait for some time below finalized data timeout.
+  - The data is still available.
+  - Wait until the data should have been purged.
+  - The data is no longer available.
+
+[RAM]: ../../types/overseer-protocol.md#runtime-api-message
+[ASM]: ../../types/overseer-protocol.md#availability-store-message
+[Assumptions]: ../../node/index.md#assumptions
```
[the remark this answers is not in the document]
montekki: removed

roadmap/implementers-guide/src/types/overseer-protocol.md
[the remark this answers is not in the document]
rphmeier: I agree with @montekki on this. The variant can be ignored if it is not wanted.

roadmap/implementers-guide/src/types/overseer-protocol.md
rphmeier, line 54: Let's rename this to `QueryAvailableData` or something like that?

roadmap/implementers-guide/src/runtime/scheduler.md
```diff
@@ -16,48 +16,64 @@ It aims to achieve these tasks with these goals in mind:
...
+		fr4 [label = "Free" shape=rectangle]
+		occ3 [label = "Occupied" shape=rectangle]
+		occ4 [label = "Occupied" shape=rectangle]
+
+		occ2 -> fr3 [label = "Availability"]
+		occ2 -> occ3 [label = "No availability"]
+		fr3 -> fr4 [label = "No backing"]
+		fr3 -> occ4 [label = "Backing"]
+		occ3 -> occ4 [label = "(no change)"]
+		fr4 -> occ4 [label = "Availability Timeout"]
```
rphmeier, line 71: I think that arrow was accidentally backwards in the old diagram, but it should be the other way around.

rphmeier: And actually i think it makes most sense to change it to `occ3 -> fr3` now that i look closer

roadmap/implementers-guide/src/node/utility/availability-store.md
```diff
@@ -50,3 +93,107 @@ On `QueryChunk` message:
...
+   - Remove the record
+
+| Relay Block N | .. | Chain Head 1 | Chain Head 2 |
+|---------------|----|--------------|--------------|
+| CandidateN_1 Included | .. | Candidate1_1 Included | Candidate2_1 Included |
+| CandidateN_2 Included | .. | Candidate1_2 Included | Candidete2_2 Included |
+|          ..           | .. |          ..           |           ..          |
+| CandidateN_M Included | .. | Candidate1_K Included | Candidate2_L Included |
+
+> TODO: It's likely we will have to have a way to go from block hash to `BlockNumber` to make this work.
```
rphmeier, line 155: This is the Chain API subsystem: #1415

roadmap/implementers-guide/src/node/utility/availability-store.md
```diff
@@ -50,3 +93,107 @@ On `QueryChunk` message:
...
+|    Prune At    | .. |    Prune At    |
+| CandidateState | .. | CandidateState |
+
+### Included blocks caching
+
+In order to process finality events correctly we need to cache the set of parablocks included into each relay block beginning with the last finalized block and up to the most recent heads. We have to cache this data since we are only able to query this info from the state for the `k` last blocks where `k` is a relatively small number (for more info see `Assumptions`)
+
+These are used to update Chunk pruning and PoV pruning records upon finality:
+When another block finality notification is received:
+ - For any record older than this block:
```
rphmeier, line 144: Is this `O(n)` in the difference in number between the last finalized block number and the new one?

rphmeier: I am worried that, especially while syncing, we may have large gaps in finality. If the node is offline for a day, or something like that. Furthermore, the races against runtime APIs might get worse in that case.

montekki: Can we really avoid linear time here? I think that this can optimize things a bit:
1. Suppose we have to store this info on blocks `[1000..2000]`.We could do two things here:
   * shrink the data kept in this one big blob to only keeping hashes and block numbers: `[(hash1000, 1000)...(hash2000, 2000)]` and query into single records by some key like `cached_availability_info_key(hash: &Hash)`. 
   * we can also keep this blob sorted at a cheap price since we are likely to only have to append new blocks to it.
2. As soon as we get a new finality notification say about `hash1500`, we can find the exact place to chop off the recent blocks at `O(log(n))` (I guess?)
3. Chop it off, update the individual records.
4. Now we have `[(hash1501, 1501)..(hash2000, 2000)]` array.

But in any case we are going to "touch" every one of them, but hopefully only once

montekki: in any case, i have to give it another thought

rphmeier: Yeah, that is a good point. I guess what I'd want to avoid is that the subsystem gets stuck in a millions-of-blocks loop. It seems better to me to only adjust pruning some number of blocks behind the head and accidentally leak data when doing larger synchronizations. Maybe there is some room for a behind-the-scenes cleanup that does not tie up the message-processing task as well.

rphmeier:
Rebase on master?

PR #1432, opened 2020-07-17
Candidate Validation Subsystem

### LINKED ISSUE #1241: Implement: Candidate Validation

rphmeier:
Closes #1241 

This is a very basic implementation; it doesn't do jobs or results-caching or anything like that. It simply wraps a pool of processes used for candidate validation (a utility we had already implemented for v0) and forwards all validation requests to it.

Still TODO:
  - [x] Fetching necessary data from chain state
  - [x] Fixing fallout in CandidateBacking
  - [x] Tests


rphmeier:
Also, I think we might need something like `tokio::blocking`. These candidate validation tasks are waiting on a response or time-out from a subprocess for a long time, but I'm not sure how we can do that without introducing a hard dependency on tokio.

bkchr:
The SpawnNamed trait provides you with a function for spawning blocking tasks. (I did not looked at the pr on how you solved this now)

rphmeier:
@bkchr OK, we need to add that to the `SubsystemContext`. I will file an issue for both.

node/subsystem/src/messages.rs
```diff
@@ -301,8 +305,28 @@ pub enum RuntimeApiRequest {
 	ValidationCode(ParaId, BlockNumber, Option<BlockNumber>, oneshot::Sender<ValidationCode>),
 	/// Get head data for a specific para.
 	HeadData(ParaId, oneshot::Sender<HeadData>),
-	/// Get a the candidate pending availability for a particular parachain by parachain / core index
+
+	// NEW runtime APIs.
```
drahnr, line 309: :+1:

node/core/candidate-validation/src/lib.rs
```diff
@@ -0,0 +1,913 @@
...
+					let res = spawn_validate_from_chain_state(
+						&mut ctx,
+						Some(pool.clone()),
+						descriptor,
+						pov,
+						spawn.clone(),
+					).await;
+
+					match res {
+						Ok(x) => { let _ = response_sender.send(x); }
```
drahnr, line 97: Should we warn if sending failed rather than silently dropping the result

node/core/candidate-validation/src/lib.rs
```diff
@@ -0,0 +1,913 @@
...
+			validation_code,
+			descriptor,
+			pov,
+			spawn,
+		);
+
+		let _ = tx.send(res);
+	};
+
+	ctx.spawn("blocking-candidate-validation-task", fut.boxed()).await?;
```
drahnr, line 315: How are these task names shown? Should we add a common prefix per subsystem?

roadmap/implementers-guide/src/node/utility/candidate-validation.md
```diff
@@ -12,12 +12,38 @@ Output: Validation result via the provided response side-channel.
 
 ## Functionality
 
-Given the hashes of a relay parent and a parachain candidate block, and either its PoV or the information with which to retrieve the PoV from the network, spawn a short-lived async job to determine whether the candidate is valid.
+This subsystem answers two types of requests. The goal of both request types is to validate a candidate. There are three possible outputs of validation: either the candidate is valid, the candidate is invalid, or an internal error occurred. Whatever the end result is, it will be returned on the response channel to the requestor.
```
drahnr, line 15: Explicitly state the both request types right there or at least in the same block.

node/core/candidate-validation/src/lib.rs
```diff
@@ -0,0 +1,913 @@
...
+			validation_pool,
+			omitted_validation,
+			validation_code,
+			descriptor,
+			pov,
+			spawn,
+		);
+
+		let _ = tx.send(res);
+	};
```
drahnr, line 313: According to the guide, this should have a timeout, or are we expecting the runtime to take care of that?

node/core/candidate-validation/src/lib.rs
rphmeier: They show up in Grafana, I think. The name is not important, only that there is a name.

node/core/candidate-validation/src/lib.rs
rphmeier: The `wasm_executor` crate below has the time-out functionality

node/core/candidate-validation/src/lib.rs
rphmeier: Sure. We might get some spurious warnings but probably OK.

rphmeier:
Thanks for the review @drahnr

rphmeier:
bot merge

rphmeier:
bot merge

PR #1433, opened 2020-07-18
Bump version, enable full governance

gavofyork:


gavofyork:
I think it's best to keep them in; successful referenda have Root privs so it ensures that referenda can always override anything else, as it should be.

PR #1437, opened 2020-07-20
Remove Sudo

gavofyork:
NOTE: To ensure minimal index changes to pre-existing pallet deployments,
this is done with a "swap_remove" style; the previous last pallet
(Purchase), which is hitherto unused, has been shifted into the old index
of Sudo.

### DIFF
--- a/runtime/polkadot/src/lib.rs
+++ b/runtime/polkadot/src/lib.rs
@@ -826,11 +826,12 @@ parameter_types! {
 /// The type used to represent the kinds of proxying allowed.
 #[derive(Copy, Clone, Eq, PartialEq, Ord, PartialOrd, Encode, Decode, RuntimeDebug)]
 pub enum ProxyType {
-	Any,
-	NonTransfer,
-	Governance,
-	Staking,
-	IdentityJudgement,
+	Any = 0,
+	NonTransfer = 1,
+	Governance = 2,
+	Staking = 3,
+	// Skip 4 as it is now removed (was SudoBalances)
+	IdentityJudgement = 5,
 }
 impl Default for ProxyType { fn default() -> Self { Self::Any } }
 impl InstanceFilter<Call> for ProxyType {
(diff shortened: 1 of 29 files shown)


runtime/polkadot/src/lib.rs
```diff
@@ -815,11 +815,6 @@ impl multisig::Trait for Runtime {
 	type WeightInfo = ();
 }
 
-impl sudo::Trait for Runtime {
```
bkchr, line 818: The crate is not removed from the `Cargo.toml`.

runtime/polkadot/src/lib.rs
```diff
@@ -835,7 +830,6 @@ pub enum ProxyType {
 	NonTransfer,
 	Governance,
 	Staking,
-	SudoBalances,
```
sorpaas, line 838: I think we need to explicitly specify the codec index for this enum now? Otherwise `IdentityJudgement` variant may be broken.

gavofyork: Yah fair enough, I'll introduce a dummy.

bkchr: Old transactions will be invalidated anyway through `CheckSpecVersion`.

And we probably don't have anything scheduled that uses the proxy feature?

bkchr: Ahh yeah.

runtime/polkadot/src/lib.rs
```diff
@@ -831,12 +826,12 @@ parameter_types! {
...
-	Governance,
-	Staking,
-	SudoBalances,
-	IdentityJudgement,
+	Any = 0,
+	NonTransfer = 1,
+	Governance = 2,
+	Staking = 3,
+	// Skip 4 as it is now removed (was SudoBalances)
+	IdentityJudgement = 5,
```
bkchr, line 834: Can we than get a test to ensure that the decoding works as expected?

Issue #1348, opened 2020-07-03
Parachain / Collator networking design proposal


rphmeier:
(text courtesy of @infinity0)

# Parachain networking

This is a high-level proposal sketch for everyone's convenient reference; details will be added to https://github.com/w3f/research/tree/infinity0/paranet, including more justification, explanation, and background.

It should be implementable in an incremental way more-or-less, e.g. going through the points below in order. Perhaps (4) can be done in a very basic form, before (1).

## Protocol workflow

Parachain networking forms the first main part of how parachains get their blocks finalised on the Polkadot relay chain. This happens in a few stages:

1. Collators find the current validator set for their parachain, and select one to connect to.
2. Validators receive the PoV blocks from collators.
3. Validators validate the PoV blocks they receive, and send these to other parachain validators.
4. Validators sign proposals and attestations over the PoV blocks they have validated, and send these onto the relay chain via gossip. At least one PoV block must have a quorum of attestations for the block-producing protocol to include it.

We go into these in more detail below. (2) is the most significant part, as it involves communication across a trust boundary - under the Polkadot model, from the perspective of the relay chain, validators are mostly-trusted since they are staked, whereas collators are entirely untrusted and even unauthenticated.

### Collators selecting validators

Collators are expected to be full-nodes of the relay chain, so have easy access to relay chain data. Specifically, which validators are assigned to a parachain at the current block.

In order to help load-balancing, the collator should shuffle this set using their own transport (TLS or QUIC) public key as a seed. Then they can try connecting to each validator in this order, stopping when the first validator accepts the connection.

For honest collators that choose their public key randomly, this will distribute these collators evenly across the set of validators. (Malicious collators that attempt to overwhelm a single validator, are dealt with in the below section.)

### Collator-validator communication

This section describes collator-validator direct communication, largely from the perspective of validators attempting to defend against potentially-malicious collators since that is the hard part.

(An honest collator being serviced by a malicious validator *is* a problem, but it is largely protected by rotating the validator groups around; our 2/3-honest assumption over the validators means that the effect of a malicious validator only lasts for a short time against any parachain.)

0. We need a pre-validation interface, a.k.a. incremental-validation interface - see [#441](https://github.com/w3f/research-internal/issues/441). This would be in addition to the existing (full) validation function interface for parachains.

[...]

PR #1452, opened 2020-07-22
guide: collator networking & subsystems

### LINKED ISSUE #1355: Guide: Collation Distribution Subsystem
This subsystem is responsible for ensuring that collators can connect to validators and that collators can distribute their collations to validators.

The Candidate Selection subsystem is the other side of this, from the validators' perspective.

It seems to make sense to establish a Collator Network Subsystem that actually describes the networking code (collations over the wire, pre-PoVs, validators circulating whitelists, etc.) and then have Candidate Selection and Collation Distribution leverage the Collator Network Subsystem.

rphmeier:
Closes #1355 

Keeping #1348 in mind, although not yet specifying the features of incremental validation, whitelist/blacklisting, heatmaps, etc. which I think bear further discussion & roll-out plans as well as lower-level design work and interfaces. So this PR is primarily setting up the basic logic and structure the code will have.

TODO:
  - [x] Note validator look-up and discovery that collators use
  - [x] Describe pull protocol for collations
  - [x] Network bridge: expand to introduce a split between underlying network protocols for validators & collators, utilities for peer discovery.


roadmap/implementers-guide/src/node/collators/collator-protocol.md
```diff
@@ -0,0 +1,232 @@
...
+
+If we receive a `ToOurSentry::RequestCollation` as a sentry, we need to act as a proxy and issue a request to the peer who made the particular advertisement being referenced. If it times out or the peer disconnects or sends invalid data, we need to respond to our validator with `ToOurValidator::RequestedCollation(..., None)`. If the request comes from a peer who is not our validator, we need to report and disconnect the peer, as this is a clear breach of protocol.
+
+As a validator, once the collation has been fetched some other subsystem will inspect and do deeper validation of the collation. The subsystem will report to this subsystem with a [`CollatorProtocolMessage`][CPM]`::ReportCollator` or `NoteGoodCollation` message. In that case, if we are connected directly to the collator, we apply a cost to the `PeerId` associated with the collator. Otherwise, if we're connected to the collator via a sentry node, we issue a `ToOurSentryMesage` of the corresponding type. As the recipient of such a message on a sentry node, we carry out the requisite action. When handling a report that a collator is bad, we'd want to cancel all requests to that collator.
```
mxinden, line 170: `s/ToOurSentryMesage/ToOurSentryMessage/gc`

roadmap/implementers-guide/src/node/utility/network-bridge.md
```diff
@@ -44,3 +49,12 @@ On `ReportPeer` message:
...
 - Issue a corresponding `ProtocolMessage` to each listed peer with given protocol ID and bytes.
+
+[NBM]: ../../types/overseer-protocol.md#network-bridge-message
+
+On `ConnectToValidators` message:
+
+- Determine the DHT keys to use for each validator based on the relay-chain state and Runtime API.
+- Recover the Peer IDs of the validators from the DHT. There may be more than one peer ID per validator.
+- Accumulate all `(ValidatorId, PeerId)` pairs and send on the response channel.
+- Feed all Peer IDs to the discovery utility the underlying network provides.
```
mxinden, line 60: What is meant with `discovery utility` here? Would the line below work as well?

```suggestion
- Add one `PeerId` per validator as a priority group to the `PeerSet`.
```

roadmap/implementers-guide/src/node/collators/collator-protocol.md
```diff
@@ -0,0 +1,133 @@
...
+enum WireMessage {
+	/// Declare the intent to advertise collations under a collator ID.
+	Declare(CollatorId),
+	/// Advertise a collation to a validator. Can only be sent once the peer has declared
+	/// that they are a collator with given ID.
+	AdvertiseCollation(Hash, ParaId),
+	/// Request the advertised collation at that relay-parent.
+	RequestCollation(RequestId, Hash, ParaId),
+	/// A requested collation.
+	Collation(RequestId, CandidateReceipt, PoV),
```
tomaka, line 39: This should instead be a request-response-style protocol, but I suppose this can be changed later?

roadmap/implementers-guide/src/types/overseer-protocol.md
```diff
@@ -126,20 +126,60 @@ enum CandidateSelectionMessage {
...
 ## Network Bridge Message
 
 Messages received by the network bridge. This subsystem is invoked by others to manipulate access
 to the low-level networking code.
 
 ```rust
+/// Peer-sets handled by the network bridge.
+enum PeerSet {
+	/// The collation peer-set is used to distribute collations from collators to validators.
+	Collation,
```
infinity0, line 163: Perhaps I am interpreting "PeerSet" incorrectly, but with collator networking for validators there are actually two types of neighbours:

- collators, a one-to-one protocol
- other parachain validators, as described in "Validator-validator communication" in #1348, a gossip protocol

Then there is "Passing to the relay chain" which is done on the main relay chain gossip protocol that already exists, with its own set of neighbours that (as you note below) can include non-validators, and also validators on another parachain.

roadmap/implementers-guide/src/node/utility/network-bridge.md
rphmeier: I guess so, but I talked with Pierre and we wanted to do away with this priority groups thing. The previous line seems more general and can be adapted based on what is actually done.

roadmap/implementers-guide/src/node/collators/collation-generation.md
```diff
@@ -1,9 +1,36 @@
...
 ## Protocol
 
+Input: None
+
+Output: CollationDistributionMessage
+
 ## Functionality
 
-## Jobs, if any
+The process of generating a collation for a parachain is very parachain-specific. As such, the details of how to do so are left beyond the scope of this description. The subsystem should be implemented as an abstract wrapper, which is aware of this configuration:
```
infinity0, line 13: Are validation/pre-validation functions mentioned anywhere else? If not they could go here.

roadmap/implementers-guide/src/node/utility/network-bridge.md
rphmeier: The guide in general leans away from implementation details of Substrate. I believe priority groups are such a detail.

roadmap/implementers-guide/src/node/collators/collation-generation.md
```diff
@@ -1,9 +1,36 @@
...
+
+```rust
+struct CollationGenerationConfig {
+	key: CollatorPair,
+	collation_producer: Fn(params) -> async (HeadData, Vec<UpwardMessage>, PoV),
+}
+```
```
infinity0, line 20: Are you intending that collator nodes extend the polkadot client & recompile? It might be easier to have a local parachain-specific process talk to a polkadot node acting as a collator via some interprocess API, but that's a discussion for much later.

roadmap/implementers-guide/src/node/collators/collator-protocol.md
```diff
@@ -0,0 +1,133 @@
...
+
+When acting on an advertisement, we issue a `WireMessage::RequestCollation`. If the request times out, we need to note the collator as being unreliable and reduce its priority relative to other collators. And then make another request - repeat until we get a response or the chain has moved on.
+
+As a validator, once the collation has been fetched some other subsystem will inspect and do deeper validation of the collation. The subsystem will report to this subsystem with a [`CollatorProtocolMessage`][CPM]`::ReportCollator` or `NoteGoodCollation` message. In that case, if we are connected directly to the collator, we apply a cost to the `PeerId` associated with the collator and potentially disconnect or blacklist it.
+
```
infinity0, line 124: As mentioned in my other comment on PeerSet, validators should also be connected to a few other validators on the same parachain, and forward received collations onto them. (The details of neighbour selection are mentioned in #1348)

Other validators are more trusted than collators, so the protocol wire message here only needs to consist of the actual `Collation`, but if it's simpler to re-use the collator-validator wire message for now, it wouldn't hurt. However in the future once we get onto passing around whitelists/blacklists of collators, the message types would have to diverge.

roadmap/implementers-guide/src/types/overseer-protocol.md
```diff
@@ -126,20 +126,60 @@ enum CandidateSelectionMessage {
 }
 ```
 
+## Collator Protocol Message
+
+Messages received by the [Collator Protocol subsystem](../node/collators/collator-protocol.md)
+
+```rust
+enum CollatorProtocolMessage {
```
infinity0, line 134: "Protocol" suggests that this message is passed to other nodes, and is somewhat stable, but as I understand this is just an internal message between subsystems and could change whenever. How about `CollatorInfoMessage` or just `CollatorInfo`?

roadmap/implementers-guide/src/types/overseer-protocol.md
rphmeier: For what's specified, we only need collator<>validator communication. The validator<>validator aspects of distributing whitelists etc. are not handled in this version.

Other aspects of parachain networking (distributing PoVs among parachain group, gossiping statements) are beyond the scope of this subsystem and do indeed use the `Validation` peerset.

roadmap/implementers-guide/src/node/collators/collation-generation.md
rphmeier: Yup, they would go here. However I've deferred that to a later point.

roadmap/implementers-guide/src/node/collators/collation-generation.md
rphmeier: > Are you intending that collator nodes extend the polkadot client & recompile

Basically, yes. That's how the current Cumulus architecture works, at least. Some IPC-based split would also be amenable to this approach, with the `collation_producer` yielding an IPC future.

roadmap/implementers-guide/src/node/collators/collator-protocol.md
rphmeier: > As mentioned in my other comment on PeerSet, validators should also be connected to a few other validators on the same parachain, and forward received collations onto them

Agreed, but this is handled by the PoV distribution and Statement distribution subsystems, not collation distribution.

roadmap/implementers-guide/src/types/overseer-protocol.md
rphmeier: The subsystem is called the Collator Protocol subsystem. Our naming convention for these message types is `format!("{}Message", subsystem_name)`

roadmap/implementers-guide/src/types/overseer-protocol.md
infinity0: In this context, PoV blocks are not supposed to be distributed to the `Validation` peerset; that is the purpose of the A&V protocol. Here they are only supposed to be distributed to other parachain validators, so that they can sign attestations for them, and is less traffic overall.

If we do not have this component, then parachain collators will need to send the same PoV block to multiple parachain validators, in order to achieve the minimum number of attestations needed for the block production protocol. By having parachain validators also pass this between each other, we alleviate this bottleneck. I think this is fairly important to have even in an early version of the protocol.

infinity0: Also, you are using both the terms "collation" and "PoV block" separately. Are you saying they are different things? Because I understood them to be the same thing.

rphmeier: > Here they are only supposed to be distributed to other parachain validators, so that they can sign attestations for them, and is less traffic overall.

Yes, I understood this. And it is covered by another part of the code, as I mentioned.

> Also, you are using both the terms "collation" and "PoV block" separately. Are you saying they are different things? Because I understood them to be the same thing.

Collation is `(CandidateReceipt, PoV)`

roadmap/implementers-guide/src/node/utility/network-bridge.md
mxinden:
```suggestion
- Feed all Peer IDs to the peer set manager the underlying network provides.
```

In case we want to keep it generic I would suggest the following. I don't think `discovery utility` is the right term here as the peer has already been discovered at this point.

roadmap/implementers-guide/src/types/overseer-protocol.md
infinity0: > Yes, I understood this. And it is covered by another part of the code, as I mentioned.

OK, what you said makes sense although I am still confused by the docstring on "Validation", it says:

> This may include nodes which are not validators, as some protocols on this peer-set are expected to be gossip.

This makes sense for the main gossip network on the relay chain, for GRANDPA/BABE etc. However it does not make sense for PoV distribution - here you are only distributing to other parachain validators, not non-validators nor validators on other parachains.

rphmeier: @infinity0 We can't guarantee at this point that we have transitive connection among the validator set. I guarantee that, in practice, if we excluded full nodes we would probably not achieve parachain liveness.

rphmeier: In this case, the data are authenticated, because they must be presented alongside a `Seconded` candidate by a validator. Barring validator equivocations, the amount of data is bounded.

roadmap/implementers-guide/src/node/utility/network-bridge.md
rphmeier: addressed in #1535

PR #1454, opened 2020-07-22
Dont bring down the node if validation has failed

### LINKED ISSUE #1388: Candidate Validation failures in backing should be non-fatal to the node

montekki:
Closes #1388

### DIFF
--- a/node/core/backing/src/lib.rs
+++ b/node/core/backing/src/lib.rs
@@ -439,7 +439,11 @@ impl CandidateBackingJob {
 			}
 			CandidateBackingMessage::Statement(_, statement) => {
 				self.check_statement_signature(&statement)?;
-				self.maybe_validate_and_import(statement).await?;
+				match self.maybe_validate_and_import(statement).await {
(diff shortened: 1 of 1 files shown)


node/core/backing/src/lib.rs
```diff
@@ -439,7 +439,11 @@ impl CandidateBackingJob {
 			}
 			CandidateBackingMessage::Statement(_, statement) => {
 				self.check_statement_signature(&statement)?;
-				self.maybe_validate_and_import(statement).await?;
+				match self.maybe_validate_and_import(statement).await {
+					Err(Error::ValidationFailed(_)) => return Ok(()),
+					Err(e) => return Err(e),
+					Ok(()) => (),
+				}
```
coriolinus, line 446: Optional

```suggestion
				match self.maybe_validate_and_import(statement).await {
					Ok(()) | Err(Error::ValidationFailed(_)) => return Ok(()),
					Err(e) => return Err(e),
				}
```

PR #1460, opened 2020-07-23
impl availability distribution

drahnr:
Implements the logic for availability distribution as described by the guide plus guide improvements.

Ref #1237


rphmeier:
Bad bot

drahnr:
Mea culpa, I accidentally created a "regular" PR and then changed it to draft, so the bot probably already got triggered.

roadmap/implementers-guide/src/node/availability/availability-distribution.md
```diff
@@ -26,9 +26,12 @@ Register on startup an event producer with  `NetworkBridge::RegisterEventProduce
 
 For each relay-parent in our local view update, look at all backed candidates pending availability. Distribute via gossip all erasure chunks for all candidates that we have to peers.
 
-We define an operation `live_candidates(relay_heads) -> Set<CommittedCandidateReceipt>` which returns a set of [`CommittedCandidateReceipt`s](../../types/candidate.md#committed-candidate-receipt) a given set of relay chain heads that implies a set of candidates whose availability chunks should be currently gossiped. This is defined as all candidates pending availability in any of those relay-chain heads or any of their last `K` ancestors. We assume that state is not pruned within `K` blocks of the chain-head.
+We define an operation `live_candidates(relay_heads) -> Set<CommittedCandidateReceipt>` which returns a set of [`CommittedCandidateReceipt`s](../../types/candidate.md#committed-candidate-receipt) a given set of relay chain heads that implies a set of candidates whose availability chunks should be currently gossiped.
```
rphmeier, line 29: The change on this line reads strangely

"... which returns a set of `CommitedCandidateReceipts`s a given set of relay chain heads that implies a set of ..."

node/network/availability-distribution/src/lib.rs
```diff
@@ -0,0 +1,1552 @@
...
+
+// Polkadot is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with Polkadot.  If not, see <http://www.gnu.org/licenses/>.
+
+//! The bitfield distribution
```
rphmeier, line 17: Comment is copied from bitfield distribution

node/network/availability-distribution/src/lib.rs
```diff
@@ -0,0 +1,1552 @@
...
+const COST_PEER_DUPLICATE_MESSAGE: Rep = Rep::new(-500, "Peer sent identical messages");
+const BENEFIT_VALID_MESSAGE_FIRST: Rep = Rep::new(15, "Valid message with new information");
+const BENEFIT_VALID_MESSAGE: Rep = Rep::new(10, "Valid message");
+
+/// Checked signed availability bitfield that is distributed
+/// to other peers.
+#[derive(Encode, Decode, Debug, Clone, PartialEq, Eq, Hash)]
+pub struct AvailabilityGossipMessage {
+	/// Ankor hash of the candidate this is associated to.
+	pub candidate_hash: H256,
```
rphmeier, line 84: We haven't imported this type as `H256` elsewhere in the codebase

node/network/availability-distribution/src/lib.rs
```diff
@@ -0,0 +1,1552 @@
...
+	/// Track all active peers and their views
+	/// to determine what is relevant to them.
+	peer_views: HashMap<PeerId, View>,
+
+	/// Our own view.
+	view: View,
+
+	/// Caches a mapping of relay parents to live candidates
+	/// and allows fast + intersection free obtaining of active heads set by unionizing.
+	/// Includes the `K` ancestors too.
```
rphmeier, line 102: As in, this includes an entry for each of the `K` ancestors, or this contains an entry for each relay-parent that includes the candidates for each of the `K` ancestors?

The first option is preferable and the comment below leads me to believe it is what is implemented

node/network/availability-distribution/src/lib.rs
```diff
@@ -0,0 +1,1552 @@
...
+	received_messages: HashMap<PeerId, HashSet<(H256, ValidatorIndex)>>,
+
+	/// Track already sent candidate hashes and the erasure chunk index to the peers.
+	sent_messages: HashMap<PeerId, HashSet<(H256, ValidatorIndex)>>,
+}
+
+#[derive(Debug, Clone, Default)]
+struct PerRelayParent {
+	/// Set of `K` ancestors for this relay parent.
+	ancestors: HashSet<H256>,
```
rphmeier, line 141: Would a `Vec` or even an array make more sense?

node/network/availability-distribution/src/lib.rs
```diff
@@ -0,0 +1,1552 @@
...
+				modify_reputation(ctx, remote, COST_MESSAGE_NOT_DECODABLE).await?;
+			}
+		}
+	}
+	Ok(())
+}
+
+fn candidate_hash_of(receipt: &CommittedCandidateReceipt) -> H256 {
+	// @todo is this correct? or is there a better candidate hash?
+	receipt.descriptor().validation_data_hash.clone()
```
rphmeier, line 324: no, this isn't correct - you should use `receipt.hash()`

node/network/availability-distribution/src/lib.rs
```diff
@@ -0,0 +1,1552 @@
...
+			.is_test(true)
+			.filter(None, log::LevelFilter::Trace)
+			.try_init();
+		let keystore = keystore::Store::new_in_memory();
+
+		let candidate_hash_af = H256::from_slice(&[0x0Fu8; 32]);
+		let candidate_hash_11 = H256::from_slice(&[0x11u8; 32]);
+
+		// @todo fails with invalid seed
+		// let validator_keypair: ValidatorPair = keystore
```
rphmeier, line 1207: Try `Sr25519Keyring` - you can see examples in other subsystem crates or in the runtime

`Sr25519Keyring::Alice.pair().into()` should work

node/network/availability-distribution/src/lib.rs
```diff
@@ -0,0 +1,1552 @@
...
+	let hint = iter.size_hint();
+
+	let capacity = hint.1.unwrap_or(hint.0) * (1 + AvailabilityDistributionSubsystem::K);
+	let mut live_candidates = HashMap::<H256, CommittedCandidateReceipt>::with_capacity(capacity);
+
+	for relay_parent in iter {
+		// direct candidates for one relay parent
+		let receipts =
+			query_live_candidates_without_ancestors(ctx, iter::once(relay_parent.clone())).await?;
+		live_candidates.extend(iter::once(relay_parent.clone()).zip(receipts.into_iter()));
```
rphmeier, line 865: is this a correct use of `zip`?  This might take only the first entry from `receipts`, which is a `HashSet`

node/network/availability-distribution/src/lib.rs
```diff
@@ -0,0 +1,1552 @@
...
+		.await
+		.map_err::<Error, _>(Into::into)?;
+	rx.await
+		.map_err::<Error, _>(Into::into)?
+		.map_err::<Error, _>(Into::into)
+}
+
+
+/// Query the hash of the `K` ancestors
+#[cfg(feature = "std")]
```
rphmeier, line 1030: is there a `no-std` version of this crate?

node/network/availability-distribution/src/lib.rs
```diff
@@ -0,0 +1,1552 @@
...
+async fn query_k_ancestors<Context>(
```
rphmeier, line 1031: Can we make this function actually do "up to k ancestors that have the same session as parent"?

rphmeier: This would use the `session_index_for_child`, so actually it's "up to k ancestors that have the same session as the parent's child". The implication is that there might just be no candidates to distribute for a given relay-parent.

This mirrors the behavior of the runtime. If you have a block B which is the last block of the session, anything pending availability will be discarded in the `on_initialize` logic of its child. So there is no use distributing anything pending availability in such a block anyway.

rphmeier: However I notice that the treatment of the relay-parent itself is separate from the ancestors in `query_live_candidates`, so it might make sense to re-structure that code somewhat to perform these two steps:
  1. get a set of relay-chain block hashes, which may be empty, that contains all of the relay-parent and its K ancestors who have the same session as the relay-parent's child. This implies making a state query at `relay_parent - (K+1)`
  2. Get live candidates for each of those, querying the cache as necessary.

node/network/availability-distribution/src/lib.rs
```diff
@@ -0,0 +1,1552 @@
...
+			.into_iter()
+			.filter(|relay_parent| {
+				// use the ones which we pulled before
+				// but keep the unknown relay parents
+				state
+					.receipts
+					.get(relay_parent)
+					.and_then(|receipts| {
+						// directly extend the live_candidates with the cached value
+						live_candidates.extend(
```
rphmeier, line 882: does the `.get` above need to be a `get_mut`?

Why would this `receipts` value ever change? I would expect it to be written once and read everywhere else

node/network/availability-distribution/src/lib.rs
```diff
@@ -0,0 +1,1552 @@
...
+		// direct candidates for one relay parent
+		let receipts =
+			query_live_candidates_without_ancestors(ctx, iter::once(relay_parent.clone())).await?;
+		live_candidates.extend(iter::once(relay_parent.clone()).zip(receipts.into_iter()));
+
+		// register one of relay parents (not the ancestors)
+		let ancestors =
+			query_k_ancestors(ctx, relay_parent, AvailabilityDistributionSubsystem::K).await?;
+
+		// ancestors might overlap, so check t	he cache too
```
rphmeier, line 871:
```suggestion
		// ancestors might overlap, so check the cache too
```

node/network/availability-distribution/src/lib.rs
```diff
@@ -0,0 +1,1552 @@
...
+						);
+						Some(())
+					})
+					.is_none()
+			})
+			.collect::<Vec<_>>();
+
+		// query the ones that were not present in the receipts cache
+		let receipts = query_live_candidates_without_ancestors(ctx, unknown.clone()).await?;
+		live_candidates.extend(unknown.into_iter().zip(receipts.into_iter()));
```
rphmeier, line 895: should we write into the receipts cache here?

drahnr: That happens at the call site, but it's better to move it into the this `fn` :+1:

node/network/availability-distribution/src/lib.rs
drahnr: That is correct, each relay parent as well as each ancestor hash is a key with an associated receipt each.

node/network/availability-distribution/src/lib.rs
drahnr: Issues I've run into, was a myriad of `std::hash::Hash` and `Hash` collisions, I've seen the `H256` in `sp_core` i.e., I am open to suggestions.

rphmeier: Our convention in Substrate is to import `std::hash::Hash` as `HashT`. We do this for other traits too, like `BlockT` and `HeaderT`.

node/network/availability-distribution/src/lib.rs
```diff
@@ -0,0 +1,1552 @@
...
+	}
+
+	fn remove_relay_parent(&mut self, relay_parent: &H256) -> Result<()> {
+		// we might be ancestor of some other relay_parent
+		if let Some(ref mut descendants) = self.ancestry.get_mut(relay_parent) {
+			// if we were the last user, and it is
+			// not explicitly set to be worked on by the overseer
+			if descendants.is_empty() {
+				// remove from the ancestry index
+				self.ancestry.remove(relay_parent);
```
rphmeier, line 246: I guess we could use the entry API here, but I actually think that as currently written it's cleaner. And the performance effect of the entry API will be negligible.

node/network/availability-distribution/src/lib.rs
```diff
@@ -0,0 +1,1552 @@
...
+			})
+			.flatten()
+			.collect::<HashSet<H256>>();
+
+		relay_parents_and_ancestors
+			.into_iter()
+			.filter_map(|relay_parent_or_ancestor| self.receipts.get(&relay_parent_or_ancestor))
+			.map(|receipt_set| receipt_set.into_iter())
+			.flatten()
+			.map(|receipt| (candidate_hash_of(receipt), receipt.clone()))
```
rphmeier, line 179: maybe better for the stored data in `self.receipts` to cache this hash, so we don't do hundreds of hashes on every relay-chain block

node/network/availability-distribution/src/lib.rs
```diff
@@ -0,0 +1,1552 @@
...
+					// a network message was received
+					if let Err(e) = handle_network_msg(&mut ctx, &mut state, event).await {
+						warn!(
+							target: TARGET,
+							"Failed to handle incomming network messages: {:?}", e
+						);
+					}
+				}
+				FromOverseer::Communication {
+					msg: AvailabilityDistributionMessage::DistributeChunk(_hash, _erasure_chunk),
```
rphmeier, line 769: This PR can remove this message

node/network/availability-distribution/src/lib.rs
```diff
@@ -0,0 +1,1552 @@
...
+				} => {
+					trace!(target: TARGET, "Processing incoming erasure chunk");
+				}
+				FromOverseer::Communication {
+					msg: AvailabilityDistributionMessage::FetchChunk(_hash, _erasure_chunk),
```
rphmeier, line 774: And this one, since it's not used

node/network/availability-distribution/src/lib.rs
```diff
@@ -0,0 +1,1552 @@
...
+	validators.iter().enumerate().find_map(|(idx, v)| {
+		keystore
+			.key_pair::<ValidatorPair>(&v)
+			.ok()
+			.map(move |_| idx as ValidatorIndex)
+	})
+}
+
+/// Handle an incoming message from a peer.
+async fn process_incoming_peer_message<Context>(
```
rphmeier, line 621: This function should also be issuing a `StoreChunk` message to the availability store

node/network/availability-distribution/src/lib.rs
drahnr: I initially had the entry API here, but that would leave a `ancestry` entries there which would hold a `HashSet`, and I did not see a way to use the `Entry` based `API` without resorting to an explicit `match` statement which would make it more convoluted than the above imo.

node/network/availability-distribution/src/lib.rs
drahnr: > However I notice that the treatment of the relay-parent itself is separate from the ancestors in `query_live_candidates`, so it might make sense to re-structure that code somewhat to perform these two steps:
> 
>     1. get a set of relay-chain block hashes, which may be empty, that contains all of the relay-parent and its K ancestors who have the same session as the relay-parent's child. This implies making a state query at `relay_parent - (K+1)`

Don't we have to do a binary search/linear traversal over `{ relay_parant, ancestor_1, ancestor_2, .. ancestor_N } ` to know which one actually is the last one in the same session? 

One simple probe would not tell us which ancestors have the same session index.

node/network/availability-distribution/src/lib.rs
drahnr: Good point, the assumption that the `Hasher` used by `HashMap` (`hashbrown`, or are we using something else/planning on using smth else?) is significantly faster than `H256` is always true :+1:

node/network/availability-distribution/src/lib.rs
rphmeier: linear traversal would make most sense given that `K` is so small

node/network/availability-distribution/src/lib.rs
rphmeier: Yeah, blake2-256 is cryptographically secure whereas AHash is meant to be fast.

However, we should also have a utility at some point where we create a `HashMap` type alias using an identity hashes. We don't need do any hashing of the keys, because the keys are already hashes with more than sufficient entropy. But that is a premature optimization.

node/network/availability-distribution/Cargo.toml
```diff
@@ -0,0 +1,35 @@
...
+sp-staking = { git = "https://github.com/paritytech/substrate", branch = "master", features = ["std"]  }
+
+[dev-dependencies]
+polkadot-subsystem = { package = "polkadot-node-subsystem", path = "../../subsystem", features = [ "test-helpers" ] }
+bitvec = { version = "0.17.4", default-features = false, features = ["alloc"] }
+sp-core = { git = "https://github.com/paritytech/substrate", branch = "master", features = ["std"] }
+sp-keyring = { git = "https://github.com/paritytech/substrate", branch = "master" }
+parking_lot = "0.11.0"
+maplit = "1.0.2"
+smol = "0.3.3"
```
ordian, line 31: what is it used for? we're using `futures::executor` for tests it seems

node/network/availability-distribution/src/lib.rs
```diff
@@ -0,0 +1,1827 @@
...
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with Polkadot.  If not, see <http://www.gnu.org/licenses/>.
+
+//! The availability distribution
+//!
+//! Transforms `AvailableData` into erasure chunks, which are distributed to peers
+//! which are intereseted in the relevant candidates.
```
ordian, line 20:
```suggestion
//! who are interested in the relevant candidates.
```

drahnr: Never sure what is a "person" and what is a "thing" :>

ordian: I think both are correct in this case, just to avoid repetition.

node/network/availability-distribution/src/lib.rs
```diff
@@ -0,0 +1,1823 @@
...
+
+	// iterate from oldest to youngest
+	let mut iter = ancestors.into_iter().rev();
+	if let Some(oldest) = iter.next() {
+		let mut child_session_index = query_session_index_for_child(ctx, oldest).await?;
+		for ancestor in iter {
+			let ancestor_session_index = child_session_index;
+			child_session_index = query_session_index_for_child(ctx, ancestor).await?;
+
+			if desired_session == ancestor_session_index {
```
ordian, line 1112: just to make sure I understand this correctly: once this condition is false, it's always false, right?

node/network/availability-distribution/src/lib.rs
```diff
@@ -0,0 +1,1852 @@
...
+					activated,
+					deactivated,
+				})) => {
+					for relay_parent in activated {
+						trace!(target: TARGET, "Start {:?}", relay_parent);
+					}
+					for relay_parent in deactivated {
+						trace!(target: TARGET, "Stop {:?}", relay_parent);
+					}
+				}
```
ordian, line 799: we don't handle this signal here, not sure what's the purpose of logs

node/network/availability-distribution/src/lib.rs
```diff
@@ -0,0 +1,1852 @@
...
+						.is_some()
+				})
+				.map(|peer| peer.clone())
+				.collect::<Vec<_>>();
+			let message = AvailabilityGossipMessage {
+				candidate_hash,
+				erasure_chunk,
+			};
+
+			// let per_candidate = state.per_candidate.entry(candidate_hash.clone()).or_default();
```
ordian, line 476: remove?

node/network/availability-distribution/src/lib.rs
```diff
@@ -0,0 +1,1852 @@
...
+	received_messages: HashMap<PeerId, HashSet<(Hash, ValidatorIndex)>>,
+
+	/// Track already sent candidate hashes and the erasure chunk index to the peers.
+	sent_messages: HashMap<PeerId, HashSet<(Hash, ValidatorIndex)>>,
+
+	/// The set of validators.
+	validators: Vec<ValidatorId>,
+
+	/// If this node is a validator, note the index in the validator set.
+	validator_index: Option<ValidatorIndex>,}
```
ordian, line 142:
```suggestion
	validator_index: Option<ValidatorIndex>,
}
```

node/network/bitfield-distribution/Cargo.toml
```diff
@@ -21,9 +21,9 @@ sc-network = { git = "https://github.com/paritytech/substrate", branch = "master
 polkadot-subsystem = { package = "polkadot-node-subsystem", path = "../../subsystem", features = [ "test-helpers" ] }
 bitvec = { version = "0.17.4", default-features = false, features = ["alloc"] }
 sp-core = { git = "https://github.com/paritytech/substrate", branch = "master" }
-parking_lot = "0.10.0"
+parking_lot = "0.11.0"
 maplit = "1.0.2"
-smol = "0.2.0"
+smol = "0.3.3"
```
ordian, line 26: I think this one is unused too :)

node/network/availability-distribution/src/lib.rs
```diff
@@ -0,0 +1,1852 @@
...
+
+		// collect the ancestors again from the hash map
+		let ancestors = candidates
+			.iter()
+			.filter_map(|(ancestor_or_relay_parent, _receipt)| {
+				if ancestor_or_relay_parent == &relay_parent {
+					None
+				} else {
+					Some(ancestor_or_relay_parent.clone())
+				}
```
ordian, line 226: feels like a `filter(ancestor_or_relay_parent != &relay_parent)` would be simpler here

drahnr: The main reason is, that I can go from `&T -> T` by means of `Clone` at the same time as filtering.

node/network/availability-distribution/src/lib.rs
drahnr: whops

node/network/availability-distribution/src/lib.rs
drahnr: It's only trace logs, but there was no particular purpose here. Can be removed.

node/network/availability-distribution/src/lib.rs
```diff
@@ -0,0 +1,1823 @@
...
+fn obtain_our_validator_index(
+	validators: &[ValidatorId],
+	keystore: KeyStorePtr,
+) -> Option<ValidatorIndex> {
+	let keystore = keystore.read();
+	validators.iter().enumerate().find_map(|(idx, v)| {
+		keystore
+			.key_pair::<ValidatorPair>(&v)
+			.ok()
+			.map(move |_| idx as ValidatorIndex)
```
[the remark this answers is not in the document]
cheme: I think it is not, would be preferable to stick to `BareCryptoStore` trait and use `has_key` here.

node/network/availability-distribution/src/lib.rs
drahnr: since we iterator starting from the farthest descendant from `relay_parent`, it will be `false`, until it will be eventually `true` and then stray `true` up to the `relay_parent` (which is not included).

roadmap/implementers-guide/src/node/availability/availability-distribution.md
```diff
@@ -26,13 +26,16 @@ Register on startup an event producer with  `NetworkBridge::RegisterEventProduce
...
+We will send any erasure-chunks that correspond to candidates in `live_candidates(peer_most_recent_view_update)`.
+Likewise, we only accept and forward messages pertaining to a candidate in `live_candidates(current_heads)`.
+Each erasure chunk should be accompanied by a merkle proof that it is committed to by the erasure trie root in the candidate receipt, and this gossip system is responsible for checking such proof.
 
 We re-attempt to send anything live to a peer upon any view update from that peer.
 
-On our view change, for all live candidates, we will check if we have the PoV by issuing a `QueryPoV` message and waiting for the response. If the query returns `Some`, we will perform the erasure-coding and distribute all messages to peers that will accept them.
+On our view change, for all live candidates, we will check if we have the PoV by issuing a `QueryAvailabileData` message and waiting for the response. If the query returns `Some`, we will perform the erasure-coding and distribute all messages to peers that will accept them.
```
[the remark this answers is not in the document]
drahnr: @rphmeier is there a rationale on how to choose `K` I could add - so far we only talked about "small"~ish

node/network/bitfield-distribution/Cargo.toml
drahnr: It's used now :> for the delays which utilize a `smol::Timer`

ordian: strange, I only see `use smol_timeout::TimeoutExt;` in code

node/network/availability-distribution/Cargo.toml
```diff
@@ -0,0 +1,34 @@
+[package]
+name = "polkadot-availability-distribution"
+version = "0.1.0"
+authors = ["Parity Technologies <admin@parity.io>"]
+edition = "2018"
+
+[dependencies]
+futures = "0.3.5"
+futures-timer = "3.0.2"
```
ordian, line 9: this one is what you're using `smol::Timer` for? :)
I guess it could be moved to dev-dependencies and replace `smol` there

roadmap/implementers-guide/src/node/availability/availability-distribution.md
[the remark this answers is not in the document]
rphmeier: There isn't much rationale, other than "small"-ish. let's set K=3 for now and if there are empirical issues then we can adjust.

node/network/availability-distribution/src/lib.rs
```diff
@@ -0,0 +1,1856 @@
...
+	*current = view;
+
+	// only contains the intersection of what we are interested and
+	// the union of all relay parent's candidates.
+	let delta_candidates = state.cached_live_candidates_unioned(delta_vec.iter());
+
+	// Send all messages we've seen before and the peer is now interested
+	// in to that peer.
+
+	for (candidate_hash, _receipt) in delta_candidates {
```
coriolinus, line 585: Same concurrency thing here

node/network/availability-distribution/src/lib.rs
```diff
@@ -0,0 +1,1856 @@
...
+				// collect all direct interests of a peer w/o ancestors
+				state
+					.cached_live_candidates_unioned(view.0.iter())
+					.contains_key(&candidate_hash)
+			})
+			.map(|(peer, _view)| peer.clone())
+			.collect();
+
+		// distribute all erasure messages to interested peers
+		for erasure_chunk in erasure_chunks {
```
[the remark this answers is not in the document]
rphmeier: sending a message requires `&mut ctx` which can't be shared due to Rust's ownership rules

node/network/availability-distribution/src/lib.rs
```diff
@@ -0,0 +1,1856 @@
...
+	peers: Vec<PeerId>,
+	message_iter: impl IntoIterator<Item = AvailabilityGossipMessage>,
+) -> Result<()>
+where
+	Context: SubsystemContext<Message = AvailabilityDistributionMessage>,
+{
+	if peers.is_empty() {
+		return Ok(())
+	}
+	for message in message_iter {
```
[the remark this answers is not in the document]
rphmeier: i think this has the same problem w.r.t. a mutable borrow

node/network/availability-distribution/src/lib.rs
```diff
@@ -0,0 +1,1856 @@
...
+
+	/// Unionize all cached entries for the given relay parents and it's ancestors ancestors.
+	/// Ignores all non existent relay parents, so this can be used directly with a peers view.
+	/// Returns a map from candidate hash -> receipt
+	fn cached_live_candidates_unioned<'a>(
+		&'a self,
+		relay_parents: impl IntoIterator<Item = &'a Hash> + 'a,
+	) -> HashMap<Hash, CommittedCandidateReceipt> {
+		let relay_parents_and_ancestors = self.extend_with_ancestors(relay_parents);
+		relay_parents_and_ancestors
```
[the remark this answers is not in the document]
drahnr: It can't, we would need a mut and a immutable binding to `self`

primitives/src/v1.rs
```diff
@@ -189,6 +189,10 @@ impl<H> CommittedCandidateReceipt<H> {
 	pub fn descriptor(&self) -> &CandidateDescriptor<H> {
 		&self.descriptor
 	}
+	/// Get a reference to the candidate descriptor.
+	pub fn commitments(&self) -> &CandidateCommitments {
```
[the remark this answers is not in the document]
drahnr: It's the same for descriptor, either we could remove both or leave both.

node/network/availability-distribution/src/lib.rs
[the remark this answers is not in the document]
drahnr: Not sure if that helps much, if you look into the impl of `SubsystemContext::send_message`, it enqueues all messages in a `mpsc::channel()`, my gut feeling says its probably counter productive pending lack of measurements :signal_strength:

node/network/availability-distribution/src/lib.rs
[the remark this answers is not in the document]
coriolinus: Good point, but `OverseerSubsystemContext<M>` has a `mpsc::Sender` which is intentionally cheap to clone.

It would be very straightforward to add a method to `Context` like `fn sender(&self) -> mpsc::Sender<ToOverseer>`, but that might be too much. It would be less straightforward, but a better abstraction, to add a message like `fn message_sender<Message: Into<AllMessages>>(&self) -> mpsc::Sender<Message>`, and then spawn a little job which forwards everything from the returned sender to the overseer, constructing a `ToOverseer` message from it on the way.

In both cases, we could add the method as a default method on the trait, which keeps things simple for the implementer.

primitives/src/v1.rs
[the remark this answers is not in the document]
coriolinus: My inclination would be to remove both.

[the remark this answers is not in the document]
drahnr: Actually `descriptor()` seems to be a broader pattern, I'll only remove `commitments()` then.

node/network/availability-distribution/src/lib.rs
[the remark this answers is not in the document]
drahnr: I think this is equiv to the other topic, the overseer processes the messages in sequence as part of a `mpsc::channel()`, so panellizing does not provide many advantages, but a few risks: It spikes the number of elements being in the channel, potentially exhausting the channels capacity. It also would only cause a very short term concurrent generation of messages, which I doubt will be any faster given the overhead of launching the task on a thread in the threadpool.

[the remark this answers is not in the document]
drahnr: > Good point, but `OverseerSubsystemContext<M>` has a `mpsc::Sender` which is intentionally cheap to clone.
> 
> It would be very straightforward to add a method to `Context` like `fn sender(&self) -> mpsc::Sender<ToOverseer>`, but that might be too much. It would be less straightforward, but a better abstraction, to add a message like `fn message_sender<Message: Into<AllMessages>>(&self) -> mpsc::Sender<Message>`, and then spawn a little job which forwards everything from the returned sender to the overseer, constructing a `ToOverseer` message from it on the way.
> 
> In both cases, we could add the method as a default method on the trait, which keeps things simple for the implementer.

I think that mandates a separate PR :)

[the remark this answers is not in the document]
coriolinus: > I think this is equiv to the other topic, the overseer processes the messages in sequence as part of a `mpsc::channel()`...

That's a very good point. I'd forgotten that we would be blocking on the overseer in that case instead of directly on the peers. That may be something to consider changing in the future, but for now, feel free to ignore these suggestions; as you say, concurrency doesn't get us much at the moment.

node/network/availability-distribution/src/lib.rs
```diff
@@ -0,0 +1,1947 @@
...
+
+		let erasure_chunks = if let Ok(erasure_chunks) = derive_erasure_chunks_with_proofs(
+			per_candidate.validators.len(),
+			validator_index,
+			&available_data,
+		) {
+			erasure_chunks
+		} else {
+			return Ok(());
+		};
```
montekki, line 427: This is a done via lazy-evaluation fashion in `av-store`. So, if you have some data available and you ask `QueryChunk` message, the store will compute the chunks, store them and return a necessary one. Probably we could add a `QueryAllChunks` variant to get them all at once but still, the idea is that breaking data into chunks should happen at most once per stored `AvailableData`.

node/network/availability-distribution/src/lib.rs
```diff
@@ -0,0 +1,1947 @@
...
+
+		if let Some(erasure_chunk) = erasure_chunks.get(validator_index as usize) {
+			match store_chunk(ctx, candidate_hash, validator_index, erasure_chunk.clone()).await {
+				Err(e) => warn!(target: TARGET, "Failed to send store message to overseer: {:?}", e),
+				Ok(Err(())) => warn!(target: TARGET, "Failed to store our own erasure chunk"),
+				Ok(Ok(())) => {}
```
montekki, line 433: Again, if you do a `QueryChunk` it will make it available. I think that in the end you can completely remove the logic with `obtain_chunks` and the dependency on `polkadot-erasure-encoding`.

node/network/availability-distribution/src/lib.rs
```diff
@@ -0,0 +1,1947 @@
...
+		relay_parent,
+		RuntimeApiRequest::AvailabilityCores(tx),
+	)))
+	.await
+	.map_err::<Error, _>(Into::into)?;
+
+	let all_para_ids: Vec<_> = rx
+		.await
+		.map_err::<Error, _>(Into::into)?
+		.map_err::<Error, _>(Into::into)?;
```
montekki, line 936: hmmm, interesting, `rx.await??` would not work?

node/network/availability-distribution/src/lib.rs
drahnr: I see, this is redundant code essentially. Removing.

node/network/availability-distribution/src/lib.rs
```diff
@@ -0,0 +1,1947 @@
...
+		target: TARGET,
+		"Reputation change of {:?} for peer {:?}",
+		rep,
+		peer
+	);
+	ctx.send_message(AllMessages::NetworkBridge(
+		NetworkBridgeMessage::ReportPeer(peer, rep),
+	))
+	.await
+	.map_err::<Error, _>(Into::into)
```
montekki, line 966: I think that i've managed to somehow dodge the explicit error conversions so far; but maybe there is a reason for this here

drahnr: It might have become obsolete, but at one point, it was required to make it compile, let me check if that is still mandatory.

drahnr: Some have, most didn't become obsolete.

node/network/availability-distribution/src/lib.rs
```diff
@@ -0,0 +1,2022 @@
...
+	let live_candidate = if let Some(live_candidate) = live_candidates.get(&message.candidate_hash)
+	{
+		live_candidate
+	} else {
+		return modify_reputation(ctx, origin, COST_NOT_A_LIVE_CANDIDATE).await;
+	};
+
+	// check the merkle proof
+	let root = &live_candidate.commitments.erasure_root;
+	let anticipated_hash = if let Ok(hash) = branch_hash(
```
montekki, line 612: This can be filed as a TODO for later I guess but we should probably add this checking logic to an `impl` on the `ErasureChunk` primitive.

rphmeier: That would probably cause a circular dependency. However, we should expose a better API from `polkadot-erasure-coding`.

node/network/availability-distribution/src/lib.rs
```diff
@@ -0,0 +1,2030 @@
+// Copyright 2020 Parity Technologies (UK) Ltd.
```
bkchr, line 1: Can this not be split up into multiple files? Reading a 2k file is really not that funny.

drahnr: It can be and it will be :) , I think tests can be pushed to a separate file quite easily, but let's see how that goes.

node/network/availability-distribution/src/lib.rs
drahnr: I'll leave that for another PR

node/network/availability-distribution/src/lib.rs
```diff
@@ -0,0 +1,1093 @@
...
+			.cloned()
+			.collect::<HashSet<_>>();
+
+		send_tracked_gossip_messages_to_peer(ctx, per_candidate, origin.clone(), messages).await?;
+	}
+	Ok(())
+}
+
+/// Obtain the first key with a signing key, which must be ours. We obtain the index as `ValidatorIndex`
+/// If we cannot find a key in the validator set, which we could use.
```
ordian, line 574: this sentence seems unfinished: "..., `None` is returned"?

node/network/availability-distribution/src/lib.rs
```diff
@@ -0,0 +1,1093 @@
...
+		if let Some(ancestor_parent) = iter.peek() {
+			let session = query_session_index_for_child(ctx, *ancestor_parent).await?;
+			if session != desired_session {
+				break;
+			}
+			if ancestor != relay_parent {
+				acc.push(ancestor);
+			}
+		} else {
+			// we reached the genesis, which does not have ancestors
```
ordian, line 1081: not necessarily the genesis, but the oldest block in `ancestors`

node/network/availability-distribution/src/lib.rs
```diff
@@ -0,0 +1,1093 @@
...
+	// iterate from youngest to oldest
+	let mut iter = ancestors.into_iter().peekable();
+
+	while let Some(ancestor) = iter.next() {
+		if let Some(ancestor_parent) = iter.peek() {
+			let session = query_session_index_for_child(ctx, *ancestor_parent).await?;
+			if session != desired_session {
+				break;
+			}
+			if ancestor != relay_parent {
```
ordian, line 1077: I wonder why this check is needed? Doesn't `query_k_ancestors` guarantee not to return `relay_parent`?

node/network/availability-distribution/src/lib.rs
drahnr: In this case it must be genesis, since we are only interested in `k` blocks session index, but we query `k+1`.
Assuming the case of `k` being chosen with a good error margin around the start-to-prune history depth, this should only happen in the genesis case.

drahnr: Actually we don't want to include the genesis block here, there is nothing to do for it anyways.

node/network/availability-distribution/src/lib.rs
```diff
@@ -0,0 +1,1093 @@
...
+				acc.push(ancestor);
+			}
+		} else {
+			// we reached the genesis, which does not have ancestors
+			acc.push(ancestor);
```
ordian, line 1082: why do we push `ancestor` here? we don't know if it's session_index is same as desired_session, right?

drahnr: Se above :)

node/network/availability-distribution/src/lib.rs
ordian: Good point, I don't know if it's guaranteed that if we good pruning depth, query will never return less than `k+1` and `>0` hashes. Could it be the case that tip of the chain and relay parent we're querying are far away from each other. This is all speculation atm, but it would be good to handle this case in case our assumption is not correct.

node/network/availability-distribution/src/lib.rs
```diff
@@ -0,0 +1,1092 @@
...
+	}
+}
+
+impl<Context> Subsystem<Context> for AvailabilityDistributionSubsystem
+where
+	Context: SubsystemContext<Message = AvailabilityDistributionMessage> + Sync + Send,
+{
+	fn start(self, ctx: Context) -> SpawnedSubsystem {
+		SpawnedSubsystem {
+			name: "availability-distribution",
```
ordian, line 776:
```suggestion
			name: "availability-distribution-subsystem",
```

PR #1465, opened 2020-07-24
move bitfield-signing crate to `node/core`

rphmeier:
The convention is to have all subsystems in `node/core` or `node/network`. This might not be the best convention, but it's the convention. I'm definitely open to a conversation on how we could reorganize. For instance, mirroring the guide, which organizes subsystems by functionality.


coriolinus:
Exception: the overseer is simply `polkadot-overseer`. It's debatable whether the overseer is itself a subsystem, but it seems plausible that it should also be a member of `node/core`. That's out of scope of this PR, though.

rphmeier:
> Exception: the overseer is simply polkadot-overseer. It's debatable whether the overseer is itself a subsystem, but it seems plausible that it should also be a member of node/core. That's out of scope of this PR, though.

I don't see the overseer as a subsystem. It oversees the subsystems.

PR #1445, opened 2020-07-21
Update comments in polkadot runtime

joepetrowski:
Was about to comment on 1443 when it got merged.

PR #1466, opened 2020-07-24
Fix bitfield signing

coriolinus:
Fixes the problems with #1364, which was merged too soon.


rphmeier:
Could use a rebase. This is 3 PRs worth of commits!

node/core/bitfield-signing/src/lib.rs
```diff
@@ -169,6 +171,12 @@ async fn get_core_availability(
 	Ok(false)
 }
 
+// delegates to the v1 runtime API
+async fn get_availability_cores(_relay_parent: Hash, _sender: &mut mpsc::Sender<FromJob>) -> Result<Vec<CoreState>, Error> {
```
rphmeier, line 175: I realized this later, but another way to stub this out would be to add the variant to the `RuntimeApiRequest` in the `subsystem` crate.

coriolinus: I like that idea; 146746c.

coriolinus:
Rebasing dramatically simplifies this PR, it turns out.

coriolinus:
An interesting fact: the failing test, `protocol::tests::erasure_fetch_drop_also_drops_gossip_sender`, succeeds instantly when run in isolation, but reliably hangs when running the full suite. This suggests that the test is malformed; context from outside the test body determines whether or not it succeeds. 

```
$ time cargo test -p polkadot-network drops_gossip_sender
Mon 27 Jul 2020 01:11:53 PM CEST
    Finished test [unoptimized + debuginfo] target(s) in 0.25s
     Running target/debug/deps/polkadot_network-24a4527cc39ffa6b

running 1 test
test protocol::tests::erasure_fetch_drop_also_drops_gossip_sender ... ok

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 22 filtered out


real	0m0.302s
user	0m0.237s
sys	0m0.086s
```

coriolinus:
This test appears to have become flaky at some point before or during #1469: the master-merge for that PR failed CI, as did the master-merge for #1445. In both of those cases, the test succeeded on the final commit before the squash-merge. However, three other master-merge CI runs in that span succeeded.

It's not obvious to me how the test is supposed to work, or why it's sometimes failing. The obvious suspicion is that the `loop` block isn't terminating, but it's not clear why it would have been reliable for some months and only become flaky now. The test itself [hasn't changed](https://github.com/paritytech/polkadot/blame/master/network/src/protocol/tests.rs#L470-L530) since #897. 

@rphmeier AFAICT you're the person who has the best chance of understanding this issue; what's your take on it?

montekki:
^^^^ Also saw this thing with the test on some CI runs, but atm it is not failing with this test. Probably fixing the build could help

node/subsystem/src/messages.rs
```diff
@@ -25,11 +25,28 @@
...
+	HeadData,
+	Id as ParaId,
+	OmittedValidationData,
+	PoV,
+	SignedAvailabilityBitfield,
+	SigningContext,
+	ValidationCode,
+	ValidatorId,
+	ValidatorIndex,
+	ValidatorSignature,
```
montekki, line 49: Not sure if thats ok; if that's the IDE sorting imports (rust-analyzer can do that I think) could that at least not do the one-per-line thing?

node/subsystem/src/messages.rs
```diff
@@ -240,19 +257,28 @@ pub enum AvailabilityStoreMessage {
...
 
 	/// Query an `ErasureChunk` from the AV store.
 	QueryChunk(Hash, ValidatorIndex, oneshot::Sender<Option<ErasureChunk>>),
 
+	/// Query whether an `ErasureChunk` exists within the AV Store.
+	///
+	/// This is useful in cases like bitfield signing, when existence
+	/// matters, but we don't want to necessarily pass around large
+	/// quantities of data to get a single bit of information.
+	QueryChunkAvailability(Hash, ValidatorIndex, oneshot::Sender<bool>),
```
montekki, line 273: I had serious doubts as I was adding `QueryDataAvaialbility` message; is it really better for performance? We anyway lift one big blob from disk, there is no way to ask `kvdb` for just availability; when we send whatever data struct that's containing the pointer to this blob, only it will be sent around, wouldn't it?

node/subsystem/src/messages.rs
```diff
@@ -240,19 +257,28 @@ pub enum AvailabilityStoreMessage {
...
+
 	/// Store an `ErasureChunk` in the AV store.
 	///
 	/// Return `Ok(())` if the store operation succeeded, `Err(())` if it failed.
 	StoreChunk(Hash, ValidatorIndex, ErasureChunk, oneshot::Sender<Result<(), ()>>),
 
+
+
```
montekki, line 281: These are probably unnecessary

node/subsystem/src/messages.rs
coriolinus: I did that intentionally: I was having trouble diffing the imports when resolving a merge conflict in which people added and removed various items in different places. Using one per line, alphabetically, made it much easier to sort things out.

node/subsystem/src/messages.rs
coriolinus: At worst, it masks a `kvdb` limitation which can be fixed in the future.

coriolinus: https://github.com/paritytech/parity-common/pull/409

coriolinus:
This passed, this time, but I do think we should resolve #1474 eventually.

node/subsystem/src/messages.rs
rphmeier: It would be better to have them be more compact. We don't use this style anywhere else.

coriolinus: As a question of policy, I'd prefer to leave them like this, and expand the use of this style. Rationale: the vast majority of the time, we're not looking at the import lines, but elsewhere in the file. Of the times we are in fact looking at the import lines, it's when adding or merging imports; in those times, it's faster when the lists are broad and alphabetical than when using basically any other taxonomy.

Nevertheless, 8e5c6da.

bkchr: Maybe the general rule should be to merge all crates import into one import, after that the compiler will ensure that you don't have duplicates in this list.

coriolinus:
@rphmeier I got a LGTM 5 days ago; could I get a formal approval so we can merge this PR please?

rphmeier:
bot merge

PR #1467, opened 2020-07-24
Refactor service to allow building full (and light) node matching chain spec

cecton:
Reviewers: looking at the code you won't see easily what I am trying to do. So please read carefully.

I'm working on https://github.com/paritytech/cumulus/issues/152

Right now cumulus only uses the `fn start_collator()` from polkadot-collator. What I need to do is to allow cumulus to run a standard polkadot node.

In the current state of polkadot we need to implement a pattern like this:
```
if chain spec is westend {
    do something
} else if chain spec is kusama {
    do the same thing but kusama-style 
} else {
    do the same thing polkadot style
}
```

This is annoying because this logic belongs to the polkadot repo, not to cumulus. The reason why we need to do this at the moment is because there is a specific Client for every chain spec type and we can't use dynamic dispatch on it (because Rust won't allow it for types that use Self in method arguments or in type parameter, etc...).

Thanks to @expenses, the situation has improved recently because before that, we had an AbstractService and it was the same story. Now there is only the Client.

This means that everywhere we wanted to create a light or full service, we had the pattern I just explained above. But we don't need that pattern anymore, we can simply return the TaskManager and put this pattern in a single place (or almost!). So I made a NodeBuilder with the functions `fn build_full()` and `fn build_light()` that does the pattern.

This will allow me, in cumulus, to build a polkadot node without re-implementing that same pattern.


service/src/lib.rs
```diff
@@ -793,26 +793,75 @@ pub struct FullNodeHandles {
...
+			config,
+		}
+	}
 
-/// Create a new Westend service for a light client.
-pub fn westend_new_light(config: Configuration, ) -> Result<
-	(TaskManager, Arc<RpcHandlers>), ServiceError
->
-{
-	new_light::<westend_runtime::RuntimeApi, KusamaExecutor, _>(config)
```
cecton, line 817: Note that I have removed the light node builders. I wish I could do the same for the full builders but we still need them for the `start_collator`.

cli/src/browser.rs
```diff
@@ -46,8 +46,8 @@ async fn start_inner(chain_spec: String, log_level: String) -> Result<Client, Bo
 	info!("👤 Role: {}", config.display_role());
 
 	// Create the service. This is the most heavy initialization step.
-	let (task_manager, rpc_handlers) = service::kusama_new_light(config)
-		.map_err(|e| format!("{:?}", e))?;
+	let builder = service::NodeBuilder::new(config);
+	let (task_manager, rpc_handlers) = builder.build_light().map_err(|e| format!("{:?}", e))?;
```
cecton, line 50: I changed this. @expenses maybe you know if this is correct?

Before: it uses only kusama
After: it uses kusama if the chain spec is kusama, westend if westend **AND fallback to polkadot if it can't say what chain spec it is**

It sounds reasonable but maybe there is a reason why we forced kusama in the browser?

expenses: No, there wasn't a reason for doing that as far as I know.

expenses: So what you've done is probably an improvement.

cecton: Purrfect

expenses: :smile_cat:

cecton:
bot merge

cecton:
@sjeohp bot error :point_up:

PR #1470, opened 2020-07-26
Add manual action to build PR binary

ddorgan:
Add a manual action to allow other services trigger a build and extract the binary file for open PR's.


ddorgan:
Removing the github actions build. Just leaving in the option for manual builds via gitlab which should be triggered via the process bot. @General-Beck any feedback?

PR #1478, opened 2020-07-27
README.md: Add clang-dev to required dependencies to install

joshtriplett:


joshtriplett:
The README could probably also use an update to make it clear that you need to install dependencies before the `Install via Cargo` section, too.

joshtriplett:
Didn't realize there was a CLA requirement. Not worth it for a one-word change. (I'd have been happy to provide a [`Signed-off-by`](https://developercertificate.org/) for provenance.)

Issue #1474, opened 2020-07-27
protocol::tests::erasure_fetch_drop_also_drops_gossip_sender is flaky


coriolinus:
See discussion beginning [here](https://github.com/paritytech/polkadot/pull/1466#issuecomment-664333834).

Demonstration: of the most recent 5 (at the time of writing) commits to `master`, 2 of them timed out on that test. Both of those passed that test as PRs.

![Screenshot from 2020-07-27 15-51-32](https://user-images.githubusercontent.com/7822926/88549915-30d53a00-d021-11ea-8de1-c43b5b9258c2.png)

rphmeier:
This can be taken care of by just removing the `polkadot-network` crate, which we'll be able to do soon.

PR #1490, opened 2020-07-28
add simple name derivation heuristic for JobManager Subsystem impl

coriolinus:
Subsystems are encouraged to either typedef themselves as appropriate
`JobManager` instances for their job type, or wrap a `JobManager`
instance and delegate the `Subsystem` impl. In both cases, we want
to use a sensible, non-repeated subsystem name for appropriate
logging and debugging.

This PR adds a heuristic: if the job name ends in the literal
"Job", then that gets stripped. Otherwise, the job name is used.
This improves on the previous situation, in which subsystems
typedef'd to or wrapping `JobManager` all got the same constant (!)
name.

### DIFF
--- a/node/subsystem/src/util.rs
+++ b/node/subsystem/src/util.rs
@@ -711,7 +711,7 @@ where
 		});
 
 		SpawnedSubsystem {
-			name: "JobManager",
+			name: Job::NAME.strip_suffix("Job").unwrap_or(Job::NAME),
 			future,
 		}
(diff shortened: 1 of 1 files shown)


coriolinus:
CI fails due to https://github.com/paritytech/polkadot/issues/1474.

rphmeier:
bot merge

rphmeier:
Thanks!

PR #1493, opened 2020-07-28
[CI] Publish draft release redux

s3krit:
This PR is an overhaul of the way we publish draft releases on pushing of a new tag.
The old method was a very cumbersome 274-line bash script in addition to some helper functions in `lib.sh`.

The new version:
* Outputs releases in exactly the same format as the bash version
* Uses a Github action as opposed to a gitlab-ci job
* Does all of the heavy-lifting in a [ruby gem](https://rubygems.org/gems/changelogerator) I wrote (which you don't need to worry about in the context of this PR)
* Uses a template for formatting the release notes, which makes it much easier to edit (and separates formatting from logic)
* Accesses Github way less so ends up running faster for releases with lots of changes

Using a *real* language rather than bash makes it a lot more readable and maintainable, even if you don't know so much Ruby (compared to the weird bash I was having to use).


scripts/github/generate_release_text.rb
```diff
@@ -0,0 +1,73 @@
...
+
+# get last polkadot version. Use handy Gem::Version for sorting by version
+last_version = pg
+              .tags
+              .map(&:name)
+              .grep(/^v\d+\.\d+\.\d+.*$/)
+              .sort_by { |v| Gem::Version.new(v.slice(1...)) }[-2]
+
+polkadot_cl = Changelog.new(
+  's3krit/polkadot', version, last_version, token: token
```
kirushik, line 28: I think there's a typo here?
```suggestion
  'paritytech/polkadot', version, last_version, token: token
```

s3krit: Good catch :) result of me testing on a fork

.github/workflows/publish-draft-release.yml
```diff
@@ -0,0 +1,68 @@
...
+      release_url: ${{ steps.create-release.outputs.html_url }}
+    steps:
+    - uses: actions/checkout@v2
+      with:
+        fetch-depth: 0
+        path: polkadot
+    - name: Set up Ruby 2.7
+      uses: actions/setup-ruby@v1
+      with:
+        ruby-version: 2.6
```
kirushik, line 35: It says "Set up Ruby 2.7" just a few lines above

scripts/github/generate_release_text.rb
```diff
@@ -0,0 +1,73 @@
...
+client_changes = Changelog.changes_with_label(all_changes, 'B5-clientnoteworthy')
+runtime_changes = Changelog.changes_with_label(all_changes, 'B7-runtimenoteworthy')
+
+release_priority = Changelog.highest_priority_for_changes(all_changes)
+
+# Pulled from the previous Github step
+rustc_stable = ENV['RUSTC_STABLE']
+rustc_nightly = ENV['RUSTC_NIGHTLY']
+
+polkadot_runtime = File.open(polkadot_path + '/runtime/polkadot/src/lib.rs')
```
dvdplm, line 64: Maybe 
```ruby 
polkadot_runtime = File.open(filepath) do |f|
  f.find {|l| l =~ /spec_version/ }.match(/[0-9]/+).first
end
```
…to make sure you close the file after reading?

s3krit: Good shout on the file handling :) thanks.
Though for whatever reason, [MatchData](https://ruby-doc.org/core-2.7.1/MatchData.html) doesn't have `first` or `last` methods...

PR #1455, opened 2020-07-22
Companion: ignore checks for companion PR

cecton:
Follow up of #1426 

To make our CI work less and improve the speed of merge between a substrate branch and the polkadot companion branch, this PR skips most of the tests.

The tests are already done in substrate check-polkadot-companion-build:

PR #1499, opened 2020-07-29
Companion PR for `Add a `DefaultQueue` type alias to remove the need to use `sp_api::TransactionFor``

expenses:
https://github.com/paritytech/substrate/pull/6761

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1370,7 +1370,7 @@ checksum = "3f9eec918d3f24069decb9af1554cad7c880e2da24a9afd88aca000531ab82c1"
 [[package]]
 name = "fork-tree"
 version = "2.0.0-rc5"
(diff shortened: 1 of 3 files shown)


expenses:
Is there a way to get the substrate PR merged without switching the branch on this one temporarily so the tests pass?

bkchr:
You don't need to have tests passed here.

andresilva:
The tests here shouldn't need to pass for the substrate companion build to pass, I think it only checks whether this PR is mergeable or not (without enforcing all CI tasks here are green).

cecton:
> Is there a way to get the substrate PR merged without switching the branch on this one temporarily so the tests pass?

I have a PR that fixes that. All the tests here will be green. (for all future companion pr) #1455

PR #1503, opened 2020-07-30
Implementer's guide: downward messages and HRMP, take 2

### LINKED ISSUE #1139: Guide: Downwards Messages

pepyakin:
Closes #1139
Subsumes #1399 

This is a refined version of #1399 that uses channels for HRMP. Changes from the previous approach:

1. HRMP uses channels now. The channels behavior mimics that of XCMP. It's important to note that the messages are removed eagerly when a channel is closed or the para being offboarded.
1. Now we somewhat consider offboarding.
1. DMQ now is reserved exclusively for VMP. HRMP messages are not routed through DMQ.
1. To specify what messages were drained from the HRMP channels we introduce a `hrmp_watermark` into the candidate.


roadmap/implementers-guide/src/runtime/paras.md
```diff
@@ -93,7 +93,9 @@ OutgoingParas: Vec<ParaId>;
 
 ## Session Change
 
-1. Clean up outgoing paras. This means removing the entries under `Heads`, `ValidationCode`, `FutureCodeUpgrades`, and `FutureCode`. An according entry should be added to `PastCode`, `PastCodeMeta`, and `PastCodePruning` using the outgoing `ParaId` and removed `ValidationCode` value. This is because any outdated validation code must remain available on-chain for a determined amount of blocks, and validation code outdated by de-registering the para is still subject to that invariant.
+1. Clean up outgoing paras.
+	1. This means removing the entries under `Heads`, `ValidationCode`, `FutureCodeUpgrades`, and `FutureCode`. An according entry should be added to `PastCode`, `PastCodeMeta`, and `PastCodePruning` using the outgoing `ParaId` and removed `ValidationCode` value. This is because any outdated validation code must remain available on-chain for a determined amount of blocks, and validation code outdated by de-registering the para is still subject to that invariant.
+	1. Call `Router::offboard_para(P)` for each offboarded para `P`.
```
rphmeier, line 98: This implies that `Router` should be before `Paras` in initialization order, doesn't it?

roadmap/implementers-guide/src/runtime/router.md
```diff
@@ -19,6 +17,91 @@ RelayDispatchQueues: map ParaId => Vec<UpwardMessage>;
...
+
+/// A description of a request to open an HRMP channel.
+struct HrmpOpenChannelRequest {
+	/// The sender and the initiator of this request.
+    sender: ParaId,
+	/// The recipient of the opened request.
+    recipient: ParaId,
+	/// Indicates if this request was confirmed by the recipient.
+    confirmed: bool,
+	/// How many session boundaries this request has seen.
```
rphmeier, line 40:
```suggestion
	/// How many session boundaries this request was seen.
```

roadmap/implementers-guide/src/runtime/router.md
```diff
@@ -19,6 +17,91 @@ RelayDispatchQueues: map ParaId => Vec<UpwardMessage>;
...
+    age: u32,
```
rphmeier, line 41: Do we use `SessionIndex` type anywhere else in guide? I guess this should be it

roadmap/implementers-guide/src/runtime/router.md
```diff
@@ -19,6 +17,91 @@ RelayDispatchQueues: map ParaId => Vec<UpwardMessage>;
...
+    limit_used_bytes: u32,
+}
+
+/// A description of a request to close an opened HRMP channel.
+struct HrmpCloseChannelRequest {
+	/// The para which initiated closing an existing channel.
+    /// invariant: equals either to sender or recipient.
+    initiator: ParaId,
+	/// The identifier of an HRMP channel to be closed.
+    id: HrmpChannelId,
```
rphmeier, line 56: tabs & spaces mixed in this file, not just here

roadmap/implementers-guide/src/runtime/router.md
```diff
@@ -19,6 +17,91 @@ RelayDispatchQueues: map ParaId => Vec<UpwardMessage>;
...
+}
+
+/// A metadata of an HRMP channel.
+struct HrmpChannel {
+    /// The amount that the sender supplied as a deposit when opening this channel.
+    sender_deposit: Balance,
+	/// The amount that the recipient supplied as a deposit when accepting opening this channel.
+    recipient_deposit: Balance,
```
rphmeier, line 64: What is the actual deposit mechanism? What actions are prevented and how might the accounting work in practice?

roadmap/implementers-guide/src/runtime/router.md
```diff
@@ -19,6 +17,91 @@ RelayDispatchQueues: map ParaId => Vec<UpwardMessage>;
...
+	/// The total size in bytes of all message payloads in the channel.
+	/// Invariant: should be less or equal to `limit_used_bytes`.
+    used_bytes: u32,
+}
+```
+HRMP related storage layout
+
+```rust,ignore
+/// Pending HRMP open channel requests.
+HrmpOpenChannelRequests: Vec<HrmpOpenChannelRequest>;
```
rphmeier, line 81: sorted in any way?

roadmap/implementers-guide/src/runtime/router.md
```diff
@@ -19,6 +17,91 @@ RelayDispatchQueues: map ParaId => Vec<UpwardMessage>;
...
+DownwardMessageQueues: map ParaId => Vec<DownwardMessage>;
+```
+
+### HRMP
+
+HRMP related structs:
+
+```rust,ignore
+/// A type used to designate a HRMP channel between a (sender, recipient).
+type HrmpChannelId = (ParaId, ParaId);
```
rphmeier, line 30: would it make sense to use a struct here with explicit `sender, recipient`?

roadmap/implementers-guide/src/runtime/router.md
```diff
@@ -27,9 +110,108 @@ No initialization routine runs for this module.
 
 ## Routines
 
+The following routines are intended to be invoked by paras' upward messages.
+
+* `init_open_channel(recipient)`:
```
rphmeier, line 115: rename to `hrmp_*` for these functions? 

also, as an architecture question, might it make sense to have a separate `Hrmp` module? Or is everything in `Router` module OK?

roadmap/implementers-guide/src/runtime/router.md
```diff
@@ -19,6 +17,91 @@ RelayDispatchQueues: map ParaId => Vec<UpwardMessage>;
...
+    limit_used_places: u32,
+	/// The maximum total size of the messages that can be pending in the channel at once.
+    limit_used_bytes: u32,
+}
+
+/// A description of a request to close an opened HRMP channel.
+struct HrmpCloseChannelRequest {
+	/// The para which initiated closing an existing channel.
+    /// invariant: equals either to sender or recipient.
+    initiator: ParaId,
```
rphmeier, line 54: maybe another variant for the relay-chain, so governance can forcibly close all channels.

And another thing - when we phase HRMP out, we can have a runtime upgrade to close all channels and stop accepting new ones.

roadmap/implementers-guide/src/runtime/router.md
```diff
@@ -27,9 +110,108 @@ No initialization routine runs for this module.
...
 
+The following routines are intended to be invoked by paras' upward messages.
+
+* `init_open_channel(recipient)`:
+  1. Check that the origin of this message is a para. We say that the origin of this message is the `sender`.
+  1. Check that the `sender` is not `recipient`.
+  1. Check that the `recipient` para exists.
+  1. Check that there is no existing open channel request from sender to recipient.
+  1. Check that the sum of the number of already opened HRMP channels by the `sender` (the size of the set found `HrmpEgressChannelsIndex` for `sender`) and the number of open requests by the `sender` (the value from `HrmpOpenChannelRequestCount` for `sender`) doesn't exceed the limit of channels (`config.hrmp_max_parachain_outbound_channels` or `config.hrmp_max_parathread_outbound_channels`) minus 1.
+  1. Reserve the deposit for the `sender` according to `config.hrmp_sender_deposit`
```
rphmeier, line 121: @alistairstewart It would be helpful to have some more guidance on deposits for parachains/parathreads and how this interacts with the broader parachain / parathread deposits, respectively

roadmap/implementers-guide/src/runtime/router.md
```diff
@@ -27,9 +110,108 @@ No initialization routine runs for this module.
...
+  1. Add a new entry to `HrmpOpenChannelRequests` and increase `HrmpOpenChannelRequestCount` by 1 for the `sender`.
+      1. Set `sender_deposit` to `config.hrmp_sender_deposit`
+      1. Set `limit_used_places` to `config.hrmp_channel_max_places`
+      1. Set `limit_limit_used_bytes` to `config.hrmp_channel_max_size`
+
+* `accept_open_channel(i)`, `i` - is the index of open channel request:
```
rphmeier, line 127: I see, so these are all `UpwardMessages`. cool. Noting that implementation depends on https://github.com/paritytech/polkadot/issues/1510

roadmap/implementers-guide/src/runtime/router.md
```diff
@@ -27,9 +110,108 @@ No initialization routine runs for this module.
...
+  1. Check that the designated open channel request exists
+  1. Check that the request's `recipient` corresponds to the origin of this message.
+  1. Reserve the deposit for the `recipient` according to `config.hrmp_recipient_deposit`
+  1. Set the request's `confirmed` flag to `true`.
+
+* `close_channel(sender, recipient)`:
+  1. Check that the channel between `sender` and `recipient` exists
+  1. Check that the origin of the message is either `sender` or `recipient`
+  1. Check that there is no existing intention to close the channel between `sender` and `recipient`.
```
rphmeier, line 136: what exactly happens here? i could imagine a case where both `sender` and `recipient` initiate at the same time & race.

could it even count as a confirmed close?

rphmeier: alternate behavior would be to ignore, and for paras to confirm close requests even when, from their perspective, they have already initiated a close.

rphmeier: either way, would be nice to specify what happens here and what the corner cases are.

roadmap/implementers-guide/src/runtime/router.md
```diff
@@ -27,9 +110,108 @@ No initialization routine runs for this module.
...
+The routines described below are for use within a session change and called by the `Paras` module.
+
+* `offboard_para(P: ParaId)`:
+  1. Remove all inbound channels of `P`, i.e. `(_, P)`,
+  1. Remove all outbound channels of `P`, i.e. `(P, _)`,
+  1. Remove all `DownwardMessageQueues` of `P`.
+  1. Remove `RelayDispatchQueueSize` of `P`.
+  1. Remove `RelayDispatchQueues` of `P`.
+  - Note that we don't remove the open/close requests since they are gon die out naturally.
+TODO: What happens with the deposits in channels or open requests?
```
rphmeier, line 185: great question! lol

roadmap/implementers-guide/src/types/candidate.md
```diff
@@ -145,6 +145,9 @@ struct LocalValidationData {
 	/// which case the code upgrade should be applied at the end of the signaling
 	/// block.
 	code_upgrade_allowed: Option<BlockNumber>,
+
+	// TODO: Does this look sensible?
+	hrmp_ingress: BTreeMap<ParaId /* sender */, Vec<InboundHrmpMessage>>,
```
rphmeier, line 150: Yeah, that seems fine. 

I would also expect a field or fields in either `GlobalValidationData` or `LocalValidationData` that can be used to ensure that not too many messages or too much data is sent.

roadmap/implementers-guide/src/runtime/router.md
pepyakin: Oh, nice to know

roadmap/implementers-guide/src/types/candidate.md
rphmeier: i guess we need such a field for upward messages too, but that seems beyond the scope of this PR

roadmap/implementers-guide/src/runtime/router.md
pepyakin: I thought that those don't need any particular sorting. 
The reason: we iterate over all of them linearly at each session boundary and we confirm them using an explicit index.

roadmap/implementers-guide/src/runtime/router.md
pepyakin: When I was prefixing everything (and apparently missed these messages) I was tempted to extract this into a separate module. But then I got thinking that then XCMP will need to be placed in a separate module only leaving VMP here which kinda ruins the whole picture. And I haven't. So this is something to think about.

roadmap/implementers-guide/src/runtime/router.md
pepyakin: good point, I'll try to incorporate that

roadmap/implementers-guide/src/runtime/router.md
pepyakin: Heh, well, that's a subject for discussion. Remember I asked you about introducing special kinds of `UpwardMessage`s instead of relying on dispatching them? That was specifically for this case. I was wondering if it makes sense to add dedicated variants for these messages so that we could reject candidates that fail any of those checks. I didn't go with that way since I thought it might have been too strict. But what do you think?

roadmap/implementers-guide/src/runtime/router.md
pepyakin: There is no notion of close confirmation in this spec, since we decided to go ahead with unilateral closing with dropping messages atm to move things forward (Right?)

So there is no difference apart the value of `initiator` which I guess doesn't make any difference in the current spec and could be safely removed. In other words, if the race happens, even though the lost party's message fails to dispatch, the result is for all practical reasons is the same.

roadmap/implementers-guide/src/runtime/router.md
AlistairStewart: The easiest thing would be to have the sender always put down a fixed deposit. While it may be able to share code or storage location with the auction deposit, it shouldn't interact in any way.

roadmap/implementers-guide/src/runtime/router.md
AlistairStewart: What happens to DOTs owned by offboarded paras more generally? A parachain with a large locked depsoit for an auction should become a parathread, which should then be able to dispose of it's free DOTs. But parathreads will in general have a small deposit for being a parathread, which means that it will own DOTs when it is offboarded.

So their probably should be an account these DOTs go to, by default the treasury. If these deposits are in the parachain account but just reserved, then we might not need to clean them up specifically here. So what happens to the para's account on the relay chain when it is offboarded?

roadmap/implementers-guide/src/runtime/router.md
rphmeier: it seems reasonable to reject candidates that fail any checks, actually.

pepyakin: Ok, in that case we should relax some requirements to avoid unnecessary rejecting candidates. I.e. [closing](https://github.com/paritytech/polkadot/pull/1503#pullrequestreview-459577136) should not fail if already requested by the other party.

rphmeier: but that property might not be easily preserved for e.g. contextual execution.

roadmap/implementers-guide/src/runtime/router.md
pepyakin: Yeah, although note that I prefer to use the `(sender, recipient)` syntax in the body.

roadmap/implementers-guide/src/runtime/router.md
pepyakin: Why not? `close_channel(C)` should be idempotent for the same C even though if it was sent by different origins (At least, as long as we don't allow contextual execution across sessions)

roadmap/implementers-guide/src/types/candidate.md
pepyakin: > but that seems beyond the scope of this PR

https://github.com/paritytech/polkadot/issues/1534

roadmap/implementers-guide/src/runtime/router.md
pepyakin: done

roadmap/implementers-guide/src/runtime/router.md
```diff
@@ -75,6 +75,14 @@ struct HrmpChannel {
     /// The total size in bytes of all message payloads in the channel.
     /// Invariant: should be less or equal to `limit_used_bytes`.
     used_bytes: u32,
+    /// A head of the Message Queue Chain for this channel. Each link in this chain has a form:
+	/// `(prev_head, B, H(M))`, where
+	/// - `prev_head`: is the previous value of `mqc_head`.
+	/// - `B`: is the [relay-chain] block number in which a message was appended
+	/// - `H(M)`: is the hash of the message being appended.
+	/// This value is initialized to a special value that consists of all zeroes which indicates
```
rphmeier, line 87: Is this comment about the value of the field when no messages are pending?

pepyakin: No, this comment refers to the initial state assigned to this field when a new fresh channel is created. One can answer a question if there are pending messages in the channel only having:
- the mqc head
- the receiver's watermark
- the preimage of the latest link (to extract the block number of the latest message)
FWIW, this is similar to what we have in XCMP

pepyakin:
Some tabs sneaked in *.md again. Changed the configuration of my editor, so I hope this won't happen again.

roadmap/implementers-guide/src/runtime/paras.md
pepyakin: Ok, now we are tracking the outgoing paras in the Router module

roadmap/implementers-guide/src/runtime/router.md
rphmeier: > What happens to DOTs owned by offboarded paras more generally?

The `slots` module maintains an Offboarding address for each parachain. This is initially set to the address of the bidder, but the parachain can submit a `Call` to change it to an address of its choosing. When the parachain is offboarded, its deposit is sent to the given address.

roadmap/implementers-guide/src/runtime/inclusion.md
```diff
@@ -68,13 +68,20 @@ All failed checks should lead to an unrecoverable error making the block invalid
   1. Check the collator's signature on the candidate data.
   1. check the backing of the candidate using the signatures and the bitfields, comparing against the validators assigned to the groups, fetched with the `group_validators` lookup.
   1. check that the upward messages, when combined with the existing queue size, are not exceeding `config.max_upward_queue_count` and `config.watermark_upward_queue_size` parameters.
+  1. call `Router::check_processed_downward_messages(para, commitments.processed_downward_messages)` to check that the DMQ is properly drained.
+  1. call `Router::check_hrmp_watermark(para, commitments.hrmp_watermark)` for each candidate to check rules of processing the HRMP watermark.
+  1. check that in the commitments of each candidate the horizontal messages are sorted by ascending recipient ParaId and there is no two horizontal messages have the same recipient.
+  1. using `Router::verify_outbound_hrmp(sender, commitments.horizontal_messages)` ensure that the each candidate's para correctly send horizontal messages.
```
rphmeier, line 74:
```suggestion
  1. using `Router::verify_outbound_hrmp(sender, commitments.horizontal_messages)` ensure that the each candidate send a valid set of horizontal messages
```

rphmeier:
We can defer the `ValidationData` refactoring to another PR (#1539). This LGTM

PR #1504, opened 2020-07-30
reduce slash defer durations

rphmeier:


runtime/kusama/src/lib.rs
```diff
@@ -318,7 +318,7 @@ parameter_types! {
 	// 28 eras for unbonding (7 days).
 	pub const BondingDuration: staking::EraIndex = 28;
 	// 28 eras in which slashes can be cancelled (7 days).
```
shawntabrizi, line 320:
```suggestion
	// 27 eras in which slashes can be cancelled (~7 days).
```

And in the other places too

runtime/kusama/src/lib.rs
```diff
@@ -318,7 +318,7 @@ parameter_types! {
...
-	pub const SlashDeferDuration: staking::EraIndex = 28;
+	pub const SlashDeferDuration: staking::EraIndex = 27;
```
pepyakin, line 321: Update the comment?

runtime/westend/src/lib.rs
```diff
@@ -295,7 +295,7 @@ parameter_types! {
 	// 28 eras for unbonding (7 days).
 	pub const BondingDuration: staking::EraIndex = 28;
 	// 28 eras in which slashes can be cancelled (7 days).
-	pub const SlashDeferDuration: staking::EraIndex = 28;
+	pub const SlashDeferDuration: staking::EraIndex = 27;
```
pepyakin, line 298: Ditto?

PR #1506, opened 2020-07-30
[CI] Build wasm blob with srtool and include prop hashes and blobs in release notes

s3krit:
Previously, building the WASM blobs was done on the cleanroom server, manually initiated by me. Since we have reproducible builds of the WASM blobs using `srtool` now, this isn't necessary.

This PR includes the building of those blobs in Github Actions - using the Github-provided runners, this only takes 30 minutes for a clean build (around 10 minutes after that depending on caching, which is implemented) - which is faster than the Gitlab pipeline takes from start to finish, so this doesn't actually delay anything.

One important change is that now, the creation of the draft-release on tagging will *only* happen after the WASM blobs are built - which means there will be that 10-35 minute gap between pushing a tag and the draft release being created (and the #polkadot matrix channel being notified).

An example of the format of the release can be found here:

https://github.com/s3krit/polkadot/releases/tag/v0.8.19


.github/workflows/publish-draft-release.yml
```diff
@@ -6,6 +6,51 @@ on:
       - v**.**.**
 
 jobs:
+  build-runtimes:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        runtime: ['polkadot', 'kusama']
+    container:
+      image: chevdor/srtool:nightly-2020-07-20
```
gabreal, line 15: afaiu according to our [Docker images maintenance policy](https://github.com/paritytech/devops/wiki/Docker-images-maintenance-policy) only official images or images based on them should be used.

PR #1516, opened 2020-07-31
Sort out validation errors

### LINKED ISSUE #1477: Refine handling of lower-level Wasm validation errors for Candidate Validation Subsystem
Currently we have a blanket `ValidationFailed` type which indicates all types of errors.

The mapping that we have from `wasm_executor::Error` to `ValidationFailed` or `ValidationResult::Invalid` also needs refinement.

It would be better to have more variants under `ValidationFailed` and to go through the `wasm_executor::Error` enum to determine which variants should be treated as the candidate being invalid and which variants imply an internal error of some kind. The rule of thumb to use is: if we treat the candidate as invalid, we should expect that validators will issue signed statements to that effect. So it shouldn't encompass any spurious errors.

arkpar:
Closes #1477

### DIFF
--- a/node/core/candidate-validation/src/lib.rs
+++ b/node/core/candidate-validation/src/lib.rs
@@ -26,8 +26,8 @@ use polkadot_subsystem::{
 };
 use polkadot_subsystem::messages::{
 	AllMessages, CandidateValidationMessage, RuntimeApiMessage, ValidationFailed, RuntimeApiRequest,
-	RuntimeApiError,
 };
+use polkadot_subsystem::errors::RuntimeApiError;
 use polkadot_node_primitives::{ValidationResult, ValidationOutputs, InvalidCandidate};
 use polkadot_primitives::v1::{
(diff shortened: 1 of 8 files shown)


node/core/candidate-validation/src/lib.rs
```diff
@@ -321,52 +322,52 @@ async fn spawn_validate_exhaustive(
...
 	pov: &PoV,
-) -> bool {
+) -> Result<(), InvalidCandidate> {
 	let encoded_pov = pov.encode();
 	let hash = pov.hash();
 
 	if let Some(max_size) = max_block_data_size {
 		if encoded_pov.len() as u64 > max_size {
-			return false;
+			return Err(InvalidCandidate::ParamsTooLarge(encoded_pov.len() as u64));
```
rphmeier, line 337: This `max_block_data_size` is not a protocol limit, but we could make it one so that there is consensus on it.

rphmeier:
bot merge

PR #1517, opened 2020-08-01
Sudo wrapper for paras

montekki:
A sudo wrapper for paras was forgotten. Should constant weights be here and return values from the functions be ignored? If not how do we go from `Weight` to `DispatchResult`?

### DIFF
--- a/runtime/parachains/src/paras.rs
+++ b/runtime/parachains/src/paras.rs
@@ -32,11 +32,9 @@ use primitives::v1::{
 use frame_support::{
 	decl_storage, decl_module, decl_error,
 	traits::Get,
-	dispatch::DispatchResult,
-	weights::{DispatchClass, Weight},
+	weights::Weight,
 };
 use codec::{Encode, Decode};
-use system::ensure_root;
 use crate::{configuration, initializer::SessionChangeNotification};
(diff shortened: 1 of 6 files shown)


runtime/parachains/src/paras.rs
```diff
@@ -242,6 +245,26 @@ decl_module! {
 	/// The parachains configuration module.
 	pub struct Module<T: Trait> for enum Call where origin: <T as system::Trait>::Origin, system = system {
 		type Error = Error<T>;
+
+		/// Schedule a para to be initialized at the start of the next session.
+		#[weight = (1_000, DispatchClass::Operational)]
+		pub fn sudo_schedule_para_initialize(
```
rphmeier, line 251: I would prefer this be in a separate module, because this is a low-level module that we shouldn't need to expose directly

rphmeier: It should be in the `common` crate or a separate crate, though, not the `parachains` crate. my rationale is that the parachains-core layer should be agnostic as to _how_ and _when_ paras get registered, only that they do.

montekki: moved to `common`

Issue #1269, opened 2020-06-15
STVF determinism


pepyakin:
As the time of writing, polkadot validation function execution is performed using the same executor that is used for executing Substrate runtimes. Substrate executor doesn't guarantee non-determinism of execution, or more precise it doesn't provide more than base guarantees of wasm execution. If the chain-writer manages to exploit these then it's their fault. 

We would like to eradicate these effects for the execution of PVF.

We should come up with a way on how to avoid that. (See [this discussion](https://matrix.to/#/!JUnvACtbMyxHFGrTrh:matrix.parity.io/$FOSCI5Zoxh_J-bqV0j9u7wut_PUZ_VSjZui67m1X7ac?via=matrix.parity.io&via=web3.foundation) for more details)

burdges:
I noticed threads mentioned in https://github.com/WebAssembly/design/blob/master/Nondeterminism.md so I guess any threads we use should come from the host, so several runtimes' entry points?

burdges:
https://github.com/paritytech/substrate/issues/1459#issuecomment-566045237

rphmeier:
Revisiting this - I'm no Wasm expert. What can we do to address this?

burdges:
I'm not overly worried about resource limits because if we've some block dispute caused by resource limits then we've a straightforward bugs in resource allocation.  We should declare the required resources in the block header so that allocations happen when we start running the block.  If a reasonable allocation fails, then validators should trigger a system reboot.  We could clean up resource usage first and/or restart the host process, but really we should just fix our own leaks, so that we can assume leaks were caused by other processes, and kill them all with a reboot. 

Aside from resources, I think the first questions is:  Are we merely avoiding floating point operations?  Or did we disable them outright in the builds of wasmtime and wasmi used by substrate?  I'd presume the latter.

After that, our next question is:  Are we really more confident in wasmi being deterministic than in wasmtime being deterministic?  What differences exist?  We want wasmtime to agree with wasmi as much as possible, but if wasmi is more deterministic in some fundemental way then..

We discussed separate separate wasmi and wasmtime validity claims for backing validators.  We cannot permit approval checks to start if backers could still retract their validity claim, so we've two plausible sequences for backing validator checks, either 
1. backing checkers run both wasmtime and wasmi in either order before claiming validity, or else
2. we pipeline wasmi beside availability, like
    - backers run wasmitime validity first and eventually announce their wasmtime validity claim, which gets posted on-chain and triggers availability distribution,
[...]

PR #1518, opened 2020-08-01
Initial guide text for approvals and especially approvals assignments 

burdges:


roadmap/implementers-guide/src/node/validity/approvals.md
```diff
@@ -0,0 +1,37 @@
...
+
+...
+
+### Gossip
+
+Any validator could send their assignment notices and/or approval votes too early.  We gossip the approval votes because they represent a major commitment by the validator.  We delay gossiping the assignment notices until they agree with our local clock.
+
+### Future work
+
+We could consider additional gossip messages with which nodes claims "slow availability" and/or "slow candidate" to fine tune the assignments "no show" system, but long enough "no show" delays suffice probably.
```
burdges, line 33: https://github.com/paritytech/polkadot/issues/1269#issuecomment-667650052

roadmap/implementers-guide/src/node/validity/approvals.md
```diff
@@ -0,0 +1,49 @@
+# Approval subsystem
```
rphmeier, line 1: Can you add this file to the `SUMMARY.md` which constructs the book? Is this the approvals subsystem or the approval subsystem? The file name should reflect that.

roadmap/implementers-guide/src/node/validity/approvals.md
```diff
@@ -0,0 +1,49 @@
...
+
+- In backing, adversaries could select when they propose invalid candidates based upon when they control the parachain's backing validators who perform the checks.
+
+- In approvals, we randomly assigns individual validators to check specific candidates without giving adversaries foreknowledge about either which honest validators get assigned to which candidates, or even how many check.  Availability prevents adversaries from choosing which validators obtain their possibly invalid candidate.
+
+As such, approval checks provide significantly more security than backing checks, so polkadot achieves some fixed security level most efficiently when we perform more approval checks per backing check or per relay chain block.  
+
+...
+
+Approval requires two gossiped message types, assignment notices created by its assignments subsystem, and approval votes sent by our approval checks subsystem when authorized by the candidate validity utility subsystem.
```
rphmeier, line 15: Oh, these are actually gossiped? Can we bound them somehow? I guess only those where the VRF is low enough & view of ourself and the peer overlaps.

In other cases, we have separate subsystems for networking and core logic. But we can handle the split of this file into individual components after this is merged.

roadmap/implementers-guide/src/node/validity/approvals.md
```diff
@@ -0,0 +1,49 @@
...
+
+As such, approval checks provide significantly more security than backing checks, so polkadot achieves some fixed security level most efficiently when we perform more approval checks per backing check or per relay chain block.  
+
+...
+
+Approval requires two gossiped message types, assignment notices created by its assignments subsystem, and approval votes sent by our approval checks subsystem when authorized by the candidate validity utility subsystem.  
+
+Approval has roughly two parts:
+
+- **Assignments** ensures that each candidates receives enough random checkers, while reducing adversaries odds for obtaining enough checkers, and limiting adversaries foreknowledge.  It tracks approval votes to identify "no show" approval check takes suspiciously long, perhaps indicating the node being under attack, and assigns more checks in this case.  It tracks relay chain equivocations to determine when adversaries possibly gained foreknowledge about assignments, and adds additional checks in this case.
```
rphmeier, line 19:
```suggestion
- **Assignments** ensures that each candidate receives enough random checkers, while reducing adversaries' odds for obtaining enough checkers, and limiting adversaries' foreknowledge.  It tracks approval votes to identify when "no show" approval check takes suspiciously long, perhaps indicating the node being under attack, and assigns more checks in this case.  It tracks relay chain equivocations to determine when adversaries possibly gained foreknowledge about assignments, and adds additional checks in this case.
```

candidates->candidate
adversaries->adversaries'
identify->identify when

and maybe we could leave out the part about equivocations for now, because I don't think we'll implement that in the first go

roadmap/implementers-guide/src/node/validity/approvals.md
```diff
@@ -0,0 +1,49 @@
...
+
+- **Approval checks** listens to the assignments subsystem for outgoing assignment notices that we shall check specific candidates.  It then performs these checks by first invoking the reconstruction subsystem to obtain the candidate, second invoking the candidate validity utility subsystem upon the candidate, and finally sending out an approval vote, or perhaps initiating a dispute.
```
rphmeier, line 21: This references an assignments subsystem which hasn't been defined.

What are "outgoing assignment notices"? Are these notifications from some other piece of code that we need to be checking some particular candidate?

Great that this references reconstruction & candidate validity : ) - that's exactly how this will be implemented.

roadmap/implementers-guide/src/node/validity/approvals.md
```diff
@@ -0,0 +1,49 @@
...
+
+There are rewards computations on-chain too, which runs the assignments code to determine approval, but does so after the fact.
+
+...
+
+### Approval keys
+
+We need two separate keys for the approval subsystem
```
rphmeier, line 29: Oh, so this implies _yet another_ session key?

not sure if we have yet migrated a Substrate chain to add more session keys, but it should be doable. However, there is a bootstrapping concern here. When migrating the chain nobody will have yet registered the extra key, but we can't just throw out the validator set.

So I think the process would be that we'd have to add the session key, make an announcement that everyone should rotate session keys, and then enable parachains.

Or is there some way that both of these keys can be the same? It would make the practicalities of upgrading the relay-chain much simpler

roadmap/implementers-guide/src/node/validity/approvals.md
```diff
@@ -0,0 +1,49 @@
...
+
+- **Approval assignment keys** are sr25519/schnorrkel keys used only for the assignment criteria VRF.  We implicitly sign assignment notices with approval assignment keys by including their relay chain context and additional data in the VRF's extra message, but exclude these from its VRF input.  
+
+- **Approval vote keys** would only sign off on candidate parblock validity.
```
rphmeier, line 33: this section doesn't say what key type should be used. Although the following text gives some background information on how such a choice might be made, a concrete choice would be best at this level. Do you suggest ed25519?

roadmap/implementers-guide/src/node/validity/approvals.md
```diff
@@ -0,0 +1,49 @@
...
+
+There are rewards computations on-chain too, which runs the assignments code to determine approval, but does so after the fact.
+
+...
+
+### Approval keys
+
+We need two separate keys for the approval subsystem
+
+- **Approval assignment keys** are sr25519/schnorrkel keys used only for the assignment criteria VRF.  We implicitly sign assignment notices with approval assignment keys by including their relay chain context and additional data in the VRF's extra message, but exclude these from its VRF input.
```
rphmeier, line 31: This could just be the `Parachains` key that we already have, right?

rphmeier: Also, can you explain what the relay chain context and additional data refer to? Or maybe rather, who the "they" is in that statement?

roadmap/implementers-guide/src/node/validity/approvals.md
```diff
@@ -0,0 +1,49 @@
...
+
+In future, we shall determine which among the several hardening techniques best benefits the netwrok as a whole.  We could provide a multi-process multi-machine architecture for validators, perhaps even reminiscent of GNUNet, or perhaps more resembling smart HSM tooling.  We might instead design a system that more resembled full systems, like like Cosmos' sentry nodes.  In either case, approval assignments might be handled by a slightly hardened machine, but not necessarily nearly as hardened as approval votes, but approval votes machines must similarly run foreign WASM code, which increases their risk, so assignments being separate sounds helpful.  
+
+### Gossip
+
+Any validator could send their assignment notices and/or approval votes too early.  We gossip the approval votes because they represent a major commitment by the validator.  We delay gossiping the assignment notices until they agree with our local clock.
```
rphmeier, line 41: OK, this is an interesting point. Thanks for including it. Is there any element of DoS we should worry about?

I also expect this to be "agree with our local clock, with a lenience of N seconds" in practice.

roadmap/implementers-guide/src/node/validity/approvals.md
rphmeier: I guess the point of confusion here is that the "approvals" page here refers to assignments, but there's a separate page on assignments that should have the information on assignments.

roadmap/implementers-guide/src/node/validity/approvals.md
```diff
@@ -0,0 +1,49 @@
...
+
+### Future work
+
+We could consider additional gossip messages with which nodes claims "slow availability" and/or "slow candidate" to fine tune the assignments "no show" system, but long enough "no show" delays suffice probably.
+
+We shall develop more practical experience with UDP once the availability system works using direct UDP connections.  In this, we should discover if reconstruction performs adequately with a complete graphs or  
+benefits from topology restrictions.  At this point, an assignment notices could implicitly request pieces from a random 1/3rd, perhaps topology restricted, which saves one gossip round.  If this preliminary fast reconstruction fails, then nodes' request alternative pieces directly.  There is an interesting design space in how this overlaps with "slow availability" claims.
```
rphmeier, line 48: oh, that's cool. cc @infinity0

roadmap/implementers-guide/src/node/validity/assignments.md
```diff
@@ -0,0 +1,101 @@
+# Approval assignments
+
+Approval assignment determines on which candidate parachain blocks each validator performs approval checks.  An approval session consideres only one relay chain block and assigns only those candidates that relay chain block declares available.
+
+Assignment balances several concerns:
+
+- limits adversaries foreknowledge about assignments,
```
rphmeier, line 7:
```suggestion
- limits adversaries' foreknowledge about assignments,
```

roadmap/implementers-guide/src/node/validity/assignments.md
```diff
@@ -0,0 +1,101 @@
+# Approval assignments
+
+Approval assignment determines on which candidate parachain blocks each validator performs approval checks.  An approval session consideres only one relay chain block and assigns only those candidates that relay chain block declares available.
```
rphmeier, line 3:
```suggestion
Approval assignment determines on which candidate parachain blocks each validator performs approval checks.  An approval session considers only one relay chain block and assigns only those candidates that relay chain block declares available.
```

consideres->considers

roadmap/implementers-guide/src/node/validity/assignments.md
```diff
@@ -0,0 +1,101 @@
...
+
+All criteria require the validator evaluate a verifiable random function (VRF) using their VRF secret key.  All criteria input specific data called "stories" about the session's relay chain block, and output candidates to check and a precedence called a `DelayTranche`.  
+
+We liberate availability cores when their candidate becomes available of course, but one approval assignment criteria continues associating each candidate with the core number it occupied when it became available. 
+
+Assignment operates in loosely timed rounds determined by this `DelayTranche`s, which proceed roughly 12 times faster than six second block production assuming half second gossip times.  If a candidate `C` needs more approval checkers by the time we reach round `t` then any validators with an assignment to `C` in delay tranche `t` gossip their send assignment notice for `C`.  We continue until all candidates have enough approval checkers assigned.  We take entire tranches together if we do not yet have enough, so we expect strictly more than enough checkers.  We also take later tranches if some checkers return their approval votes too slow (see no shows below).
```
rphmeier, line 17: I wonder if this leans too heavily on the half-second gossip requirement.

How many `DelayTranches` do we run per relay-chain block?

What should validators do who are offline for some amount of time and are re-syncing the chain?

roadmap/implementers-guide/src/node/validity/assignments.md
```diff
@@ -0,0 +1,101 @@
...
+
+Assignment ensures validators check those relay chain blocks for which they have delay tranche zero aka the highest high precedence, so that adversaries always face honest checkers equal to the expected number of assignments with delay tranche zero.  
+
+Among these criteria, the BABE VRF output provides the story for two, which reduces how frequently adversaries could position their own checkers.  We have one criteria whose story consists of the candidate's block hash plus external knowledge that a relay chain equivocation exists with a conflicting candidate.  It provides unforeseeable assignments when adversaries gain foreknowledge about the other two by committing an equivocation in relay chain block production.
```
rphmeier, line 21:
```suggestion
Among these criteria, the BABE VRF output provides the story for two, which reduces how frequently adversaries could position their own checkers.  We have one criterion whose story consists of the candidate's block hash plus external knowledge that a relay chain equivocation exists with a conflicting candidate.  It provides unforeseeable assignments when adversaries gain foreknowledge about the other two by committing an equivocation in relay chain block production.
```

criteria->criterion (singular)

roadmap/implementers-guide/src/node/validity/assignments.md
```diff
@@ -0,0 +1,101 @@
...
+
+We liberate availability cores when their candidate becomes available of course, but one approval assignment criteria continues associating each candidate with the core number it occupied when it became available. 
+
+Assignment operates in loosely timed rounds determined by this `DelayTranche`s, which proceed roughly 12 times faster than six second block production assuming half second gossip times.  If a candidate `C` needs more approval checkers by the time we reach round `t` then any validators with an assignment to `C` in delay tranche `t` gossip their send assignment notice for `C`.  We continue until all candidates have enough approval checkers assigned.  We take entire tranches together if we do not yet have enough, so we expect strictly more than enough checkers.  We also take later tranches if some checkers return their approval votes too slow (see no shows below).  
+
+Assignment ensures validators check those relay chain blocks for which they have delay tranche zero aka the highest high precedence, so that adversaries always face honest checkers equal to the expected number of assignments with delay tranche zero.
```
rphmeier, line 19: What does highest high precedence mean?

roadmap/implementers-guide/src/node/validity/assignments.md
rphmeier, line 17: Another point on delay-tranches. It seems that there is no consensus on which delay-tranches should be used.

For reconstruction and gossip, this seems important. If I receive a reconstruction request, I want it to be legitimized by an assignment proof.

And as I gossip assignments, I will only want to gossip assignments from tranches that I believe should be active. However, how are my peers supposed to know what I accept and what I drop?

The common thread here is to make sure that there is no way for a single validator to create an unbounded amount of assignment proofs that other nodes are forced to circulate or respond to for some reason.

roadmap/implementers-guide/src/node/validity/assignments.md
```diff
@@ -0,0 +1,101 @@
...
+
+Among these criteria, the BABE VRF output provides the story for two, which reduces how frequently adversaries could position their own checkers.  We have one criteria whose story consists of the candidate's block hash plus external knowledge that a relay chain equivocation exists with a conflicting candidate.  It provides unforeseeable assignments when adversaries gain foreknowledge about the other two by committing an equivocation in relay chain block production.
+
+## Announcements / Notices
+
+We gossip assignment notices among nodes so that all validators know which validators should check each candidate, and if any candidate requires more checkers.  
+
+Assignment notices consist of a relay chain context given by a block hash, an assignment criteria, consisting of the criteria identifier and optionally a criteria specific field, an assignee identifier, and a VRF signature by the assignee, which itself consists of a VRF pre-output and a DLEQ proof.  Its VRF input consists of the criteria, usually including a criteria specific field, and a "story" about its relay chain context block.
```
rphmeier, line 27: What is the context? Is it the block where the candidate was backed, the block where the candidate was included, or some other block?

roadmap/implementers-guide/src/node/validity/assignments.md
```diff
@@ -0,0 +1,101 @@
...
+
+We never include stories inside the gossip messages containing assignment notices, but require each validator reconstruct them.  We never care about assignments in the disputes process, so this does not complicate remote disputes.
+
+In a Schnorr VRF, there is an extra signed message distinct from this input, which we set to the relay chain block hash.  As a result, assignment notices are self signing and can be "politely" gossiped without additional signatures, meaning between nodes who can compute the story from the relay chain context.  In other words, if we cannot compute the story required by an assignment notice's VRF part then our self signing property fails and we cannot verify its origin.  We could fix this with either another signature layer (64 bytes) or by including the VRF input point computed from the story (32 bytes), but doing so appears unhelpful.
```
rphmeier, line 31: This seems related to my question above about bounding notices

roadmap/implementers-guide/src/node/validity/assignments.md
```diff
@@ -0,0 +1,101 @@
...
+
+Any validator could send their assignment notices and/or approval votes too early.  We gossip the approval votes early because they represent a major commitment by the validator.  We delay gossiping the assignment notices until they agree with our local clock however.  We also impose a politness condition that the recipient knows the relay chain context used by the assignment notice.
```
rphmeier, line 33:
```suggestion
Any validator could send their assignment notices and/or approval votes too early.  We gossip the approval votes early because they represent a major commitment by the validator.  We delay gossiping the assignment notices until they agree with our local clock however.  We also impose a politeness condition that the recipient knows the relay chain context used by the assignment notice.
```

roadmap/implementers-guide/src/node/validity/assignments.md
```diff
@@ -0,0 +1,101 @@
...
+
+## Stories
+
+We based assignment criteria upon two possible "stories" about the relay chain block `R` that declared the candidate available.  All stories have an output that attempts to minimize adversarial influence, which then acts as the VRF input for an assignment criteria.
+
+We first have a `RelayVRFStory` that outputs the randomness from another VRF output known only to the relay chain block producer who created `R`, and unknown even to them two epochs previously.  
+
+In BABE, we create this value calling `schnorrkel::vrf::VRFInOut::make_bytes` with a context "A&V RC-VRF", with the `VRFInOut` coming from either the VRF that authorized block production for primary blocks, or else form the secondary block VRF for the secondary block type.
```
rphmeier, line 41: This is a piece of data that is then submitted to the chain? The writing here is unclear; it might also be interpreted as this story (and thus assignments based on it) is only instantiable by the block author of the context of the story, who must persist this private VRF output for some time.

rphmeier, line 41:
```suggestion
In BABE, we create this value calling `schnorrkel::vrf::VRFInOut::make_bytes` with a context "A&V RC-VRF", with the `VRFInOut` coming from either the VRF that authorized block production for primary blocks, or else from the secondary block VRF for the secondary block type.
```

roadmap/implementers-guide/src/node/validity/assignments.md
```diff
@@ -0,0 +1,101 @@
...
+
+Any validator could send their assignment notices and/or approval votes too early.  We gossip the approval votes early because they represent a major commitment by the validator.  We delay gossiping the assignment notices until they agree with our local clock however.  We also impose a politness condition that the recipient knows the relay chain context used by the assignment notice.
+
+## Stories
+
+We based assignment criteria upon two possible "stories" about the relay chain block `R` that declared the candidate available.  All stories have an output that attempts to minimize adversarial influence, which then acts as the VRF input for an assignment criteria.
+
+We first have a `RelayVRFStory` that outputs the randomness from another VRF output known only to the relay chain block producer who created `R`, and unknown even to them two epochs previously.
```
rphmeier, line 39: to be clear, the VRF output is known only to the relay-chain block producer, but the randomness is public? rephrasing would help clarify

roadmap/implementers-guide/src/node/validity/assignments.md
```diff
@@ -0,0 +1,101 @@
...
+
+Assignment criteria come in three flavors, `RelayVRFModulo`, `RelayVRFDelay` and `RelayEquivocation`.  Among these, both `RelayVRFModulo` and `RelayVRFDelay` run a VRF whose input is the output of a `RelayVRFStory`, while `RelayEquivocation` runs a VRF whose input is the output of a `RelayEquivocationStory`.
+
+Among these, we have two distinct VRF output computations:
+
+`RelayVRFModulo` runs several distinct samples whose VRF input is the `RelayVRFStory` and the sample number.  It computes the VRF output with `schnorrkel::vrf::VRFInOut::make_bytes` using the context "core", reduces this number modulo the number of availability cores, and outputs the candidate leaving that availability core.  We drop any samples that return no candidate because no candiate was leaving the sampled availability core in this relay chain block.  We choose three samples initially, but we could make polkadot more secure and efficient by increasing this to four or five, and reducing the backing checks accordingly.  All successful `RelayVRFModulo` samples are assigned delay tranche zero.
```
rphmeier, line 59:
```suggestion
`RelayVRFModulo` runs several distinct samples whose VRF input is the `RelayVRFStory` and the sample number.  It computes the VRF output with `schnorrkel::vrf::VRFInOut::make_bytes` using the context "core", reduces this number modulo the number of availability cores, and outputs the candidate leaving that availability core.  We drop any samples that return no candidate because no candidate was leaving the sampled availability core in this relay chain block.  We choose three samples initially, but we could make polkadot more secure and efficient by increasing this to four or five, and reducing the backing checks accordingly.  All successful `RelayVRFModulo` samples are assigned delay tranche zero.
```

candiate->candidate

rphmeier, line 59: can you clarify what is meant by "leaving" some availability core?

These are blocks that are now _included_ because they have become available? So those returned by the `CandidateEvents` API?

We could have `CandidateEvents` give the core index that a candidate used to occupy, but we don't expose this information currently.

roadmap/implementers-guide/src/node/validity/assignments.md
```diff
@@ -0,0 +1,101 @@
...
+
+There is no sampling process for `RelayVRFDelay` and `RelayEquivocation`.  We instead run them on specific candidates and they compute a delay from their VRF output.  `RelayVRFDelay` inputs a `RelayVRFStory` runs for all candidates entering availability, equivalently all candidates leaving some availability core.   `RelayEquivocation` run only candidate lock equivocations, and inputs their block hashes via the `RelayEquivocation` story.
```
rphmeier, line 61: > `RelayVRFDelay` inputs a `RelayVRFStory` runs for all candidates entering availability

Kind of hard to parse this line. This is run for all candidates entering _and_ leaving availability?

rphmeier, line 61:
```suggestion
There is no sampling process for `RelayVRFDelay` and `RelayEquivocation`.  We instead run them on specific candidates and they compute a delay from their VRF output.  `RelayVRFDelay` inputs a `RelayVRFStory` runs for all candidates entering availability, equivalently all candidates leaving some availability core.   `RelayEquivocation` runs only on candidate block equivocations, and inputs their block hashes via the `RelayEquivocation` story.
```

I fixed some typos in the last sentence but unsure if I preserved meaning

roadmap/implementers-guide/src/node/validity/approvals.md
```diff
@@ -0,0 +1,37 @@
+# Approval subsystem
+
+The approval subsystem determines whether a relay chain block can be considered for finality.  It does so by running validity checks on the candidates declared available in that relay chain block.  
+
+These approval validity checks differ from the backing validity checks performed before starting availability:
+
+- In backing, adversaries could select when they propose invalid candidates based upon when they control the parachain's backing validators who perform the checks.
+
+- In approvals, we randomly assigns individual validators to check specific candidates without giving adversaries foreknowledge about either which honest validators get assigned to which candidates, or even how many check.  Availability prevents adversaries from choosing which validators obtain their possibly invalid candidate.
```
hndnklnc, line 9:
```suggestion
- In approvals, we randomly assign individual validators to check specific candidates without giving adversaries' foreknowledge about either which honest validators get assigned to which candidates, or even how many check.  Availability prevents adversaries from choosing which validators obtain their possibly invalid candidate.
```

roadmap/implementers-guide/src/node/validity/approvals.md
```diff
@@ -0,0 +1,37 @@
...
+
+As such, approval checks provide significantly more security than backing checks, so polkadot achieves some fixed security level most efficiently when we perform more approval checks per backing check or per relay chain block.
```
hndnklnc, line 11:
```suggestion
As such, approval checks provide significantly more security than backing checks, so Polkadot achieves some fixed security level most efficiently when we perform more approval checks per backing check or per relay chain block.  
```

roadmap/implementers-guide/src/node/validity/approvals.md
```diff
@@ -0,0 +1,37 @@
...
+
+- **Approval checks** listens to the assignments subsystem for outgoing assignment notices that we shall check specific candidates.  It then performs these checks by first invoking the reconstruction subsystem to obtain the candidate, second invoking the candidate validity utility subsystem upon the candidate, and finally sending out an approval vote, or perhaps initiating a dispute.
+
+There are rewards computations on-chain too, which runs the assignments code to determine approval, but does so after the fact.
+
+...
+
+### Gossip
+
+Any validator could send their assignment notices and/or approval votes too early.  We gossip the approval votes because they represent a major commitment by the validator.  We delay gossiping the assignment notices until they agree with our local clock.
```
hndnklnc, line 41: What do you mean by agreeing with our clock?

roadmap/implementers-guide/src/node/validity/assignments.md
```diff
@@ -0,0 +1,101 @@
...
+
+We announce, and start checking for, our own assignments when their delay is reached, but only if the tracker says the assignee candidate requires more approval checkers.  We never announce an assignment we believe unnecessary because early announcements gives an adversary information.  All delay tranche zero assignments always get announced, which includes all `RelayVRFModulo` assignments.
+
+We have a "no show" timeout longer than one relay chain slot, so at least six seconds, during which we expect approval checks should succeed in reconstructing the candidate block, in redoing its erasure coding to check the candidate receipt, and in rechecking the candidate block itself.  
+
+We consider a validator a "no show" if they do not approve or dispute within this "no show" timeout from our receiving their assignment notice.  We worry "no shows" represent a validator under denial of service attack, presumably to prevent it from reconstructing the candidate, but perhaps delaying it form gossiping a dispute too.  We always replace "no shows" by adding one entire extra delay tranche worth of validators, so such attacks always result in additional checkers.
```
rphmeier, line 77: What does replacement mean in this context? This happens off-chain, right. What is the significance of any single validator adding an extra tranche of validators? Does that make its way on-chain somehow?

roadmap/implementers-guide/src/node/validity/assignments.md
```diff
@@ -0,0 +1,101 @@
...
+
+We consider a validator a "no show" if they do not approve or dispute within this "no show" timeout from our receiving their assignment notice.  We worry "no shows" represent a validator under denial of service attack, presumably to prevent it from reconstructing the candidate, but perhaps delaying it form gossiping a dispute too.  We always replace "no shows" by adding one entire extra delay tranche worth of validators, so such attacks always result in additional checkers.  
+
+As an example, imagine we need 20 checkers, but tranche zero produces only 14, and tranche one only 4, then we take all 5 from tranche two, and thus require 23 checkers for that candidate.  If one checker Charlie from tranche one-two does not respond within say 6 seconds, then we add all 7 checkers from tranche three.  If one checker Cindy from tranche three does not respond within 6 seconds then we take all 3 checkers from tranche four.  We now have 33 checkers working on the candidate, so this escalated quickly.
```
rphmeier, line 79: what is tranche one-two?

rphmeier: This is written from a confusing perspective. It might be easier to understand given the perspective of a single node. It seems that the actual operation that occurs is that if I as a validator observe that there is a no-show and I appear in the next delay-tranche, then I submit my assignment and begin approval.

roadmap/implementers-guide/src/node/validity/assignments.md
```diff
@@ -0,0 +1,101 @@
...
+
+As an example, imagine we need 20 checkers, but tranche zero produces only 14, and tranche one only 4, then we take all 5 from tranche two, and thus require 23 checkers for that candidate.  If one checker Charlie from tranche one-two does not respond within say 6 seconds, then we add all 7 checkers from tranche three.  If one checker Cindy from tranche three does not respond within 6 seconds then we take all 3 checkers from tranche four.  We now have 33 checkers working on the candidate, so this escalated quickly.  
+
+We escalated so quickly because we worried that Charlie and Cindy might be the only honest checkers assigned to that candidate.  If therefore either Charlie or Cindy finally return an approval, then we can conclude approval, and abandon the checkers from tranche four.
+
+We require the "no show" timeout to be longer than a relay chain slot so that we can witness "no shows" on-chain and reward.  We avoid slashing for "no shows" per se, although being "no show" could enter into some computation that punishes repeated poor performance, presumably replaces ImOnline, and we could reduce their rewards and further rewards those who filled in.
```
rphmeier, line 83: Oh, so no-shows actually are on-chain?

roadmap/implementers-guide/src/runtime/validity.md
```diff
@@ -17,8 +17,37 @@ We account for these requirements by having the validity module handle two kinds
...
+
+We liberate availability cores when their candidate becomes available of course, but one approval assignment criteria continues associating each candidate with the core number it occupied when it became available. 
+
+Assignment proceeds in loosely timed rounds called `DelayTranche`s roughly 12 times faster than block production, in which validators send assignment notices until all candidates have enough checkers assigned.  Assignment tracks when approval votes arrive too and assigns more checkers if some checkers run late.
+
+Approval checks provide more security than backing checks, so polkadot becomes more efficient when validators perform more approval checks per backing check.  If validators run 4 approval checks for every backing check, and run almost one backing check per relay chain block, then validators actually check almsot 6 blocks per relay chain block.
```
rphmeier, line 30:
```suggestion
Approval checks provide more security than backing checks, so polkadot becomes more efficient when validators perform more approval checks per backing check.  If validators run 4 approval checks for every backing check, and run almost one backing check per relay chain block, then validators actually check almost 6 blocks per relay chain block.
```

roadmap/implementers-guide/src/node/validity/assignments.md
```diff
@@ -0,0 +1,101 @@
...
+
+Approval assignment determines on which candidate parachain blocks each validator performs approval checks.  An approval session consideres only one relay chain block and assigns only those candidates that relay chain block declares available.
+
+Assignment balances several concerns:
+
+- limits adversaries foreknowledge about assignments,
+- ensures enough checkers, and 
+- distributes assignments relatively equitably.  
+
+Assignees determine their own assignments to check specific candidates using two or three assignment criteria.  Assignees never reveal their assignments until relevant, and gossip delays assignments sent early, which limits others' foreknowledge.  Assignees learn their assignment only with the relay chain block.
```
hndnklnc, line 11: When do they reveal their assignments?

roadmap/implementers-guide/src/runtime/validity.md
```diff
@@ -17,8 +17,37 @@ We account for these requirements by having the validity module handle two kinds
...
+
+We might track the assignments and approvals together as pairs in a simple rewards system.  There are however two reasons to track assignments and approvals separately on-chain:
+
+- An approval that arrives slowly prompts assigning extra checkers.  Yet, we consider a block valid if the earlier checker completes their work, even if the extra checkers never quite finish, which complicates rewarding these extra checkers.  We could support more nuanced rewards for extra checkers if assignments are placed on-chain earlier.  Assignment delay tranches progress 12ish times faster than the relay chain, but no shows could still be witness by the relay chain because the no show delay takes longer than a relay chain slot. 
+
+- We know off-chain when the approval process completes based upon all gossiped assignment notices, not just the approving ones.  We need not-yet-approved assignment notices to appear on-chain if the chain should know about the validity of recently approved blocks.  Integration with grandpa then requires only that grandpa messages identify some future relay chain block that approves the block for which they vote.
```
rphmeier, line 38: Can you elaborate on GRANDPA here?

roadmap/implementers-guide/src/runtime/validity.md
```diff
@@ -17,8 +17,37 @@ We account for these requirements by having the validity module handle two kinds
...
+
+- We know off-chain when the approval process completes based upon all gossiped assignment notices, not just the approving ones.  We need not-yet-approved assignment notices to appear on-chain if the chain should know about the validity of recently approved blocks.  Integration with grandpa then requires only that grandpa messages identify some future relay chain block that approves the block for which they vote. 
+
+Aside from inherents for assignment notices and approval votes, we provide an "Approved" inherent that triggers the on-chain logic in a relay chain block `R1` to rerun the assignment and approval tracker logic for its ancestor `R0`, which then declares `R0` approved.  We could integrate with GRANDPA by gossiping messates that list the descendent `R1`, but then map this into the approved ancestor `R0` for GRANDPA itself.
```
rphmeier, line 40: "inherent" is probably not the right terminology if you're referring to an automated action taken by the chain automatically

rphmeier: > triggers the on-chain logic in a relay chain block `R1` to rerun the assignment and approval tracker logic for its ancestor `R0`

This might have a lot of hidden complexity due to state pruning

rphmeier:
```suggestion
Aside from inherents for assignment notices and approval votes, we provide an "Approved" inherent that triggers the on-chain logic in a relay chain block `R1` to rerun the assignment and approval tracker logic for its ancestor `R0`, which then declares `R0` approved.  We could integrate with GRANDPA by gossiping messages that list the descendent `R1`, but then map this into the approved ancestor `R0` for GRANDPA itself.
```

roadmap/implementers-guide/src/runtime/validity.md
```diff
@@ -17,8 +17,37 @@ We account for these requirements by having the validity module handle two kinds
...
+
+Assignment notices and/or approval votes should be recorded on-chain only when relevant.  Any sent too early are retained but ignore until relevant by our off-chain assignment system.  Assignments are ignored completely by the dispute system because any dispute immediately escalates into all validators checking, but disputes count existing approval votes of course.
```
rphmeier, line 42: Don't we have to put approval votes on-chain eagerly otherwise we might get lots of false-positive no-shows?

roadmap/implementers-guide/src/node/validity/assignments.md
hndnklnc, line 39: Why is the randomness of `R` known by only its creator? After creating the block, it should be public to check whether he has a right to produce, right? Or do you talk about another randomness generated with the epoch randomness but not included to the block R?

roadmap/implementers-guide/src/node/validity/assignments.md
hndnklnc, line 41: What is secondary block type?

roadmap/implementers-guide/src/node/validity/assignments.md
```diff
@@ -0,0 +1,101 @@
...
+
+We first have a `RelayVRFStory` that outputs the randomness from another VRF output known only to the relay chain block producer who created `R`, and unknown even to them two epochs previously.  
+
+In BABE, we create this value calling `schnorrkel::vrf::VRFInOut::make_bytes` with a context "A&V RC-VRF", with the `VRFInOut` coming from either the VRF that authorized block production for primary blocks, or else form the secondary block VRF for the secondary block type.
+
+In Sassafras, we shall always use the non-anonymized recycling VRF output, never the anonymized ring VRF that authorizes block production.  We do not currently know if Sassafras shall have a separate schnorrkel key, but if it reuses its ring VRF key there is an equivalent `ring_vrf::VRFInOut::make_bytes`.
+
+We like that `RelayVRFStory` admits relatively few choices, but an adversary who equivocates in relay chain block production might assignments that depend upon it too early.
```
hndnklnc, line 45: I could not understand this sentence.

roadmap/implementers-guide/src/node/validity/assignments.md
burdges: I'm afraid this first pass focused primarily upon the assignment module's interface, so I forgot that part presumably, but..

We should clarify that assignment noticed go out only when their delay tranche gets reached.  It's more complex actually since they could be delayed further, either by waiting for no shows, or else by legitimately waiting to reduce their workload.  I've spent today coding that portion, which became obnoxiously subtle.  I'll add some text once I've cleaned up this new code.

roadmap/implementers-guide/src/runtime/validity.md
burdges: I meant this approach simplifies integration with GRANDPA, but yes we could clarify the conditional aspect of that sentence.

roadmap/implementers-guide/src/node/validity/approvals.md
burdges: Assignment notices should always arrive several seconds before their approval vote. 

Assignment notices are bounded by them being VRF signatures:  Any validator could gossip out maybe num_relay_vrf_modulo + num_cores assignment notices but those num_cores would not be gossiped onwards by other validators until their delay_tranche output by the VRF gets reached, which lies beyond the senders' control.  You could make this more polite if you liked, but not sure that's necessarily. 

Approval votes are always gossiped so yes every validator could send out num_cores approval votes, but doing so sounds mildly risky.

roadmap/implementers-guide/src/node/validity/approvals.md
burdges: We could make the approval vote key be the grandpa key.  We could always separate them if some distant future super-sentry node iteration would support consensus running on a separate machine from candidate worker nodes or whatever. 

We need either the assignments key to be some new key immune to large slashes, or else some validators would ask that assignments by run in a remote signer, which sounds absolutely nightmarish.

roadmap/implementers-guide/src/node/validity/approvals.md
burdges: It's fine either way really, but ed25519 enables both the grandpa key doing double duty, and HSM support.

roadmap/implementers-guide/src/node/validity/approvals.md
burdges: Assignments key should not be associated with any slashing conditions.

roadmap/implementers-guide/src/node/validity/approvals.md
burdges: Yes, some leniency here

roadmap/implementers-guide/src/node/validity/assignments.md
burdges: We want delay tranches to run at about our gossip speed, while no show timeouts should be long enough to reconstruct and check, and presumably longer than one relay chain block. 

If they're resyncing then they should not send out assignment notices or do approval checks, which should enable them to resync the chain much faster than if they're validating those 6+ candidates per relay chain block .

roadmap/implementers-guide/src/node/validity/assignments.md
burdges: It's the relay chain block where it was declared available.  I suppose that's included then?

roadmap/implementers-guide/src/node/validity/assignments.md
burdges: We need the "story" to validate this signature, but there are two modes of operation:  As a consensus process, we keep this information live in memory.  As an on-chain check, we need to store the assignment notices in some reasonable way.

roadmap/implementers-guide/src/node/validity/assignments.md
burdges: As a consensus process, we file incoming notices inside a data structure that contains the story.  As an on-chain check, yes I suppose each block creates a similar data structure inside the state that contains the story, along with any notices received. 

I've avoided thinking too much about the on-chain data structure because I'm worried first about the security aspects of the consensus process, but..  I've been considering some refactoring which as a by product permits doing the on-chain work more efficiently, so doing that tonight I guess..

roadmap/implementers-guide/src/node/validity/assignments.md
burdges: I should maybe be clearer here:  As a security property, an honest upcoming relay chain block producer is the only one who knows this value until they actually produce their block.  We need everyone to know all stories about an existing block of course.

roadmap/implementers-guide/src/node/validity/assignments.md
burdges: Yes "leaving" a core means included from that core.  I avoided the term included because it's confusing with the who stage system, but we can start migrating terminology here.

roadmap/implementers-guide/src/node/validity/assignments.md
burdges: No this is just gibberish right now.  oops

roadmap/implementers-guide/src/node/validity/approvals.md
burdges: Yes same thing.   It appears these comments gets folded in when mixed with a merged suggestion.

roadmap/implementers-guide/src/node/validity/approvals.md
burdges: It needs some leniency if that's what you mean?

roadmap/implementers-guide/src/node/validity/assignments.md
burdges: I'd move any more details discussion about the asymptotics to https://github.com/w3f/research-internal/issues/515 

We need the same logic on-chain as off-chain to know if a candidate is approved on-chain, but.. 

I'd expect no shows would usually resolve themselves fast, so the on-chain logic could abandon the replacements too.  We'll discuss incentives for replacements later, but obviously they impact our abandoning replacements.

roadmap/implementers-guide/src/node/validity/assignments.md
burdges: Yes. :)

roadmap/implementers-guide/src/node/validity/assignments.md
burdges: Yes somewhat, for two reasons: incentives, and correctly computing the replacements when the no show stays a no show.

roadmap/implementers-guide/src/runtime/validity.md
burdges: Yes I suppose so..

burdges:
> The interaction between disputes and assignment/approval is unclear to me. Conflicting approval votes should clearly start a dispute, but shouldn't a dispute also select more checkers?

Any dispute among validators requires that _all_ validators check the candidate because we never reach consensus to slash based upon only a subset of validators.  Any dispute thus turns off assignments for that candidate, but assignments worries only about the current chain, while disputes handles remote disputes too.

> One detail that was removed from the paper is what should happen when a dispute is initiated

We've always wanted everyone to checks any dispute.  I figured that came under the disputes part, but maybe worth mentioning here too.

roadmap/implementers-guide/src/node/validity/assignments.md
burdges: Aura-babe-ous blocks.  I hope I got this language right form the code.

roadmap/implementers-guide/src/node/validity/assignments.md
burdges: I'm discussing the foreknowledge, but yes this could be more clear.

roadmap/implementers-guide/src/node/validity/approvals.md
burdges: db2a8eec

I'll place them under validity for now, but I've no idea where these go really, nor even if disputes and approvals go into the same crate, although I suspect no..

roadmap/implementers-guide/src/node/validity/approvals.md
burdges: Assignments is kinda a subthing, but maybe partially useful on-chain, but..

roadmap/implementers-guide/src/node/validity/approvals.md
burdges: 94f65ab1

roadmap/implementers-guide/src/node/validity/approvals.md
burdges: 94f65ab1

roadmap/implementers-guide/src/node/validity/approvals.md
burdges: 1d4b69b0

roadmap/implementers-guide/src/node/validity/approvals.md
burdges: 1d4b69b0

roadmap/implementers-guide/src/node/validity/assignments.md
burdges: just highest
dd1d4592

burdges: so checks first

roadmap/implementers-guide/src/node/validity/assignments.md
burdges: > Another point on delay-tranches. It seems that there is no consensus on which delay-tranches should be used.

It's one tranche every k seconds after the relay chain block's slot.  I've two numbers in the code:  delay tranches start from zero with the relay chain block's slot, while AnV slots are `12 * relay_chain_slot + delay_tranche` give an absolute close.  I'll let someone else figure out which should be more or less exposed in the interface, etc.

> For reconstruction and gossip, this seems important. If I receive a reconstruction request, I want it to be legitimized by an assignment proof.

Yes and no, we could let validators reconstruct anything, but prioritize approval assignments. 

> And as I gossip assignments, I will only want to gossip assignments from tranches that I believe should be active. However, how are my peers supposed to know what I accept and what I drop?

You need not drop anything:
- Approval votes are a huge deal, so gossip them always.  
- Assignment notices are inherently somewhat limited in number due to being VRFs, so merely save them, and regossip them only when you believe they become viable.

We still need politeness for relay chain block knowledge of course.

roadmap/implementers-guide/src/node/validity/assignments.md
burdges: dd1d459276fc

roadmap/implementers-guide/src/node/validity/assignments.md
burdges: https://github.com/w3f/research-internal/issues/519

burdges: 3bec0174

roadmap/implementers-guide/src/node/validity/approvals.md
rphmeier: I see, that makes a lot of sense. Expounding on the "risk" these keys have as the reason for the separation would make sense in this section. I'm fine with reusing the GRANDPA key for that, although we will be upgrading that to BLS at some point. Maybe makes most sense to just upgrade and add the extra key type, put out an announcement that every validator should rotate their session keys, and then enable parachains a few weeks later

roadmap/implementers-guide/src/node/validity/assignments.md
rphmeier: Yep, that's what we call "included". We have a `CandidateEvents` runtime API we can use to gather that information.

roadmap/implementers-guide/src/node/validity/assignments.md
rphmeier: I see; it makes sense to mirror it in the chain state, at least up to some amount of time.

Can you clarify though, that my interpretation is correct and this story can be used by any validator and not just the author of the block? The language of the post here is confusing as it says "from another VRF output known only to the relay-chain block producer". So the VRF output is private, but the randomness is public - right?

rphmeier: oh, I see you answered it below

rphmeier: Yep

roadmap/implementers-guide/src/node/validity/approvals.md
burdges: We actually cannot wholly replace Schnorr with BLS because BLS verification seems just too slow.  We'll likely add a BLS12-381 G1 point with a Schnorr proof-of-possession as either the only or as a second GRANDPA public key, but we've then two choices:

1.  We could sign GRANDPA messages first with this BLS public key and then sign that signed message with the Ed25519 public key.  We could even replace Ed25519 with Rabin-Williams here, which gets shockingly fast.  We'll need slashing condition for when something messy happens.

2.  We sign GRANDPA messages with a Schnorr VRF using this BLS public key, so verification runs much slower than Ed25519, but still vastly faster than BLS signatures.  At this point the VRF pre-output actually is a BLS signature however, so we can transition smoothly to BLS verification whenever we gain enough signatures for aggregation to help, but doing this avoids any slashing conditions.

roadmap/implementers-guide/src/node/validity/assignments.md
burdges: d3413e3e

roadmap/implementers-guide/src/node/validity/assignments.md
burdges: bf0b6fc5

burdges: Resolved?  Or should I be more careful about using the word context?

roadmap/implementers-guide/src/node/validity/assignments.md
burdges: bf0b6fc5

roadmap/implementers-guide/src/node/validity/assignments.md
burdges: Included in 2b2e4f90

burdges:
Added notes on parameters in 2b2e4f90

roadmap/implementers-guide/src/runtime/validity.md
burdges: ae675438

roadmap/implementers-guide/src/node/validity/assignments.md
burdges: 6bd47582

burdges:
Added draft code PR in https://github.com/paritytech/polkadot/pull/1558  :)

burdges:
We should add a discussion about this scenario: 

A validator with a tranche zero (or other low) assignment never makes their announcement, like because they postponed their work (which is allowed). Yet, they then made this announcement later right around finality. If this announcement gets on-chain (also allowed), then yes it delays finality. If it does not get on-chain, then yes we've one announcement that the off-chain consensus system says is valid, but the chain says was too slow.  In this case, the chain wins.

roadmap/implementers-guide/src/node/validity/approvals.md
burdges: I just say to use ed25519 for approval vote keys now, which along with previous tweaks seemingly finishes this one.  Anything else?

burdges:
@rphmeier We should chat about the equivocation symmetry:  If X and Y are equivocation than differ in parachain rho, so included candidates X[rho] and Y[rho] differ.  We risk some subsystem deciding X does not warrant work because Y looks better, but maybe X is an attack, X[rho] is invalid, and Y exists to distract from X.  We could say all inclusions get checked, meaning no subsystem could decide X does not warrant work.  We might need this for other chain distraction, like maybe X and Y are not even equivocations, but not necessarily.  We could alternatively say the candidate equivocations X[rho] and Y[rho] should always be checked, even if we give up on X for other reasons.

PR #1525, opened 2020-08-03
Implementer's guide: notes on contextual execution

pepyakin:
plus clarification that `validation_data_hash` consists of the global and local validation data.

### DIFF
--- a/roadmap/implementers-guide/src/runtime/inclusion.md
+++ b/roadmap/implementers-guide/src/runtime/inclusion.md
@@ -63,7 +63,7 @@ All failed checks should lead to an unrecoverable error making the block invalid
   1. check that `scheduled` is sorted ascending by `CoreIndex`, without duplicates.
   1. check that there is no candidate pending availability for any scheduled `ParaId`.
   1. check that each candidate's `validation_data_hash` corresponds to a `(LocalValidationData, GlobalValidationData)` computed from the current state.
-    > NOTE: With contextual execution in place, local and global validation data will be obtained as of the state of the context block. However, only the current state will be used for such a query.
+    > NOTE: With contextual execution in place, local and global validation data will be obtained as of the state of the context block. However, only the state of the current block can be used for such a query.
   1. If the core assignment includes a specific collator, ensure the backed candidate is issued by that collator.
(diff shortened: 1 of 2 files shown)


roadmap/implementers-guide/src/runtime/inclusion.md
```diff
@@ -63,6 +63,7 @@ All failed checks should lead to an unrecoverable error making the block invalid
   1. check that `scheduled` is sorted ascending by `CoreIndex`, without duplicates.
   1. check that there is no candidate pending availability for any scheduled `ParaId`.
   1. check that each candidate's `validation_data_hash` corresponds to a `(LocalValidationData, GlobalValidationData)` computed from the current state.
+    > NOTE: With contextual execution in place, local and global validation data will be obtained as of the state of the context block. However, only the current state will be used for a such query.
```
bkchr, line 66:
```suggestion
    > NOTE: With contextual execution in place, local and global validation data will be obtained as of the state of the context block. However, only the current state will be used for such a query.
```

> However, only the current state will be used for such a query

I also don't understand this. What does this means?

rphmeier: It means the current state needs to be able to provide `LocalValidationData` and `GlobalValidationData` for the last `k` ancestors.

Although, it can be done indirectly with a header-chain proof and a state proof from there.

pepyakin: Well, it refers the same "current state" as the statement above. I.e. the current relay chain state. Stated differently: Let's only consider `LocalValidationData`. Currently, we have a function that looks roughly like this `local_validation_data(para_id)` that implicitly also takes the current relay-chain state. However, to support the contextual execution we will have to a function `local_validation_data_at(para_id, ctx_block_num)` where `ctx_block_num` is some past block. Note that this function also accepts the current relay-chain state implicitly. That basically translates that we need to be able to maintain required information around to be able to reconstruct that `LocalValidationData`.

Does it make sense now? If so, do you have an idea how to improve the wording?

pepyakin: Ok, I

1. changed the wording so that it says that we can only access the latest state (at most)
2. I also added @rphmeier 's suggestion about the merkle proofs.

rphmeier:
bot merge

PR #1530, opened 2020-08-03
Unalias Substrate Imports

shawntabrizi:
This PR updates a bunch of substrate crates to use their original crate name rather than aliases.

For example instead of `balances`, we now use `pallet_balances`, just like in Substrate.


validation/Cargo.toml
```diff
@@ -32,7 +32,7 @@ block-builder = { package = "sc-block-builder", git = "https://github.com/parity
 trie = { package = "sp-trie", git = "https://github.com/paritytech/substrate", branch = "master" }
 runtime_primitives = { package = "sp-runtime", git = "https://github.com/paritytech/substrate", branch = "master" }
 bitvec = { version = "0.17.4", default-features = false, features = ["alloc"] }
-runtime_babe = { package = "pallet-babe", git = "https://github.com/paritytech/substrate", branch = "master" }
+pallet-babe = { git = "https://github.com/paritytech/substrate", branch = "master" }
```
bkchr, line 35:
```suggestion
```

I think this crate is not even used when you don't needed to change any code?

shawntabrizi: Yes there are many crates that are not used in these cargo toml files.

We for example list every pallet, even ones we dont use.

shawntabrizi: How to clean them up?

bkchr: You can activate some lint in rustc, but don't ask me how. :D

kianenigma: `cargo udpes` is supposed to do this but I have tried it on substrate and it was quite useless. Might be more lucky here. If you are down the path of removing udeps, keep an eye out for `sp_io` and `sp_core` as afaik parity devs are quite fond of sprinkling them around crates :D

shawntabrizi:
bot merge

shawntabrizi:
bot merge

shawntabrizi:
bot merge

PR #1531, opened 2020-08-04
Rewrite client handling

bkchr:
We are supporting muliple polkadot-like chains and all have different
client types. This pr reworks the client handling by having all of them
in one enum combined. Besides that, there is added a special trait
`ExecuteWithClient` to use the internal client.


service/src/client.rs
```diff
@@ -14,40 +14,133 @@
...
-				StateBackend = Backend ::State
-			>
+				StateBackend = Backend::State
+			>,
+		Client::Api: RuntimeApiCollection<StateBackend = Backend::State>,
 {}
+
+/// Execute something with the client instance.
+///
+/// As there exists multiple chains inside Polkadot, like Polkadot itself, Kusama, Westend etc,
```
rphmeier, line 98:
```suggestion
/// As there exist multiple chains inside Polkadot, like Polkadot itself, Kusama, Westend etc,
```

service/src/client.rs
```diff
@@ -14,40 +14,133 @@
...
+/// there can exists different kinds of client types. As these client types differ in the generics
```
rphmeier, line 99:
```suggestion
/// there can exist different kinds of client types. As these client types differ in the generics
```

service/src/client.rs
```diff
@@ -14,40 +14,133 @@
...
+}
+
+/// A client instance of Polkadot.
+///
+/// See [`ExecuteWithClient`] for more information.
+#[derive(Clone)]
+pub enum Client {
+	Polkadot(Arc<crate::FullClient<polkadot_runtime::RuntimeApi, crate::PolkadotExecutor>>),
+	Westend(Arc<crate::FullClient<westend_runtime::RuntimeApi, crate::WestendExecutor>>),
+	Kusama(Arc<crate::FullClient<kusama_runtime::RuntimeApi, crate::KusamaExecutor>>),
```
rphmeier, line 128: Will you add a Rococo enum here?

bkchr: Yeah, the pr was not yet merged :)

service/src/lib.rs
```diff
@@ -52,7 +52,7 @@ pub use polkadot_runtime;
 pub use kusama_runtime;
 pub use westend_runtime;
 use prometheus_endpoint::Registry;
-pub use self::client::PolkadotClient;
+pub use self::client::*;
```
rphmeier, line 55: it would be better to explicitly export. glob re-exports scare me :)

PR #1532, opened 2020-08-04
Revert enabling authority discovery by default

andresilva:
Reverts https://github.com/paritytech/polkadot/pull/1395.

The authority discovery module ends up consuming a very high amount of sockets which could lead to exhaustion of file descriptors.

Screenshot_2020-08-04_10-54-01

Here's the difference on my local node with and without authority discovery enabled. On the left we can see peaks of >4k sockets and low is at ~1.5k. Without authority discovery we still see an initial peak (presumably due to regular DHT discovery) but the absolute number of sockets taken is an order of magnitude lower.


bkchr:
Do we still want this?

tomaka:
So what is the criteria for re-enabling authority-discovery again? For all we know it's working as intended, and I don't know what would need to change.

PR #1535, opened 2020-08-04
Network Bridge Refactoring

rphmeier:
This PR makes a few major changes to the structure of our networking subsystems and the network bridge subsystem that interacts with the lower-level substrate-network crate.

The largest change is reducing the amount of abstraction over message types. The picture of how to do protocol upgrades via the network bridge was quite fuzzy before because each subsystem was effectively its own subprotocol that accepted bytes and sent bytes out.

My proposal is that we don't actually need abstraction over the message types of the network at this level, and that the only purpose the network bridge should serve is abstracting over how exactly those messages are sent and received.

So this PR explicitly defines the message types sent and received by the network in v1 and forwards them to subsystems as concrete types. This also reduces the amount of places where we'd have to decode bytes from the network and allows our network subsystems to deal in concrete types.

The other major change this PR makes is the elaboration on multiple peer-sets in the network bridge. This change was first introduced in #1452 and I've expounded on it further here.


roadmap/implementers-guide/src/SUMMARY.md
```diff
@@ -48,6 +48,7 @@
     - [Runtime](types/runtime.md)
     - [Chain](types/chain.md)
     - [Messages](types/messages.md)
+	- [Network](types/network.md)
```
tomaka, line 51:
```suggestion
    - [Network](types/network.md)
```

roadmap/implementers-guide/src/types/network.md
```diff
@@ -0,0 +1,119 @@
...
+	/// An erasure chunk for a given candidate hash.
+	Chunk(Hash, ErasureChunk),
+}
+```
+
+### Bitfield Distribution V1
+
+```rust
+enum BitfieldDistributionV1Message {
+	/// A signed availability bifield for a given relay-parent hash.
```
pepyakin, line 34:
```suggestion
	/// A signed availability bitfield for a given relay-parent hash.
```

roadmap/implementers-guide/src/types/network.md
```diff
@@ -0,0 +1,119 @@
+# Network Types
+
+These types are those that are actually sent over the network to subsystems.
+
+## Universal Types
+
+```rust
+type RequestId = u64;
+type ProtocolVersion = u32;
```
pepyakin, line 9: FWIW, this is not used anywhere currently

rphmeier: That's fine, because we don't have more than one protocol version yet.

rphmeier:
bot merge

roadmap/implementers-guide/src/node/utility/network-bridge.md
```diff
@@ -6,55 +6,118 @@ The most notable challenge is coordinating and eliminating race conditions of pe
...
 
-There are two types of network messages this sends and receives:
+### Overseer Signal: ActiveLeavesUpdate
 
-- ProtocolMessage(ProtocolId, Bytes)
-- ViewUpdate(View)
+The `activated` and `deactivated` lists determine the evolution of our local view over time. A `ProtocolMessage::ViewUpdate` is issued to each connected peer on each peer-set, and a `NetworkBridgeEvent::OurViewChange` is issued to each event handler for each protocol.
 
-Each of these network messages is associated with a particular peer-set. If we are connected to the same peer on both peer-sets, we will receive two `ViewUpdate`s from them every time they change their view.
+If we are connected to the same peer on both peer-sets, we will send the peer two view updates as a result.
```
drahnr, line 62: The base assumption here is, that both protocol versions operate completely independently? So this assumes one subsystem accepting both protocol versions will never exist, since that might cause two incompatible views being received for version `A` and `A+1`

roadmap/implementers-guide/src/types/network.md
```diff
@@ -0,0 +1,119 @@
...
+```
+
+## V1 Network Subsystem Message Types
+
+### Availability Distribution V1
+
+```rust
+enum AvailabilityDistributionV1Message {
+	/// An erasure chunk for a given candidate hash.
+	Chunk(Hash, ErasureChunk),
```
drahnr, line 26: :+1:

roadmap/implementers-guide/src/node/utility/network-bridge.md
tomaka: It's not about protocol versions here, but about whether a node is connected to the local node as a collator or as a fellow validator.
It indeed means that you can't implement any logic specific to the situation where you are connected to a node is both a collator and a validator, but I don't think that's a problem.

rphmeier: Subsystems might be able to receive network bridge notifications based on both the collation and validation peer-sets. But we currently make no guarantee that the views provided by a peer on one peer-set are the same as the views provided by a peer on the other peer-set.

At the moment, we have no subsystem which bridges both peer-sets. We can cross that bridge when we come to it, maybe by adding guarantees about views.

PR #1537, opened 2020-08-04
Network bridge refactoring impl

rphmeier:
Based on guide updates from #1535


node/subsystem/src/messages.rs
```diff
@@ -136,45 +137,57 @@ impl CandidateValidationMessage {
...
-	PeerMessage(PeerId, Vec<u8>),
-
-	/// Peer's `View` has changed.
-	PeerViewChange(PeerId, View),
 
-	/// Our `View` has changed.
-	OurViewChange(View),
+/// Messages received by the Collator Protocol subsystem.
+#[derive(Debug)]
+pub enum CollatorProtocolMessage {
```
coriolinus, line 143: Worth adding a `relay_parent` function for parity with the rest of the subsystem messages, or does it seem unlikely that such a function would actually be useful?

node/network/protocol/src/lib.rs
```diff
@@ -0,0 +1,255 @@
...
+
+	/// Peer's `View` has changed.
+	PeerViewChange(PeerId, View),
+
+	/// Our `View` has changed.
+	OurViewChange(View),
+}
+
+macro_rules! impl_try_from {
+	($m:ident, $variant:ident, $out:ty) => {
```
coriolinus, line 98:
```suggestion
	($m:ty, $variant:ident, $out:ty) => {
```

I'm sure `ident` works here, but `ty` is a bit clearer about how it's used.

coriolinus, line 98: It would be nice to have a more descriptive metavariable name than `m`.

node/network/bridge/src/lib.rs
```diff
@@ -282,37 +324,211 @@ fn construct_view(live_heads: &[Hash]) -> View {
...
+			net.report_peer(peer, UNCONNECTED_PEERSET_COST).await?;
+
+			return Ok(Vec::new());
+		},
+		Some(d) => d,
+	};
+
+	let mut outgoing_messages = Vec::with_capacity(messages.len());
+	for message in messages {
+		outgoing_messages.push(match message {
```
coriolinus, line 390: This loop might be cleaner as

```rust
messages.iter().filter_map(|message| ...).collect()
```

node/network/bridge/src/lib.rs
```diff
@@ -282,37 +324,211 @@ fn construct_view(live_heads: &[Hash]) -> View {
...
+
+	Ok(outgoing_messages)
+}
+
+async fn send_validation_message<I>(
+	net: &mut impl Network,
+	peers: I,
+	message: WireMessage<protocol_v1::ValidationProtocol>,
+) -> SubsystemResult<()>
+	where I: IntoIterator<Item=PeerId>, I::IntoIter: ExactSizeIterator
```
coriolinus, line 424: Personal style preference: it's much easier for me to understand the set of constraints on generic items when each constraint appears on its own line. I get that you're grouping related items, but I'd prefer to just make the code a bit taller. This appears in several places in this PR.

node/network/bridge/src/lib.rs
```diff
@@ -282,37 +324,211 @@ fn construct_view(live_heads: &[Hash]) -> View {
...
+	message: WireMessage<protocol_v1::CollationProtocol>,
+) -> SubsystemResult<()>
+	where I: IntoIterator<Item=PeerId>, I::IntoIter: ExactSizeIterator
+{
+	send_message(net, peers, PeerSet::Collation, message).await
+}
 
-	let notifications = peers.keys().cloned()
-		.map(move |peer| Ok(NetworkAction::WriteNotification(peer, message.clone())));
+async fn send_message<M: Encode + Clone, I>(
```
coriolinus, line 439: Given that there's already a `where` clause for this function, it's easier to read and understand the set of generic constraints if all of them appear within that clause.

node/network/bridge/src/lib.rs
```diff
@@ -282,37 +324,211 @@ fn construct_view(live_heads: &[Hash]) -> View {
...
+{
+	let mut message_producer = stream::iter({
+		let peers = peers.into_iter();
+		let n_peers = peers.len();
+		let mut message = Some(message.encode());
+
+		peers.enumerate().map(move |(i, peer)| {
+			// optimization: avoid cloning the message for the last peer in the
+			// list. The message payload can be quite large. If the underlying
+			// network used `Bytes` this would not be necessary.
```
coriolinus, line 455: I like the mechanism of this optimization, but I don't understand why the `Bytes` type would make this unnecessary. It's not necessary for the purpose of this PR, but it would be nice if you could link me to the type you mean, so I could better understand this statement.

node/network/bridge/src/lib.rs
```diff
@@ -647,6 +854,39 @@ mod tests {
 		executor::block_on(future::select(test_fut, network_bridge));
 	}
 
+	async fn assert_sends_validation_event_to_all(
```
coriolinus, line 857:
```suggestion
	#[test]
	async fn assert_sends_validation_event_to_all(
```

node/network/protocol/src/lib.rs
rphmeier: Actually, I tried `ty` in my first approach, but it gives some path errors with the match statement because you don't typically use a type in those situations

node/network/bridge/src/lib.rs
rphmeier: another thing I tried at first :) 

The problem is the `.await` statements within the loop

node/network/bridge/src/lib.rs
rphmeier: this is not new code from the PR, it's just been moved around

rphmeier: Check out https://docs.rs/bytes/0.5.6/bytes : it's a fancy `Arc<Vec<u8>>`

node/network/bridge/src/lib.rs
rphmeier: this isn't a test, but it's a test helper.

rphmeier:
bot merge

PR #1545, opened 2020-08-05
Update .editorconfig to what we have in practice

pepyakin:
While editing the impl guide markdowns I tried to be inline with what seemingly more
common way to indent them: spaces. However, despite that I changed it kept reseting.
Turned out the culprit is the .editorconfig file.

This commit addresses this issue. I didn't try to deduplicate the rules since
I found that the formal specification is a bit ambigious and it is not a big
deal anyway.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,12 +1,13 @@
 root = true
-[*]
+
+[*.rs]
 indent_style=tab
 indent_size=tab
 tab_width=4
+max_line_length=120
 end_of_line=lf
 charset=utf-8
 trim_trailing_whitespace=true
-max_line_length=120
 insert_final_newline=true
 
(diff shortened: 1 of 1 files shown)


pepyakin:
me: creates a PR for .editorconfig that stayed in its original form for 2 years straight.
.editorconfig: changes in a way to conflict with my PR

> Seems fine to me, but I'm somewhat surprised that you mention in the description that the motivating incident involved markdown files, but you didn't add a *.md section to the file.

Ah yeah that's confusing, sorry. My motivation was to improve the status quo by deregulating how the *.md files are formatted. I didn't risk to add actual rules for the *.md to not spark a bikeshed discussion and removing all discrepancies that will arise from that.

PR #1554, opened 2020-08-07
Align for "Starting parachain attestation session"

jasl:
This is not a big problem.

You can find "✍️ Starting parachain attestation" isn't alignment with other output

image

I do a little experiment, I copy this string to terminal and I found the space between `✍️` and `Starting` disappeared, so give one more space would help for alignment


bkchr:
bot merge

PR #1556, opened 2020-08-07
Implementer's Guide: Flesh out more details for upward messages

### LINKED ISSUE #1534: Guide: Impose limits on the number of messages that a candidate can send
See this [comment](https://github.com/paritytech/polkadot/pull/1503#discussion_r463953374)

pepyakin:
Closes #1534 

This PR changes the way how we process the upward messages.

### Upward Message Kind

Upward messages are split to different kinds, with `Dispatchable` being only one of them. The upside is that the unlike plain opaque dispatchables, explicit message kinds allows us to impose some validity criteria upon the upward messages. E.g. we can reject candidates that we know will fail upfront.

However, we only introduce the scaffolding for different upward message kinds without actually introducing any new.

### Weight

Then, in the current construction the weight consumed by messages is not limited. In the proposed changes, each upward message of dispatchable kind consumes some predefined weight and the total weight consumed during dispatching messages is limited, leaving a guaranteed amount of weight to other extrinsics. 

This proposal doesn't yet address charging fees for the weight.

### Round-robin dispatching

Since it can happen that some messages won't fit in the weight budget devoted for a block, we need to fairly distribute the weight among the pending dispatchables.

To address this, we go over all queues in a round robin fashion dequeuing a message each time. When we reach the weight budget for executing dispatchables or there are none left we finish processing. There can be multiple iterations. If any unused weight left it can be leveraged by signed extrinsics.

### Dispatch order

The current construction dispatches messages during finalization. The proposed construction moves this processing to be just right at the end of inclusion inherent giving the dispatchables an advantage over the regular signed extrinsics. That alleviates the front-running the parachain upward messages by the validators to some degree.

Optionally, perhaps we can add another round of dispatching during the finalization stage to leverage unused weight, but that would worsen the property against front-running

### DIFF
--- a/roadmap/implementers-guide/src/messaging.md
+++ b/roadmap/implementers-guide/src/messaging.md
@@ -36,7 +36,7 @@ The weight that processing of the dispatchables can consume is limited by a prec
 that some dispatchables will be left for later blocks. To make the dispatching more fair, the queues are processed turn-by-turn
 in a round robin fashion.
 
-Other kinds of upward messages can be introduced in the future as well. Most likely the HRMP channel managem
+Other kinds of upward messages can be introduced in the future as well. Most likely the HRMP channel management.
 
 ## Horizontal Message Passing
 
--- a/roadmap/implementers-guide/src/messaging.md
+++ b/roadmap/implementers-guide/src/messaging.md
@@ -17,9 +17,7 @@ digraph {
 }
 ```
 
-Downward Message Passing (DMP) is a mechanism for delivering messages to parachains from the relay chain. Downward
-messages may originate not from the relay chain, but rather from another parachain via a mechanism
-called HRMP (Will be described later).
+Downward Message Passing (DMP) is a mechanism for delivering messages to parachains from the relay chain.
 
 Each parachain has its own queue that stores all pending inbound downward messages. A parachain
 doesn't have to process all messages at once, however, there are rules as to how the downward message queue
@@ -28,16 +26,17 @@ The downward message queue doesn't have a cap on its size and it is up to the re
 that prevent spamming in place.
 
 Upward Message Passing (UMP) is a mechanism responsible for delivering messages in the opposite direction:
-from a parachain up to the relay chain. Upward messages are dispatched to Runtime entrypoints and
-typically used for invoking some actions on the relay chain on behalf of the parachain.
+from a parachain up to the relay chain. Upward messagess can serve different purposes and can be of different
+ kinds.
 
-> NOTE: It is conceivable that upward messages will be divided to more fine-grained kinds with a dispatchable upward message
-being only one kind of multiple. That would make upward messages inspectable and therefore would allow imposing additional
-validity criteria for the candidates that contain these messages.
+One kind of messages is `Dispatchable`. They could be thought of similar to extrinsics sent to a relay chain: they also
+invoke exposed runtime entrypoints, they consume weight and require fees. The difference is that they originate from
+a parachain. Each parachain has a queue of dispatchables to be executed. There can be only so many dispatchables at a time.
+The weight that processing of the dispatchables can consume is limited by a preconfigured value. Therefore, it is possible
+that some dispatchables will be left for later blocks. To make the dispatching more fair, the queues are processed turn-by-turn
+in a round robin fashion.
 
-Semantically, there is a queue of upward messages queue where messages from each parachain are stored. Each parachain
-can have a limited number of messages and the total sum of pending messages is also limited. Each parachain can dispatch
-multiple upward messages per candidate.
+Other kinds of upward messages can be introduced in the future as well. Most likely the HRMP channel managem
 
 ## Horizontal Message Passing
 
--- a/roadmap/implementers-guide/src/messaging.md
+++ b/roadmap/implementers-guide/src/messaging.md
@@ -36,7 +36,9 @@ The weight that processing of the dispatchables can consume is limited by a prec
 that some dispatchables will be left for later blocks. To make the dispatching more fair, the queues are processed turn-by-turn
 in a round robin fashion.
 
-Other kinds of upward messages can be introduced in the future as well. Most likely the HRMP channel management.
+Other kinds of upward messages can be introduced in the future as well. Potential candidates are channel management for
+horizontal message passing (XCMP and HRMP, both are to be described later), new validation code signalling, or other
+requests to the relay chain.
 
 ## Horizontal Message Passing
 
--- a/roadmap/implementers-guide/src/messaging.md
+++ b/roadmap/implementers-guide/src/messaging.md
@@ -37,7 +37,7 @@ that some dispatchables will be left for later blocks. To make the dispatching m
 in a round robin fashion.
 
 Other kinds of upward messages can be introduced in the future as well. Potential candidates are channel management for
-horizontal message passing (XCMP and HRMP, both are to be described later), new validation code signalling, or other
+horizontal message passing (XCMP and HRMP, both are to be described below), new validation code signalling, or other
 requests to the relay chain.
 
 ## Horizontal Message Passing
--- a/roadmap/implementers-guide/src/messaging.md
+++ b/roadmap/implementers-guide/src/messaging.md
@@ -26,10 +26,10 @@ The downward message queue doesn't have a cap on its size and it is up to the re
 that prevent spamming in place.
 
 Upward Message Passing (UMP) is a mechanism responsible for delivering messages in the opposite direction:
-from a parachain up to the relay chain. Upward messagess can serve different purposes and can be of different
+from a parachain up to the relay chain. Upward messages can serve different purposes and can be of different
  kinds.
 
-One kind of messages is `Dispatchable`. They could be thought of similar to extrinsics sent to a relay chain: they also
+One kind of message is `Dispatchable`. They could be thought of similarly to extrinsics sent to a relay chain: they also
 invoke exposed runtime entrypoints, they consume weight and require fees. The difference is that they originate from
 a parachain. Each parachain has a queue of dispatchables to be executed. There can be only so many dispatchables at a time.
 The weight that processing of the dispatchables can consume is limited by a preconfigured value. Therefore, it is possible
--- a/roadmap/implementers-guide/src/runtime/router.md
+++ b/roadmap/implementers-guide/src/runtime/router.md
@@ -13,7 +13,7 @@ OutgoingParas: Vec<ParaId>;
 /// Messages ready to be dispatched onto the relay chain. The messages are processed in FIFO order.
 /// This is subject to `max_upward_queue_count` and
 /// `watermark_queue_size` from `HostConfiguration`.
-RelayDispatchQueues: map ParaId => Vec<UpwardMessage>;
+RelayDispatchQueues: map ParaId => Vec<RawDispatchable>;
 /// Size of the dispatch queues. Caches sizes of the queues in `RelayDispatchQueue`.
 /// First item in the tuple is the count of messages and second
 /// is the total length (in bytes) of the message payloads.
@@ -165,8 +165,6 @@ Candidate Acceptance Function:
   1. Checks that there are at most `config.max_upward_msg_num_per_candidate` messages.
   1. Checks each upward message individually depending on its kind:
   1. If the message kind is `Dispatchable`:
-      1. Check that `data` actually decodes to a valid entrypoint / dispatchable
-      1. Check that the entrypoint weight doesn't exceed `config.max_parachain_ump_dispatch_weight`
       1. Verify that `RelayDispatchQueueSize` for `P` has enough capacity for the message (NOTE that should include all processed
       upward messages of the `Dispatchable` kind up to this point!)
 * `check_processed_downward_messages(P: ParaId, processed_downward_messages)`:
@@ -201,7 +199,6 @@ Candidate Enactment:
   1. Set `HrmpWatermarks` for `P` to be equal to `new_hrmp_watermark`
 * `prune_dmq(P: ParaId, processed_downward_messages)`:
   1. Remove the first `processed_downward_messages` from the `DownwardMessageQueues` of `P`.
-
 * `enact_upward_messages(P: ParaId, Vec<UpwardMessage>)`:
   1. Process all upward messages in order depending on their kinds:
   1. If the message kind is `Dispatchable`:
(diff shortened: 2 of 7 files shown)


roadmap/implementers-guide/src/runtime/router.md
```diff
@@ -158,6 +161,14 @@ The following routines are intended to be invoked by paras' upward messages.
 
 Candidate Acceptance Function:
 
+* `check_upward_messages(P: ParaId, Vec<UpwardMessage>`:
+  1. Checks that there are at most `config.max_upward_msg_num_per_candidate` messages.
+  1. Checks each upward message individually depending on its kind:
+  1. If the message kind is `Dispatchable`:
+      1. Check that `data` actually decodes to a valid entrypoint / dispatchable
+      1. Check that the entrypoint weight doesn't exceed `config.max_parachain_ump_dispatch_weight`
```
rphmeier, line 169: so this weight is per-message and not total in queue?

rphmeier: given the validation data refactoring, i wonder how parachain block production functions are supposed to know that they are not producing too heavy of calls.

Maybe it would make sense to put a weight matrix of some kind into the `NonPersistedValidationData`. However the exact format doesn't seem obvious, given relay-chain runtime upgrades, for instance. relay-chain runtime upgrades can occur at any block and can change the weight. This may pose a barrier for contextual execution.

pepyakin: that's right.

pepyakin: yeah, the problem is that such a matrix can be invalidated on a chain upgrade anyway, i.e. actual weights could change. (although not sure if it will be a big of a deal in practice).

Can we delay upgrades to the session boundary?

pepyakin: 1. We want the acceptance criteria to be pre-evaluable by parachain validators only having access to `ValidationData` (the persisted and non persisted parts) for a candidate.
1. The acceptance function says that the weight of each dispatchable should not exceed a given maximum.
1. that maximum is easy to convey, we just put `max_parachain_ump_dispatch_weight` into `NonPersistedValidationData`.
1. the "the weight of each dispatchable" part is tricky. That implies (a) the need to know how to decode a dispatchable (also worth keeping in mind that we might need to duplicate this logic, if we don't go with the shared library approach) to extract the entrypoint and arguments and (b) a way to associate a dispatchable with a concrete weight cost (I guess the weight matrix you mentioned). However, some static matrix I guess won't do without further constraints because the weight is not a fixed value but rather a function of the entrypoint and its arguments.
1. The encoding and the matrix can be potentially invalidated during the upgrade.

Yeah, that's tricky.

pepyakin:
I pushed a commit that gives up a strict enforcement that dispatchables should be decodable and that they are not exceeding some critical weight per dispatchable.

Instead, I introduced a downward message that is sent in response for each dispatchable upward message that reports the status of execution, basically `Executed { success: bool } | DecodeFailed | CriticalWeightExceeded`

That allows not only to perform handling of failure on the parachain side, e.g. by retrying or releasing the assets that could be potentially locked, but  it also mitigates a more broad problem. It's not only the runtime code can invalidate a dispatchable but some dispatchable fails just because of some racy but nonetheless legit condition.

roadmap/implementers-guide/src/messaging.md
```diff
@@ -28,16 +26,17 @@ The downward message queue doesn't have a cap on its size and it is up to the re
 that prevent spamming in place.
 
 Upward Message Passing (UMP) is a mechanism responsible for delivering messages in the opposite direction:
-from a parachain up to the relay chain. Upward messages are dispatched to Runtime entrypoints and
-typically used for invoking some actions on the relay chain on behalf of the parachain.
+from a parachain up to the relay chain. Upward messagess can serve different purposes and can be of different
```
rphmeier, line 29:
```suggestion
from a parachain up to the relay chain. Upward messages can serve different purposes and can be of different
```

roadmap/implementers-guide/src/messaging.md
```diff
@@ -28,16 +26,17 @@ The downward message queue doesn't have a cap on its size and it is up to the re
...
+ kinds.
 
-> NOTE: It is conceivable that upward messages will be divided to more fine-grained kinds with a dispatchable upward message
-being only one kind of multiple. That would make upward messages inspectable and therefore would allow imposing additional
-validity criteria for the candidates that contain these messages.
+One kind of messages is `Dispatchable`. They could be thought of similar to extrinsics sent to a relay chain: they also
```
rphmeier, line 32:
```suggestion
One kind of message is `Dispatchable`. They could be thought of similarly to extrinsics sent to a relay chain: they also
```

roadmap/implementers-guide/src/messaging.md
```diff
@@ -28,16 +26,17 @@ The downward message queue doesn't have a cap on its size and it is up to the re
...
+invoke exposed runtime entrypoints, they consume weight and require fees. The difference is that they originate from
+a parachain. Each parachain has a queue of dispatchables to be executed. There can be only so many dispatchables at a time.
+The weight that processing of the dispatchables can consume is limited by a preconfigured value. Therefore, it is possible
+that some dispatchables will be left for later blocks. To make the dispatching more fair, the queues are processed turn-by-turn
+in a round robin fashion.
 
-Semantically, there is a queue of upward messages queue where messages from each parachain are stored. Each parachain
-can have a limited number of messages and the total sum of pending messages is also limited. Each parachain can dispatch
-multiple upward messages per candidate.
+Other kinds of upward messages can be introduced in the future as well. Most likely the HRMP channel managem
```
rphmeier, line 39: unfinished sentence?

roadmap/implementers-guide/src/messaging.md
rphmeier, line 32: Since this is the only kind right now, phrasing as "One kind, and the only current kind, of message..."

roadmap/implementers-guide/src/runtime/router.md
```diff
@@ -10,16 +10,19 @@ Storage layout:
 /// Paras that are to be cleaned up at the end of the session.
 /// The entries are sorted ascending by the para id.
 OutgoingParas: Vec<ParaId>;
-/// Messages ready to be dispatched onto the relay chain.
+/// Messages ready to be dispatched onto the relay chain. The messages are processed in FIFO order.
```
rphmeier, line 13: Maybe reframe this as "Dispatchable objects ready to be...".

There is no definition of `RawDispatchable` elsewhere in the guide, which is fine, but the comment should give some context.

roadmap/implementers-guide/src/runtime/router.md
```diff
@@ -190,15 +199,37 @@ Candidate Enactment:
...
+`process_upward_dispatchables()`:
+  1. Initialize a cumulative weight counter `T` to 0
+  1. Initialize a local in memory dictionary `R` that maps `ParaId` to a vector of `DispatchResult`.
+  1. Iterate over items in `NeedsDispatch` cyclically, starting with `NextDispatchRoundStartWith`. If the item specified is `None` start from the beginning. For each `P` encountered:
+      1. Dequeue `D` the first dispatchable `D` from `RelayDispatchQueues` for `P`
+      1. Decrement the size of the message from `RelayDispatchQueueSize` for `P`
+      1. Decode `D` into a dispatchable. If failed append `DispatchResult::DecodeFailed` into `R` for `P`. Otherwise, if succeeded:
+          1. If `weight_of(D) > config.dispatchable_upward_message_critical_weight` then append `DispatchResult::CriticalWeightExceeded` into `R` for `P`. Otherwise:
+            1. Execute `D` and add the actual amount of weight consumed to `T`. Add the `DispatchResult` into `R` for `P`.
+      1. If `weight_of(D) + T > config.preferred_dispatchable_upward_messages_step_weight`, set `NextDispatchRoundStartWith` to `P` and finish processing.
```
rphmeier, line 226: Perhaps should be `weight_of(D) + bookkeeping_weight + T` because of the downward message result.

roadmap/implementers-guide/src/runtime/router.md
rphmeier: I spoke too soon, there is a definition of `RawDispatchable`.

roadmap/implementers-guide/src/types/messages.md
```diff
@@ -22,11 +22,24 @@ enum ParachainDispatchOrigin {
...
+	/// This upward message is meant to schedule execution of a provided dispatchable.
+	Dispatchable {
+		/// The origin with which the dispatchable should be executed.
+		origin: ParachainDispatchOrigin,
+		/// The dispatchable to be executed in its raw form.
+		dispatchable: RawDispatchable,
+	},
+	// Examples:
+	// HrmpOpenChannel { .. },
+	// HrmpCloseChannel { .. },
```
rphmeier, line 42: What are these examples of?

roadmap/implementers-guide/src/types/messages.md
```diff
@@ -53,11 +66,25 @@ struct InboundHrmpMessage {
 
 ## Downward Message
 
-A message that go down from the relay chain to a parachain. Such a message could be initiated either
-as a result of an operation took place on the relay chain.
+`DownwardMessage`- is a message that go down from the relay chain to a parachain. Such a message
```
rphmeier, line 69:
```suggestion
`DownwardMessage`- is a message that goes down from the relay chain to a parachain. Such a message
```

roadmap/implementers-guide/src/types/messages.md
```diff
@@ -53,11 +66,25 @@ struct InboundHrmpMessage {
...
+could be seen as a notification, however, it is concievable that they might be used by the relay
```
rphmeier, line 70:
```suggestion
could be seen as a notification, however, it is conceivable that they might be used by the relay
```

roadmap/implementers-guide/src/types/messages.md
pepyakin: Of other potential kinds of messages, I can remove that if you think that's unnecessary given that those will become real messages

roadmap/implementers-guide/src/messaging.md
pepyakin: 🙈 yeah, I lost a battle with git-rebase there...

roadmap/implementers-guide/src/runtime/router.md
pepyakin: Well, then we would need to account well beyond the downward messages, namely accessing `RelayDispatchQueues`, `NeedsDispatch`.

TBH I think we need to start the weightification separately and together since literally nothing else account for it.

But I can leave a note here for future us though

roadmap/implementers-guide/src/messaging.md
```diff
@@ -28,16 +26,17 @@ The downward message queue doesn't have a cap on its size and it is up to the re
...
+invoke exposed runtime entrypoints, they consume weight and require fees. The difference is that they originate from
+a parachain. Each parachain has a queue of dispatchables to be executed. There can be only so many dispatchables at a time.
+The weight that processing of the dispatchables can consume is limited by a preconfigured value. Therefore, it is possible
+that some dispatchables will be left for later blocks. To make the dispatching more fair, the queues are processed turn-by-turn
+in a round robin fashion.
 
-Semantically, there is a queue of upward messages queue where messages from each parachain are stored. Each parachain
-can have a limited number of messages and the total sum of pending messages is also limited. Each parachain can dispatch
-multiple upward messages per candidate.
+Other kinds of upward messages can be introduced in the future as well. Most likely the HRMP channel management.
```
drahnr, line 39: Was `HRMP` declared before hands?

pepyakin: Thanks, fixed!

drahnr:
```suggestion
Other kinds of upward messages can be introduced in the future as well. Most likely the Horizontal Relay-chain Message Passing (HRMP) channel management.
```

roadmap/implementers-guide/src/types/messages.md
drahnr: I would be in favour of removing them, I find it unclear what usage they show.

roadmap/implementers-guide/src/types/runtime.md
```diff
@@ -39,7 +39,23 @@ struct HostConfiguration {
 	/// Total size of messages allowed in the parachain -> relay-chain message queue before which
 	/// no further messages may be added to it. If it exceeds this then the queue may contain only
 	/// a single message.
-	pub watermark_upward_queue_size: u32,
+	pub max_upward_queue_size: u32,
+	/// The amount of weight we wish to devote to the processing the dispatchable upward messages
+	/// stage.
+	///
+	/// NOTE that this is a soft limit and could be exceeded.
```
drahnr, line 46: Should we consider removing notes and making it part of the regular text?
```suggestion
	/// It is a soft limit and could be exceeded.
```

roadmap/implementers-guide/src/messaging.md
```diff
@@ -36,7 +36,9 @@ The weight that processing of the dispatchables can consume is limited by a prec
 that some dispatchables will be left for later blocks. To make the dispatching more fair, the queues are processed turn-by-turn
 in a round robin fashion.
 
-Other kinds of upward messages can be introduced in the future as well. Most likely the HRMP channel management.
+Other kinds of upward messages can be introduced in the future as well. Potential candidates are channel management for
+horizontal message passing (XCMP and HRMP, both are to be described later), new validation code signalling, or other
```
drahnr, line 40: "Later" as in at a later point of time in this document, or as in further down in this document, if so, a link might be helpful

pepyakin: Later, is actually as in later PRs : ) I am preparing one as we speak

roadmap/implementers-guide/src/types/runtime.md
pepyakin: I believe that this note is worth attention hence the tag. I am not married with this particular tag however. Not sure what is a common way to do a similar thing in the guide

roadmap/implementers-guide/src/messaging.md
pepyakin: I think this one is outdated

roadmap/implementers-guide/src/messaging.md
pepyakin: Ah sorry I misinterpreted. Further down the document

pepyakin: Addressed

roadmap/implementers-guide/src/types/runtime.md
drahnr: Frankly, I have no strong opinion about it, I just did not see much usage in other parts of the guide.

pepyakin:
bot merge

PR #1559, opened 2020-08-10
Parachains v1 registrar module.

### LINKED ISSUE #1508: Write a registrar module
This supersedes the functionality previously implemented in the `registrar` module of the v0 runtime. The ways in which it is superseded are described in #1501 .

This is point 1: This Registrar should provide a small wrapper around the Paras for handling deposits.

Things this module needs to do:
  * Accept parathread registrations and forward to the `Paras` module. Hold deposits for parathread
  * Accept parachain registrations and forward to the `Paras` module
  * Handle de-registrations of parathreads and parachains
  * Expose a `Call` for `Parachain` origins to trigger parachain/thread swaps (#1510, #1509)

The old registrar code (prior to #1501) may provide a reference for some of the logic

montekki:
Mostly a wrapper around `paras` module, however

* how does it work together with `paras_sudo_wrapper`? It seems that they may conflict.
* how do we manage `ParaId`s? Atm it's done by hand and I am not sure that the approach existing in the previous registrar would work.
* The module has to look out for what's going on in `paras` modules. For instance, it should not allow registration for `id`s that are outgoing in the `Paras` module. It's done now by looking into `paras`module but probably I've overlooked something.

This also has to build on #1562 

Closes #1508


runtime/common/src/paras_registrar.rs
```diff
@@ -0,0 +1,675 @@
...
+		/// Invalid para head data size.
+		HeadDataTooLarge,
+	}
+}
+
+decl_module! {
+	pub struct Module<T: Trait> for enum Call where origin: <T as frame_system::Trait>::Origin {
+		type Error = Error<T>;
+
+		#[weight = (5_000_000_000, DispatchClass::Operational)]
```
montekki, line 98: TODO: add docs and review weights.

montekki:
Also it's not quite clear if this should check the limits such as validation code length and head data length.

runtime/common/src/paras_registrar.rs
```diff
@@ -0,0 +1,675 @@
...
+
+			let debtor = <Debtors<T>>::take(id);
+			let _ = <T as Trait>::Currency::unreserve(&debtor, T::ParathreadDeposit::get());
+
+			<paras::Module<T>>::schedule_para_cleanup(id);
+
+			Ok(())
+		}
+
+		/// Swap a parachain with another parachain or parathread. The origin must be a `Parachain`.
```
drahnr, line 187: What's the reasoning for this limitation? Is it not enough that the types of `ParaId` are not equivalent?

rphmeier:
> how does it work together with paras_sudo_wrapper? It seems that they may conflict.

Yeah, they are not meant to be used together. This registrar is for permissionless deployments, whereas the sudo wrapper is for permissioned (testnet) deployments.

runtime/common/src/paras_registrar.rs
```diff
@@ -0,0 +1,681 @@
...
+decl_module! {
+	pub struct Module<T: Trait> for enum Call where origin: <T as frame_system::Trait>::Origin {
+		type Error = Error<T>;
+
+		/// Register a parathread with given code for immediate use.
+		///
+		/// Must be sent from a Signed origin that is able to have `ParathreadDeposit` reserved.
+		/// `gensis_head` and `validation_code` are used to initalize the parathread's state.
+		#[weight = (5_000_000_000, DispatchClass::Operational)]
+		pub fn register_parathread(
```
rphmeier, line 96: These calls may not even need to be calls, per-se. They could just be free functions that are invoked by e.g. the `slots` module

runtime/common/src/paras_registrar.rs
rphmeier: This is about the origin of the `Call` - parachains can send messages to the relay chain that are executed as calls to modules. So this requirement means that the parachains themselves must initiate the swap

runtime/parachains/src/lib.rs
```diff
@@ -42,8 +42,8 @@ pub mod runtime_api_impl;
 mod mock;
 
 /// Origin for the parachains.
-#[derive(PartialEq, Eq, Clone, Encode, Decode)]
-#[cfg_attr(feature = "std", derive(Debug))]
+#[derive(Debug, PartialEq, Eq, Clone, Encode, Decode)]
+// TODO: #[cfg_attr(feature = "std", derive(Debug))]
```
rphmeier, line 46: TODO needs to be resolved. Maybe `derive(RuntimeDebug)` is an option?

runtime/common/src/paras_registrar.rs
```diff
@@ -0,0 +1,674 @@
...
+
+decl_module! {
+	pub struct Module<T: Trait> for enum Call where origin: <T as frame_system::Trait>::Origin {
+		type Error = Error<T>;
+
+		/// Register a parathread with given code for immediate use.
+		///
+		/// Must be sent from a Signed origin that is able to have `ParathreadDeposit` reserved.
+		/// `gensis_head` and `validation_code` are used to initalize the parathread's state.
+		#[weight = 0]
```
rphmeier, line 94: did the previous registrar code have 0 weight also?

runtime/common/src/paras_registrar.rs
```diff
@@ -0,0 +1,674 @@
...
+				PendingSwap::insert(id, other);
+			}
+		}
+	}
+}
+
+impl<T: Trait> Module<T> {
+	/// Register a parachain with given code. Must be called by root.
+	/// Fails if given ID is already used.
+	pub fn register_parachain(
```
rphmeier, line 185: I note this doesn't set `Debtors`. The previous logic used something called `SwapAux` because deposits for parachains are in the `slots` module. I would prefer to have all deposits in one place, and this includes XCMP/HRMP channel deposits. So at some point we should have a `deposits` module where we can implement swapping over. I'm fine with the code committed as is for now, though.

runtime/common/src/paras_registrar.rs
montekki: I think [so](https://github.com/paritytech/polkadot/blob/a2b008ddd127652a98838dd6e28e651f82bb58af/runtime/common/src/registrar.rs#L317)

runtime/common/src/paras_registrar.rs
```diff
@@ -0,0 +1,668 @@
...
+decl_module! {
+	pub struct Module<T: Trait> for enum Call where origin: <T as frame_system::Trait>::Origin {
+		type Error = Error<T>;
+
+		/// Register a parathread with given code for immediate use.
+		///
+		/// Must be sent from a Signed origin that is able to have `ParathreadDeposit` reserved.
+		/// `gensis_head` and `validation_code` are used to initalize the parathread's state.
+		#[weight = 0]
+		fn register_parathread(
```
rphmeier, line 95: I would like a way to disable parathread registration so we can activate it later.

montekki: I added a bool to storage for this that can be toggled by `root` but I am not sure that this is a great solution.

rphmeier: It's not a great solution but it is OK for now.

rphmeier (approved):

Issue #1509, opened 2020-07-31
Parachain/thread swapping


rphmeier:
The old registrar removed in #1501 allowed the ability to swap parachains/threads. This kept the ID stable but swapped ownership of all deposits and removed any parathread scheduling.

The `Paras` and `Scheduler` modules need to be amended to support swapping. Swaps should occur only at session boundaries and only one swap per `ParaId`. It could be implemented by simply swapping within the `Paras` module and then making sure the `Scheduler`'s `initializer_on_new_session` handler clears all entries from the `ParathreadQueue` which are not parachains - previously, this was not possible, but a thread<->chain swap makes it a case to account for.

rphmeier:
This is addressed by #2406 to an extent?

shawntabrizi:
From what I can read in the description of this issue, this should be fixed by #2406

PR #1562, opened 2020-08-10
support swapping the para type

drahnr:
Impl of swapping the para type from thread to chain and vice versa while keeping all modalities.

Ref #1509


rphmeier:
The old behavior was something like this:

Two paras agree to swap. This can be chain<>chain, thread<>thread, or chain<>thread. Practically, we only really need to support the chain<>thread case. Then the swap is carried out: the chains' IDs stay the same, but their kind (thread/chain) are swapped, as well as their deposits.

drahnr:
No it doesn't, it was only the very beginning of swapping `thread <> chain` since, but since that #1559 + assumptions got falsified.

burdges:
I'd assume the swap only takes effect two epochs later or whatever?

rphmeier:
@burdges At the end of the current session (epoch) was what I was thinking

burdges:
It's fine I think.   I worried about something that looks not really too problematic anyways.

PR #1574, opened 2020-08-12
Make parachain validation wasm executor functional

bkchr:
- Increase the size of the validation result in the shared memory. The
validation result holds the new runtime when a runtime upgrade is
scheduled. So, we need to give it enough memory to send the data between
the validator and the wasm execution host.
- Add the `CallInWasmExt`. This is required when doing a runtime upgrade
to check that we upgrade to something meaningful.

### DIFF
--- a/parachain/src/wasm_executor/mod.rs
+++ b/parachain/src/wasm_executor/mod.rs
@@ -35,7 +35,7 @@ mod validation_host;
 // maximum memory in bytes
 const MAX_RUNTIME_MEM: usize = 1024 * 1024 * 1024; // 1 GiB
 const MAX_CODE_MEM: usize = 16 * 1024 * 1024; // 16 MiB
-const MAX_VALIDATION_RESULT_HEADER_MEM: usize = MAX_CODE_MEM + 1024; // 16.1 MiB
+const MAX_VALIDATION_RESULT_HEADER_MEM: usize = MAX_CODE_MEM + 1024; // 16.01 MiB
 
 /// A stub validation-pool defined when compiling for Android or WASM.
 #[cfg(any(target_os = "android", target_os = "unknown"))]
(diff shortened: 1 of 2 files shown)


parachain/src/wasm_executor/mod.rs
```diff
@@ -35,6 +35,7 @@ mod validation_host;
 // maximum memory in bytes
 const MAX_RUNTIME_MEM: usize = 1024 * 1024 * 1024; // 1 GiB
 const MAX_CODE_MEM: usize = 16 * 1024 * 1024; // 16 MiB
+const MAX_VALIDATION_RESULT_HEADER_MEM: usize = MAX_CODE_MEM + 1024; // 16.1 MiB
```
rphmeier, line 38: seems lower than 16.1

bkchr: Yeah, my math doesn't check out xD

bkchr, line 38:
```suggestion
const MAX_VALIDATION_RESULT_HEADER_MEM: usize = MAX_CODE_MEM + 1024; // 16.01 MiB
```

parachain/src/wasm_executor/mod.rs
```diff
@@ -35,6 +35,7 @@ mod validation_host;
 // maximum memory in bytes
 const MAX_RUNTIME_MEM: usize = 1024 * 1024 * 1024; // 1 GiB
 const MAX_CODE_MEM: usize = 16 * 1024 * 1024; // 16 MiB
+const MAX_VALIDATION_RESULT_HEADER_MEM: usize = MAX_CODE_MEM + 1024; // 16.01 MiB
```
NikVolf, line 38:
```suggestion
const MAX_VALIDATION_RESULT_HEADER_MEM: usize = MAX_CODE_MEM + 1024; // 16 MiB + 1KiB
```

Still lower than 16.01 🤔

bkchr:
bot merge

PR #1577, opened 2020-08-13
Move Claim to Existed Address

AurevoirXavier:
If we move claim to some existed address, we should also update the `Total`.

Due to the statement, I can't accumulate two addresses' claims. So I decide to overwrite the target address's claim.

And here's another solution: 

```rust
ensure!(Self::claims(address).is_none(), "Can NOT Move Claim to an EXISTED Address");
```

---

- [ ] I set the `on_runtime_upgrade`'s weight to `0`, I don't know how to deal with it.

### DIFF
--- a/runtime/common/src/claims.rs
+++ b/runtime/common/src/claims.rs
@@ -204,11 +204,10 @@ decl_module! {
 		fn deposit_event() = default;
 
 		fn on_runtime_upgrade() -> frame_support::weights::Weight {
-			use frame_support::migrations::*;
+			use frame_support::migration::*;
 
 			let total = <StorageIterator<BalanceOf<T>>>::new(b"Claims", b"Claims")
-				.iter()
-				.fold(0.into(), |acc, x| acc + x);
+				.fold(BalanceOf::<T>::zero(), |acc, (_, x)| acc + x);
 
 			put_storage_value(b"Claims", b"Total", &[], total);
 
--- a/runtime/common/src/claims.rs
+++ b/runtime/common/src/claims.rs
@@ -406,7 +406,7 @@ decl_module! {
 		}
 
 		#[weight = (
-			T::DbWeight::get().reads_writes(4, 4) + 100_000_000_000,
+			T::DbWeight::get().reads_writes(4, 5) + 100_000_000_000,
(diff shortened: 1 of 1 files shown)


kianenigma:
cc @thiolliere as I _think_ you wrote parts of this module?

bkchr:
Ping @shawntabrizi @thiolliere

runtime/common/src/claims.rs
```diff
@@ -405,7 +416,13 @@ decl_module! {
...
+			Claims::<T>::take(&old).map(|c| Claims::<T>::mutate(&new, |maybe_overwrite| {
+				if let Some(overwrite) = maybe_overwrite {
+					Total::<T>::mutate(|total| *total = *total - *overwrite + c);
+				}
+
+				*maybe_overwrite = Some(c);
+			}));
 			Vesting::<T>::take(&old).map(|c| Vesting::<T>::insert(&new, c));
 			Signing::take(&old).map(|c| Signing::insert(&new, c));
 			maybe_preclaim.map(|preclaim| Preclaims::<T>::mutate(&preclaim, |maybe_o|
```
thiolliere, line 428: what if the new ethereum account is also preclaim ? (i.e. it exist an account where Preclaim(account) == new)

thiolliere (commented):
Sorry for the delay, I forgot about this PR.

Considering new_ethereum account could also be preclaim, I can't really understand what is correct.

AFAICT moving claim to existing claim adress is not usefull I would rather do as you suggest:
```
ensure!(Self::claims(address).is_none(), "Can NOT Move Claim to an EXISTED Address");
```

I don't expect any of polkadot/kusama runtime to have moved claim to existing claim address, thus I think runtime upgrade is not necessary.

thiolliere:
considering this https://github.com/paritytech/polkadot/pull/1577#discussion_r496765959 preclaim doesn't work well with move claim to existed address, also is there any needs to move claim to existing address ?
I would prefer the second solution:
```
ensure!(Self::claims(address).is_none(), "Can NOT Move Claim to an EXISTED Address");
```

bkchr:
Do we still need this?

rphmeier:
stale

PR #1579, opened 2020-08-13
Implementer's Guide: Move `ParaId` from candidate.md to runtime.md

pepyakin:
Treat my OCD a bit


rphmeier:
The `Runtime` types section says that it's only for types that are exclusively used within the Runtime. However, the `ParaId` is used in subsystems as well.

pepyakin:
ah ok, I misunderstood the "exclusively" part.

PR #1585, opened 2020-08-13
Implement validation data refactor

### LINKED ISSUE #1539: ValidationData refactoring

rphmeier:
Closes #1539 

This also includes a couple auxiliary changes:
  * A few more guide updates that were missed before
  * Reducing the fields in `ValidationParams`


roadmap/implementers-guide/src/types/candidate.md
```diff
@@ -219,8 +219,8 @@ This struct encapsulates the outputs of candidate validation.
 struct ValidationOutputs {
 	/// The head-data produced by validation.
 	head_data: HeadData,
-	/// The validation data, persisted and transient.
-	validation_data: ValidationData,
+	/// The validation data, persisted.
+	validation_data: PersistedValidationData,
```
pepyakin, line 223: Can you elaborate on this change?

primitives/src/v1.rs
```diff
@@ -114,7 +106,7 @@ pub struct CandidateDescriptor<H = Hash> {
 	pub relay_parent: H,
 	/// The collator's sr25519 public key.
 	pub collator: CollatorId,
-	/// The blake2-256 hash of the validation data. This is extra data derived from
+	/// The blake2-256 hash of the persisted validation data. This is extra data derived from
```
montekki, line 109: A persisted validation data hash? Or the hash of the whlole `ValidationData`? Field name should probably be `persisted_validation_data_hash`.

primitives/src/v1.rs
```diff
@@ -168,10 +160,10 @@ impl<H> CandidateReceipt<H> {
...
 	/// The inner candidate receipt.
 	pub inner: CandidateReceipt<H>,
-	/// The global validation schedule.
-	pub global_validation: GlobalValidationData<N>,
-	/// The local validation data.
-	pub local_validation: LocalValidationData<N>,
+	/// The validation data derived from the relay-chain state at that
+	/// point. Should match the `validation_data_hash` in the descriptor
+	/// of the receipt.
+	pub validation_data: ValidationData<N>,
```
montekki, line 166: Again, the hash of the whole struct should match the one in the descriptor or only the hash of `persisted` field?

node/subsystem/src/messages.rs
```diff
@@ -372,15 +374,21 @@ pub enum RuntimeApiRequest {
 	ValidatorGroups(RuntimeApiSender<(Vec<Vec<ValidatorIndex>>, GroupRotationInfo)>),
 	/// Get information on all availability cores.
 	AvailabilityCores(RuntimeApiSender<Vec<CoreState>>),
-	/// Get the global validation data.
-	GlobalValidationData(RuntimeApiSender<GlobalValidationData>),
-	/// Get the local validation data for a particular para, taking the given
+	/// Get the full validation data for a particular para, taking the given
```
montekki, line 377: the "full validation data" is the next variant `FullValidationData` isn't it? This one is persisted?
```suggestion
	/// Get the persisted validation data for a particular para, taking the given
```

roadmap/implementers-guide/src/types/candidate.md
rphmeier: The reason is that the `ValidateCandidateExhaustive` API supports not providing the `TransientValidationData` if not desired. Then there would be no way to return it with the validation outputs.

However, this change also doesn't impact our other primary user of the `CandidateValidation` subsystem: candidate backing. Since Candidate Validation checks the transient validation data when validating from the runtime state, there isn't anything else that candidate backing could meaningfully do with that data.

node/primitives/src/lib.rs
```diff
@@ -153,6 +151,8 @@ pub enum InvalidCandidate {
 	NewCodeTooLarge(u64),
 	/// Head-data is over the limit.
 	HeadDataTooLarge(u64),
+	/// Code upgrade triggered but not allowed.
+	CodeUpgradeNotAllowed,
```
drahnr, line 155: Snuck that in :wink:

parachain/src/primitives.rs
```diff
@@ -228,22 +230,16 @@ pub struct UpwardMessage {
...
+	/// The collation body.
+	pub block_data: BlockData,
 	/// The current relay-chain block number.
-	pub relay_chain_height: polkadot_core_primitives::BlockNumber,
-	/// Whether a code upgrade is allowed or not, and at which height the upgrade
-	/// would be applied after, if so. The parachain logic should apply any upgrade
-	/// issued in this block after the first block
-	/// with `relay_chain_height` at least this value, if `Some`. if `None`, issue
-	/// no upgrade.
-	pub code_upgrade_allowed: Option<polkadot_core_primitives::BlockNumber>,
```
coriolinus, line 246: Did you intentionally remove `code_upgrade_allowed`? If yes, then ~~it's not clear why; there's no obvious connection to the validation data refactor~~.

[edit] Maybe this is related to the `code_upgrade_allowed` field in `TransientValidationData`? If so, then this is fine.

primitives/src/v1.rs
```diff
@@ -224,17 +216,78 @@ impl Ord for CommittedCandidateReceipt {
 	}
 }
 
-/// Extra data that is needed along with the other fields in a `CandidateReceipt`
-/// to fully validate the candidate. These fields are parachain-specific.
+/// The validation data provide information about how to validate both the inputs and
+/// outputs of a candidate.
```
coriolinus, line 220: 👍 for thorough documentation here.

roadmap/implementers-guide/src/node/utility/candidate-validation.md
```diff
@@ -40,7 +40,7 @@ Once we have all parameters, we can spin up a background task to perform the val
   * The collator signature is valid
   * The PoV provided matches the `pov_hash` field of the descriptor
 
-After that, we can invoke the validation function. Lastly, we do some final checks on the output:
+After that, we can invoke the validation function. Lastly, we do some final checks on the output using the `TransientValidationData`:
```
coriolinus, line 43:
```suggestion
After that, we can invoke the validation function. Lastly, if available, we do some final checks on the output using the `TransientValidationData`:
```

node/network/availability-distribution/src/tests.rs
```diff
@@ -434,8 +420,7 @@ fn reputation_verification() {
 			validator_groups,
 			availability_cores,
 			head_data: _,
-			local_validation_data: _,
-			global_validation_data: _,
+			persisted_validation_data: _,
```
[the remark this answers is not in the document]
drahnr: That's on me

parachain/src/primitives.rs
rphmeier: Yep, I did intentionally do it. Checking whether a code upgrade was submitted only when allowed is done by the relay-chain at inclusion, so it doesn't need to be kept as part of the persisted validation data.

node/network/availability-distribution/src/tests.rs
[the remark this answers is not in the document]
rphmeier: This is just me fixing fallout and preserving consistency with the code that was already there. But the syntax you mention is better for this, yeah.

rphmeier:
bot merge

rphmeier:
bot merge

PR #1576, opened 2020-08-12
guide: validation data refactoring

rphmeier:
related to #1539 

implementation following in another PR

PR #1588, opened 2020-08-14
Implementer's Guide: Incorporate HRMP to TransientValidationData 

pepyakin:
Updates `TransientValidationData` according to my comment here https://github.com/paritytech/polkadot/issues/1539#issuecomment-669256920

Related to https://github.com/paritytech/polkadot/pull/1576
Follow-up #1629 

**What to look for**

It is important to ensure that `TransientValidationData` contains all information to perform checks equivalent to those in Router's acceptance criteria.

### DIFF
--- a/roadmap/implementers-guide/src/types/candidate.md
+++ b/roadmap/implementers-guide/src/types/candidate.md
@@ -197,10 +197,15 @@ struct HrmpTransientValidationData {
 	/// A vector of open requests in which the para participates either as sender or recipient. The
 	/// items are ordered ascending by `HrmpChannelId`. The vector doesn't contain two entries
 	/// with the same `HrmpChannelId`.
-	open_requests: Vec<(HrmpChannelId, HrmpOpenChannelRequest)>,
+	open_requests: Vec<(HrmpChannelId, HrmpAbridgedOpenChannelRequest)>,
 	/// A vector of close requests in which the para participates either as sender or recipient.
 	/// The vector doesn't contain two entries with the same `HrmpChannelId`.
 	close_requests: Vec<HrmpChannelId>,
+}
+
+/// A shorter version of `HrmpOpenChannelRequest`.
+struct HrmpAbridgedOpenChannelRequest {
+	confirmed: bool,
 }
 
 struct HrmpChannelLimits {
--- a/roadmap/implementers-guide/src/types/candidate.md
+++ b/roadmap/implementers-guide/src/types/candidate.md
@@ -191,12 +191,15 @@ struct HrmpTransientValidationData {
 	/// mapping. The number elements in this vector corresponds to the number of egress channels.
 	/// Since it's a mapping there can't be two items with same `ParaId`.
 	egress_limits: Vec<(ParaId, HrmpChannelLimits)>,
-	/// The vector of paras that have a channel to this para. The number of elements in this vector
+	/// A vector of paras that have a channel to this para. The number of elements in this vector
 	/// correponds to the number of egress channels.
 	ingress_senders: Vec<ParaId>,
-	/// A list of open requests in which the para participates either as sender or recipient.
+	/// A vector of open requests in which the para participates either as sender or recipient. The
+	/// items are ordered ascending by `HrmpChannelId`. The vector doesn't contain two entries
+	/// with the same `HrmpChannelId`.
 	open_requests: Vec<(HrmpChannelId, HrmpOpenChannelRequest)>,
-	/// A list of open requests in which the para participates either as sender or recipient.
+	/// A vector of close requests in which the para participates either as sender or recipient.
+	/// The vector doesn't contain two entries with the same `HrmpChannelId`.
 	close_requests: Vec<HrmpChannelId>,
 }
 
--- a/roadmap/implementers-guide/src/types/candidate.md
+++ b/roadmap/implementers-guide/src/types/candidate.md
@@ -167,7 +167,7 @@ struct TransientValidationData {
 	///
 	/// This informs a relay-chain backing check and the parachain logic.
 	code_upgrade_allowed: Option<BlockNumber>,
-	/// A vector that enumerates the list of blocks in which there were at least one message was
+	/// A vector that enumerates the list of blocks in which there was at least one message
 	/// received. The first number is always after the watermark.
 	hrmp_digest: Vec<BlockNumber>,
 	/// The watermark of the HRMP. That is, the block number up to which (inclusive) all HRMP messages
--- a/roadmap/implementers-guide/src/types/candidate.md
+++ b/roadmap/implementers-guide/src/types/candidate.md
@@ -167,22 +167,37 @@ struct TransientValidationData {
 	///
 	/// This informs a relay-chain backing check and the parachain logic.
 	code_upgrade_allowed: Option<BlockNumber>,
-	/// A vector that enumerates the list of blocks in which there was at least one message
-	/// received. The first number is always after the watermark.
-	hrmp_digest: Vec<BlockNumber>,
-	/// The watermark of the HRMP. That is, the block number up to which (inclusive) all HRMP messages
-	/// sent to the parachain are processed.
-	hrmp_watermark: BlockNumber,
-	/// A mapping that specifies if the parachain can send an HRMP message to the given recipient
-	/// channel. A candidate can send a message only to the recipients that are present in this
-	/// mapping.
-	/// Since it's a mapping there can't be two items with same `ParaId`.
-	hrmp_egress_limits: Vec<(ParaId, HrmpChannelLimits)>,
 	/// A copy of `config.max_upward_message_num_per_candidate` for checking that a candidate doesn't
 	/// send more messages that permitted.
 	config_max_upward_message_num_per_candidate: u32,
 	/// The number of messages pending of the downward message queue.
 	dmq_length: u32,
+	/// A part of transient validaiton data related to HRMP.
+	hrmp: HrmpTransientValidationData,
+}
+
+struct HrmpTransientValidationData {
+	/// A vector that enumerates the list of blocks in which there was at least one HRMP message
+	/// received.
+	///
+	/// The first number in the vector, if any, is always greater than the HRMP watermark. The
+	/// elements are ordered by ascending the block number. The vector doesn't contain duplicates.
+	digest: Vec<BlockNumber>,
+	/// The watermark of the HRMP. That is, the block number up to which (inclusive) all HRMP messages
+	/// sent to the parachain are processed.
+	watermark: BlockNumber,
+	/// A mapping that specifies if the parachain can send an HRMP message to the given recipient
+	/// channel. A candidate can send a message only to the recipients that are present in this
+	/// mapping. The number elements in this vector corresponds to the number of egress channels.
+	/// Since it's a mapping there can't be two items with same `ParaId`.
+	egress_limits: Vec<(ParaId, HrmpChannelLimits)>,
+	/// The vector of paras that have a channel to this para. The number of elements in this vector
+	/// correponds to the number of egress channels.
+	ingress_senders: Vec<ParaId>,
+	/// A list of open requests in which the para participates either as sender or recipient.
+	open_requests: Vec<(HrmpChannelId, HrmpOpenChannelRequest)>,
+	/// A list of open requests in which the para participates either as sender or recipient.
+	close_requests: Vec<HrmpChannelId>,
 }
 
 struct HrmpChannelLimits {
--- a/roadmap/implementers-guide/src/types/candidate.md
+++ b/roadmap/implementers-guide/src/types/candidate.md
@@ -192,7 +192,8 @@ struct HrmpTransientValidationData {
 	/// Since it's a mapping there can't be two items with same `ParaId`.
 	egress_limits: Vec<(ParaId, HrmpChannelLimits)>,
 	/// A vector of paras that have a channel to this para. The number of elements in this vector
-	/// correponds to the number of ingress channels.
+	/// correponds to the number of ingress channels. The items are ordered ascending by `ParaId`.
+	/// The vector doesn't contain two entries with the same `ParaId`.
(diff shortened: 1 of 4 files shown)


rphmeier:
Will there be a follow-up for `PersistedValidationData`? For instance, to store the incoming (downward) messages?

roadmap/implementers-guide/src/types/candidate.md
```diff
@@ -159,6 +161,29 @@ struct TransientValidationData {
 	/// vector is sorted ascending by the para id and doesn't contain multiple entries with the same
 	/// sender. This informs the collator.
 	hrmp_mqc_heads: Vec<(ParaId, Hash)>,
+	/// A vector that enumerates the list of blocks in which there was at least one message
+	/// received. The first number is always after the watermark.
+	hrmp_digest: Vec<BlockNumber>,
+	/// The watermark of the HRMP. That is, the block number up to which (inclusive) all HRMP messages
+	/// sent to the parachain are processed.
+	hrmp_watermark: BlockNumber,
```
rphmeier, line 169: @bkchr One thing to watch out for here, is that because this is in the `TransientValidationData`, it is not a direct parameter to the validation function and instead will be provided by inhernet. That means that a malicious collator can provide something different (e.g. 100 instead of 90) and the relay-chain will still accept the block. So it is possible for this value to move _backwards_ from the parachain's perspective and Cumulus needs to account for that. Similarly with other values in this struct.

rphmeier: Actually, after conversation in DM with Sergei, our conclusion was that this watermark should not be provided via inherent and instead Cumulus should be tracking the watermark internally.

pepyakin:
> Will there be a follow-up for PersistedValidationData? For instance, to store the incoming (downward) messages?

FWIW, a follow up is here https://github.com/paritytech/polkadot/pull/1629

roadmap/implementers-guide/src/types/candidate.md
```diff
@@ -159,6 +167,29 @@ struct TransientValidationData {
...
+	/// A copy of `config.max_upward_message_num_per_candidate` for checking that a candidate doesn't
+	/// send more messages that permitted.
+	config_max_upward_message_num_per_candidate: u32,
+	/// The number of messages pending of the downward message queue.
+	dmq_length: u32,
+}
+
+struct HrmpChannelLimits {
+	/// Indicates if the channel is already full and cannot accept any more messages.
+	overfilled: bool,
```
bkchr, line 190:
```suggestion
	is_full: bool,
```

Sounds much simpler to me :P

roadmap/implementers-guide/src/types/candidate.md
bkchr: Not sure I understand this here, why isn't the relay chain checking that the watermark is `>= old_watermark`?

roadmap/implementers-guide/src/types/candidate.md
pepyakin: Indeed!

roadmap/implementers-guide/src/types/candidate.md
pepyakin: The relay-chain actually does check that the watermark is `> old_watermark`.

```
check_hrmp_watermark(P: ParaId, new_hrmp_watermark):
- new_hrmp_watermark should be strictly greater than the value of HrmpWatermarks for P (if any).
- new_hrmp_watermark must not be greater than the context's block number.
- in HrmpChannelDigests for P an entry with the block number equal to new_hrmp_watermark should exist.
```

pepyakin:
I grouped and split into a new struct `HrmpTransientValidationData` all data related to HRMP. I also added all data that was missing to check the acceptance criteria for a candidate, making this PR good to go.

I used quite some bit of git-fu (between this and some not-yet-published PR) so there might be some rebase artifacts.

Note, that some data is superfluous there. If we had had #1594 we wouldn't have needed those to live in `TransientValidationData` because these data do not inform the collator.

Take for example `open_requests`/`close_requests`. The decision for opening or closing a channel is taken by the PVF and not by a collator. If the PVF decided to open another channel beyond the allowed number of channels anyway, then all bets are off, and the collator won't be able to do anything.

One piece that is still missing is how do the messages get inside. The follow-up #1629 doesn't directly address this yet.

roadmap/implementers-guide/src/types/candidate.md
bkchr: What happens if the parachain can not process all downward messages of a given relay chain block and thus could not increase the value?

pepyakin: FWIW, in this particular place, we are discussing the HRMP watermark. For XCMP and by extension HRMP we have agreed on a simplifying assumption: we assume that a parachain can receive and process all the incoming messages within a block. I think that might be fine for HRMP, I don't think that we will reach the usage levels where they could be a problem and we are going to put aggressive limits on HRMP anyway (I see it happening within upcoming #1632).

For XCMP I think we might want to revisit this assumption. Under circumstances when the parachain is bombarded by the stream of messages over all the channels and processing of all these messages just don't fit into the PVF time or PoV space budget. Although that doesn't seem likely, I also feel that this simplifying assumption doesn't give us a lot - maybe it is not that hard to start with them?

Then, regarding downward messages. In the current version that we are discussing here, downward messages are implemented the same way you (@bkchr) introduced them: i.e. instead of a watermark we just used a simple `processed_downward_messages`. I liked this approach because, well, I didn't see a lot of reasons to use watermarks, but then today I finally broke down and switched to watermarks in https://github.com/paritytech/polkadot/pull/1629 (see the PR description to find out why).

But now I think maybe it would be a good idea to revert it back to `processed_downward_messages`. I'll think about that.

bkchr: Thank you for the explanation :) 

> Although that doesn't seem likely, I also feel that this simplifying assumption doesn't give us a lot - maybe it is not that hard to start with them?

I'm still not sure that this is true, IMHO we can reach this limit relative easy. In the end the parachain wants to process some of its own transactions as well. Just alone think about a parathread that collects messages over quite some time, the parachain will probably not be able to process them in one block.

pepyakin: I think in a big part it depends on the parametrization and there are two parameters in play: weight and pov size. I don't think that it is productive to speak about the weight here, because it is highly dependent on the exact code of PVF at hand and more primarily it's PVF which will be designed to not violate any constraints put onto the it by polkadot (although we need to come up with a less costraining system overall, about this later). Therefore, let's assume for now that messages received will always fit into the weight constraints.

Regarding the PoV size, this could be enforced simply. 

Since: 

1. there can be only so many paras, `N`, executed at the same time, and
1. every parachain can target the same single parachain, and
1. every parachain can only send one XCMP/HRMP message per block

we arrive to the worst case of `N` messages can be received at the same time, then we can say that we want to devote a half of PoV at most for messages. Dividing the terms gives us the maximum size of the message which would make it impossible to overwhelm the parachain as long as it's PVF doesn't use more than half of its PVF (and once again, it should follow the constraints unless it accepts possibility of getting in the aforementioned unfortunate situation).


Now, regarding the parathread, I don't think they are too special here

bkchr: >     3\. every parachain can only send one XCMP/HRMP message per block

Okay, I never heard from this before. However, this clearly lowers the amount of messages :)

pepyakin: I beg your pardon! It is actually

> every parachain can only send one XCMP/HRMP message per block **per recipient**

roadmap/implementers-guide/src/types/candidate.md
```diff
@@ -159,6 +167,52 @@ struct TransientValidationData {
...
+
+struct HrmpTransientValidationData {
+	/// A vector that enumerates the list of blocks in which there was at least one HRMP message
+	/// received.
+	///
+	/// The first number in the vector, if any, is always greater than the HRMP watermark. The
+	/// elements are ordered by ascending the block number. The vector doesn't contain duplicates.
+	digest: Vec<BlockNumber>,
+	/// The watermark of the HRMP. That is, the block number up to which (inclusive) all HRMP messages
+	/// sent to the parachain are processed.
```
rphmeier, line 187: The docs are a bit unclear. Is this the _previous_ watermark of the para or the _new_ watermark?

roadmap/implementers-guide/src/types/candidate.md
```diff
@@ -159,6 +167,52 @@ struct TransientValidationData {
...
+	watermark: BlockNumber,
+	/// A mapping that specifies if the parachain can send an HRMP message to the given recipient
+	/// channel. A candidate can send a message only to the recipients that are present in this
+	/// mapping. The number elements in this vector corresponds to the number of egress channels.
+	/// Since it's a mapping there can't be two items with same `ParaId`.
+	egress_limits: Vec<(ParaId, HrmpChannelLimits)>,
+	/// A vector of paras that have a channel to this para. The number of elements in this vector
+	/// correponds to the number of egress channels.
+	ingress_senders: Vec<ParaId>,
```
rphmeier, line 196: is this sorted in any way?

roadmap/implementers-guide/src/types/candidate.md
```diff
@@ -159,6 +167,52 @@ struct TransientValidationData {
...
+	/// A vector of open requests in which the para participates either as sender or recipient. The
+	/// items are ordered ascending by `HrmpChannelId`. The vector doesn't contain two entries
+	/// with the same `HrmpChannelId`.
+	open_requests: Vec<(HrmpChannelId, HrmpAbridgedOpenChannelRequest)>,
+	/// A vector of close requests in which the para participates either as sender or recipient.
+	/// The vector doesn't contain two entries with the same `HrmpChannelId`.
+	close_requests: Vec<HrmpChannelId>,
```
rphmeier, line 203: Does a channel appear in this vector only during one parablock, or does it appear in this vector as long as the channel has not yet been closed?

roadmap/implementers-guide/src/types/candidate.md
```diff
@@ -159,6 +167,52 @@ struct TransientValidationData {
...
+	/// mapping. The number elements in this vector corresponds to the number of egress channels.
+	/// Since it's a mapping there can't be two items with same `ParaId`.
+	egress_limits: Vec<(ParaId, HrmpChannelLimits)>,
+	/// A vector of paras that have a channel to this para. The number of elements in this vector
+	/// correponds to the number of egress channels.
+	ingress_senders: Vec<ParaId>,
+	/// A vector of open requests in which the para participates either as sender or recipient. The
+	/// items are ordered ascending by `HrmpChannelId`. The vector doesn't contain two entries
+	/// with the same `HrmpChannelId`.
+	open_requests: Vec<(HrmpChannelId, HrmpAbridgedOpenChannelRequest)>,
```
rphmeier, line 200: I wonder if this and close requests need to be part of `PersistedValidationData`. We don't want the collator to be able to forge this data.

Or maybe the hash of that data should be in the `PersistedValidationData`.

roadmap/implementers-guide/src/types/candidate.md
pepyakin: Yeah, it is the previous. I didn't want to repeat myself that all of those are collected before the execution of the PVF so I left a note above:

> It's worth noting that all the data is collected **before** the candidate execution.

Do you think it's better to clarify that they are from the current state explictily?

roadmap/implementers-guide/src/types/candidate.md
pepyakin: Those are channel close **requests** and these two fields contain all the registered at the context block before the execution of PVF. You can think basically as a copy of the storage list found in the router, but only filtered to be relevant to the parachain at hand.
As the open/close requests found in the router module storage, they live up until they are handled. For the case of close requests - it's the first session boundary.

roadmap/implementers-guide/src/types/candidate.md
pepyakin: Right, the PVF needs a way to authenticate that the following events:

roadmap/implementers-guide/src/types/candidate.md
```diff
@@ -159,6 +167,52 @@ struct TransientValidationData {
 	///
 	/// This informs a relay-chain backing check and the parachain logic.
 	code_upgrade_allowed: Option<BlockNumber>,
+	/// A copy of `config.max_upward_message_num_per_candidate` for checking that a candidate doesn't
+	/// send more messages that permitted.
+	config_max_upward_message_num_per_candidate: u32,
+	/// The number of messages pending of the downward message queue.
+	dmq_length: u32,
+	/// A part of transient validaiton data related to HRMP.
```
drahnr, line 175:
```suggestion
	/// A part of transient validation data related to HRMP.
```

roadmap/implementers-guide/src/types/candidate.md
pepyakin: If we assume that with #1642 we limit the total number of channels to a 2 digit number and then retire HRMP completely may be it makes sense to actually go with 1 for the time being.

pepyakin: I created #1663 to address this

roadmap/implementers-guide/src/types/candidate.md
pepyakin: Addressed

pepyakin:
I think all points were addressed or created issues for follow-ups, so I am going ahead and merging this to make my life less painful because of the rebases of the next PRs.

That said, feel free to leave any comments and I will address them in the follow up PRs.

pepyakin:
bot merge

pepyakin:
Hey, @drahnr can help me to please the bot by your thumbs up here? (Unless you have objections for merging ofc)

drahnr (approved):

drahnr:
bot merge

pepyakin:
bot merge

PR #1607, opened 2020-08-17
Propose parachain pallet for Rococo

bkchr:
Adds simple pallet for easy proposing of parachains for Rococo.

### DIFF
--- a/runtime/rococo/src/lib.rs
+++ b/runtime/rococo/src/lib.rs
@@ -658,6 +658,18 @@ impl proxy::Trait for Runtime {
 	type WeightInfo = ();
 }
 
+parameter_types! {
+	pub const ProposeDeposit: Balance = 1000 * DOLLARS;
+	pub const MaxNameLength: u32 = 20;
+}
+
+impl propose_parachain::Trait for Runtime {
+	type Event = Event;
+	type MaxNameLength = MaxNameLength;
+	type ProposeDeposit = ProposeDeposit;
+	type PriviledgedOrigin = EnsureRoot<AccountId>;
+}
+
 construct_runtime! {
 	pub enum Runtime where
 		Block = Block,
@@ -710,11 +722,14 @@ construct_runtime! {
 		// Sudo.
 		Sudo: sudo::{Module, Call, Storage, Event<T>, Config<T>},
 
-		// Proxy module. Late addition.
+		// Proxy module.
 		Proxy: proxy::{Module, Call, Storage, Event<T>},
 
-		// Multisig module. Late addition.
+		// Multisig module.
 		Multisig: multisig::{Module, Call, Storage, Event<T>},
+
+		// Propose parachain pallet.
+		ProposeParachain: propose_parachain::{Module, Call, Storage, Event},
 	}
 }
 
--- a/runtime/rococo/src/lib.rs
+++ b/runtime/rococo/src/lib.rs
@@ -681,6 +681,8 @@ impl EnsureOrigin<Origin> for PriviledgedOrigin {
 			_ => Err(origin),
 		}
 	}
+
+	fn successful_origin() -> OuterOrigin { Origin::root() }
 }
 
 impl propose_parachain::Trait for Runtime {
--- a/runtime/rococo/src/lib.rs
+++ b/runtime/rococo/src/lib.rs
@@ -59,6 +59,7 @@ use authority_discovery_primitives::AuthorityId as AuthorityDiscoveryId;
 use transaction_payment_rpc_runtime_api::RuntimeDispatchInfo;
 use session::historical as session_historical;
 use system::EnsureRoot;
+use constants::{time::*, currency::*, fee::*};
 
 #[cfg(feature = "std")]
 pub use staking::StakerStatus;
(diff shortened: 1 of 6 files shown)


runtime/rococo/src/propose_parachain.rs
```diff
@@ -0,0 +1,836 @@
...
+	type Event: From<Event> + Into<<Self as system::Trait>::Event>;
+
+	/// The maximum name length of a parachain.
+	type MaxNameLength: Get<u32>;
+
+	/// The amount that should be deposited when creating a proposal.
+	type ProposeDeposit: Get<BalanceOf<Self>>;
+
+	/// Priviledged origin that can approve/cancel/deregister parachain and proposals.
+	type PriviledgedOrigin: EnsureOrigin<<Self as system::Trait>::Origin, Success = ()>;
```
danforbes, line 53: Naming nitpick: could be improved as, e.g., `ParachainAdmin`.

runtime/rococo/src/propose_parachain.rs
```diff
@@ -0,0 +1,836 @@
...
+}
+
+/// A proposal for adding a parachain to the relay chain.
+#[derive(codec::Encode, codec::Decode)]
+struct Proposal<AccountId, ValidatorId, Balance> {
+	/// The account that proposed this parachain.
+	proposer: AccountId,
+	/// The validation WASM code of the parachain.
+	validation_function: ValidationCode,
```
danforbes, line 62: Is this different than the Wasm for the runtime? If so, how is this generated?

runtime/rococo/src/propose_parachain.rs
```diff
@@ -0,0 +1,836 @@
...
+				balance,
+			};
+
+			Proposals::<T>::insert(para_id, proposal);
+
+			Self::deposit_event(Event::ParachainProposed(name, para_id));
+		}
+
+		#[weight = 100_000]
+		fn approve_proposal(
```
shawntabrizi, line 178: This function doesnt do ANY verification that a proposal with this `para_id` exists. Looks like it would be a simple enough sanity check

runtime/rococo/src/propose_parachain.rs
```diff
@@ -0,0 +1,836 @@
...
+				if prop.validators.iter().all(|v| *v != validators[0] && *v != validators[1]) {
+					Ok(())
+				} else {
+					Err(Error::<T>::ValidatorAlreadyRegistered)
+				}
+			)?;
+
+			T::Currency::reserve(&who, T::ProposeDeposit::get())?;
+
+			let proposal = Proposal {
```
shawntabrizi, line 163: A best practice thing to do would be to include this deposit information into the Proposal struct, and then use that value to unreserve from the user.

This would allow us to update the `ProposeDeposit` a little easier in the future, and make sure we are not messing with any state.

This is fine if we dont plan to update this value throughout the lifetime of the chain

runtime/rococo/src/propose_parachain.rs
```diff
@@ -0,0 +1,836 @@
...
+
+			let proposal = Proposals::<T>::get(&para_id).ok_or(Error::<T>::ProposalNotFound)?;
+
+			if let Some(who) = who {
+				ensure!(who == proposal.proposer, Error::<T>::NotAuthorized);
+			}
+
+			Proposals::<T>::remove(&para_id);
+
+			T::Currency::unreserve(&proposal.proposer, T::ProposeDeposit::get());
```
shawntabrizi, line 204: Depending on who has access to Rococo, you should consider slashing the deposit here if called by the privileged origin. (assuming the proposal was rejected in that case)

Otherwise, there is no real disincentive to just ask for a million proposals... But maybe this kind of thinking only applies to a production chain

runtime/rococo/src/propose_parachain.rs
bkchr: No, it is exactly the same. However, `validation_function` is the term that is being used in Polkadot and I wanted to stick to this.

runtime/rococo/src/propose_parachain.rs
bkchr: My current idea is to require a high deposit to prevent spam xD And the tokens will come from us :P

So, yeah, I'm ignoring this all here.

runtime/rococo/src/propose_parachain.rs
bkchr: Yeah, I thought about this as well. And yeah, I also decided to never update the value throughout lifetime and if we would do, it would also be no problem.

runtime/rococo/src/lib.rs
```diff
@@ -657,6 +658,37 @@ impl proxy::Trait for Runtime {
...
+
+/// Priviledged origin used by propose parachain.
+pub struct PriviledgedOrigin;
+
+impl EnsureOrigin<Origin> for PriviledgedOrigin {
+	type Success = ();
+
+	fn try_origin(o: Origin) -> Result<Self::Success, Origin> {
+		let allowed = [
+			hex_literal::hex!("b44c58e50328768ac06ed44b842bfa69d86ea10f60bc36156c9ffc5e00867220")
```
danforbes, line 674:
```suggestion
			hex_literal::hex!("b44c58e50328768ac06ed44b842bfa69d86ea10f60bc36156c9ffc5e00867220"),
			hex_literal::hex!("762a6a38ba72b139cba285a39a6766e02046fb023f695f5ecf7f48b037c0dd6b")
```

PR #1609, opened 2020-08-18
Remove v0 node-side parachains code

rphmeier:
This removes the following crates:
  - polkadot-network
  - polkadot-collator
  - polkadot-availability-store
  - (most of) polkadot-validation

polkadot-validation still remains as a simple stub that provides an empty set of candidates for inclusion in the set-heads inherent. Although v0 parachains runtime code was removed in #1501, there are still live networks (Polkadot, Kusama, Westend) which have it deployed, and they will require the set-heads inherent in order to proceed until a runtime upgrade enacts the removal on those networks.

At that point, we will be able to remove the polkadot-validation and polkadot-service crates entirely, replacing with v1 code.


rphmeier:
@gavofyork Yeah, it has burned in on Westend and Kusama correctly.

rphmeier:
bot merge

PR #1614, opened 2020-08-19
Update Kusama Council seats & Runners Up number

rrtti:
Update Kusama council seats to 19 and `DesiredRunnersUp` to 19 as well, increasing representation of passive stakeholders in the council and giving more visibility to runners up in Kusama Network (from 7 to 19): this will be more visibility to new candidates passing the threshold to expand options for stakeholders when voting.


shawntabrizi (approved):
LGTM

PR #1622, opened 2020-08-21
Make ValidationPool accepts execution mode to run custom command or in process validation

cecton:
Related to https://github.com/paritytech/cumulus/issues/199
Also related to https://github.com/paritytech/cumulus/pull/201 for the actual use

~~This change will make the polkadot test service use the "test mode" of the validation pool.~~

I ~~also~~ moved the test flag from the validate_candidate to the pool directly which looks like a refactoring but my actual goal ~~was~~ _is_ to pass the test mode all the way down to the Validation pool. This will be done for v1 of the validation.


arkpar (approved):

service/src/lib.rs
```diff
@@ -391,6 +393,15 @@ pub fn new_full<RuntimeApi, Executor>(
 
 		polkadot_network_service.register_availability_store(availability_store.clone());
 
+		let execution_mode = if test_mode {
```
bkchr, line 396:
```suggestion
		let execution_mode = if !test_mode {
```

service/src/lib.rs
```diff
@@ -391,6 +393,15 @@ pub fn new_full<RuntimeApi, Executor>(
...
+			ValidationExecutionMode::Remote {
+				binary: std::env::current_exe().unwrap(),
```
bkchr, line 398:
```suggestion
				binary: std::env::current_exe().map_err(|_| "Could not get current executable name")?,
```

bkchr (approved):

service/src/lib.rs
cecton: Wait what. Why? The whole point was to get this on the worker args for the tests. Am I missing something?

parachain/src/wasm_executor/validation_host.rs
```diff
@@ -242,12 +259,17 @@ impl ValidationHost {
...
-		let mut args = if test_mode { WORKER_ARGS_TEST.to_vec() } else { WORKER_ARGS.to_vec() };
-		args.push(memory.get_os_path());
-		let worker = process::Command::new(self_path)
+		let (cmd, args) = match execution_mode {
+			ValidationExecutionMode::Local => (
+				env::current_exe()?,
+				WORKER_ARGS.iter().map(|x| x.to_string()).collect()
+			),
+			ValidationExecutionMode::Remote { binary, args } => (binary, args),
+		};
```
bkchr, line 268: Wait. You clear did not understand what Local and Remote mean here.

Local => Run the verification inside the current process
Remote => Use the given external command and args to start a worker.

service/src/lib.rs
bkchr: See my other comment, this makes no sense. In test mode you don't want to start the external worker and do it as thread instead. You mixed the concepts here totally. You should read the code of the validation again that decides on whether to start a validation worker or to run it inside the process.

bkchr (changes requested):

parachain/src/wasm_executor/validation_host.rs
arkpar: I actually would prefer an enum option to use default args and executable name. Defaults should work in all cases other than testing, and requiring the user to always specify executable/args would be bad API design. It is very much an internal implementation details and it only leaks away because tests require special handling.

arkpar:
> See my other comment, this makes no sense. In test mode you don't want to start the external worker and do it as thread instead. 

Some tests actually rely on the interruptable execution support. I would prefer all test to run as close to the real world usage as possible. The current version of this PR supports out of process execution for tests and I see no problem with that.

parachain/src/wasm_executor/validation_host.rs
bkchr: What you mean by enum option?

bkchr:
> > See my other comment, this makes no sense. In test mode you don't want to start the external worker and do it as thread instead.
> 
> Some tests rely on the interruptable execution support. I would prefer all test to run as close to the real world usage as possible. The current version of this PR supports out of process execution for tests and I see no problem with that.

Yeah and that is totally find and is still supported. I also don't want to remove support for this. However, @cecton had the problem that she wrote an integration tests and was not aware of the remote test execution. And this lead to some nested loop of always starting the same process over and over again.

For our integration test usage it is totally fine to run the validation in the process and that is what this pr is about.

So, what I imagined is the following:
- You can specify if you want to run the validation worker "local" aka in the process or "remote" aka as a new process
- This is configured when setting up the validation pool.
- We can make the internals easier that used `Local`, `Remote` and `RemoteTest`, by merging the later ones. So, when specifying the `Remote` we give the path to the executable and the arguments.

service/src/lib.rs
cecton: Oh no it's fine I get it now!

So Local would be use by the integration test on cumulus because we don't need it to be interruptible (or any more sophisticated), just need it to run for a few seconds during the test and everything is stopped anyway (because tests have their own process in Rust which is great I think).

Remote is the currently supported case in polkadot (and polkadot's tests!) where a background tasks is ran by spawning a new child process.

ez pz

parachain/src/wasm_executor/validation_host.rs
arkpar: An enum value that does not require setting executable name for out-of-process validation. E.g. `RemoteDefault` or simply `Remote`.  Basically something like this:
```rust
pub enum ValidationExecutionMode {
	InProcess,
        ExternalProcessSelfHost, 
	ExternalProcessCustomHost {
		binary: PathBuf,
		args: Vec<String>,
	},
```

arkpar:
> Yeah and that is totally find and is still supported. I also don't want to remove support for this. However, @cecton had the problem that she wrote an integration tests and was not aware of the remote test execution. And this lead to some nested loop of always starting the same process over and over again.

Right, but this PR fixes that. It allows for tests to run in external process.

> For our integration test usage it is totally fine to run the validation in the process and that is what this pr is about.

Are you sure about that? This might not be important for unit tests, but **integration** tests should be as close to real life behaviour as possible.

parachain/src/wasm_executor/validation_host.rs
bkchr: Okay. Sounds reasonable.

bkchr:
Yeah I know, but it is also weird to replicate always this extra test that is being called as the worker. The worker only exists to make validation interruptable and more observabale. I'm not that convinced that this feature is that important to test.

cecton:
> > For our integration test usage it is totally fine to run the validation in the process and that is what this pr is about.
> 
> Are you sure about that? This might not be important for unit tests, but **integration** tests should be as close to real life behaviour as possible.

Yes I know it sounds weird but as far as cumulus is concerned we don't really test that polkadot can run its validation in a subprocess properly. That is a test done in polkadot. We trust that polkadot is working and is tested. In cumulus we only care that both chains interact properly. Thus the simplest (and with less pitfalls) implementation is better.

> However, @cecton had the problem that she wrote an integration tests and was not aware of the remote test execution.

No no I was very aware of that. I did made the subcommand for the original integration tests. I just forgot about it and this time no error message warned me of my mistake (1). What I mean is: this is a human mistake and this is a pitfall.

(1) in the original integration test the command failed when received arguments it couldn't parse. Here it's the test binary that is ran and it doesn't fail easily, it tries to find a test called "validation-worker", filters out everything, and exit with a status 0. This is pure evil.

cecton:
@bkchr @arkpar ok this is ready for review

arkpar (approved):

bkchr (approved):

cecton:
bot merge

PR #1629, opened 2020-08-24
Implementer's Guide: Integrate DMP into PersistentValidationData + DMQ watermarks

### LINKED ISSUE #1643: Guide: Pass DMP messages into the PVF
Work out the way on how the PVF receives the messages from DMQ.

A straight way is to put them into `PersistentValidationData`. We need to put all messages in because `PersistentValidationData` is used to form the inputs and is collected before the execution of the PVF. 

An alternative is to only put data that allows to authorize the messages and let the collator include them with a proof. That would allow to have only a hash in `PersistentValidationData` and all received messages could appear only once in PoV.

pepyakin:
Based on top of https://github.com/paritytech/polkadot/pull/1588
Closes #1643 

### Summary

- Main theme of this PR is integration of DMP into the `PersistentValidationData` ("ground truth") through a DMQ MQC head.
- Introduce DMQ watermarks

This PR is accompanied by some drive-by wording fixes here and there.

### Description

I decided to take the approach of putting only the MQC head into `PersistentValidationData` to allow the authorization of the messages that a collator provides while building a parablock. 

<details>
<summary>Background</summary>
The alternative is to put the DMQ contents, all messages, into `PersistentValidationData`. We need to put all messages in because `PersistentValidationData` is used to form the inputs and is collected before the execution of the PVF.
Putting only data to authorize the sequence of messages into `PersistentValidationData` allows the collator to include the messages that were actually consumed into the PoV.

All in all, this should positively affect the amount of data made available.
</details>

While working on it I realized that it might be useful to supply the block number in which a particular downward message was put in the DMQ. That also makes it similar to HRMP/XCMP MQC which may be good from the implementation standpoint.

It seemed to me that it would be harder for the PVF to return `processed_downward_messages` and trying to keep track of the on-the-relay-chain state of DMQ rather than dealing with watermarks. So, i finally introduced **DMQ watermarks**. They cost a bit more, since we need to keep track the digest (the list of all blocks where the DMP messages were sent to correctly check the DMQ watermark)

Those DMQ watermarks, however, are still different from HRMP watermarks. I am still not confident about merging them.

### What to look for

Here are some clues how to approach this PR

- The `TransientValidationData` has all data to check if the resulting candidate passes the acceptance criteria (by collator or primary checker)
- The acceptance criteria establishes sensible invariants
    - That the enactment function can rely on
    - Won't change in between backing and the enactment
- Unused data is properly cleaned and doesn't pile up
- Among other things from areas of safety, consistency, etc

### DIFF
--- a/roadmap/implementers-guide/src/runtime/inclusion.md
+++ b/roadmap/implementers-guide/src/runtime/inclusion.md
@@ -69,7 +69,7 @@ All failed checks should lead to an unrecoverable error making the block invalid
   1. Check the collator's signature on the candidate data.
   1. check the backing of the candidate using the signatures and the bitfields, comparing against the validators assigned to the groups, fetched with the `group_validators` lookup.
   1. call `Router::check_upward_messages(para, commitments.upward_messages)` to check that the upward messages are valid.
-  1. call `Router::check_processed_downward_messages(para, commitments.processed_downward_messages)` to check that the DMQ is properly drained.
+  1. call `Router::check_dmq_watermark(para, commitments.dmq_watermark)` for each canddiate to check rules of processing the DMQ watermark.
   1. call `Router::check_hrmp_watermark(para, commitments.hrmp_watermark)` for each candidate to check rules of processing the HRMP watermark.
   1. check that in the commitments of each candidate the horizontal messages are sorted by ascending recipient ParaId and there is no two horizontal messages have the same recipient.
   1. using `Router::verify_outbound_hrmp(sender, commitments.horizontal_messages)` ensure that the each candidate send a valid set of horizontal messages
@@ -82,7 +82,7 @@ All failed checks should lead to an unrecoverable error making the block invalid
   1. call `Router::enact_upward_messages` for each backed candidate, using the [`UpwardMessage`s](../types/messages.md#upward-message) from the [`CandidateCommitments`](../types/candidate.md#candidate-commitments).
   1. call `Router::queue_outbound_hrmp` with the para id of the candidate and the list of horizontal messages taken from the commitment,
   1. call `Router::prune_hrmp` with the para id of the candiate and the candidate's `hrmp_watermark`.
-  1. call `Router::prune_dmq` with the para id of the candidate and the candidate's `processed_downward_messages`.
+  1. call `Router::prune_dmq` with the para id of the candidate and the candidate's `dmq_watermark`.
   1. Call `Paras::note_new_head` using the `HeadData` from the receipt and `relay_parent_number`.
 * `collect_pending`:
 
--- a/roadmap/implementers-guide/src/runtime/inclusion.md
+++ b/roadmap/implementers-guide/src/runtime/inclusion.md
@@ -69,7 +69,7 @@ All failed checks should lead to an unrecoverable error making the block invalid
   1. Check the collator's signature on the candidate data.
   1. check the backing of the candidate using the signatures and the bitfields, comparing against the validators assigned to the groups, fetched with the `group_validators` lookup.
   1. call `Router::check_upward_messages(para, commitments.upward_messages)` to check that the upward messages are valid.
-  1. call `Router::check_dmq_watermark(para, commitments.dmq_watermark)` for each canddiate to check rules of processing the DMQ watermark.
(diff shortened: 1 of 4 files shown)


roadmap/implementers-guide/src/runtime/router.md
```diff
@@ -23,15 +23,35 @@ NeedsDispatch: Vec<ParaId>;
 /// This is the para that gets will get dispatched first during the next upward dispatchable queue
 /// execution round.
 NextDispatchRoundStartWith: Option<ParaId>;
+```
+
+### DMP
```
drahnr, line 28:
```suggestion
### Downward Message Passing (DMP)
```

roadmap/implementers-guide/src/types/candidate.md
```diff
@@ -159,6 +172,56 @@ struct TransientValidationData {
 	///
 	/// This informs a relay-chain backing check and the parachain logic.
 	code_upgrade_allowed: Option<BlockNumber>,
+	/// A copy of `config.max_upward_message_num_per_candidate` for checking that a candidate doesn't
+	/// send more messages that permitted.
```
drahnr, line 176:
```suggestion
	/// send more messages than permitted.
```

roadmap/implementers-guide/src/types/candidate.md
```diff
@@ -159,6 +172,56 @@ struct TransientValidationData {
...
+	config_max_upward_message_num_per_candidate: u32,
+	/// A vector that enumerates the list of blocks in which there was at least one DMQ messages
+	/// enqueued.
+	///
+	/// The first number in the vector, if any, is always greater than the DMQ watermark. The
+	/// elements are ordered by ascending the block number. The vector doesn't contain duplicates.
+	dmq_digest: Vec<BlockNumber>,
+	/// A part of transient validaiton data related to HRMP.
```
drahnr, line 184:
```suggestion
	/// A part of transient validation data related to HRMP.
```

roadmap/implementers-guide/src/runtime/inclusion.md
```diff
@@ -69,7 +69,7 @@ All failed checks should lead to an unrecoverable error making the block invalid
   1. Check the collator's signature on the candidate data.
   1. check the backing of the candidate using the signatures and the bitfields, comparing against the validators assigned to the groups, fetched with the `group_validators` lookup.
   1. call `Router::check_upward_messages(para, commitments.upward_messages)` to check that the upward messages are valid.
-  1. call `Router::check_processed_downward_messages(para, commitments.processed_downward_messages)` to check that the DMQ is properly drained.
+  1. call `Router::check_dmq_watermark(para, commitments.dmq_watermark)` for each canddiate to check rules of processing the DMQ watermark.
```
drahnr, line 72:
```suggestion
  1. call `Router::check_dmq_watermark(para, commitments.dmq_watermark)` for each candidate to check rules of processing the DMQ watermark.
```

drahnr: The sentence here is not entirely clear to me, are the messages drained down to the watermark?

drahnr (commented):
Looks mostly good as far as I can tell, it would be nice to avoid using acronyms in headers,

roadmap/implementers-guide/src/runtime/inclusion.md
pepyakin: No, this call is part of the `process_candidates` invocation and that is happening at the inclusion time. That implies that we only checking the validity of the candidates at this time and we are not enacting any changes by this candidate as of yet.

pepyakin:
On the second thought, the DMQ watermarks are not such a great idea. The problem with them is that they are coarse grained and require reading all messages within a block. Earlier, we decided that we don't want to restrict the relay-chain on how many messages to send to a parachain. With that, we can potentially end up in a situation the relay-chain sent a too many messages to even fit into a parablock.

We could potentially leverage a compound watermark, similar to what we are going to use for XCMP where the watermark consists of the block number and para id, which allows to read only a part of messages in a block.

In the DMP we have only the single sender - the relay-chain and only meaningful tie-breaker for the messages within the block is the order in which they were sent, e.g. `(block number 5, message number 3 of 5)`.

That, however, seems to be an over-engineered version of the original `processed_downward_messages` approach. I think we would better to leave it.

Therefore, I am closing this PR

PR #1636, opened 2020-08-25
Remove Purchase Pallet from Polkadot + Westend

shawntabrizi:
This PR removes the Purchase pallet from Polkadot + Westend.

* A new function is introduced in the Purchase pallet that simply removes all storage items.
* The purchase pallet is replaced with the `dummy` runtime.
* The `remove_pallet` function is called on the next `CustomOnRuntimeUpgrade`.

### DIFF
--- a/runtime/common/src/purchase.rs
+++ b/runtime/common/src/purchase.rs
@@ -361,7 +361,7 @@ impl<T: Trait> Module<T> {
 
 	/// WARNING: Executing this function will clear all storage used by this pallet.
 	/// Be sure this is what you want...
-	fn remove_pallet() -> Weight{
+	pub fn remove_pallet() -> Weight{
 		Accounts::<T>::remove_all();
 		PaymentAccount::<T>::remove();
 		Statement::remove();
--- a/runtime/common/src/purchase.rs
+++ b/runtime/common/src/purchase.rs
@@ -358,17 +358,6 @@ impl<T: Trait> Module<T> {
 			false => Err(Error::<T>::InvalidSignature)?,
 		}
 	}
-
-	/// WARNING: Executing this function will clear all storage used by this pallet.
-	/// Be sure this is what you want...
-	pub fn remove_pallet() -> Weight{
-		Accounts::<T>::remove_all();
-		PaymentAccount::<T>::remove();
-		Statement::remove();
-		UnlockBlock::<T>::remove();
-
-		T::MaximumBlockWeight::get()
-	}
 }
 
 // This function converts a 32 byte AccountId to its byte-array equivalent form.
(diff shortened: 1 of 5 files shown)


runtime/common/src/purchase.rs
```diff
@@ -358,6 +358,17 @@ impl<T: Trait> Module<T> {
 			false => Err(Error::<T>::InvalidSignature)?,
 		}
 	}
+
+	/// WARNING: Executing this function will clear all storage used by this pallet.
+	/// Be sure this is what you want...
+	fn remove_pallet() -> Weight{
```
bkchr, line 364:
```suggestion
	pub fn remove_pallet() -> Weight{
```

shawntabrizi:
Is it strange the dummy pallet does not expose `Events`? Isn't this also something that would change ordering between runtime upgrade?

Obviously it would not be as impactful as Call, but feels zero cost to keep around here. @rphmeier

rphmeier:
@shawntabrizi Good point. Yeah, the dummy pallet should expose events. That was an oversight on my part when introducing it.

runtime/polkadot/src/lib.rs
```diff
@@ -915,10 +917,10 @@ construct_runtime! {
...
 
 		// Old parachains stuff. All dummies to avoid messing up the transaction indices.
-		DummyParachains: dummy::<Instance0>::{Module, Call},
-		DummyAttestations: dummy::<Instance1>::{Module, Call},
-		DummySlots: dummy::<Instance2>::{Module, Call},
-		DummyRegistrar: dummy::<Instance3>::{Module, Call},
+		DummyParachains: dummy::<Instance0>::{Module, Call, Event<T>},
+		DummyAttestations: dummy::<Instance1>::{Module, Call, Event<T>},
+		DummySlots: dummy::<Instance2>::{Module, Call, Event<T>},
+		DummyRegistrar: dummy::<Instance3>::{Module, Call, Event<T>},
```
rphmeier, line 923: Did all of these have events before?

shawntabrizi: good call. this was a bit mindless

shawntabrizi: Just slots and registrar:

```
		// Parachains stuff; slots are disabled (no auctions initially). The rest are safe as they
		// have no public dispatchables. Disabled `Call` on all of them, but this should be
		// uncommented once we're ready to start parachains.
		Parachains: parachains::{Module, Call, Storage, Config, Inherent, Origin},
		Attestations: attestations::{Module, Call, Storage},
		Slots: slots::{Module, Call, Storage, Event<T>},
		Registrar: registrar::{Module, Call, Storage, Event, Config<T>},

File: runtime/polkadot/src/lib.rs [hunk 3]
```diff
@@ -872,150 +874,10 @@ impl pallet_proxy::Trait for Runtime {
...
-			if let Direct { ref mut votes, .. } = voting {
-				if let Some((0, Standard { ref mut vote, .. })) = votes.first_mut() {
-					vote.conviction = Conviction::None
-				}
-			}
-			VotingOf::<Runtime>::insert(who, voting);
-		}
-
-		<Runtime as frame_system::Trait>::MaximumBlockWeight::get()
+		purchase::remove_pallet::<Runtime>()
```
gavofyork, line 877: executed both here and from the main lib.rs - which should it be?

shawntabrizi: Main `lib.rs`? Right now it is executed in the Polkadot runtime and the Westend runtime. Westend probably doesnt have much in there, but we could test this simple cleanup on there first.

gavofyork (approved):

PR #1645, opened 2020-08-26
implement candidate selection subsystem

coriolinus:


node/core/candidate-selection/src/lib.rs
```diff
@@ -0,0 +1,409 @@
...
+async fn second_candidate(
+	relay_parent: Hash,
+	candidate_receipt: CandidateReceipt,
+	pov: PoV,
+	sender: &mut mpsc::Sender<FromJob>,
+	metrics: &Metrics,
+) -> Result<(), Error> {
+	// CandidateBackingMessage::Second(Hash, CandidateReceipt, PoV),
+	match sender
+		.send(FromJob::Backing(CandidateBackingMessage::Second(
```
coriolinus, line 324: Fairly sure that seconding a candidate just involves sending this message, but if there's more or if the requirements are different, then I'll need to update this function.

coriolinus:
Also: worth noting that this one has no test suite. My idea was that this is factored pretty finely; few functions are both non-boilerplate and more than a handful of statements long. I didn't think that writing a suite was worth it in this case. It does feel odd not to write tests, though, so I would need very little convincing that it's worth the time investment to add one.

coriolinus:
As I will be gone until Thursday, if this PR is acceptable as-is, I hope someone else will merge it.

node/core/candidate-selection/src/lib.rs
```diff
@@ -0,0 +1,409 @@
...
+				pov.clone(),
+				self.sender.clone(),
+			)
+			.await
+			{
+				return;
+			}
+
+			let pov = Arc::try_unwrap(pov)
+				.expect("only clone went to fn candidate_is_valid, which is complete; qed");
```
rphmeier, line 229: This is kind of shaky. Ownership of the PoV has been transferred to the candidate validation subsystem, and we don't know what it's done with it in this section of the code.

node/core/candidate-selection/src/lib.rs
```diff
@@ -0,0 +1,409 @@
...
+}
+
+async fn second_candidate(
+	relay_parent: Hash,
+	candidate_receipt: CandidateReceipt,
+	pov: PoV,
+	sender: &mut mpsc::Sender<FromJob>,
+	metrics: &Metrics,
+) -> Result<(), Error> {
+	// CandidateBackingMessage::Second(Hash, CandidateReceipt, PoV),
```
rphmeier, line 322: commented out code should be removed

node/core/candidate-selection/src/lib.rs
rphmeier: Yeah, seconding the candidate shouldn't require anything beyond that.

node/core/candidate-selection/src/lib.rs
```diff
@@ -0,0 +1,409 @@
...
+			{
+				Err(err) => {
+					log::warn!(target: "candidate_selection", "failed to second a candidate: {:?}", err)
+				}
+				Ok(()) => self.seconded_candidate = Some(collator_id),
+			}
+		}
+	}
+
+	async fn handle_invalid(&mut self, candidate_receipt: CandidateReceipt) {
```
rphmeier, line 248: Should this function clear `self.seconded_candidate` and select a new collation?

rphmeier: (probably not needed for this basic version)

rphmeier (approved):
Approach looks generally good. Some tests would be nice for sanity. In particular, these three come to mind:
  * Ensure that, when nothing is seconded so far, the collation is fetched and seconded
  * Ensure that, when something is seconded so far, further collation notifications are ignored 
  * Reports of invalidity from candidate backing are propagated

node/core/candidate-selection/src/lib.rs
```diff
@@ -0,0 +1,409 @@
...
+		let received_from = match &self.seconded_candidate {
+			Some(peer) => peer,
+			None => {
+				log::warn!(target: "candidate_selection", "received invalidity notice for a candidate we don't remember seconding");
+				return;
+			}
+		};
+		log::info!(target: "candidate_selection", "received invalidity note for candidate {:?}", candidate_receipt);
```
montekki, line 256: Nit: using a constant `TARGET` would make things a bit more compact and guarantee that spelling errors would not lead to log entries going to different targets.

node/core/candidate-selection/src/lib.rs
```diff
@@ -0,0 +1,409 @@
...
+		let succeeded;
+		if let Err(err) = forward_invalidity_note(received_from, &mut self.sender).await {
+			log::warn!(target: "candidate_selection", "failed to forward invalidity note: {:?}", err);
+			succeeded = false;
+		} else {
+			succeeded = true;
+		}
```
montekki, line 263: maybe:
```rust
let succeeded = match forward_invalidity_note
```
?

montekki (commented):
Some nits

Probably should have some tests?

node/core/candidate-selection/src/lib.rs
coriolinus: IMO it should neither clear `self.seconded_candidate` nor select a new collation: it doesn't have the information necessary to recover from that, and it would take a lot of state to retain that information. Given that the collator who sent the invalid candidate is dropped and blocked, I'd expect other collators to try again on the next block, and it should pick one of those.

coriolinus:
bot merge

PR #1657, opened 2020-08-29
XCM v0

gavofyork:


parachain/src/xcm.rs
```diff
@@ -0,0 +1,238 @@
...
+	}
+
+	#[derive(Clone, Eq, PartialEq, Encode, Decode, RuntimeDebug)]
+	pub enum MultiNetwork {
+		Wildcard,
+		Identified(Vec<u8>),
+	}
+
+	#[derive(Clone, Eq, PartialEq, Encode, Decode, RuntimeDebug)]
+	pub enum MultiLocation {
```
shawntabrizi, line 109: This is pseudo code?

parachain/src/xcm.rs
```diff
@@ -0,0 +1,238 @@
...
+		Null,
+		X1(Junction),
+		X2(Junction, Junction),
+		X3(Junction, Junction, Junction),
+		X4(Junction, Junction, Junction, Junction),
+	}
+
+	#[derive(Clone, Eq, PartialEq, Encode, Decode, RuntimeDebug)]
+	pub enum Junction {
```
shawntabrizi, line 118: Doesn't match spec?

parachain/src/xcm.rs
```diff
@@ -0,0 +1,238 @@
...
+		Index128 { #[codec(compact)] id: u128 },
+		Array4([u8; 4]),
+		Array8([u8; 8]),
+		Array16([u8; 16]),
+		Array32([u8; 32]),
+		Blob(Vec<u8>),
+	}
+
+	#[derive(Clone, Eq, PartialEq, Encode, Decode, RuntimeDebug)]
+	pub enum MultiAsset {
```
shawntabrizi, line 155: potentially need to update spec here too

parachain/src/xcm.rs
```diff
@@ -0,0 +1,238 @@
...
+		ConcreteFungible { id: MultiLocation, #[codec(compact)] amount: u128 },
+		ConcreteNonFungible { class: MultiLocation, instance: AssetInstance },
+		Each(Vec<MultiAsset>),
+	}
+
+	#[derive(Clone, Eq, PartialEq, Encode, Decode)]
+	pub enum Ai {
+		Each(Vec<Ai>),
+		DepositAsset { asset: MultiAsset, dest_: MultiLocation },
+		ExchangeAsset { give: MultiAsset, receive: MultiAsset },
```
shawntabrizi, line 174: nit: spec says `for` instead of `receive`

parachain/src/xcm.rs
```diff
@@ -0,0 +1,238 @@
...
+		AbstractNonFungible { class: Vec<u8>, instance: AssetInstance },
+		ConcreteFungible { id: MultiLocation, #[codec(compact)] amount: u128 },
+		ConcreteNonFungible { class: MultiLocation, instance: AssetInstance },
+		Each(Vec<MultiAsset>),
+	}
+
+	#[derive(Clone, Eq, PartialEq, Encode, Decode)]
+	pub enum Ai {
+		Each(Vec<Ai>),
+		DepositAsset { asset: MultiAsset, dest_: MultiLocation },
```
shawntabrizi, line 173:
```suggestion
		DepositAsset { asset: MultiAsset, dest: MultiLocation },
```

TODO: Clean these up once scale codec is fixed https://github.com/paritytech/parity-scale-codec/pull/222

thiolliere: parity-scale-codec 1.3.5 has been released and should fix it.

parachain/src/xcm.rs
```diff
@@ -0,0 +1,238 @@
...
+		fn from(x: Junction) -> Self {
+			MultiLocation::X1(x)
+		}
+	}
+
+	#[derive(Clone, Eq, PartialEq, Encode, Decode, RuntimeDebug)]
+	pub enum AssetInstance {
+		Undefined,
+		Index8(u8),
+		Index16 { #[codec(compact)] id: u16 },
```
xlc, line 143: Not sure if we want to use compact encoding for u16, it can only save up to one byte and can use two more bytes in some cases.

parachain/src/xcm.rs
```diff
@@ -0,0 +1,238 @@
...
+		WildConcreteNonFungible { class: MultiLocation },
+		AbstractFungible { id: Vec<u8>, #[codec(compact)] amount: u128 },
+		AbstractNonFungible { class: Vec<u8>, instance: AssetInstance },
+		ConcreteFungible { id: MultiLocation, #[codec(compact)] amount: u128 },
+		ConcreteNonFungible { class: MultiLocation, instance: AssetInstance },
+		Each(Vec<MultiAsset>),
+	}
+
+	#[derive(Clone, Eq, PartialEq, Encode, Decode)]
+	pub enum Ai {
```
xlc, line 171: Do we want to use the full name here? i.e. `AssetInstruction`

parachain/src/xcm.rs
```diff
@@ -0,0 +1,238 @@
...
+		WildNonFungible,
+		WildAbstractFungible { id: Vec<u8> },
+		WildAbstractNonFungible { class: Vec<u8> },
+		WildConcreteFungible { id: MultiLocation },
+		WildConcreteNonFungible { class: MultiLocation },
+		AbstractFungible { id: Vec<u8>, #[codec(compact)] amount: u128 },
+		AbstractNonFungible { class: Vec<u8>, instance: AssetInstance },
+		ConcreteFungible { id: MultiLocation, #[codec(compact)] amount: u128 },
+		ConcreteNonFungible { class: MultiLocation, instance: AssetInstance },
+		Each(Vec<MultiAsset>),
```
xlc, line 167: Move this to first one to match with `Ai::Each`

parachain/src/xcm.rs
```diff
@@ -0,0 +1,238 @@
...
+	pub enum Xcm {
+		WithdrawAsset { asset: MultiAsset, effect: Ai },
+		// Equivalent to WithdrawAsset{asset, Ai::InitiateReserveTransfer{asset, dest, effect}
+		ReserveAssetTransfer { asset: MultiAsset, dest: MultiLocation, effect: Ai },
+		ReserveAssetCredit { asset: MultiAsset, effect: Ai },
+		TeleportAsset { asset: MultiAsset, effect: Ai },
+		Balances { query_id: Vec<u8>, assets: Vec<MultiAsset> },
+		Transact { origin_type: MultiOrigin, call: Vec<u8> },
+		// these won't be staying here for long. only v0 parachains with HRMP.
+		ForwardToParachain { id: u32, inner: Box<VersionedXcm> },
```
xlc, line 190: Not sure if it is possible to set explicit index here to 250 so remove those in later version won't leave a hole

xlc (commented):
The current XCM implementation will failed to decode messages contains future version.
Do we want better future compatibility? i.e. Better handle `VersionedXXX`, decode it into some `Unknown(OpaqueMessage)` type. That also requires the messages to be length prefixed.

parachain/src/xcm.rs
```diff
@@ -0,0 +1,238 @@
...
+	#[derive(Clone, Eq, PartialEq, Encode, Decode, RuntimeDebug)]
+	pub enum Junction {
+		Parent,
+		Parachain { #[codec(compact)] id: u32 },
+		OpaqueRemark(Vec<u8>),
+		AccountId32 { network: MultiNetwork, id: [u8; 32] },
+		AccountIndex64 { network: MultiNetwork, #[codec(compact)] index: u64 },
+		AccountKey20 { network: MultiNetwork, key: [u8; 20] },
+		/// An instanced Pallet on a Frame-based chain.
+		PalletInstance { id: u8 },
```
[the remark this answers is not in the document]
shawntabrizi: Yes, this is the index from `construct_runtime`. Will see if I can rename it.

shaopengw (commented):
Some feedback while integrating XCM v0 to ORML.

#### Decimal conversion.

We might need a mechanism to handle decimal differences between chains, like DOT in Acala is 18. It's not a problem for sending XCM, as the appropriate amounts could be set on different part of XCM while building it. But while executing DMP, currently there is no way to identify if it's from a chain with different decimals.

A straightforward solution might be to introduce `origin` parameter in `deposit_asset` and `withdraw_asset` in `TransactAsset` traits. So that amounts could be adjusted based on different origins.

#### More error handling

There is no detailed error info if something goes wrong in `CurrencyAdapter`. Maybe we could add a customizable variant into `xcm::traits::Error`, to contain details, instead of returning `Error::Undefined`.

#### Xcm/Effect execution events

It would be helpful to have events indicating what Xcm/Effects got executed.

PR #1659, opened 2020-08-30
Collator protocol subsystem

montekki:


node/network/collator-protocol/src/validator_side.rs
```diff
@@ -0,0 +1,1369 @@
...
+};
+use polkadot_node_network_protocol::{
+	v1 as protocol_v1, View, PeerId, ReputationChange as Rep, RequestId,
+	NetworkBridgeEvent,
+};
+
+use super::{modify_reputation, TARGET, Result};
+
+const COST_UNEXPECTED_MESSAGE: Rep = Rep::new(-10, "An unexpected message");
+const COST_REQUEST_TIMEDOUT: Rep = Rep::new(-20, "A collation request has timed out");
```
coriolinus, line 46:
```suggestion
const COST_REQUEST_TIMED_OUT: Rep = Rep::new(-20, "A collation request has timed out");
```

clarify the intent: this is _not_ a typo of "timeout"

node/network/collator-protocol/src/validator_side.rs
```diff
@@ -0,0 +1,1369 @@
...
+	timeout: Delay,
+
+	// The id of this request.
+	request_id: RequestId,
+}
+
+impl Future for CollationRequest {
+	type Output = CollationRequestResult;
+
+	fn poll(mut self: Pin<&mut Self>, cx: &mut Context) -> Poll::<Self::Output> {
```
coriolinus, line 74: Might be worth adding a comment to the effect that you can't just `select!` because this function is not async.

node/network/collator-protocol/src/validator_side.rs
```diff
@@ -0,0 +1,1369 @@
...
+	///  - reply with collations to other subsystems.
+	requests_info: HashMap<RequestId, PerRequest>,
+
+	/// Collation requests that are currently in progress.
+	requests_in_progress: FuturesUnordered<CollationRequest>,
+
+	/// Delay after which a collation request would time out.
+	request_timeout: Duration, 
+
+	/// Posessed collations.
```
coriolinus, line 140:
```suggestion
	/// Possessed collations.
```

node/network/collator-protocol/src/validator_side.rs
```diff
@@ -0,0 +1,1369 @@
...
+	// First take a look if we have already stored some of the relevant collations.
+	if let Some(collations) = state.collations.get(&(relay_parent, para_id)) {
+		for collation in collations.iter() {
+			if let Err(e) = tx.send(collation.clone()).await {
+				// We do not want this to be fatal because the receving subsystem
+				// may have closed the results channel for some reason.
+				trace!(
+					target: TARGET,
+					"Failed to send collation: {:?}", e,
+				);
```
coriolinus, line 164: Should probably `break` at this point to avoid spamming the traces.

node/network/collator-protocol/src/validator_side.rs
```diff
@@ -0,0 +1,1369 @@
...
+	*current = view;
+
+	if let Some(advertisments) = state.advertisments.get_mut(&peer_id) {
+		advertisments.retain(|(_, relay_parent)| !removed.contains(relay_parent));
+	}
+
+	let mut requests_to_cancel = Vec::new();
+
+	for removed in removed.into_iter() {
+		state.requested_collations.retain(|k, v| {
```
coriolinus, line 248: `O(m*n)` where `m` is `removed.len()` and `n` is `state.requested_collations.len()`.

We could reduce that algorithmic complexity if we require that both collections be sorted, and then do a double-pointer merge traversal, but I don't have a sense of the sizes involved here. Is it even worth worrying about that?

node/network/collator-protocol/src/validator_side.rs
```diff
@@ -0,0 +1,1369 @@
...
+						.or_default()
+						.push((collator_id.clone(), receipt, pov));
+				}
+			}
+		}
+	} else {
+		// TODO: This is tricky. If our chain has moved on, we have already canceled
+		// the relevant request and removed it from the map; so and we are not expecting
+		// this reply although technically it is not a malicious behaviur.
+		modify_reputation(ctx, origin, COST_UNEXPECTED_MESSAGE).await?;
```
coriolinus, line 307: The current behavior seems plausible to me. Even so, should remove the TODO or link an issue for future discussion of what the behavior should be.

node/network/collator-protocol/src/validator_side.rs
```diff
@@ -0,0 +1,1369 @@
...
+		}
+	}
+
+	state.collations.retain(|k, _| {
+		if k.0 == relay_parent {
+			false
+		} else {
+			true
+		}
+	});
```
coriolinus, line 455:
```suggestion
	state.collations.retain(|k, _| k.0 != relay_parent);
```

node/network/collator-protocol/src/validator_side.rs
```diff
@@ -0,0 +1,1369 @@
...
+) -> Result<()> {
+	let mut remove_these = Vec::new();
+
+	state.requested_collations.retain(|k, v| {
+		if k.0 == relay_parent {
+			remove_these.push(*v);
+			false
+		} else {
+			true
+		}
```
coriolinus, line 440:
```suggestion
		if k.0 == relay_parent {
			remove_these.push(*v);
		}
		k.0 != relay_parent
```

node/network/collator-protocol/src/validator_side.rs
```diff
@@ -0,0 +1,1369 @@
...
+
+/// Our view has changed.
+async fn handle_our_view_change(
+	state: &mut State,
+	view: View,
+) -> Result<()> {
+	let old_view = std::mem::replace(&mut (state.view), view);
+
+	let cloned_view = state.view.clone();
+	let removed = old_view.difference(&cloned_view).collect::<Vec<_>>();
```
coriolinus, line 468: Why do we need `cloned_view`, if we just take a reference?

```suggestion
	let removed = old_view.difference(&state.view).collect::<Vec<_>>();
```

node/network/collator-protocol/src/collator_side.rs
```diff
@@ -0,0 +1,1009 @@
...
+
+/// Distribute a collation.
+/// 
+/// Figure out the core our para is assigned to and the relevant validators.
+/// Issue a connection request to these validators.
+/// If the para is not scheduled or next up on any core, at the relay-parent,
+/// or the relay-parent isn't in the active-leaves set, we ignore the message
+/// as it must be invalid in that case - although this indicates a logic error
+/// elsewhere in the node.
+async fn distribute_collation<Context>(
```
coriolinus, line 81: This doesn't implement the guide section described [here](https://github.com/paritytech/polkadot/pull/1645/files?file-filters%5B%5D=.md#diff-91227dbbde6d46a6e998ada67f12f84bR114), which is fair, as #1645 is not yet merged. 

It's fine to just say that that's out of scope for this PR and create an issue to handle it later, but we shouldn't forget about it.

coriolinus (commented):
Lots of little things; should be approvable with a few touchups.

node/network/collator-protocol/src/collator_side.rs
```diff
@@ -0,0 +1,1009 @@
...
+	msg: CollatorProtocolMessage,
+) -> Result<()>
+where
+	Context: SubsystemContext<Message = CollatorProtocolMessage>
+{
+	use CollatorProtocolMessage::*;
+
+	match msg {
+		CollateOn(id) => {
+			state.collating_on = Some(id);
```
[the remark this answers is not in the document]
montekki: I believe that it is something done by collator node's operator as part of node configuration. Collators are unlikely to skip between paras

node/network/collator-protocol/src/validator_side.rs
montekki: I highly doubt that this number ever going to be huge enough to worry about complexity here. Even if we have a para with like 500 collators it is unlikely that a huge number of those will come up with blocks at each point in time because consensus reasons, is it?

node/network/collator-protocol/src/validator_side.rs
montekki: yep, mostly TODOs exist in the PR at places I have no idea what to do about, so

node/network/collator-protocol/src/validator_side.rs
montekki: Uhhhh in fact I will now try to remove this `Future`

node/subsystem/src/messages.rs
```diff
@@ -156,8 +156,11 @@ pub enum CollatorProtocolMessage {
 	CollateOn(ParaId),
 	/// Provide a collation to distribute to validators.
 	DistributeCollation(CandidateReceipt, PoV),
-	/// Fetch a collation under the given relay-parent for the given ParaId.
-	FetchCollation(Hash, ParaId, oneshot::Sender<(CandidateReceipt, PoV)>),
+	/// As a validator may be collated more than one collator on a given `ParaId`
+	/// there may be more than one collation per relay parent. This message allows
+	/// the user to be updated with all collations as they are harvested from the
+	/// relevant collators.
+	FetchCollations(Hash, ParaId, mpsc::Sender<(CollatorId, CandidateReceipt, PoV)>),
```
rphmeier, line 163: Since collations are heavy, we should not be actively fetching more than one collation unless we need to. It would be better to expose the collators to the candidate selection subsystem and allow it to make a choice of which collations to fetch.

node/network/collator-protocol/src/validator_side.rs
rphmeier: It might be possible to replace with an `async fn`

node/network/collator-protocol/src/lib.rs
```diff
@@ -0,0 +1,136 @@
...
+}
+
+/// The collator protocol subsystem.
+pub struct CollatorProtocolSubsystem {
+	protocol_side: ProtocolSide, 
+}
+
+impl CollatorProtocolSubsystem {
+	/// Instantiate the collator protocol side of this subsystem
+	pub fn new_collator_side(id: CollatorId) -> Self {
```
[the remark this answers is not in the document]
montekki: You mean `CollatorId`? That's necessary when collations are advertised isn't it?

node/subsystem/src/messages.rs
montekki: ok, but the guide says this:

> When acting on an advertisement, we issue a WireMessage::RequestCollation. If the request times out, we need to note the collator as being unreliable and reduce its priority relative to other collators. And then make another request - repeat until we get a response or the chain has moved on.

this needs to be changed and what needs to happen is this:

1) We need to let candidate selection subsystem know about advertised collations in an async manner (with a `mpsc`) since these advertisements may arrive at any time
1.5) No collations are actually requested by the protocol subsystem on its own.
2) Selection makes its choice and issues a `FetchCollation` and at this point protocol subsystem makes a request.

?

node/network/collator-protocol/src/validator_side.rs
```diff
@@ -0,0 +1,1360 @@
...
+
+	/// Derive RequestIds from this.
+	next_request_id: RequestId,
+
+	/// The collations we have requested by relay parent and para id.
+	///
+	/// For each relay parent and para id we may be connected to a number
+	/// of collators each of those may have advertised a different collation.
+	/// So we group such cases here.
+	requested_collations: HashMap<(Hash, ParaId, PeerId), RequestId>,
```
rphmeier, line 127: Also worth noting that we don't need to fetch any of the collations by collators.

node/network/collator-protocol/src/validator_side.rs
rphmeier: I am fine with it as-is for now. The iteration should be fast on good hardware.

node/network/collator-protocol/src/validator_side.rs
rphmeier: One possible behavior would be to keep around requests for one block and then drop them. If a request arrives from an old block we know it is expired but don't need to report as the response is genuine.

However I can see the current jobs system making that hard. @coriolinus This is the kind of stuff I was referring to when I mentioned that the jobs system is kind of fragile. We should have a notion of transitions (parent -> child) and global data in the subsystem library.

node/network/collator-protocol/src/validator_side.rs
```diff
@@ -0,0 +1,1360 @@
...
+	let mut state = State {
+		request_timeout,
+		..Default::default()
+	};
+
+	loop {
+		if let Poll::Ready(msg) = futures::poll!(ctx.recv()) {
+			let msg = msg?;
+			trace!(
+				"> Received a message {:?}", msg,
```
rphmeier, line 601: target?

node/network/collator-protocol/src/validator_side.rs
```diff
@@ -0,0 +1,1360 @@
...
+			match msg {
+				Communication { msg } => process_msg(&mut ctx, msg, &mut state).await?,
+				Signal(BlockFinalized(_)) => {}
+				Signal(ActiveLeaves(_)) => {}
+				Signal(Conclude) => { break }
+			}
+			continue;
+		}
+
+		if let Poll::Ready(Some(request)) = futures::poll!(state.requests_in_progress.next()) {
```
rphmeier, line 613: I believe this should be a `while let`.

node/network/collator-protocol/src/validator_side.rs
```diff
@@ -0,0 +1,1360 @@
...
+	// We have to go backwards in the map, again.
+	if let Some(key) = find_val_in_map(&state.requested_collations, &id) {
+		if let Some(_) = state.requested_collations.remove(&key) {
+			if let Some(request_info) = state.requests_info.remove(&id) {
+				let (relay_parent, para_id, peer_id) = key;
+
+				modify_reputation(ctx, peer_id.clone(), COST_REQUEST_TIMEDOUT).await?;
+
+				// the callee will check if the parent is still in view, so do no checks here.
+				request_collation(
```
rphmeier, line 486: Why does this immediately request the collation again?

node/network/collator-protocol/src/collator_side.rs
```diff
@@ -0,0 +1,1009 @@
...
+			);
+
+			return Ok(());
+		}
+	};
+
+	state.our_validators_groups.insert(relay_parent, our_validators.clone());
+
+	// Issue a discovery request for the validators of the current group and
+	// TODO: the next group.
```
rphmeier, line 139: Likewise; this should have an issue filed.

node/network/collator-protocol/src/collator_side.rs
```diff
@@ -0,0 +1,1009 @@
...
+			warn!(
+				target: TARGET,
+				"Looks like no core is assigned to {:?} at {:?}", id, relay_parent,
+			);
+			return Ok(());
+		}
+	};
+
+	// Determine the group on that core and
+	// TODO: the next group on that core.
```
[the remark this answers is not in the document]
rphmeier: This should be resolvable given current information from the runtime API. 

`fn validator_groups` gives a `Vec<Vec<ValidatorIndex>>` and `GroupRotationInfo`. I'm assuming we've gotten the core index from `group_rotation_info.group_for_core`.

So then it's just

```rust
let next_group_index = GroupIndex((current_group_index.0 + 1) % n_groups);
```

Find the validators of the group the same way and connect to them the same way.

Note that this does cause an error at the end of sessions, where you're prospectively connecting to a "next" group which never happens, but that's fine because 99% of the time this works.

node/network/collator-protocol/src/collator_side.rs
```diff
@@ -0,0 +1,1009 @@
...
+	let mut validators_ids = rx.await?;
+
+	for id in validators_ids.drain(..) {
+		state.known_validators.insert(id.1, id.0);
+	}
+
+	Ok(())
+}
+
+/// Advertise collation to a set of peers.
```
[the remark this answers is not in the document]
rphmeier: definitely relay-chain peers. We don't currently do any parachain sub-networking outside of what happens in Cumulus.

rphmeier (commented):
The code is really orderly.

There are a few stray TODOs and questions that need to be resolved and I don't agree with the change of `RequestCollation` to `RequestCollations`.

node/network/collator-protocol/src/validator_side.rs
montekki: > If the request times out, we need to note the collator as being unreliable and reduce its priority relative to other collators. And then make another request - repeat until we get a response or the chain has moved on.

node/subsystem/src/messages.rs
coriolinus: Yes, I agree: let candidate selection know about advertised collations with a `mpsc::Sender` with all the information it needs to fetch a particular collation, but don't actually transfer `(CandidateReceipt, PoV)` without an explicit `Fetch`.

montekki: Ok, I did that

node/network/collator-protocol/src/collator_side.rs
montekki: I think that it does now.

node/network/collator-protocol/src/validator_side.rs
coriolinus: @rphmeier agree that it would be handy to have parent relation information in the active set update broadcasts, so it's possible to do things like "keep information for N blocks". However, I think that's a feature of its own which should probably be moved to its own issue so as not to delay this PR.

coriolinus (approved):
Looks good to me, pending resolution of the final TODO.

node/network/collator-protocol/src/lib.rs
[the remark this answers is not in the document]
rphmeier: I mean the choice of which side of the protocol to engage in.

node/network/collator-protocol/src/validator_side.rs
rphmeier: Oh, yeah, that was coming from the perspective of there being no `CandidateSelection` subsystem. If there is `CandidateSelection`, it should be `CandidateSelection` that decides to make another request. It doesn't really make sense to request from the collator who just timed out over and over again - the implication of the line quoted from the guide is that we should make another request of another collator.

montekki:
bot merge

PR #1666, opened 2020-08-31
Implementer's Guide: Integrate DMP into PersistentValidationData

### LINKED ISSUE #1643: Guide: Pass DMP messages into the PVF

pepyakin:
Closes #1643

Basically the same as https://github.com/paritytech/polkadot/pull/1629, but without DMQ watermarks.


pepyakin:
bot merge

PR #1668, opened 2020-09-01
Bump Nominator Reward Limits

kianenigma:
Polkadot is already passed this limit and there are half a dozen or so validators who have already been saturated, causing the reward of at least a few hundred nominators to remain unclaimed. Since this limit was arbitrarily set before benchmarks, now we can increase it. 

Will re-benchmark to apply the changes to the staking weights. 

Companion for: https://github.com/paritytech/substrate/pull/7007


shawntabrizi:
I would also suggest we reduce the number of historical eras we keep. Right now, the number is quite large, which imo is not needed given the current mechanisms and incentives.

wpank:
I think the current Historical Eras (84) is actually very useful for deriving metrics on performance. I would be in favor of keeping it for now.

thiolliere:
who will decide on this ?
what is the reason for having more nominator per validator in polkadot than kusama, because of the amount of stakers ?

kianenigma:
> what is the reason for having more nominator per validator in polkadot than kusama, because of the amount of stakers ?

Yes, although I am open to other suggestions. The main point for now is to finish https://github.com/paritytech/substrate/pull/7007 sot that we _can_ do it. The value is less relevant to me.

kianenigma:
also made to be the companion for https://github.com/paritytech/substrate/pull/7007

kianenigma:
With 64: 
`Call staking_payout_stakers (with default args) takes 0.017354368 of the block weight`
With 256: 
`Call staking_payout_stakers (with default args) takes 0.069417472 of the block weight`


I think 256 is a good number to go for. 

--- 

@jacogr Pinging you to update your 64 as well. Although if it is dynamic, then nothing to do from your side.

jacogr:
Dynamic. However with the weights adjusted would need some tweaks since at the previous there were a max of 40 payout txs in a batch based on tests. So at say 128 it would be (128 / 64) * 40 that can be included in a single batch tx from the UI. (batch is only for a single block)

With weights adjusted will re-calculate that (via testing), but initially it would just adjust as per above, probably fitting in less than the new weights would suggest. (the new "40 number" would need to be determined inside a batch)

runtime/kusama/src/lib.rs
```diff
@@ -317,7 +317,7 @@ parameter_types! {
 	// 27 eras in which slashes can be cancelled (slightly less than 7 days).
 	pub const SlashDeferDuration: pallet_staking::EraIndex = 27;
 	pub const RewardCurve: &'static PiecewiseLinear<'static> = &REWARD_CURVE;
-	pub const MaxNominatorRewardedPerValidator: u32 = 64;
+	pub const MaxNominatorRewardedPerValidator: u32 = 128;
```
kianenigma, line 320:
```suggestion
	pub const MaxNominatorRewardedPerValidator: u32 = 256;
```

I don't see a reason why not to.

shawntabrizi (approved):

redzsina:
We are done with our review for this PR and don't see any issues with it.

PR #1673, opened 2020-09-02
Implementer's Guide: Allow negotiating the HRMP limits.

### LINKED ISSUE #1641: Implementer's Guide: HRMP/XCMP channel limits
XCMP tries it's best on delivering messages. However, that requires some degree of effort applied by the recipient parachain. Specifically, XCMP relies on the availability system as a fallback to obtain the message preimages. But availability data is only stored for only so much time and pruned within a day or less. Therefore, a parachain must consume the messages lest let them slip beyond the availability window and lose them forever.

That might be problematic for parathreads, since they don't produce a lot of blocks. However, if one of the parachains start spamming messages, the para will have to process them under a risk of losing other messages.

It seems a good idea to give control of how many messages a parathread is willing to process at most. Similarly, it seems a good idea to allow to control the maximum size of any message put into the recipient's channel. 

That would allow the recipient to tune the settings so that it can definitely handle the sudden surge of messages, won't allow to send unexpectedly large messages, would allow to correctly prioritise some channels over others (or the opposite, for suspicious paras it could set lower limits).

## Approach

This concern was already raised in the original research draft: [Opening/closing XCMP Channel](https://github.com/w3f/research/commit/a52335510de4e4b3d453e5f32613287ae418a055#diff-3a27e98dcdaf0fdab2577d7d5ce867cbR26).

My current vision is similar to that. The sender proposes the parameters of the channel in the open channel request and then the recipient confirms them by accepting that request. That requires the sender to know the appropriate parameters of the request, but that's fine: after all the sender is required to know the protocol the recipient understands.

pepyakin:
Closes #1641

Incorporate ability to specify the maximum number of pending messages and maximum size of a message that could be put into the channel. See more info at #1641 

I don't particularly like naming scheme, but this could be addressed later on

### DIFF
--- a/roadmap/implementers-guide/src/runtime/router.md
+++ b/roadmap/implementers-guide/src/runtime/router.md
@@ -56,6 +56,8 @@ struct HrmpChannel {
     limit_used_places: u32,
     /// The maximum total size of the messages that can be pending in the channel at once.
     limit_used_bytes: u32,
+    /// The maximum message size that could be put into the channel.
+    limit_message_size: u32,
     /// The current number of messages pending in the channel.
     /// Invariant: should be less or equal to `limit_used_places`.
     used_places: u32,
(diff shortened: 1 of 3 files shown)


pepyakin:
bot merge

pepyakin:
bot merge

pepyakin:
bot merge

PR #1575, opened 2020-08-12
Allow using any polkadot client instead of enum Client

cecton:
Related to https://github.com/paritytech/cumulus/issues/199

PR #1675, opened 2020-09-02
Backport: Make ValidationPool accepts execution mode to run custom command or in process validation (#1622)

cecton:
Related to #1622 
Related to https://github.com/paritytech/cumulus/issues/199
Also related to https://github.com/paritytech/cumulus/pull/201 for the actual use

This change will make the polkadot test service use the "in-process" mode of the validation pool.

Also:
* backporting: Allow using any polkadot client instead of enum Client (#1575)
* Replace impl Trait by generic parameter (code does not exist in master anymore)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -8623,8 +8623,11 @@ dependencies = [
  "polkadot-collator",
  "polkadot-parachain",
  "polkadot-primitives",
+ "polkadot-service",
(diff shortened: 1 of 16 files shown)


cecton:
The commit 93f0029184a6c1749ce9420f4df1628a313a117d seems to make the CI fail (cc @bkchr )

PR #1676, opened 2020-09-02
Add deb and RPM repository config and documentation

s3krit:
This change adds some documentation to the README.md on how to install the latest version of Polkadot using our package repositories, and some additional config to Cargo.toml for building .deb and .rpm packages with [cargo-deb](https://github.com/mmstick/cargo-deb) and [cargo-rpm](https://github.com/iqlusioninc/cargo-rpm)
Internal documentation on how the repos are maintained can be found [here](https://www.notion.so/paritytechnologies/Package-repositories-a02fe24f796d4645bfb1eac4694d468c).


README.md
```diff
@@ -12,12 +12,43 @@ information about installing the `polkadot` binary and developing on the codebas
...
+### Debian-based (Debian, Ubuntu)
+
+Currently supports Debian 10 (Buster) and Ubuntu 20.04 (Focal), and derivatives.
 
-### Use a Provided Binary
+```
+# Import the security@parity.io GPG key
+curl -fsSL 'https://keys.mailvelope.com/pks/lookup?op=get&search=security%40parity.io&options=mr&exact=on' | apt-key add -
+# Add the Parity repository and update the package index
+echo 'deb https://releases.parity.io/deb buster main' >> /etc/apt/sources.list.d/parity.list
```
kirushik, line 30: Can we please suggest putting the key into `/usr/share/keyrings/` and using the `deb [signed-by=<path to key>] URL…` format instead?
There's no real reason to add a system-level apt key for a non-system repo.
(Warning: the key should be _dearmored_ for this to work)

kirushik (commented):
By the way, do we package systemd units for polkadot?
I think we probably should (at least eventually), and bonus points for setting the necessary security mechanisms up: running `polkadot` as a separate unprivileged user, setting up limited capabilities for the daemon, properly handling logs...

s3krit:
> By the way, do we package systemd units for polkadot?
> I think we probably should (at least eventually), and bonus points for setting the necessary security mechanisms up: running `polkadot` as a separate unprivileged user, setting up limited capabilities for the daemon, properly handling logs...

We don't yet no, but there's no reason not to. It would be nicer if we could configure the node with a config file rather than having to edit the systemd service file each time. There are two outstanding issues for this in the Substrate repo (below).

https://github.com/paritytech/substrate/issues/6856
https://github.com/paritytech/substrate/issues/5887

kirushik:
@s3krit the common way to handle this with systemd is to put your options into a env variable in `/etc/default/polkadot`, something like this:
```
POLKADOT_CLI_ARGS="--database paritydb --wasm-execution Compiled"
```

Then load those ENV vars into your Systemd unit with 
```
[Service]
EnvironmentFile=/etc/default/polkadot
```

and then maybe using variable expansion when composing a string for the `Exec=` entry in the systemd unit.

I've seen quite a lot of daemons doing so, including `docker` for example.

README.md
s3krit: Resolved this with the latest commit. *Please* let me know if there's any method of getting the key from a keyserver to pipe to a file (unarmored), rather than importing it into their keyring and exporting it (which isn't the worst thing in the world...). There are only so many lines of the gpg manpage I can read in one day ;)

README.md
```diff
@@ -12,12 +12,53 @@ information about installing the `polkadot` binary and developing on the codebas
...
+## Installation
+
+If you just wish to run a Polkadot node without compiling it yourself, you may
+either run the latest binary from our
+[releases](https://github.com/paritytech/polkadot/releases) page, or install
+Polkadot from one of our package repositories.
+
+Installation from the debian or rpm repositories will create a `systemd`
+service that can be used to run a Polkadot node. This is disabled by default,
+and can be started by running `systemctl start polkadot`. By default, it will
```
kirushik, line 24:
```suggestion
and can be started by running `systemctl start polkadot` on demand (use `systemctl enable polkadot` to make it auto-start after reboot). By default, it will
```

README.md
```diff
@@ -12,12 +12,53 @@ information about installing the `polkadot` binary and developing on the codebas
...
+
+Currently supports Debian 10 (Buster) and Ubuntu 20.04 (Focal), and
+derivatives. Run the following commands as the `root` user.
+
+```
+# Import the security@parity.io GPG key
+gpg --recv-keys --keyserver hkps://keys.mailvelope.com FF0812D491B96798
+gpg --export security@parity.io > /usr/share/keyrings/parity.gpg
+# Add the Parity repository and update the package index
+echo 'deb [signed-by=/usr/share/keyrings/parity.gpg] https://releases.parity.io/deb buster main' > /etc/apt/sources.list.d/parity.list
```
kirushik, line 40: It's a really tiny nit, but since we're using the same repo for both `buster` and `focal`, maybe let's use some more generic name here in the path? I can easily imagine myself changing the distro name here, and ending up with an incorrect repository URL.
It's not a big issue (trivial to both check and figure out, and is probably only affects a small fraction of potential users anyway) — but I thought I'd better leave it as a note.

kirushik (approved):
This is awesome!
Like, I'm really-really happy we're finally doing this.

It's up to you if you're going to fix my two remaining comments, those are really tiny nits.

Side note: maybe let's see if there some additional features of systemd we can use to make this setup more secure by default?
I would suspect @ddorgan would have a lot of first-hand experience in running polkadot services in production, and @demimarie to know as much about capability filters and other Linux isolation tools as humanely possible.

README.md
```diff
@@ -12,12 +12,53 @@ information about installing the `polkadot` binary and developing on the codebas
...
+the command-line.
+
+### Debian-based (Debian, Ubuntu)
+
+Currently supports Debian 10 (Buster) and Ubuntu 20.04 (Focal), and
+derivatives. Run the following commands as the `root` user.
+
+```
+# Import the security@parity.io GPG key
+gpg --recv-keys --keyserver hkps://keys.mailvelope.com FF0812D491B96798
```
DemiMarie, line 37: This needs to be a full (40 hex character) fingerprint, to prevent preimage attacks.  If `keys.mailvelope.com` is a pool, we should also point to an individual server, as GPG can’t receive keys from a pool if it cannot resolve DNS.

README.md
```diff
@@ -12,12 +12,53 @@ information about installing the `polkadot` binary and developing on the codebas
...
+gpg --export security@parity.io > /usr/share/keyrings/parity.gpg
```
DemiMarie, line 38: This also needs to specify the fingerprint.

README.md
```diff
@@ -12,12 +12,53 @@ information about installing the `polkadot` binary and developing on the codebas
...
 
-If you want to connect to one of the networks supported by this repo, you can go to the latest
-release and download the binary that is provided.
+Currently supports Fedora 32 and CentOS 8, and derivatives.
+
+```
+# Install dnf-plugins-core (This might already be installed)
+dnf install dnf-plugins-core
+# Add the repository and enable it
+dnf config-manager --add-repo https://releases.parity.io/rpm/polkadot.repo
```
DemiMarie, line 55: We need to specify the GPG key fingerprint somewhere here.

scripts/packaging/deb-maintainer-scripts/postinst
```diff
@@ -0,0 +1,17 @@
...
+
+if [ "$action" = "configure" ]; then
+  # Make user and group
+  getent group polkadot >/dev/null 2>&1 || addgroup --system polkadot
+  getent passwd polkadot >/dev/null 2>&1 ||
+    adduser --system --home /home/polkadot --disabled-password \
+    --ingroup polkadot polkadot
+  if [ ! -e "$config_file" ]; then
+    echo 'POLKADOT_CLI_ARGS=""' > /etc/default/polkadot
+  fi
```
DemiMarie, line 16:
```suggestion
```

This isn’t actually needed, as systemd will treat an unset environment variable as an empty string.

README.md
DemiMarie, line 55: The GPG key fingerprint needs to be included somewhere, preferably in-line.

README.md
DemiMarie, line 38: GPG needs to be passed the full 40-character key fingerprint on both of these lines, to prevent preimage and spoofing attacks.

.rpm/polkadot.spec
```diff
@@ -0,0 +1,50 @@
+%define __spec_install_post %{nil}
+%define __os_install_post %{_dbpath}/brp-compress
```
DemiMarie, line 2: I have written my own RPM spec files and have never seen either of these lines, so I recommend adding a comment explaining what they do.

.gitignore
```diff
@@ -7,5 +7,7 @@ runtime/wasm/target/
 .idea
 .vscode
 polkadot.*
+!polkadot.service
```
DemiMarie, line 10: Why should systemd service units be in `.gitignore`?

scripts/packaging/polkadot.service
```diff
@@ -0,0 +1,17 @@
...
+
+[Service]
+EnvironmentFile=-/etc/default/polkadot
+ExecStart=/usr/bin/polkadot $POLKADOT_CLI_ARGS
+User=polkadot
+Group=polkadot
+Restart=always
+RestartSec=120
+MemoryHigh=5400M
+MemoryMax=5500M
```
DemiMarie, line 14: There are several sandboxing options that should be used here.  `PrivateTmp=yes`, `ProtectSystem=yes`, `ProtectHome=strict`, `NoNewPrivileges=yes`, and `SystemCallFilter=@system-service` come to mind, but there are others.  Ideally, we should use `Type=notify` and make Polkadot support the systemd notification API.

DemiMarie (commented):
The Debian and Fedora package install directions have a security vulnerability, and there are places where additional documentation could be added.  The systemd unit files should be locked down more as well; `systemd-analyze security` will help with that.

.gitignore
s3krit: This is actually a negation to counteract the `polkadot.*` line directly above - I didn't want to simply remove that line without understanding the potential consequences.

README.md
s3krit: 40-hex-character fingerprint = :+1: 

Could you clarify on your second point? Do you mean specifying the keyserver with a single IP address in case of DNS failure? If so, is that a better solution than not being able to receive keys in the event of DNS issues?

scripts/packaging/deb-maintainer-scripts/postinst
s3krit: My intent was to populate the blank config file so that users can edit it as necessary without having to consult the README again to see what the name of the CLI arg environment variable was named.

README.md
DemiMarie: GPG does support using an HTTP proxy, but only for standalone keyservers.  It cannot receive keys from a pool without DNS.  This has caused problems for me in the past when installing software in QubesOS, since QubesOS TemplateVMs can only access the Internet via a proxy and have no ability to resolve DNS.

scripts/packaging/deb-maintainer-scripts/postinst
DemiMarie: Good point.  That said, `/etc/default/polkadot` should be included in the package as a configuration file.

README.md
DemiMarie: GPG does support using an HTTP proxy, but only for standalone keyservers.  It cannot receive keys from a pool without DNS.  This has caused problems for me in the past when installing software in QubesOS, since QubesOS TemplateVMs can only access the Internet via a proxy and have no ability to resolve DNS.

README.md
s3krit: I've updated the polkadot.repo file in the repository to contain the full fingerprint, and added some documentation to this section specifying what the fingerprint should be

scripts/packaging/polkadot.service
```diff
@@ -12,6 +12,27 @@ Restart=always
...
+PrivateTmp=true
+PrivateUsers=true
+ProtectClock=true
+ProtectControlGroups=true
+ProtectHostname=true
+ProtectKernelModules=true
+ProtectKernelTunables=true
+ProtectSystem=strict
+RemoveIPC=true
+RestrictAddressFamilies=AF_INET AF_INET6 AF_NETLINK AF_PACKET
```
DemiMarie, line 29: We should not need `AF_PACKET`.  In this case, it is harmless, as it requires `CAP_NET_RAW` to use, but it is still good practice to avoid it.

README.md
```diff
@@ -54,7 +55,8 @@ dnf install dnf-plugins-core
 # Add the repository and enable it
 dnf config-manager --add-repo https://releases.parity.io/rpm/polkadot.repo
 dnf config-manager --set-enabled polkadot
-# Install polkadot (You may have to confirm the import of the GPG key)
+# Install polkadot (You may have to confirm the import of the GPG key, which
+# should have the following fingerprint: 9D4B2B6EB8F97156D19669A9FF0812D491B96798
```
DemiMarie, line 59:
```suggestion
# should have the following fingerprint: 9D4B2B6EB8F97156D19669A9FF0812D491B96798)
```

DemiMarie (commented):
Looks good!

scripts/packaging/polkadot.service
s3krit: Please see my latest commit for the updated `polkadot.service` file. I took your suggestions and the output of `systemd-analyze security`, being careful to not actually break the functionality of polkadot. It's looking a lot better ^_^ Thanks so much for the recommendations. I think this new service file is a good middle-ground between security and usability.

```
before changes: → Overall exposure level for polkadot.service: 9.2 UNSAFE 😨
after changes: → Overall exposure level for polkadot.service: 1.9 OK 🙂
```

Its remaining nags are:

```
 # Required to communicate over the network
PrivateNetwork=                            Service has access to the host's network        0.5
IPAddressDeny=                                              Service does not define an IP address allow list

# Required to communicate over the network and IPC
RestrictAddressFamilies=~AF_(INET|INET6)                    Service may allocate Internet sockets
RestrictAddressFamilies=~AF_PACKET                          Service may allocate packet sockets
RestrictAddressFamilies=~AF_NETLINK                         Service may allocate netlink sockets
RestrictAddressFamilies=~AF_UNIX                            Service may allocate local sockets

# Access to special devices like /dev/random, /dev/null and I think stdout and stdin
DeviceAllow=                                                Service has a device ACL with some special devices

# We currently write the db to $HOME/.local/share/polkadot. We could change this (I attempted to change it with StateDirectory but ran into problems nesting env vars in /etc/default. And this isn't particularly egregious anyway)
ProtectHome=                                                Service has full access to home directories

# Unsure of the impact of the two following entries, so left out to be safe
ProtectKernelLogs=                                          Service may read from or write to the kernel log ring buffer
RestrictRealtime=                                           Service may acquire realtime scheduling

# We're not doing any namespacing really
RootDirectory=/RootImage=                                   Service runs within the host's root directory

# Required because our code is polymorphic, crashes if we run with the compiled runtime with this option set
MemoryDenyWriteExecute=                                     Service may create writable executable memory mappings

# I figure it's possible some monitoring tools or *something* might need to read the database ¯\_(ツ)_/¯
UMask=                                                      Files created by service are group-readable by default
```

s3krit:
Final nags fixed :) Should be good to merge once green. I'll also manually add some info to the next release notes regarding our shiny new package repos

DemiMarie (approved):
LGTM!

PR #1679, opened 2020-09-02
The Router Module

### LINKED ISSUE #1400: Implement HRMP for parachains
https://github.com/paritytech/polkadot/pull/1399 describes the changes to the runtime and types that should be made.

We probably don't want to open up HRMP to parathreads without imposing a fee scheme on top of it.

### LINKED ISSUE #1663: Implementer's Guide: HRMP: Reliable way to notify PVF about incoming channels
The PVF needs a way to authenticate that the following events:

_Originally posted by @pepyakin in https://github.com/paritytech/polkadot/pull/1588#discussion_r477185909_

### LINKED ISSUE #1674: Router module
Implement the router skeleton and all the relevant boilerplate.

### LINKED ISSUE #1702: Implementer's Guide: Dekindification / XCM
During the XCM retreat it was decided to remove kinds from the upward and downward messages and basically replace them with XCM messages.

Right now upward messages are described by this struct

```rust
enum UpwardMessage {
    Dispatchable(...),
    HrmpInitOpenChannel(...),
    HrmpInitAcceptChannel(...),
    HrmpCloseChannel(...),
}
```

Dekindification of upward messages involves changing this definition directly into an XCM message. Therefore, the set of upward messages will contain either byte blobs which are to be interpreted as `VersionedXcm` struct or as `VersionedXcm` directly. At the moment of writing, [this](https://github.com/paritytech/polkadot/blob/6565989608bc9a416d7e6be421ff2cb44a6fd31a/xcm/src/lib.rs#L27-L30) is how `VersionedXcm` defined and `Xcm` for v0 is defined like the following

https://github.com/paritytech/polkadot/blob/6565989608bc9a416d7e6be421ff2cb44a6fd31a/xcm/src/v0.rs#L312-L323

Consequently, should there be there a need to add new message kinds - they will go directly into the XCM standard.

Apart from that, it was decided that the upward message kinds devoted to channel management are demoted to regular dispatchables (aka entrypoints). See https://github.com/paritytech/polkadot/issues/1704

Downward messages undergo similar changes.

### LINKED ISSUE #1704: Implementer's Guide: Demote HRMP channel management to regular dispatchables
Demote the HRMP channel management upward message kinds to regular entrypoints.

First reaction is that these entrypoints should go into the `Router` module. 

I think it might also imply changing `ValidationData` structs. The reason is the following. Currently if the collator provides invalid data to the PVF and then PVF does something nonsensical then the candidate is rejected and collator misses out. With dispatchables the candidate becomes valid and we need to pass the required data through the ground truth (if any) to inform the PVF. On the other hand, we may want to remove some stuff from the `TransientValidationData`.

Related https://github.com/paritytech/polkadot/issues/1702

### LINKED ISSUE #1806: HRMP: Provide a collator access to incoming messages
Since a collator needs to access the raw downward messages and incoming HRMP messages, there should be a means for doing that.

For the start, we might not need something sophisticated. I.e. we might get away with just a runtime API function or two that would just return all messages. We might extend that API down the line, e.g. return N last messages or return messages since a given block. 

The constraints for this API are:

1. The HRMP messages and downward messages must be separate.
1. The order of messages should be preserved.
1. HRMP messages must carry the sender.
1. HRMP messages and downward messages should carry the block number they were sent.

Note, that since HRMP will retire at some point it might be better to have a separate runtime API for it

### LINKED ISSUE #1869: UMP, HRMP: CandidateCommitments size limits
basically, currently, at the moment `CandidateCommitments` is mostly of constant size with exception of messages. HRMP and UMP outbound messages inflate the candidate commitments. HRMP has some upper bound of size: a parachain can open a limited number of channels and the sizes of messages are limited as well. UMP has only a limit on number of messages and these messages can be of any size. 

Therefore, if I am not mistaken, this means that `CandidateCommitments` can be inflated as well. Not sure, which would be the next limit (perhaps networking? or do we have a special limit for `CandidateCommitments`?), but I think we are better off introducing a dedicated limit or limits, that would effectively limit the overall size taken in candidate commitments. 

Now, apart from the size concern there is also the weight.  But now after dekindification (#1702, also tackled in this PR), all the messages are byte blobs anyway, so UMP delegates this to a higher level. UMP dispatching aside, we need to weigh all this machinery that propels it. 

I had this in mind and I was planning to discuss this on the second call with you (Rob) and Al, but that never happened and then I forgot until I was writing [these docs](https://github.com/paritytech/polkadot/pull/1679/commits/0d1835c7a2cef750828a14f2caf2a0793191c4bc#diff-65cfe3153bc3f3af755d35c872a99d790a940efd54223693a492f4bcf7b2a447R26).

_Originally posted by @pepyakin in https://github.com/paritytech/polkadot/pull/1679#discussion_r513073585_

pepyakin:
This is a placeholder PR that implements the router module, specifically UMP, DMP and HRMP. Along the way it updates the implementer's guide.

This PR has structured commits. Therefore:

1. expect a lot of rebases
1. the recommended way to check this PR is commit-by-commit basis.
1. if commits are good to go, they could be extracted into separate PRs and landed in advance.

Closes #1400
Closes #1663 
Closes #1702
Closes #1704
Closes #1806
Closes #1869

Regarding #1663 I decided to go with a simple implementation: just to send a DM to notify paras. For now, it makes sense to start with a simpler approach to see how it works and then extend it as needed. My concern about this solution was that these notifications should be processed in a timely manner but a downward message queue can be clogged. Well, that's a bit too forward looking since most likely the queues won't be filled in the foreseeable future (assuming we can upgrade later). Apart from that, if that becomes an issue, it is possible for the PVF to enact messages out-of-order therefore getting some sort of QoS. But again I doubt that this will be required for now.

TODO

- [x] Rebase this PR on after #1842
- [x] After #1842 was merged there were some TODOs left in code. I should address them in here.

### DIFF
--- a/core-primitives/src/lib.rs
+++ b/core-primitives/src/lib.rs
@@ -94,6 +94,25 @@ pub struct InboundDownwardMessage<BlockNumber = crate::BlockNumber> {
 	pub msg: DownwardMessage,
 }
 
+/// An HRMP message seen from the perspective of a recipient.
+#[derive(codec::Encode, codec::Decode, Clone, sp_runtime::RuntimeDebug, PartialEq)]
+pub struct InboundHrmpMessage<BlockNumber = crate::BlockNumber> {
+	/// The block number at which this message was sent.
+	/// Specifically, it is the block number at which the candidate that sends this message was
+	/// enacted.
+	pub sent_at: BlockNumber,
+	/// The message payload.
+	pub data: sp_std::vec::Vec<u8>,
+}
+
+#[derive(codec::Encode, codec::Decode, Clone, sp_runtime::RuntimeDebug, PartialEq, Eq, Hash)]
+pub struct OutboundHrmpMessage<Id> {
+	/// The para that will get this message in its downward message queue.
+	pub recipient: Id,
+	/// The message payload.
+	pub data: sp_std::vec::Vec<u8>,
+}
+
 /// V1 primitives.
 pub mod v1 {
 	pub use super::*;
--- a/parachain/src/primitives.rs
+++ b/parachain/src/primitives.rs
@@ -127,6 +127,14 @@ impl Id {
 	/// Use IsSystem instead.
 	#[deprecated]
 	pub fn is_system(&self) -> bool { self.0 < USER_INDEX_START }
+
+	/// Returns the para id represented as a u32.
+	///
+	/// An escape hatch meant to be used only in exceptional situations. For instance, when you need
+	/// to put the ID on wire but have no luxury of referencing this type directly.
+	pub fn reveal_inner_u32(self) -> u32 {
+		self.0
+	}
 }
 
 pub trait IsSystem {
--- a/parachain/src/primitives.rs
+++ b/parachain/src/primitives.rs
@@ -28,7 +28,7 @@ use serde::{Serialize, Deserialize};
 #[cfg(feature = "std")]
 use sp_core::bytes;
 
-use polkadot_core_primitives::Hash;
+use polkadot_core_primitives::{Hash, OutboundHrmpMessage};
 
 /// Block number type used by the relay chain.
 pub use polkadot_core_primitives::BlockNumber as RelayChainBlockNumber;
@@ -206,6 +206,16 @@ impl<T: Encode + Decode + Default> AccountIdConversion<T> for Id {
 	}
 }
 
+/// A type that uniquely identifies an HRMP channel. An HRMP channel is unidirectional
+#[derive(Clone, PartialEq, Eq, Encode, Decode, RuntimeDebug)]
+#[cfg_attr(feature = "std", derive(Hash))]
+pub struct HrmpChannelId {
+	/// The para that acts as the sender in this channel.
+	pub sender: Id,
+	/// The para that acts as the recipient in this channel.
+	pub recipient: Id,
+}
+
 /// A message from a parachain to its Relay Chain.
 pub type UpwardMessage = Vec<u8>;
 
@@ -232,7 +242,7 @@ pub struct ValidationParams {
 }
 
 /// The result of parachain validation.
-// TODO: egress and balance uploads (https://github.com/paritytech/polkadot/issues/220)
+// TODO: balance uploads (https://github.com/paritytech/polkadot/issues/220)
 #[derive(PartialEq, Eq, Encode)]
 #[cfg_attr(feature = "std", derive(Debug, Decode))]
 pub struct ValidationResult {
@@ -242,8 +252,12 @@ pub struct ValidationResult {
 	pub new_validation_code: Option<ValidationCode>,
 	/// Upward messages send by the Parachain.
 	pub upward_messages: Vec<UpwardMessage>,
+	/// Outbound horizontal messages sent by the parachain.
+	pub horizontal_messages: Vec<OutboundHrmpMessage<Id>>,
 	/// Number of downward messages that were processed by the Parachain.
 	///
 	/// It is expected that the Parachain processes them from first to last.
 	pub processed_downward_messages: u32,
+	/// The mark which specifies the block number up to which all inbound HRMP messages are processed.
+	pub hrmp_watermark: RelayChainBlockNumber,
 }
--- a/parachain/src/primitives.rs
+++ b/parachain/src/primitives.rs
@@ -198,44 +198,8 @@ impl<T: Encode + Decode + Default> AccountIdConversion<T> for Id {
 	}
 }
 
-/// Which origin a parachain's message to the relay chain should be dispatched from.
-#[derive(Clone, PartialEq, Eq, Encode, Decode)]
-#[cfg_attr(feature = "std", derive(Debug, Hash))]
-#[repr(u8)]
-pub enum ParachainDispatchOrigin {
-	/// As a simple `Origin::Signed`, using `ParaId::account_id` as its value. This is good when
-	/// interacting with standard modules such as `balances`.
-	Signed,
-	/// As the special `Origin::Parachain(ParaId)`. This is good when interacting with parachain-
-	/// aware modules which need to succinctly verify that the origin is a parachain.
-	Parachain,
-	/// As the simple, superuser `Origin::Root`. This can only be done on specially permissioned
-	/// parachains.
-	Root,
-}
-
-impl sp_std::convert::TryFrom<u8> for ParachainDispatchOrigin {
-	type Error = ();
-	fn try_from(x: u8) -> core::result::Result<ParachainDispatchOrigin, ()> {
-		const SIGNED: u8 = ParachainDispatchOrigin::Signed as u8;
-		const PARACHAIN: u8 = ParachainDispatchOrigin::Parachain as u8;
-		Ok(match x {
-			SIGNED => ParachainDispatchOrigin::Signed,
-			PARACHAIN => ParachainDispatchOrigin::Parachain,
-			_ => return Err(()),
-		})
-	}
-}
-
 /// A message from a parachain to its Relay Chain.
-#[derive(Clone, PartialEq, Eq, Encode, Decode)]
-#[cfg_attr(feature = "std", derive(Debug, Hash))]
-pub struct UpwardMessage {
-	/// The origin for the message to be sent from.
-	pub origin: ParachainDispatchOrigin,
-	/// The message data.
-	pub data: Vec<u8>,
-}
+pub type UpwardMessage = Vec<u8>;
 
 /// Validation parameters for evaluating the parachain validity function.
 // TODO: balance downloads (https://github.com/paritytech/polkadot/issues/220)
--- a/roadmap/implementers-guide/src/messaging.md
+++ b/roadmap/implementers-guide/src/messaging.md
@@ -26,20 +26,28 @@ The downward message queue doesn't have a cap on its size and it is up to the re
 that prevent spamming in place.
 
 Upward Message Passing (UMP) is a mechanism responsible for delivering messages in the opposite direction:
-from a parachain up to the relay chain. Upward messages can serve different purposes and can be of different
- kinds.
+from a parachain up to the relay chain. Upward messages are essentially byte blobs. However, they are interpreted
+by the relay-chain according to the XCM standard.
 
-One kind of message is `Dispatchable`. They could be thought of similarly to extrinsics sent to a relay chain: they also
-invoke exposed runtime entrypoints, they consume weight and require fees. The difference is that they originate from
-a parachain. Each parachain has a queue of dispatchables to be executed. There can be only so many dispatchables at a time.
+The XCM standard is a common vocabulary of messages. The XCM standard doesn't require a particular interpretation of
+a message. However, the parachains host (e.g. Polkadot) guarantees certain semantics for those.
+
+Moreover, while most XCM messages are handled by the on-chain XCM interpreter, some of the messages are special
+cased. Specifically, those messages can be checked during the acceptance criteria and thus invalid
+messages would lead to rejecting the candidate itself.
+
+One kind of such a message is `Xcm::Transact`. This upward message can be seen as a way for a parachain
+to execute arbitrary entrypoints on the relay-chain. `Xcm::Transact` messages resemble regular extrinsics with the exception that they
+originate from a parachain.
+
+The payload of `Xcm::Transact` messages is referred as to `Dispatchable`. When a candidate with such a message is enacted
+the dispatchables are put into a queue corresponding to the parachain. There can be only so many dispatchables in that queue at once.
 The weight that processing of the dispatchables can consume is limited by a preconfigured value. Therefore, it is possible
 that some dispatchables will be left for later blocks. To make the dispatching more fair, the queues are processed turn-by-turn
 in a round robin fashion.
 
-Upward messages are also used by a parachain to request opening and closing HRMP channels (HRMP will be described below).
-
-Other kinds of upward messages can be introduced in the future as well. Potential candidates are
-new validation code signalling, or other requests to the relay chain.
+The second category of special cased XCM messages are for horizontal messaging channel management,
+namely messages meant to request opening and closing HRMP channels (HRMP will be described below).
 
 ## Horizontal Message Passing
 
--- a/roadmap/implementers-guide/src/runtime/router.md
+++ b/roadmap/implementers-guide/src/runtime/router.md
@@ -278,6 +278,12 @@ the parachain executed the message.
         1. Set `limit_used_places` to `max_places`
         1. Set `limit_message_size` to `max_message_size`
         1. Set `limit_used_bytes` to `config.hrmp_channel_max_size`
+    1. Send a downward message to `recipient` notifying about an inbound HRMP channel request.
+        - The DM is sent using `queue_downward_message`.
+        - The DM is represented by the `HrmpNewChannelOpenRequest`  XCM message.
+            - `sender` is set to `origin`,
+            - `max_message_size` is set to `max_message_size`,
+            - `max_capacity` is set to `max_places`.
 * `hrmp_accept_open_channel(sender)`:
     1. Check that there is an existing request between (`sender`, `origin`) in `HrmpOpenChannelRequests`
         1. Check that it is not confirmed.
@@ -290,12 +296,23 @@ the parachain executed the message.
     1. Reserve the deposit for the `origin` according to `config.hrmp_recipient_deposit`
     1. For the request in `HrmpOpenChannelRequests` identified by `(sender, P)`, set `confirmed` flag to `true`.
     1. Increase `HrmpAcceptedChannelRequestCount` by 1 for `origin`.
+    1. Send a downward message to `sender` notifying that the channel request was accepted.
+        - The DM is sent using `queue_downward_message`.
+        - The DM is represented by the `HrmpChannelAccepted` XCM message.
+            - `recipient` is set to `origin`.
 * `hrmp_close_channel(ch)`:
     1. Check that `origin` is either `ch.sender` or `ch.recipient`
     1. Check that `HrmpChannels` for `ch` exists.
     1. Check that `ch` is not in the `HrmpCloseChannelRequests` set.
     1. If not already there, insert a new entry `Some(())` to `HrmpCloseChannelRequests` for `ch`
     and append `ch` to `HrmpCloseChannelRequestsList`.
+    1. Send a downward message to the opposite party notifying about the channel closing.
+        - The DM is sent using `queue_downward_message`.
+        - The DM is represented by the `HrmpChannelClosing` XCM message with:
+            - `initator` is set to `origin`,
+            - `sender` is set to `ch.sender`,
+            - `recipient` is set to `ch.recipient`.
+        - The opposite party is `ch.sender` if `origin` is `ch.recipient` and `ch.recipient` if `origin` is `ch.sender`.
 
 ## Session Change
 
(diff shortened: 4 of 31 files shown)


roadmap/implementers-guide/src/messaging.md
```diff
@@ -26,20 +26,28 @@ The downward message queue doesn't have a cap on its size and it is up to the re
...
 Upward Message Passing (UMP) is a mechanism responsible for delivering messages in the opposite direction:
-from a parachain up to the relay chain. Upward messages can serve different purposes and can be of different
- kinds.
+from a parachain up to the relay chain. Upward messages are essentially byte blobs. However, they are interpreted
+by the relay-chain according to the XCM standard.
 
-One kind of message is `Dispatchable`. They could be thought of similarly to extrinsics sent to a relay chain: they also
-invoke exposed runtime entrypoints, they consume weight and require fees. The difference is that they originate from
-a parachain. Each parachain has a queue of dispatchables to be executed. There can be only so many dispatchables at a time.
+XCM standard is a common vocabulary of messages. The XCM standard doesn't require a particular interpretation of
```
coriolinus, line 32: Suggest inserting the word "The" before "XCM" at the start of the line.

roadmap/implementers-guide/src/messaging.md
```diff
@@ -26,20 +26,28 @@ The downward message queue doesn't have a cap on its size and it is up to the re
...
+a message. However, the parachains host (e.g. Polkadot) guarantees certain semantics for those.
+
+Moreover, while most XCM messages handled by the on-chain XCM interpreter, some of the messages are special
```
coriolinus, line 35: Suggest adding the word "are" between "XCM messages" and "handled".

core-primitives/src/lib.rs
```diff
@@ -79,20 +79,19 @@ pub use sp_runtime::OpaqueExtrinsic as UncheckedExtrinsic;
...
-	/// was given with the transfer.
-	TransferInto(AccountId, Balance, Remark),
-	/// An opaque blob of data. The relay chain must somehow know how to form this so that the
-	/// destination parachain does something sensible.
-	///
-	/// NOTE: Be very careful not to allow users to place arbitrary size information in here.
-	Opaque(sp_std::vec::Vec<u8>),
-	/// XCMP message for the Parachain.
-	XCMPMessage(sp_std::vec::Vec<u8>),
+pub struct InboundDownwardMessage<BlockNumber = crate::BlockNumber> {
```
coriolinus, line 90: Is `InboundDownwardMessage` named in contrast to `DownwardMessage`, or to an `OutboundDownwardMessage` (presumably defined elsewhere and used on the relay chain)?

If possible, it would be good to change these names; I would have trouble remembering which of `InboundDownwardMessage` and `DownwardMessage` I wanted. Possibly rename `DownwardMessage` => `DownwardMessageContent` and `InboundDownwardMessage` => `DownwardMessage`.

runtime/parachains/src/configuration.rs
```diff
@@ -58,6 +58,13 @@ pub struct HostConfiguration<BlockNumber> {
 	pub thread_availability_period: BlockNumber,
 	/// The amount of blocks ahead to schedule parachains and parathreads.
 	pub scheduling_lookahead: u32,
+	/// The maximum size of a message that can be put in a downward message queue.
+	///
+	/// Since we require receiving at least one DMP message the obvious upper bound of the size is
+	/// the PoV size. Of course, there is a lot of other different things that a parachain may
+	/// decide to do with its PoV so this value in practice will be picked as a fraction of the PoV
+	/// size.
+	pub critical_downward_message_size: u32,
```
coriolinus, line 73: Suggest naming this `max_downward_message_size`; `critical` is not descriptive of the limit here.

runtime/parachains/src/inclusion.rs
```diff
@@ -155,6 +155,8 @@ decl_error! {
 		InternalError,
 		/// The downward message queue is not processed correctly.
 		IncorrectDownwardMessageHandling,
+		/// At least one upward message sent does not pass the acceptance criteria.
+		InvalidUpwardMessages,
```
coriolinus, line 159: I imagine it would be helpful for debugging if this identified which particular upward messages were invalid by index, like `InvalidUpwardMessages(Vec<u32>), // vector has indices of invalid messages`.

runtime/parachains/src/router/ump.rs
```diff
@@ -0,0 +1,59 @@
...
+
+impl<T: Trait> Module<T> {
+	/// Check that all the upward messages sent by a candidate pass the acceptance criteria. Returns
+	/// false, if any of the messages doesn't pass.
+	pub(crate) fn check_upward_messages(
+		config: &HostConfiguration<T::BlockNumber>,
+		para: ParaId,
+		upward_messages: &[UpwardMessage],
+	) -> bool {
+		drop(para);
```
coriolinus: I don't understand why this crate's functions drop the `ParaId` without using it. Just reserving the function signature for future use, maybe?

runtime/parachains/src/router/ump.rs
```diff
@@ -41,32 +60,582 @@ impl<T: Trait> Module<T> {
...
+	/// Removes the item under the cursor.
+	fn remove(&mut self) {
+		if self.needs_dispatch.is_empty() {
+			return;
+		}
+		let _ = self.needs_dispatch.remove(self.cur_idx);
+	}
+
+	/// Flushes the dispatcher state into the persistent storage.
+	fn flush<T: Trait>(self) {
```
coriolinus, line 303: Would it be useful to have a `Drop` impl which calls `self.flush()`? That's the behavior I expect from `Write` trait instances, at least.

roadmap/implementers-guide/src/runtime/router.md
```diff
@@ -277,6 +277,9 @@ the parachain executed the message.
         1. Set `limit_used_places` to `max_places`
         1. Set `limit_message_size` to `max_message_size`
         1. Set `limit_used_bytes` to `config.hrmp_channel_max_size`
+    1. Send a downward message to `recipient` notifying about an inbound HRMP channel request.
+        - The DM is sent using `queue_downward_message`.
+        - The DM is represented by an appropriate XCM message.
```
rphmeier, line 282: what is the appropriate XCM message? how is it determined?

is the language vague here to avoid replicating the entire XCM spec in the guide?

runtime/parachains/src/router/dmp.rs
```diff
@@ -0,0 +1,212 @@
...
+	pub fn queue_downward_message(para: ParaId, msg: DownwardMessage) {
+		let inbound = InboundDownwardMessage {
+			msg,
+			sent_at: <frame_system::Module<T>>::block_number(),
+		};
+
+		// obtain the new link in the MQC and update the head.
+		<Self as Store>::DownwardMessageQueueHeads::mutate(para, |head| {
+			let new_head =
+				BlakeTwo256::hash_of(&(*head, inbound.sent_at, T::Hashing::hash_of(&inbound.msg)));
```
rphmeier, line 59: Is the internal hash (`hash_of(&inbound.msg)`) to be able to traverse hash-chain links without having to provide (potentially large) message content?

runtime/parachains/src/router/hrmp.rs
```diff
@@ -62,3 +72,359 @@ pub struct HrmpChannel {
...
+	pub(super) fn process_hrmp_open_channel_requests(config: &HostConfiguration<T::BlockNumber>) {
+		let mut open_req_channels = <Self as Store>::HrmpOpenChannelRequestsList::get();
+		if open_req_channels.is_empty() {
+			return;
+		}
+
+		// iterate the vector starting from the end making our way to the beginning. This way we
+		// can leverate `swap_remove` to efficiently remove an item during iteration.
+		let mut idx = open_req_channels.len();
+		loop {
```
coriolinus, line 112: Why `loop` instead of `for (idx, channel_id) in open_req_channels.iter().enumerate().rev() { let channel_id = channel_id.clone(); ... }`? Manual index handling feels like an opportunity for bugs.

runtime/parachains/src/router/ump.rs
rphmeier: I assume filled in following commits, but I'm reading the same story as you :)

runtime/parachains/src/inclusion.rs
rphmeier: I believe this is a limitation of `decl_error!`.

runtime/parachains/src/router/ump.rs
coriolinus: Yeah, that was it.

runtime/parachains/src/router/ump.rs
```diff
@@ -15,15 +15,34 @@
...
 use super::{Trait, Module, Store};
-use crate::configuration::HostConfiguration;
+use crate::configuration::{self, HostConfiguration};
 use sp_std::prelude::*;
-use frame_support::{StorageMap, StorageValue, weights::Weight};
+use sp_std::collections::{btree_map::BTreeMap, vec_deque::VecDeque};
+use frame_support::{StorageMap, StorageValue, weights::Weight, traits::Get};
 use primitives::v1::{Id as ParaId, UpwardMessage};
 
+pub trait UmpSink {
```
rphmeier, line 24: some docs on this please :)

runtime/parachains/src/router/ump.rs
```diff
@@ -41,32 +60,582 @@ impl<T: Trait> Module<T> {
...
+		}
+		self.cur_idx = (self.cur_idx + 1) % self.needs_dispatch.len();
+	}
+
+	/// Removes the item under the cursor.
+	fn remove(&mut self) {
+		if self.needs_dispatch.is_empty() {
+			return;
+		}
+		let _ = self.needs_dispatch.remove(self.cur_idx);
```
rphmeier, line 299: If the last member of `needs_dispatch` is `remove`d, `self.cur_idx` will be out-of-bounds

xcm/src/v0/mod.rs
```diff
@@ -158,6 +158,51 @@ pub enum Xcm {
...
 	RelayedFrom { superorigin: MultiLocation, inner: Box<VersionedXcm> },
+
+	/// A message to notify about a new incoming HRMP channel. This message is meant to be sent by the
+	/// relay-chain to a para.
+	///
+	/// - `sender`: The sender in the to-be opened channel. Also, the initiator of the channel opening.
+	/// - `max_message_size`: The maximum size of a message proposed by the sender.
+	/// - `max_capacity`: The maximum number of messages that can be queued in the channel.
+	///
+	/// Safety: The message should originate immediatelly from the relay-chain.
```
rphmeier, line 169: "immediately" is the correct spelling (here and other `Hrmp` messages).

Although "directly" may be a more suitable word, if I understand the intended meaning correctly.

parachain/src/primitives.rs
```diff
@@ -232,7 +242,7 @@ pub struct ValidationParams {
 }
 
 /// The result of parachain validation.
-// TODO: egress and balance uploads (https://github.com/paritytech/polkadot/issues/220)
+// TODO: balance uploads (https://github.com/paritytech/polkadot/issues/220)
```
rphmeier, line 245: I think this can be removed now - the functionality of "balance uploads" is handled by upward messages and balances module, right?

pepyakin: Perhaps, I just don't know what is a balance upload

node/core/runtime-api/src/lib.rs
```diff
@@ -279,6 +281,23 @@ mod tests {
 			fn validator_discovery(ids: Vec<ValidatorId>) -> Vec<Option<AuthorityDiscoveryId>> {
 				vec![None; ids.len()]
 			}
+
+			// TODO: Cover these with tests
```
rphmeier, line 285: These TODOs will be removed later?

rphmeier (approved):
Looked through all commits so far! Which parts do you want to merge?

node/core/runtime-api/src/lib.rs
pepyakin: sure!

runtime/parachains/src/inclusion.rs
pepyakin: yup, that's right.

Also, as an aside. In runtime engineering we try to make the on-chain runtime code as slim as possible. That's why sometimes we tend to use less diagnostics (you might have noticed attempts at evading `Debug`).

It kinda makes sense, because most of the times, when we run a global consensus, it's not necessarily to bring detailed information. It's inefficient and eats execution time from this super constrained resource called block limit on one hand, and it's not 100% required on the other hand: one can always re-execute a block locally.

Now, it could've been more humane, like if we had something like  https://github.com/paritytech/substrate/issues/2082 or similar, we could potentially split these two cases and get two binaries. But that would require us some work, including to `decl_error` to leverage that.

runtime/parachains/src/router/dmp.rs
pepyakin: Yeah, so that the collator can provide the MQC proof to the latest (or perhaps nearest) MQC head beacon.

After all, one DMP message can be up to `{max, critical}_downward_message_size`, and I think it would be a good option to have ability to assign a value that is a big portion of a PoV: all downward messages are under control of the relay-chain anyway and more flexibility won't harm on this.

roadmap/implementers-guide/src/runtime/router.md
pepyakin: > is the language vague here to avoid replicating the entire XCM spec in the guide?

yeah, exactly...  OTOH that was written some time ago when there were no message and now at least I can state the name of the message.

runtime/parachains/src/router/ump.rs
pepyakin: While I understand that it's more idiomatic to lean on RAII generally, runtime code is exception, I'd say. The reason is, when the runtime is used in native mode, i.e. it's linked against the main binary as a normal rust crate, we use `panic=unwind` there to handle potential panics.

A big caveat with that, if this drop impl was invoked during panicking within runtime then that would lead to a double panic. And double panic leads to a process abort which is a disaster. Because of that, and because the code here is local and used only in a function with straghtforward control flow, I thought I'd go to an explicit function.

That said, I thought I had left a comment about this but apparently I didn't. I will leave one

runtime/parachains/src/configuration.rs
pepyakin: No objection here!

core-primitives/src/lib.rs
pepyakin: Ok, so first HRMP: I defined two structs for HRMP messages, for inbound and outbound. The inbound is from the pov of  the receiver and the outbound is from the pov of the sender.

At that time there were only `DownwardMessage`, which is an alias to `Vec<u8>`.

But then it turned out that I needed a specific view of downward messages from pov of the receiver, specifically the one that includes the block number. So because of that, I went with the `InboundDownwardMessage` because that's a very similar case. 

Now, I left `DownwardMessage` mostly because I just didn't think about this situation. I also don't think that this naming scheme is ideal. But I don't think that the suggested naming is better for the same reason. Specifically, if I am coding some stuff that works with the downward messages, what should I take: downward message or a downward message contents? Not clear without diving into docs reading the difference between those. While `InboundDownwardMessage` suggests that it's a downward message that is used in the context of receiving.

runtime/parachains/src/router/hrmp.rs
pepyakin: I actually tried to address this in the comment. The reason for the explicit `loop` is that we modify it while iterating on it.

runtime/parachains/src/router/ump.rs
pepyakin: but that's guarded just one line above?

pepyakin:
Good question, I think DMP is the most stable part so it can go first. But now I realized that perhaps landing all the data structures, plumbing and configuration would be more effective.

roadmap/implementers-guide/src/messaging.md
pepyakin: Done, thanks!

roadmap/implementers-guide/src/messaging.md
pepyakin: Done

core-primitives/src/lib.rs
pepyakin: Moving this discussion to https://github.com/paritytech/polkadot/pull/1859

runtime/parachains/src/router/ump.rs
rphmeier: Now that I read again, it is guarded by `peek`, so everything is OK.
I mean last as in the one at the back, not the only member.

Here is the scenario I am thinking:

1. `needs_dispatch.len() = n`, `cur_idx = n - 1`.
2. Call `remove`
3. `needs_dispatch.len() = n - 1`, `cur_idx = n - 1`.
4. `peek()` now returns `None` as `cur_idx` is out of bounds. The dispatch loop breaks and `flush` writes `StartWith = None`.
5. Resumes from index `0` next time.

runtime/parachains/src/router/ump.rs
pepyakin: Added, and realized that we still haven't figured out the limits on the `CandidateCommitments`

runtime/parachains/src/router/ump.rs
pepyakin: Well, then it's OK for some definition of OK, because yeah it doesn't blow up but the iteration would be finished prematurely.

pepyakin: I think it should be addressed in https://github.com/paritytech/polkadot/pull/1679/commits/041e379fc6560bfe4e59d1819ded55c5ee79fdd0

runtime/parachains/src/router/ump.rs
rphmeier: which limits?

pepyakin: basically, currently, at the moment `CandidateCommitments` is mostly of constant size with exception of messages. HRMP and UMP outbound messages inflate the candidate commitments. HRMP has some upper bound of size: a parachain can open a limited number of channels and the sizes of messages are limited as well. UMP has only a limit on number of messages and these messages can be of any size. 

Therefore, if I am not mistaken, this means that `CandidateCommitments` can be inflated as well. Not sure, which would be the next limit (perhaps networking? or do we have a special limit for `CandidateCommitments`?), but I think we are better off introducing a dedicated limit or limits, that would effectively limit the overall size taken in candidate commitments. 

Now, apart from the size concern there is also the weight.  But now after dekindification (#1702, also tackled in this PR), all the messages are byte blobs anyway, so UMP delegates this to a higher level. UMP dispatching aside, we need to weigh all this machinery that propels it. 

I had this in mind and I was planning to discuss this on the second call with you and Al, but that never happened and then I forgot until I was writing [these docs](https://github.com/paritytech/polkadot/pull/1679/commits/0d1835c7a2cef750828a14f2caf2a0793191c4bc#diff-65cfe3153bc3f3af755d35c872a99d790a940efd54223693a492f4bcf7b2a447R26).

roadmap/implementers-guide/src/runtime/router.md
pepyakin: should be updated now

runtime/parachains/src/router/ump.rs
pepyakin: moved this to issue https://github.com/paritytech/polkadot/issues/1869. Feel free to reply there

pepyakin:
Subsumed by https://github.com/paritytech/polkadot/pull/1900

PR #1689, opened 2020-09-08
Implementer's Guide: Allow the watermark to always land on the relay parent

### LINKED ISSUE #1671: Implementer's Guide: XCMP/HRMP watermark in case of no pending messages
There are really two options what we do here and in either case the solution would be trivial or at least close to it. Here I am just leaving breadcrumbs for motivation

Options:

1. Allow the watermark to stay the same - basically circumventing the rule of the watermark advancement - if there are no pending messages
1. Always allow the watermark to land on the relay parent block (in addition to the blocks found in the digest)

The first approach requires answering the question: if there are pending messages to a parachain. The second doesn't require any additional checks.

I am inclined towards the second approach

pepyakin:
Closes #1671 

The HRMP watermark can now land not only on an entry in the digest but also on the relay-parent block.

First, it fixes a problem that there is no way to satisfy the following two rules if there is no pending messages

1. The candidate needs to always move the watermark
1. However, the watermark can only land on a digest.

Implication of these changes is that the watermark now can point on a block where no messages were sent to the para (i.e. no entry in digest). Don't see why this could be a problem.

### DIFF
--- a/roadmap/implementers-guide/src/runtime/router.md
+++ b/roadmap/implementers-guide/src/runtime/router.md
@@ -185,7 +185,9 @@ Candidate Acceptance Function:
 * `check_hrmp_watermark(P: ParaId, new_hrmp_watermark)`:
   1. `new_hrmp_watermark` should be strictly greater than the value of `HrmpWatermarks` for `P` (if any).
   1. `new_hrmp_watermark` must not be greater than the context's block number.
-  1. in `HrmpChannelDigests` for `P` an entry with the block number equal to `new_hrmp_watermark` should exist.
+  1. `new_hrmp_watermark` should be either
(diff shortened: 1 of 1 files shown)


drahnr (approved):

pepyakin:
bot merge

pepyakin:
bot merge

burdges:
It's not impossible we'd want this to be the block before the relay parent or even further, and to forbid reading messages further ahead than this limit.  It'd address the pipelined version of the bug this OPR fixes:  You could be blocked because you needed to advance your watermark, but you cannot obtain them message yet.  We'll see what shakes out, but I wanted to caution that acting upon messages too fast could create problems.

PR #1690, opened 2020-09-08
Implementer's Guide: Limit the maximum size of a downward message

### LINKED ISSUE #1644: Guide: Limit the size of a DMP message
It should be non-controversial to claim that an individual downward message must be smaller than the size limit of PoV, given that downward messages are passed using PoV, so that at least one DM can be processed in a parablock.

That, of course, would be a bit overwhelming so perhaps the maximum size of a message should be only a fraction of the maximum PoV size.

pepyakin:
Closes #1644

### DIFF
--- a/roadmap/implementers-guide/src/runtime/router.md
+++ b/roadmap/implementers-guide/src/runtime/router.md
@@ -260,6 +260,7 @@ any of dispatchables return an error.
 Utility routines.
 
 `queue_downward_message(P: ParaId, M: DownwardMessage)`:
+    1. Check if the serialized size of `M` exceeds the `config.critical_downward_message_size`. If so, return an error.
     1. Wrap `M` into `InboundDownwardMessage` using the current block number for `sent_at`.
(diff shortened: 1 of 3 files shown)


roadmap/implementers-guide/src/runtime/router.md
```diff
@@ -260,6 +260,7 @@ any of dispatchables return an error.
 Utility routines.
 
 `queue_downward_message(P: ParaId, M: DownwardMessage)`:
+    1. Check if the serialized size of `M` exceeds the `config.critical_downward_message_size`. If so, return an error.
```
drahnr, line 263: What would be the consequence of this error? Silently dropping the message?

drahnr (approved):

roadmap/implementers-guide/src/runtime/router.md
pepyakin: This utility routine is supposed to be called only by the runtime, so basically the caller will be notified about the error. I'd expect that the caller should make sure that it doesn't try to send anything that will not fit into the queue.

rphmeier (approved):

pepyakin:
bot merge

PR #1691, opened 2020-09-09
Implementer's guide: Approval Voting Subsystem

### LINKED ISSUE #1601: Guide: Approval Voting Subsystem
This subsystem is responsible acting on an assignment to be a secondary checker of some particular candidate.

This means gathering the data necessary to re-execute the candidate via the Availability Reconstruction Subsystem (#1597), plugging it into Candidate Validation, and then either issuing an approval vote or initiating a dispute. 

Once the approval vote has been signed, it should be fed to the Approval Vote Distribution subsystem (#1603) to be distributed to other validators.

It also needs to make its way to the provisioner somehow since these votes need to make their way on-chain. It's not clear whether the approval voting subsystem or approval distribution subsystem should be responsible for that. It probably doesn't matter much, but we should choose one and be clear about it.

### LINKED ISSUE #1604: Guide: Approval Module
The Approval module is the runtime component responsible for tracking, on-chain, which blocks in the current chain (that observed by this fork of the runtime state) have had all candidates sufficiently approved by secondary checkers.

This module needs to accept all candidates pending approval from the inclusion module after inclusion. It then has two inherent types: Informing and Clearing. The Informing Inherent contains new assignment announcements by validators (generated by #1598 and distributed by #1600 ) and approval votes by validators (generated by #1601 and distributed by #1603). The Clearing Inherent triggers relay-chain checking logic to check sufficiency of approval for a relay-chain block. There may be more than one Clearing Inherent in a single block. The relay-chain block is invalid if any of the Clearing Inherents within the block do not refer to a relay-chain block which was pending approval and met the requirements for approval.

In practice, we will probably implement each of the inherents as their own module, as we have done with the Inclusion Inherent. 

One other thing that this module is in charge of is detecting no-shows. This needs to be done on-chain so the runtime logic is aware of how many validators need to approve. A no-show is a validator who has broadcasted an assignment but not an approval vote or dispute initiation within some time afterwards, which will be a non-zero multiple of the slot-time.

And lastly, this needs to account for the fact that the candidates for which it is expecting approval may belong to some prior session than the current one.

rphmeier:
Closes #1601 

This subsystem is responsible for 3 things:
  * Issuing and tracking assignments to participate in approval checks for different candidates in blocks. It cleanly handles forks, as the same candidate may appear in two different competing relay-chain blocks but the assignment selection criteria are based on the relay-chain block hash.
  * Issuing and tracking approval votes on different candidates in blocks. This involves invoking the availability recovery subsystem and the candidate validation subsystem
  * Providing an API to a custom GRANDPA voting rule that allows it to select the best block which has all candidates approved. This will allow us to have fast finality after approval checks complete.


roadmap/implementers-guide/src/node/approval/approval-voting.md
```diff
@@ -0,0 +1,234 @@
...
+  - `ApprovalNetworkingMessage::DistributeAssignment`
+  - `ApprovalNetworkingMessage::DistributeApproval`
+  - `RuntimeApiMessage::Request`
+  - `ChainApiMessage`
+  - `AvailabilityRecoveryMessage::Recover`
+  - `CandidateExecutionMessage::ValidateFromExhaustive`
+
+## Functionality
+
+The approval voting subsystem is responsible for casting votes and determining approval of candidates and as a result, blocks.
```
drahnr, line 20: Missing a verb here? `creates blocks.` or `approves blocks`?

roadmap/implementers-guide/src/node/approval/approval-voting.md
```diff
@@ -0,0 +1,234 @@
...
+  1. To provide an easy index from unfinalized blocks to candidates
+  1. To provide a lookup from candidate hash to approval status
+  1. To be easy to clear on start-up. What has happened while we were offline is unimportant.
+  1. To be fast to clear entries outdated by finality
+
+Structs:
+
+```rust
+struct TrancheEntry {
+    tranche: DelayTranche,
```
drahnr, line 36: The definition of delay tranche should be added here, rather than the disputes module:

> Assignment proceeds in loosely timed rounds called DelayTranches roughly 12 times faster than block production, in which validators send assignment notices until all candidates have enough checkers assigned. Assignment tracks when approval votes arrive too and assigns more checkers if some checkers run late.

roadmap/implementers-guide/src/node/approval/approval-voting.md
```diff
@@ -0,0 +1,234 @@
...
+  validator_index: ValidatorIndex,
+  triggered: bool,
+}
+
+struct ApprovalEntry {
+    tranches: Vec<TrancheEntry>, // sorted ascending by tranche number.
+    backing_group: GroupIndex,
+    // When the next wakeup for this entry should occur. This is either to
+    // check a no-show or to check if we need to broadcast an assignment.
+    next_wakeup: Tick,
```
drahnr, line 54: Should we not rather make this a push driven system with timeouts? That usually reduces complexity in the implementation.

roadmap/implementers-guide/src/node/approval/approval-voting.md
```diff
@@ -0,0 +1,234 @@
...
+    received_late_by: Duration,
+    // random bytes derived from the VRF submitted within the block by the block
+    // author as a credential and used as input to approval assignment criteria.
+    relay_vrf_story: [u8; 32],
+    // The candidates included as-of this block and the index of the core they are
+    // leaving. Sorted ascending by core index.
+    candidates: Vec<(CoreIndex, Hash)>,
+    // A bitfield where the i'th bit corresponds to the i'th candidate in `candidates`.
+    // The i'th bit is `true` iff the candidate has been approved in the context of
+    // this block. The block can be considered approved if the bitfield is equal to !0.
```
drahnr, line 82:
```suggestion
    // this block. The block can be considered approved if the bitfield is at least one bit set to `1`.
```

roadmap/implementers-guide/src/node/approval/approval-voting.md
```diff
@@ -0,0 +1,234 @@
...
+
+struct TrackerEntry 
+
+struct StoredBlockRange(BlockNumber, BlockNumber)
+```
+
+In the schema, we map
+
+```
+"StoredBlocks" => StoredBlockRange
```
drahnr, line 100: Do the `"` have a meaning?

roadmap/implementers-guide/src/node/approval/approval-voting.md
```diff
@@ -0,0 +1,234 @@
...
+  candidate_index: u32,
+}
+
+struct State {
+    earliest_session: SessionIndex,
+    session_info: Vec<SessionInfo>,
+    keystore: KeyStorePtr,
+    wakeups: BTreeMap<Tick, Vec<(Hash, Hash)>>, // Tick -> [(Relay Block, Candidate Hash)]
+    
+    // These are connected to each other.
```
drahnr, line 123: Would be nice to get a reasoning here, why we store rx and tx in the same struct.

roadmap/implementers-guide/src/node/approval/approval-voting.md
```diff
@@ -0,0 +1,234 @@
...
+    approval_vote_tx: mpsc::Sender<ApprovalVoteRequest>,
+    approval_vote_rx: mpsc::Receiver<ApprovalVoteRequest>,
+}
+```
+
+[`SessionInfo`](../../runtime/session_info.md)
+
+On start-up, we clear everything currently stored by the database. This is done by loading the `StoredBlockRange`, iterating through each block number, iterating through each block hash, and iterating through each candidate referenced by each block. Although this is O(n^3), we don't expect to have more than a few unfinalized blocks at any time and in extreme cases, a few thousand. The clearing operation should be relatively fast as a result.
```
coriolinus, line 131: It's not really `O(n^3)`, because the number of stored blocks in the range, the number of hashes for each block, and the number of candidates referenced per block all have differing cardinality. AFAICT these are entirely distinct, which makes the bound `O(n*o*p)`, and I (possibly naively) expect `o` and `p` to remain pretty small all the time.

roadmap/implementers-guide/src/node/approval/approval-voting.md
```diff
@@ -0,0 +1,234 @@
...
+
+Main loop:
+  * Each iteration, select over all of
+    * The next `Tick` in `wakeups`: trigger `wakeup_process` for each `(Hash, Hash)` pair scheduled under the `Tick` and then remove all entries under the `Tick`.
+    * The next message from the overseer: handle the message as described in the [Incoming Messages section](#incoming-messages)
+    * The next request from `approval_vote_rx`: handle with `issue_approval`
+
+### Incoming Messages
+
+On receiving an `OverseerSignal::BlockFinalized(h)`, we fetch the block number `b` of that block from the ChainApi subsystem. We update our `StoredBlockRange` to begin at `b+1`. Additionally, we remove all block entries and candidates referenced by them up to and including `b`. Lastly, we prune out all descendents of `h` transitively: when we remove a `BlockEntry` with number `b` that is not equal to `h`, we recursively delete all the `BlockEntry`s referenced as children. We remove the `block_assignments` entry for the block hash and if `block_assignments` is now empty, remove the `CandidateEntry`.
```
coriolinus, line 141: `h` is used before definition; I'm not sure what you mean there, precisely.

roadmap/implementers-guide/src/node/approval/approval-voting.md
```diff
@@ -0,0 +1,234 @@
...
+
+On start-up, we clear everything currently stored by the database. This is done by loading the `StoredBlockRange`, iterating through each block number, iterating through each block hash, and iterating through each candidate referenced by each block. Although this is O(n^3), we don't expect to have more than a few unfinalized blocks at any time and in extreme cases, a few thousand. The clearing operation should be relatively fast as a result.
+
+Main loop:
+  * Each iteration, select over all of
+    * The next `Tick` in `wakeups`: trigger `wakeup_process` for each `(Hash, Hash)` pair scheduled under the `Tick` and then remove all entries under the `Tick`.
+    * The next message from the overseer: handle the message as described in the [Incoming Messages section](#incoming-messages)
+    * The next request from `approval_vote_rx`: handle with `issue_approval`
+
+### Incoming Messages
```
coriolinus, line 139: I really like the detailed descriptions of what to do on each incoming message, but they tend to blend together. Might be worth adding h4 headers above each message's section.

roadmap/implementers-guide/src/node/approval/approval-voting.md
```diff
@@ -0,0 +1,234 @@
...
+    * If the cert kind is `RelayVRFDelay`, then we check if the VRF is valid for the validator's key with the input `block_entry.relay_vrf_story ++ cert.core_index.encode()` as described in [the approvals protocol section](../../protocol-approval.md#assignment-criteria). The cert can be ignored if the block did not cause inclusion of a candidate on that core index. Otherwise, this is a valid assignment for the included candidate. The delay tranche for the assignment is determined by reducing `(vrf.make_bytes().to_u64() % (session_info.n_delay_tranches + session_info.zeroth_delay_tranche_width)).saturating_sub(session_info.zeroth_delay_tranche_width)`.
+    * `import_checked_assignment`
+    * return the appropriate `VoteCheckResult` on the response channel.
+
+On receiving a `CheckAndImportApproval(indirect_approval_vote, response_channel)` message:
+  * Fetch the `BlockEntry` from the indirect approval vote's `block_hash`. If none, return `VoteCheckResult::Bad`.
+  * Fetch the `CandidateEntry` from the indirect approval vote's `candidate_index`. If the block did not trigger inclusion of enough candidates, return `VoteCheckResult::Bad`.
```
drahnr, line 165: We should consider adding more variants to `VoteCheckResult` to allow better logging with a `fn is_bad(&self)` convenience method.

roadmap/implementers-guide/src/node/approval/approval-voting.md
```diff
@@ -0,0 +1,234 @@
...
+  * Construct a `SignedApprovalVote` using the candidate hash and check against the validator's approval key, based on the session info of the block. If invalid or no such validator, return `VoteCheckResult::Bad`.
+  * Send `VoteCheckResult::Accepted`,
+  * `import_checked_approval(BlockEntry, CandidateEntry, ValidatorIndex)`
+
+On receiving an `ApprovedAncestor(Hash, BlockNumber, response_channel)`:
+  * Iterate over the ancestry of the hash all the way back to block number given, starting from the provided block hash.
+  * Keep track of an `all_approved_max: Option<Hash>`.
+  * For each block hash encountered, load the `BlockEntry` associated. If any are not found, return `None` on the response channel and conclude.
+  * If the block entry has `approval_bitfield == !0` and `all_approved_max == None`, set `all_approved_max = Some(current_hash)`.
```
drahnr, line 174: We should clearly define the notation if we want to use it (or maybe it is and I am just not aware).

roadmap/implementers-guide/src/node/approval/approval-voting.md
```diff
@@ -0,0 +1,234 @@
...
+  * If the block entry as `approval_bitfield != !0`, set `all_approved_max = None`.
+  * After iterating all ancestry, return `all_approved_max`.
+
+### Utility
+
+`import_checked_assignment`
```
drahnr, line 180: Could we make these h4 headlines too?

roadmap/implementers-guide/src/node/approval/approval-voting.md
drahnr: Reason is that the rx is used to `clone()` from.

roadmap/implementers-guide/src/runtime/session_info.md
```diff
@@ -0,0 +1,55 @@
...
+```rust
+struct SessionInfo {
+    // validators in canonical ordering.
+    validators: Vec<ValidatorId>,
+    // validators' authority discovery keys for the session in canonical ordering.
+    discovery_keys: Vec<DiscoveryId>,
+    // The assignment and approval keys for validators.
+    approval_keys: Vec<(AssignmentId, ApprovalId)>,
+    // validators in shuffled ordering
+    validator_groups: Vec<Vec<ValidatorIndex>>,
```
drahnr, line 18: what's the meaning of the index of the outer variant? what does it correspond to?

roadmap/implementers-guide/src/types/approval.md
```diff
@@ -0,0 +1,128 @@
+# Approval Types
```
drahnr, line 1: :heart: this section is so very badly needed!

roadmap/implementers-guide/src/runtime/session_info.md
coriolinus: I'm not certain, but AFAIK the outer index doesn't mean anything in particular; it's just a way to separate the various groups of sets of validators.

roadmap/implementers-guide/src/types/approval.md
```diff
@@ -0,0 +1,128 @@
...
+    // The index of the candidate in the list of candidates fully included as-of the block.
+    candidate_index: u32,
+    validator: ValidatorIndex,
+    signature: ApprovalSignature,
+}
+```
+
+## CheckedAssignmentCert
+
+An assignment cert which has been completely checked. This type should be declared in such a way as to be instantiable only when the checks have actually been done. Fields should be accessible via getters, not direct struct access.
```
drahnr, line 76: Elaborating on what `completely checked` entails would be nice.

drahnr (approved):
A few nits, a few minor elaboration requests, but overall this looks very nice!

roadmap/implementers-guide/src/node/approval/approval-voting.md
drahnr: I find it hard to justify introducing ticks (got to re-read it, I might have missed some details) - we are essentially implementing a coarse polling when futures and channels are available.

roadmap/implementers-guide/src/node/README.md
```diff
@@ -10,7 +10,14 @@ The architecture of the node-side behavior aims to embody the Rust principles of
 
 Many operations that need to be carried out involve the network, which is asynchronous. This asynchrony affects all core subsystems that rely on the network as well. The approach of hierarchical state machines is well-suited to this kind of environment.
 
-We introduce a hierarchy of state machines consisting of an overseer supervising subsystems, where Subsystems can contain their own internal hierarchy of jobs. This is elaborated on in the next section on Subsystems.
+We introduce
```
burdges, line 13: sentence?

roadmap/implementers-guide/src/node/approval/approval-voting.md
rphmeier: How would you do it instead?

rphmeier: The main thing I'm working around here is that we are dealing with these data:
  * Potentially many unfinalized blocks, forked with each other
  * Many candidates within those blocks, potentially with the same candidate appearing in multiple blocks
  * Assignments appearing for specific candidates in the context of particular blocks

For any particular `ApprovalEntry` we only want to poll the _first_ assignment that would time out as a no-show and not have a wakeup for each one.

rphmeier: I suppose we could rewrite each `ApprovalEntry` to be its own `Future`, but the other thing I'm juggling is that we want this stuff to be persisted to disk as well, and we should prefer to write from a single task

roadmap/implementers-guide/src/node/approval/approval-voting.md
apopiak: maybe add a comma like so?
"[...] approval of candidates and, as a result, blocks."

roadmap/implementers-guide/src/node/approval/approval-voting.md
drahnr: Writing from a single task is either using a [`FuturesUnordered`](https://docs.rs/futures/0.3.7/futures/stream/struct.FuturesUnordered.html) which then guarantees that we are on the same task assuming either chaining of futures or sync writes. We could also use a write queue if that is a concern.

Iiuc `ApprovalEntry` could impl `Future` iteself, which then lives in a `ApprovalEntrySet` which acts like a timerwheel and consume the `rx` end of `mpsc` for additional entries or implement `Sink` to consume additional `Entry` items.

roadmap/implementers-guide/src/node/approval/approval-voting.md
```diff
@@ -0,0 +1,247 @@
...
+
+This subsystem wraps a database which is used to store metadata about unfinalized blocks and the candidates within them. Candidates may appear in multiple blocks, and assignment criteria are chosen differently based on the hash of the block they appear in.
+
+## Database Schema
+
+The database schema is designed with the following goals in mind:
+  1. To provide an easy index from unfinalized blocks to candidates
+  1. To provide a lookup from candidate hash to approval status
+  1. To be easy to clear on start-up. What has happened while we were offline is unimportant.
+  1. To be fast to clear entries outdated by finality
```
burdges, line 32: It's maybe implicit when you talk about the unfinalized relay chain block here, but I needed maps between paraids and cores-when-included aka when declared available in preliminary draft: https://github.com/paritytech/polkadot/pull/1558/files#diff-e9db7842138e4616e3dd5824d7df2f9947e2dcc7570c5fc20d755ec6947ab98fR97-R112

roadmap/implementers-guide/src/node/approval/approval-voting.md
```diff
@@ -0,0 +1,247 @@
...
+}
+
+struct BlockEntry {
+    block_hash: Hash,
+    session: SessionIndex,
+    slot: SlotNumber,
+    received_late_by: Duration,
+    // random bytes derived from the VRF submitted within the block by the block
+    // author as a credential and used as input to approval assignment criteria.
+    relay_vrf_story: [u8; 32],
```
burdges, line 78: We'll eventually want `relay_equivocations_story` too https://github.com/paritytech/polkadot/pull/1558/files#diff-e9db7842138e4616e3dd5824d7df2f9947e2dcc7570c5fc20d755ec6947ab98fR193 but that's separate from the relay_vrf_story field in this `BlockEntry`.  It's fine if we just delay that for now.

roadmap/implementers-guide/src/node/approval/approval-voting.md
```diff
@@ -0,0 +1,247 @@
...
+  1. To be fast to clear entries outdated by finality
+
+Structs:
+
+```rust
+struct TrancheEntry {
+    tranche: DelayTranche,
+    // assigned validators who have not yet approved, and the instant we received
+    // their assignment.
+    assignments: Vec<(ValidatorIndex, Tick)>,
```
burdges, line 41: Good.  :)  It took me quite a while to figure out that we needed this `Tick`, which corresponds to `AssignmentSignature::recieved` in https://github.com/paritytech/polkadot/pull/1558/files#diff-16fadbdcb87a7ecb8febad865d8ca383f4c913f4bf272307bdd93b7a0a3ab26cR229

roadmap/implementers-guide/src/types/approval.md
```diff
@@ -0,0 +1,128 @@
...
+
+The public key of a keypair used by a validator for approval voting.
+
+## AssignmentId
+
+The private key of a keypair used by a validator for approval voting.
+
+## AssignmentCert
+
+An `AssignmentCert`, short for Assignment Certificate, is a piece of data provided by a validator to prove that they have been selected to perform secondary approval checks on an included candidate.
```
burdges, line 13: I'm actually okay with the name certificate here..  It's initially confusing since a certificate is normally a signature by one key on another key, but in this case it's a VRF by a node's own key that delegates them the authority to sign as an approval checker.  I'll let you know if I think of anything better, but let's keep the name cert for now.

roadmap/implementers-guide/src/types/approval.md
```diff
@@ -0,0 +1,128 @@
...
+        core_index: CoreIndex,
+    }
+}
+
+struct AssignmentCert {
+    // The criterion which is claimed to be met by this cert.
+    kind: AssignmentCertKind,
+    // The VRF showing the criterion is met.
+    vrf: VRFInOut,
+}
```
burdges, line 34: You've the `vrf_inout` living in two places here. 

I'd normally have a `vrf_preout` and `vrf_proof` together ala https://github.com/paritytech/polkadot/pull/1558/files#diff-16fadbdcb87a7ecb8febad865d8ca383f4c913f4bf272307bdd93b7a0a3ab26cR233 but a vrf_inout would live in the type either before signing or after verification ala https://github.com/paritytech/polkadot/pull/1558/files#diff-16fadbdcb87a7ecb8febad865d8ca383f4c913f4bf272307bdd93b7a0a3ab26cR161

I jumped through polymorphism hoops to unify the notion of an assignment before signing or after verification with the idea that I'd unify some data structures and make fewer mistakes.  It did not go as well as I'd envisioned though, so I ended up with the expected three final types largely separate:
- our assignment with `vrf_inout` computed to find delay tranche, but not yet signed because maybe never used
- actually signed assignments, and
- verified assignments by anyone, ala your `TrancheEntry`.

roadmap/implementers-guide/src/types/approval.md
```diff
@@ -0,0 +1,128 @@
...
+```rust
+struct IndirectSignedApprovalVote {
+    // A block hash where the candidate appears.
+    block_hash: Hash,
+    // The index of the candidate in the list of candidates fully included as-of the block.
+    candidate_index: u32,
+    validator: ValidatorIndex,
+    signature: ApprovalSignature,
+}
+```
```
burdges, line 72: Yes, the signature is always over `ApprovalVote`, but we handle it with `block_hash` and `candidate_index` sometimes.  Would we gossip in the `SignedApprovalVote` form?  We could use an abbreviation of `IndirectSignedApprovalVote` when placing on-chain perhaps, although maybe not initially?.

roadmap/implementers-guide/src/types/approval.md
```diff
@@ -0,0 +1,128 @@
...
+```rust
+struct ExecutionTimePair {
+    // The absolute time in milliseconds that the validator claims to have taken
+    // with the block.
+    absolute: u32,
+    // The validator's believed ratio in execution time to the average, expressed as a fixed-point
+    // 16-bit unsigned integer with 8 bits before and after the point.
+    ratio: FixedU16,
+}
+```
```
burdges, line 128: Yes exactly.  ref https://github.com/paritytech/polkadot/pull/1656

roadmap/implementers-guide/src/types/overseer-protocol.md
```diff
@@ -33,6 +35,63 @@ struct ActiveLeavesUpdate {
...
+	// This should occur primarily on startup.
+	Ignore,
+}
+
+enum ApprovalVotingMessage {
+	/// Check if the assignment is valid and can be accepted by our view of the protocol.
+	/// Should not be sent unless the block hash is known.
+	CheckAndImportAssignment(
+		Hash, 
+		AssignmentCert,
```
burdges, line 58: You've already checked the assignment cert here?  I'd expect the opposite

roadmap/implementers-guide/src/types/overseer-protocol.md
```diff
@@ -33,6 +35,63 @@ struct ActiveLeavesUpdate {
...
+		AssignmentCert, 
+		ValidatorIndex,
+		ResponseChannel<VoteCheckResult>,
+	),
+	/// Check if the approval vote is valid and can be accepted by our view of the
+	/// protocol.
+	///
+	/// Should not be sent unless the block hash within the indirect vote is known.
+	CheckAndImportApproval(
+		IndirectSignedApprovalVote,
```
burdges, line 67: Are we gossiping this form or the other?

roadmap/implementers-guide/src/node/approval/approval-voting.md
rphmeier: https://github.com/paritytech/polkadot/issues/1901

roadmap/implementers-guide/src/node/approval/approval-voting.md
rphmeier: We use `CandidateEvents` for that.

roadmap/implementers-guide/src/types/approval.md
rphmeier: Let me know what you come up with! I couldn't think of anything good either.

roadmap/implementers-guide/src/types/approval.md
rphmeier: I'd expect we gossip in the `Indirect` form, because it allows us to gauge politeness more easily. That's the main purpose of this struct, actually.

rphmeier: We don't have a good way of knowing if a peer has a candidate, except for knowing if they have the block that caused inclusion of the candidate.

roadmap/implementers-guide/src/types/approval.md
rphmeier: This struct isn't yet used, but we'll get around to it at some point.

roadmap/implementers-guide/src/types/overseer-protocol.md
rphmeier: We haven't checked it at this point, no.

roadmap/implementers-guide/src/types/overseer-protocol.md
rphmeier: Probably gossiping this form, but that might change as I work on the network protocol.

roadmap/implementers-guide/src/node/approval/approval-voting.md
rphmeier: No matter what, we still need the `Tick` because that's our basic unit of time for this protocol. But I will explore different architectures. I still think that what's laid out in this doc sounds easiest to implement from an ownership perspective.

roadmap/implementers-guide/src/types/approval.md
rphmeier: Can you clarify what you would expect to have for the _unverified_ type, which is this `AssignmentCert`?

burdges: An unverified would contain a `VRFPreOut` and `VRFProof` because although you could compute the `VRFInOut` earlier it'll make you code robust if you never compute the `VRFInOut` until you call `vrf_verify`/`dleq_verifiy`. 

An assignment of our own that is not yet announced would contain a `VRFInOut` for which we did not yet bother to compute the `VRFProof`.  I'd originally meant to unify this into our own unannounced being handled in parallel with validated ones from others, but that did not go perfectly.

I'm unsure how you want to buffer the unsent gossip messages.  I ended up holding them in the same data structures as things containing both a `VRFProof` and a `VRFInOut`, from which one can easily compute the `VRFPreOut` required in gossip.  That kinda centralized all the timing/tranche accounting, but I suspect you guys have separate modules for such things.

roadmap/implementers-guide/src/node/approval/approval-voting.md
```diff
@@ -0,0 +1,247 @@
...
+  * Each iteration, select over all of
+    * The next `Tick` in `wakeups`: trigger `wakeup_process` for each `(Hash, Hash)` pair scheduled under the `Tick` and then remove all entries under the `Tick`.
+    * The next message from the overseer: handle the message as described in the [Incoming Messages section](#incoming-messages)
+    * The next request from `approval_vote_rx`: handle with `issue_approval`
+
+### Incoming Messages
+
+#### `OverseerSignal::BlockFinalized`
+
+On receiving an `OverseerSignal::BlockFinalized(h)`, we fetch the block number `b` of that block from the ChainApi subsystem. We update our `StoredBlockRange` to begin at `b+1`. Additionally, we remove all block entries and candidates referenced by them up to and including `b`. Lastly, we prune out all descendents of `h` transitively: when we remove a `BlockEntry` with number `b` that is not equal to `h`, we recursively delete all the `BlockEntry`s referenced as children. We remove the `block_assignments` entry for the block hash and if `block_assignments` is now empty, remove the `CandidateEntry`.
```
burdges, line 145: We just need to ensure anything we want remembered makes it on-chain by this point, but I think that's done here somewhere.

rphmeier: haven't done the on-chain reporting stuff yet. Disputes will track a DB of all votes we've seen by validators.

burdges: I'd agree disputes provides a better venue for tracking this, thanks.

roadmap/implementers-guide/src/types/approval.md
burdges: Also, I'd envisioned using the VRF as the gossip message signature, which saves doing another signature, but complicates the layers since we'd verify the VRF even to know if the message makes sense.  We could avoid this verification triggering any chain historical lookups by incluidng the `VRFPreOut` along with the message, and then validating that later.  It's too much code subtelty to do this maybe..

roadmap/implementers-guide/src/node/approval/approval-voting.md
```diff
@@ -0,0 +1,247 @@
...
+
+#### `tranches_to_approve(approval_entry) -> tranches`
+  * Determine the amount of tranches `n_tranches` our view of the protocol requires of this approval entry
+    * First, take tranches until we have at least `session_info.needed_approvals`. Call the number of tranches taken `k`
+    * Then, count no-shows in tranches `0..k`. For each no-show, we require another checker. Take new tranches until each no-show is covered, so now we've taken `l` tranches. e.g. if there are 2 no-shows, we might only need to take 1 additional tranche with >= 2 assignments. Or we might need to take 3 tranches, where one is empty and the other two have 1 assignment each.
+    * Count no-shows in tranches `k..l` and for each of those, take tranches until all no-shows are covered. Repeat so on until either
+      * We run out of tranches to take, having not received any assignments past a certain point. In this case we set `n_tranches` to a special value `ALL` which indicates that new assignments are needed.
+      * All no-shows are covered. Set `n_tranches` to the number of tranches taken
+    * return `n_tranches`
```
burdges, line 213: You've written the algorithm as I defined it, but we should probably change this to add one tranche per no-show.

roadmap/implementers-guide/src/node/approval/approval-voting.md
```diff
@@ -0,0 +1,247 @@
...
+  * For each block hash encountered, load the `BlockEntry` associated. If any are not found, return `None` on the response channel and conclude.
+  * If the block entry's `approval_bitfield` has all bits set to 1 and `all_approved_max == None`, set `all_approved_max = Some(current_hash)`.
+  * If the block entry's `approval_bitfield` has any 0 bits, set `all_approved_max = None`.
+  * After iterating all ancestry, return `all_approved_max`.
+
+### Utility
+
+#### `import_checked_assignment`
+  * Load the candidate in question and access the `approval_entry` for the block hash the cert references.
+  * Ensure the validator index is not part of the backing group for the candidate.
```
burdges, line 195: I suppose backers vote with their original backing announcement message since we do not distinguish the first backer.

roadmap/implementers-guide/src/node/approval/approval-voting.md
```diff
@@ -0,0 +1,247 @@
...
+    * Then, count no-shows in tranches `0..k`. For each no-show, we require another checker. Take new tranches until each no-show is covered, so now we've taken `l` tranches. e.g. if there are 2 no-shows, we might only need to take 1 additional tranche with >= 2 assignments. Or we might need to take 3 tranches, where one is empty and the other two have 1 assignment each.
+    * Count no-shows in tranches `k..l` and for each of those, take tranches until all no-shows are covered. Repeat so on until either
+      * We run out of tranches to take, having not received any assignments past a certain point. In this case we set `n_tranches` to a special value `ALL` which indicates that new assignments are needed.
+      * All no-shows are covered. Set `n_tranches` to the number of tranches taken
+    * return `n_tranches`
+
+#### `check_approval(block_entry, approval_entry, n_tranches) -> bool`
+  * If `n_tranches` is ALL, return false
+  * Otherwise, if all validators in `n_tranches` have approved, return `true`. If any validator in these tranches has not yet approved but is not yet considered a no-show, return `false`.
```
burdges, line 217: I needed `tranches_to_approve` and `check_approval` to be one method because `check_approval` knows about no-shows.

PR #1697, opened 2020-09-10
Prepare Polkadot to be used by Cumulus

bkchr:
This begins to make Polkadot usable from Cumulus.

### DIFF
--- a/node/service/src/lib.rs
+++ b/node/service/src/lib.rs
@@ -22,13 +22,12 @@ mod client;
 
 use std::sync::Arc;
 use std::time::Duration;
-use polkadot_primitives::v1::{AccountId, Nonce, Balance};
-use service::{error::Error as ServiceError};
+use service::{error::Error as ServiceError, RpcHandlers};
 use grandpa::{self, FinalityProofProvider as GrandpaFinalityProofProvider};
 use sc_executor::native_executor_instance;
 use log::info;
 use sp_blockchain::HeaderBackend;
-use polkadot_overseer::{self as overseer, AllSubsystems, BlockInfo, Overseer, OverseerHandler};
+use polkadot_overseer::{AllSubsystems, BlockInfo, Overseer, OverseerHandler};
 use polkadot_subsystem::DummySubsystem;
 use polkadot_node_core_proposer::ProposerFactory;
 use sp_trie::PrefixedMemoryDB;
@@ -55,7 +54,7 @@ pub use polkadot_runtime;
 pub use kusama_runtime;
 pub use westend_runtime;
 use prometheus_endpoint::Registry;
-pub use self::client::PolkadotClient;
+pub use self::client::{AbstractClient, Client, RuntimeApiCollection};
 
 native_executor_instance!(
 	pub PolkadotExecutor,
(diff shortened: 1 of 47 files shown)


parachain/src/primitives.rs
```diff
@@ -225,21 +226,102 @@ pub struct UpwardMessage {
...
+/// relay-chain, secondary checkers can rely on the invariant that the relay-chain
+/// only includes para-blocks for which these checks have already been done. As such,
+/// there is no need for the validation data used to inform validators and collators about
+/// the checks the relay-chain will perform to be persisted by the availability system.
+/// Nevertheless, we expose it so the backing validators can validate the outputs of a
+/// candidate before voting to submit it to the relay-chain and so collators can
+/// collate candidates that satisfy the criteria implied these transient validation data.
+#[derive(PartialEq, Eq, Clone, Encode, Decode)]
+#[cfg_attr(feature = "std", derive(Debug, Default))]
+pub struct ValidationData<N = RelayChainBlockNumber> {
```
bkchr, line 259: @rphmeier Essentially I will need all this data when validating the parachain PovBlock, so I copied that here. You think that this is okay? Or should I create some new struct for this? (Which would essentially contain all this data anyway).

node/subsystem/src/messages.rs
```diff
@@ -123,12 +123,11 @@ pub enum CandidateValidationMessage {
...
 	///
-	/// Explicitly provide the `PersistedValidationData` and `ValidationCode` so this can do full
-	/// validation without needing to access the state of the relay-chain. Optionally provide the
-	/// `TransientValidationData` for further checks on the outputs.
+	/// Explicitly provide the `PersistedValidationData`, `TransientValidationData` and `ValidationCode`
+	/// so this can do full validation without needing to access the state of the relay-chain.
 	ValidateFromExhaustive(
 		PersistedValidationData,
-		Option<TransientValidationData>,
+		TransientValidationData,
```
rphmeier, line 130: The transient validation data is by definition transient. It literally won't be guaranteed available when this is called. That is why it is an `Option`.

Why does Cumulus need this?

node/subsystem/src/messages.rs
```diff
@@ -123,12 +123,11 @@ pub enum CandidateValidationMessage {
 	),
 	/// Validate a candidate with provided, exhaustive parameters for validation.
 	///
-	/// Explicitly provide the `PersistedValidationData` and `ValidationCode` so this can do full
-	/// validation without needing to access the state of the relay-chain. Optionally provide the
-	/// `TransientValidationData` for further checks on the outputs.
+	/// Explicitly provide the `PersistedValidationData`, `TransientValidationData` and `ValidationCode`
+	/// so this can do full validation without needing to access the state of the relay-chain.
```
rphmeier, line 127: This docs change doesn't make sense, because the only way to access `TransientValidationData` is to access the state of the relay-chain.

parachain/src/primitives.rs
rphmeier: You should not need the `TransientValidationData` within the parachain runtime, at least not directly.

parachain/src/primitives.rs
```diff
@@ -225,21 +226,102 @@ pub struct UpwardMessage {
...
 	pub block_data: BlockData,
-	/// The current relay-chain block number.
-	pub relay_chain_height: RelayChainBlockNumber,
-	/// The list of MQC heads for the inbound HRMP channels paired with the sender para ids. This
-	/// vector is sorted ascending by the para id and doesn't contain multiple entries with the same
-	/// sender.
-	pub hrmp_mqc_heads: Vec<(Id, Hash)>,
+	/// The validation data that was also available to the collator when
+	/// building the PoVBlock.
+	pub validation_data: ValidationData,
```
rphmeier, line 324: I don't believe all of this info should go into ground truth.

rphmeier (changes requested):
This completely erodes the distinction between `Persisted` and `Transient` validation data and alters the abstract parachain execution model without a clear reason.

node/service/src/lib.rs
```diff
@@ -310,13 +275,14 @@ fn real_overseer<S: SpawnNamed>(
...
-	_authority_discovery_enabled: bool,
-	_slot_duration: u64,
+	authority_discovery_enabled: bool,
 	grandpa_pause: Option<(u32, u32)>,
 ) -> Result<(
 	TaskManager,
 	Arc<FullClient<RuntimeApi, Executor>>,
+	Arc<sc_network::NetworkService<Block, <Block as BlockT>::Hash>>,
+	RpcHandlers,
+	OverseerHandler,
```
rphmeier, line 285: :D

runtime/kusama/src/lib.rs
```diff
@@ -1044,55 +1045,43 @@ sp_api::impl_runtime_apis! {
...
-			// It doesn't matter, but we shouldn't return a zero-sized vector here.
-			// As there are no parachains
-			Session::validators()
-				.into_iter()
-				.map(|k| k.using_encoded(|s| Decode::decode(&mut &s[..]))
-					.expect("correct size and raw-bytes; qed"))
-				.collect()
+	impl primitives::v1::ParachainHost<Block, Hash, BlockNumber> for Runtime {
+		fn validators() -> Vec<ValidatorId> {
+			unimplemented!()
```
rphmeier, line 1050: I'm kind of worried what will happen if this is pushed to the chain and we have subsystems that query this runtime API. I think those subsystems may just fail and the node will go down.

rphmeier (changes requested):
Looks great except I'm worried about the `unimplemented!()` in the runtime.

bkchr:
> Looks great except I'm worried about the `unimplemented!()` in the runtime.

I have done that, but now I'm more worried than you :P There will be 100% a timeframe where the new service will need to work with the old runtimes and if that is just for syncing. The services should **not** stop or print tons of data. We already had this with the availability system which printed a lot of data because it could not find required data and that is really bad for node operators, because they get spammed with useless warnings etc.

rphmeier (approved):

PR #1714, opened 2020-09-14
Add a note about memory requirements

pepyakin:
Add a note about memory requirements for performing `cargo build --release`

### DIFF
--- a/README.md
+++ b/README.md
@@ -100,6 +100,8 @@ git checkout <latest tagged release>
 cargo build --release
 ```
 
+Note that compilation can require several GiB of memory. Please make sure that the machine has that amount of physical memory or swap has enough capacity (keep in mind that if builds hit swap it tends to be very slow).
+
 ## Networks
 
 This repo supports runtimes for Polkadot, Kusama, and Westend.


ordian (approved):
I would add some examples of errors that can occur during build if we're out of memory. Or links to relevant issues.

pepyakin:
Agreed, the error messages could go to a troubleshooting section. Not sure about the issues though, do you mean issues of other people who experienced the same?

Anyway, I think those can be addressed in follow-ups and this particular PR improves the status quo which warrants merging as is.

pepyakin:
bot merge

README.md
```diff
@@ -100,6 +100,8 @@ git checkout <latest tagged release>
 cargo build --release
 ```
 
+Note that compilation can require several GiB of memory. Please make sure that the machine has that amount of physical memory or swap has enough capacity (keep in mind that if builds hit swap it tends to be very slow).
```
bkchr, line 103: You should have said that more than 4GB  are required :P 

But people will not read this anyway and will continue to open prs xD

pepyakin: Ah, 4GiB is also what I had in the back of my mind. I'll file another PR 

> But people will not read this anyway and will continue to open issues xD

Well, maybe, but at least those who reads this can save themselves a few back and forths

PR #1720, opened 2020-09-14
Update Substrate and bump version to 0.8.24

tomaka:
Ran `cargo update -p sp-io` and updated all versions from 0.8.23 to 0.8.24.

In terms of the network especially, it would be great if we could release a new version in order for our changes to be deployed.
I'm not exactly sure how to start a client release process, but this PR is probably a good first step.


s3krit (approved):

tomaka:
bot merge

PR #1726, opened 2020-09-16
Skeleton of the Router module

pepyakin:
Closes https://github.com/paritytech/polkadot/issues/1674

The purpose of this commit is to establish a foundation of the router module, it paves the way for the implementation of other features.

This commit is split from https://github.com/paritytech/polkadot/pull/1679

### DIFF
--- a/runtime/parachains/src/router.rs
+++ b/runtime/parachains/src/router.rs
@@ -1,0 +1,80 @@
+// Copyright 2020 Parity Technologies (UK) Ltd.
+// This file is part of Polkadot.
+
+// Polkadot is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+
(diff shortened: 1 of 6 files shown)


runtime/parachains/src/router.rs
```diff
@@ -0,0 +1,80 @@
...
+			if let Err(i) = v.binary_search(&id) {
+				v.insert(i, id);
+			}
+		});
+	}
+}
+
+#[cfg(test)]
+mod tests {
+}
```
pepyakin, line 80: This is used by the following commits

ordian (approved):

pepyakin:
We believe that `gitlab-check-runtime` is a false-positive and is triggered because of the change in the `paras_sudo_wrapper` pallet.

Practically this pallet is not used anywhere except rococo. It seems that rococo is explictly not including into the check so bumping the version there won't help.

Thus, I am going to go ahead and merge this one

pepyakin:
bot merge

PR #1733, opened 2020-09-18
Test helper to register a parachain

cecton:
Following comment https://github.com/paritytech/cumulus/pull/220#discussion_r490782885

### DIFF
--- a/node/test-service/src/lib.rs
+++ b/node/test-service/src/lib.rs
@@ -318,9 +318,7 @@ where
 			)),
 		)));
 
-		self.call_function(function, Sr25519Keyring::Alice).await?;
-
-		Ok(())
+		self.call_function(function, Sr25519Keyring::Alice).await.map(drop)
 	}
 }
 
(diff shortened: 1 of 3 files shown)


node/test-service/src/lib.rs
```diff
@@ -295,9 +298,33 @@ where
...
+					scheduling: Scheduling::Always,
+				},
+				binary,
+				head_data,
+			)),
+		)));
+
+		self.call_function(function, Sr25519Keyring::Alice).await?;
+
+		Ok(())
```
bkchr, line 323:
```suggestion
		self.call_function(function, Sr25519Keyring::Alice).await
```

bkchr (approved):

node/test-service/src/lib.rs
cecton: The result is different. This function returns `()` because the return value is not very useful. (`call_function` returns the RPC result which could (?) be potentially useful)

bkchr: Than do a `map`

cecton: :thinking: not convinced

bkchr: Why?

cecton: Different people have different taste. I find the other one more readable. But yours is a bit more expliclit which is good too

PR #1735, opened 2020-09-20
Companion PR for #7136 (WeightInfo for Session / Offences)

shawntabrizi:
This is a simple PR that removes the `WeightInfo` trade for pallet_offences (as the weight information should come from the different implementations of `OnOffence`.

paritytech/substrate#7136

### DIFF
--- a/runtime/polkadot/tests/weights.rs
+++ b/runtime/polkadot/tests/weights.rs
@@ -33,7 +33,6 @@ use polkadot_runtime::{self, Runtime};
 use runtime_common::MaximumBlockWeight;
 
 use pallet_elections_phragmen::Call as PhragmenCall;
-use pallet_session::Call as SessionCall;
 use frame_system::Call as SystemCall;
 
 type DbWeight = <Runtime as frame_system::Trait>::DbWeight;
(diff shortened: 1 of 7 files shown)


runtime/polkadot/tests/weights.rs
```diff
@@ -57,31 +56,6 @@ fn weight_of_system_set_code_is_correct() {
 	assert_eq!(weight, expected_weight);
 }
 
-#[test]
```
shawntabrizi, line 60: We remove these tests because the weights no longer come from downstream Substrate, but we define them locally in the Repo.

ordian (approved):

andresilva (approved):

expenses (approved):

andresilva (commented):
Shouldn't we be defining `WeightInfo` for `pallet_session` in this PR?

thiolliere:
> Shouldn't we be defining WeightInfo for pallet_session in this PR?

At least it should be done before next release

expenses:
> Shouldn't we be defining `WeightInfo` for `pallet_session` in this PR?

@shawntabrizi :grimacing: lets do this and get this merged, please!

shawntabrizi:
It will be done all in one final PR to avoid back to back merge conflicts. Even if we dont, the values we get for substrate would be fine anyway. (but it will be done)

shawntabrizi:
bot merge

PR #1736, opened 2020-09-21
Companion for 7155 (WeightInfo for Babe and Grandpa)

shawntabrizi:
This simply adds the `WeightInfo` associated type to the Babe and Grandpa traits.

It uses the default values from Substrate which are still custom written.

paritytech/substrate#7155

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1429,7 +1429,7 @@ checksum = "3f9eec918d3f24069decb9af1554cad7c880e2da24a9afd88aca000531ab82c1"
 [[package]]
 name = "fork-tree"
 version = "2.0.0-rc6"
-source = "git+https://github.com/paritytech/substrate#8f19521d3ac1487afa7f1ceb2c64627faf7b9102"
+source = "git+https://github.com/paritytech/substrate#33562b122277f0d49e8864725ddc1dd1161be151"
 dependencies = [
  "parity-scale-codec",
 ]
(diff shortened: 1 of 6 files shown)


thiolliere:
I feels slightly afraid of using `type WeightInfo = ();` in polkadot, on the substrate side nothing says that those default weight are used in production.
Maybe we could add a comment in substrate that default implementation is used in production or otherwise we can copy them here again. And when automatic weight implementation is done, manual weight implementation must be checked.

I'm not sure what is best, anyway I'm ok to use `= ()` in polkadot

andresilva (approved):

andresilva:
I agree with @thiolliere that this could be a bit dangerous. I think just copying the weights from substrate here would probably be a sensible thing to do, or maybe keep the same weights in substrate but implemented for some struct rather than `()`. Either way I think @shawntabrizi intends to deal with all custom weights in a follow-up PR so this can be addressed there.

thiolliere (approved):
yes I'm ok to address it in follow-up PR

shawntabrizi:
What is the difference between using `= ()` and `= SomeStruct`?

In general, these weights were hardcoded into substrate before. The way it is written, nothing about the behavior of how these weights would update has changed.

andresilva:
There is no pratical difference, but maybe `= SomeStruct` could communicate better that this is used in Polkadot and that these weights are critical 🤷 I don't really have a strict opinion on this, as you said these weights were already hardcoded in Substrate in the first place and could already be changed easily there without it getting noticed in Polkadot. But for example we added some tests for weights in Polkadot to make sure that some invariants were not broken (i.e. if some weight changed in Substrate we'd be aware of it in Polkadot).

kianenigma (approved):

PR #1737, opened 2020-09-21
Bump jsonrpc-core to v15

maciejhirsz:
Depends on https://github.com/paritytech/substrate/pull/7157

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1429,7 +1429,7 @@ checksum = "3f9eec918d3f24069decb9af1554cad7c880e2da24a9afd88aca000531ab82c1"
 [[package]]
 name = "fork-tree"
 version = "2.0.0-rc6"
-source = "git+https://github.com/paritytech/substrate#16474ee9ed8f75ad578f4539ae4fc016ecf8b3d1"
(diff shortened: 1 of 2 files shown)


Cargo.lock
```diff
@@ -2414,7 +2414,7 @@ checksum = "ecbdaacc17243168d9d1fa6b2bd7556a27e1e60a621d8a2a6e590ae2b145d158"
 dependencies = [
  "failure",
  "futures 0.1.29",
- "jsonrpc-core",
+ "jsonrpc-core 14.2.0",
```
gnunicorn, line 2417: Why does this still contain `14.2.0` versions? We updated all jsonrpc to `15.0`, right? By here some dependencies are specified as still bound to `jsonrpc-core 14.2.0`, is that intentional?

gnunicorn (approved):
But some version mixing I'd like to see explained.

Cargo.lock
maciejhirsz: Ah, I commited this after building with current substrate master, that's why it's there.

gnunicorn:
bot merge

ordian (approved):

PR #1749, opened 2020-09-24
Parachains: Introduce a dummy module to include the Origin.

pepyakin:
That's because `construct_runtime` doesn't support a free-standing Origin. See the comment in parachains/src/origin.rs

Split off from #1679

### DIFF
--- a/runtime/parachains/src/origin.rs
+++ b/runtime/parachains/src/origin.rs
@@ -1,0 +1,54 @@
+// Copyright 2020 Parity Technologies (UK) Ltd.
+// This file is part of Polkadot.
+
+// Polkadot is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+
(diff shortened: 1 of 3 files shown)


montekki (approved):

runtime/parachains/src/origin.rs
```diff
@@ -0,0 +1,54 @@
...
+}
+
+/// The origin module.
+pub trait Trait: frame_system::Trait {}
+
+frame_support::decl_module! {
+	/// There is no way to register an origin type in `construct_runtime` without a pallet the origin
+	/// belongs to.
+	///
+	/// This module fulfills only the single purpose of housing the `Origin` in `construct_runtime`.
```
bkchr, line 50: Sounds like you want to open an issue in Substrate :D

bkchr (approved):

montekki:
bot merge

runtime/parachains/src/origin.rs
pepyakin: Thanks for the suggestion, I left an issue https://github.com/paritytech/substrate/issues/7203

PR #1754, opened 2020-09-25
Automation for new release process

s3krit:
This PR adds the necessary automation for smoothly transitioning to a git-flow-like model for releases. Initial discussion of switching to this model can be found [here](https://forum.parity.io/t/polkadot-release-model-workflow/473). Releases using this model will work as follows:
0. The current release is always the most recent commit of the `release` branch.
1. A new release-candidate branch is forked, named like `v0.8.25`. This automatically tags that new branch with `v0.8.25-rc1` - this is the first release candidate. Subsequent pushes to this branch automatically create a new release candidate.
2. Testing and QA is performed on the release candidate, following a release checklist issue that is automatically created
3. Once happy with the release candidate, the branch is merged to the `release` branch, which constitutes a final release of that version

To achieve this, the following has been produced:
- A Github workflow that will detect pushes to release-candidate branches (named something like `v1.2.3`) and tag those branches with `v1.2.3-rc1`, with the rc incrementing on every subsequent push
- Automatic creation of a Github issue that tracks the QA performed during the release process, adapted from the current RELEASE.md
- [WIP] A shell script for cutting the release from the release-candidate branch - needs more safety checks added
- Documentation for how to perform a release from start to finish

Notes:
- There's a 4 month old branch already called `release` that should be deleted, and the new one created + protected such that only core devs can push to it.
- We need to add a new token 'RELEASE_BRANCH_TOKEN' with only the `public_repo` oauth scope in order to get around [this](https://docs.github.com/en/free-pro-team@latest/actions/reference/events-that-trigger-workflows#triggering-new-workflows-using-a-personal-access-token) issue.

TODO:
- [x]  Detect the last published release from Github rather than relying on git tags
- [x] Change rc branch name to 'release-v1.2.3' to avoid confusion with the final release tag


TriplEight (approved):
GJ!

shawntabrizi (approved):
Looks reasonable.

How/when can we try it out?

s3krit:
> How/when can we try it out?

Basically as soon as it's merged - I plan to use this process on the very next release after merging.

scripts/github/generate_release_text.rb
```diff
@@ -19,12 +19,13 @@
...
-# get last polkadot version. Use handy Gem::Version for sorting by version
+# get last full polkadot version. Use handy Gem::Version for sorting by version
+# Note: This means ^v1.2.3$, explicitly ignoring -rc versions
 last_version = pg
               .tags
               .map(&:name)
-              .grep(/^v\d+\.\d+\.\d+.*$/)
-              .sort_by { |v| Gem::Version.new(v.slice(1...)) }[-2]
+              .grep(/^v\d+\.\d+\.\d+$/)
+              .max_by { |v| Gem::Version.new(v.slice(1...)) }
```
s3krit, line 28: TODO: Replace this with a snippet to detect the last published release rather than relying on git tags

gabreal (approved):
looks very good

gabreal:
one thing that was mentioned that would be good was to make the release binaries available for download without authentication. and eventually include the link to the binary in the notification message. this way it could be used directly in the deployments which are needed for testing and qa.

TriplEight:
@gabreal  can easily be done by adding a publish to s3 job after each build job. Should I?

s3krit:
> @gabreal can easily be done by adding a publish to s3 job after each build job. Should I?

As you and I discussed on Element last week, this will be handled by a job on the cleanroom server that will build and upload release-candidates automatically to S3 (and the draft release) using the publish-s3-release gitlab job (tracked in [this](https://github.com/paritytech/devops/issues/653) internal issue). I don't think they are entirely within the scope of this PR so I'd be happy to merge it as-is.

PR #1759, opened 2020-09-28
Register listeners in statement distribution

### LINKED ISSUE #1708: Support registering listeners in the statement distribution subsystem
The statement distribution system should support registration of listeners. The listener should be called for each incoming statement. This is required by Cumulus to send out block announcements for PoVBlock's which are seconded by a validator.

montekki:
Closes #1708

### DIFF
--- a/node/network/statement-distribution/src/lib.rs
+++ b/node/network/statement-distribution/src/lib.rs
@@ -480,13 +480,16 @@ fn check_statement_signature(
 ///
 /// Removes all closed listeners.
 async fn inform_statement_listeners(
-	statement: SignedFullStatement,
+	statement: &SignedFullStatement,
 	listeners: &mut Vec<mpsc::Sender<SignedFullStatement>>,
 ) {
 	// Ignore the errors since these will be removed later.
-	for listener in listeners.iter_mut() {
-		let _ = listener.send(statement.clone()).await;
-	}
+	stream::iter(listeners.iter_mut()).for_each_concurrent(
+		None,
+		|listener| async move {
+			let _ = listener.send(statement.clone()).await;
+		}
+	).await;
 	// Remove any closed listeners.
 	listeners.retain(|tx| !tx.is_closed());
 }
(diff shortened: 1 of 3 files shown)


coriolinus (approved):

node/network/statement-distribution/src/lib.rs
```diff
@@ -476,6 +476,21 @@ fn check_statement_signature(
...
+/// Informs all registered listeners about a newly received statement.
+///
+/// Removes all closed listeners.
+async fn inform_statement_listeners(
+	statement: SignedFullStatement,
+	listeners: &mut Vec<mpsc::Sender<SignedFullStatement>>,
+) {
+	// Ignore the errors since these will be removed later.
+	for listener in listeners.iter_mut() {
+		let _ = listener.send(statement.clone()).await;
```
pepyakin, line 488: a DQ: shouldn't we fan out sending and await the joined future?

node/network/statement-distribution/src/lib.rs
```diff
@@ -874,14 +890,19 @@ async fn run(
 			}
 			FromOverseer::Signal(OverseerSignal::Conclude) => break,
 			FromOverseer::Communication { msg } => match msg {
-				StatementDistributionMessage::Share(relay_parent, statement) =>
+				StatementDistributionMessage::Share(relay_parent, statement) => {
+					inform_statement_listeners(
+						statement.clone(),
```
pepyakin, line 895: If it's cloned inside the function should we pass `statement` there as a reference to save us one `clone`?

node/network/statement-distribution/src/lib.rs
coriolinus: Something like `futures::stream::iter(listeners.iter_mut()).for_each_concurrent(None, |listener| listener.send(statement.clone())` is probably a good idea. Then again, I have the impression that we're rarely going to have more than a few listeners at a time, so it may not make any practical difference.

pepyakin:
bot merge

PR #1760, opened 2020-09-28
XCM: Add the XCM primitives crate.

pepyakin:
This is the first of many commits meant to land https://github.com/paritytech/polkadot/pull/1657 on master.

This is mostly verbatim except that all references to sp-runtime and sp-io are stripped. Now it is a plain old no-std crate.

### DIFF
--- a/xcm/src/v0/mod.rs
+++ b/xcm/src/v0/mod.rs
@@ -1,0 +1,176 @@
+// Copyright 2020 Parity Technologies (UK) Ltd.
+// This file is part of Cumulus.
+
+// Substrate is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+
+// Substrate is distributed in the hope that it will be useful,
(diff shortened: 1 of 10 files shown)


xcm/src/v0/mod.rs
```diff
@@ -0,0 +1,176 @@
...
+	/// A message (`inner`) was sent to `origin` from `superorigin` with the intention of being relayed.
+	///
+	/// - `superorigin`: The location of the `inner` message origin, **relative to `origin`**.
+	/// - `inner`: The message sent by the super origin.
+	///
+	/// Safety: `superorigin` must express a sub-consensus only; it may *NEVER* contain a `Parent` junction.
+	///
+	/// Kind: *Trusted Indication*.
+	///
+	/// Errors:
```
shawntabrizi, line 159: Maybe remove for now?

shawntabrizi (approved):

xcm/src/v0/mod.rs
pepyakin: I'd prefer to keep them if you don't mind, to save us work updating this now and later during rebasing while this doesn't seem to be actively harmful

pepyakin:
bot merge

PR #1769, opened 2020-09-30
WIP: remove deprecated only/except clauses, build is now manual on PRs

TriplEight:
Companion for https://github.com/paritytech/substrate/pull/7229

- changes somewhat broken/deprecated only and accept keys to not very readable rules. The upside is without only, DAGs will be accessible.
- build binary is now manually available in PRs


gabreal (approved):
haven't tested but looks good!

PR #1782, opened 2020-10-02
Adjust max nominators down to 128 (from 256)

jacogr:


kianenigma (approved):

bkchr (approved):

Swader:
Is there a link to a discussion on reasoning behind this?

kianenigma:
> Is there a link to a discussion on reasoning behind this?

Nothing external, we decided that the initial move to 256 was maybe a bit too much. Rest assured, 128 is more sensible now as no polkadot validator will become oversubscribed with it. Over time, the governance can and should adjust this value based on the network status.

PR #1786, opened 2020-10-05
Registrar v1 follow-ups

pepyakin:
A follow-up to recently merged https://github.com/paritytech/polkadot/pull/1559

This PR comprised of two changes:

1. Do not add alias for `polkadot-runtime-parachains::Origin` to `polkadot-runtime-parachains::paras::Origin`. 
1. Call the `schedule_para_cleanup` for the router module as well.

Re: origin. TBH, I am not sure why it was made this way, but I don't see a reason why we should not use the original declaration directly. Otherwise, it spawns questions like: why do we add this origin into the mock but not into the runtime, are there any edge-cases if you register the essentially the same origin twice but using different paths, and so on.

I actually made it this way first in #1679, but failed to land the mock part in https://github.com/paritytech/polkadot/pull/1749.

Re: `schedule_para_cleanup`. Right now, you are expected to call all `schedule_para_cleanup`. I don't remember the reasons why we decided to go with additional function instead of introducing a single entrypoint (i.e. call the router module from paras) or introducing another top-level entrypoint, but this PR might be a sign that we might need to reconsider this. Although I don't think it's critical though. cc @rphmeier 

This PR was split from https://github.com/paritytech/polkadot/pull/1679

### DIFF
--- a/runtime/common/src/paras_registrar.rs
+++ b/runtime/common/src/paras_registrar.rs
@@ -33,6 +33,7 @@ use runtime_parachains::{
 		self,
 		ParaGenesisArgs,
 	},
+	router,
 	ensure_parachain,
 	Origin,
 };
@@ -40,7 +41,7 @@ use runtime_parachains::{
 type BalanceOf<T> =
 	<<T as Trait>::Currency as Currency<<T as frame_system::Trait>::AccountId>>::Balance;
 
-pub trait Trait: paras::Trait {
+pub trait Trait: paras::Trait + router::Trait {
 	/// The aggregated origin type must support the `parachains` origin. We require that we can
 	/// infallibly convert between this origin and the system origin, but in reality, they're the
 	/// same type, we just can't express that to the Rust type system without writing a `where`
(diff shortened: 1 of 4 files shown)


montekki (approved):

runtime/common/src/paras_registrar.rs
```diff
@@ -150,6 +151,7 @@ decl_module! {
 			let _ = <T as Trait>::Currency::unreserve(&debtor, T::ParathreadDeposit::get());
 
 			<paras::Module<T>>::schedule_para_cleanup(id);
+			<router::Module::<T>>::schedule_para_cleanup(id);
```
bkchr, line 154: You need to call the same in line 245?

bkchr (approved):

runtime/common/src/paras_registrar.rs
pepyakin: oh good spot!

pepyakin:
bot merge

PR #1787, opened 2020-10-06
Make collation an optional return

bkchr:
This pr changes the collator interface function to return an optional
collation instead of a collation. This is required as the parachain
itself can fail to generate a valid collation for various reason. Now if
the collation fails it will return `None`.

Besides that the pr adds some `RuntimeDebug` derive for `ValidationData`
and removes some whitespaces.


coriolinus (approved):
This is as we discussed in Element. I do not object to using `Option` here instead of `Result`.

PR #1789, opened 2020-10-06
Pass relay parent as argument when collating

bkchr:
This pr changes the collation function to also pass the current relay
parent the parachain block should be build on.


montekki (approved):

montekki:
bot merge

PR #1799, opened 2020-10-09
[CI] Docker image updates

s3krit:
This PR makes a number of changes:

1. Do not push the parity/polkadot:latest docker image on new *tag* pushes
2. Introduce a new Github Action that will push to parity/polkadot:latest when a new release is *published*, that will build the image based on our deb package repo (this is done with `scripts/docker/Dockerfile.release`
3. Update the documentation in `doc/docker.md` to specify using our official docker images, rather than ones based on phusion. As per our [Docker images maintenance policy](https://github.com/paritytech/devops/wiki/Docker-images-maintenance-policy): Use only official (in the _ namespace) third-party DockerHub images are allowed, both in direct use and inherited with FROM.
4. Update `docker/Dockerfile` to be based on `paritytech/ci-linux:production` as per the aforementioned Docker images maintenance policy.

### DIFF
--- a/.github/workflows/publish-docker-release.yml
+++ b/.github/workflows/publish-docker-release.yml
@@ -12,7 +12,7 @@ jobs:
       - uses: actions/checkout@v2
       - name: Build
         id:   docker_build
-        uses: docker/build-push-action@v2
+        uses: docker/build-push-action@v1
         with:
           username: ${{ secrets.DOCKERHUB_USERNAME }}
           password: ${{ secrets.DOCKERHUB_TOKEN }}
--- a/.github/workflows/publish-docker-release.yml
+++ b/.github/workflows/publish-docker-release.yml
@@ -10,10 +10,6 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
-      - name: Set up QEMU
-        uses: docker/setup-qemu-action@v1
-      - name: Set up Docker Buildx
-        uses: docker/setup-buildx-action@v1
       - name: Build
         id:   docker_build
         uses: docker/build-push-action@v2
--- a/.github/workflows/publish-docker-release.yml
+++ b/.github/workflows/publish-docker-release.yml
@@ -23,6 +23,8 @@ jobs:
         with:
           push: true
           file: scripts/docker/Dockerfile.release
-          tags: martinparity/polkadot:latest
+          tags: |
+            martinparity/polkadot:latest
+            martinparity/polkadot:${{ github.event.release.tag_name }}
       - name: Image digest
         run: echo ${{ steps.docker_build.outputs.digest }}
--- a/.github/workflows/publish-docker-release.yml
+++ b/.github/workflows/publish-docker-release.yml
@@ -10,17 +10,16 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
-      - name: Login to Dockerhub
-        uses: docker/login-action@v1
-        with:
-          username: ${{ secrets.DOCKERHUB_USERNAME }}
-          password: ${{ secrets.DOCKERHUB_TOKEN }}
       - name: Build
         id:   docker_build
         uses: docker/build-push-action@v2
         with:
+          username: ${{ secrets.DOCKERHUB_USERNAME }}
+          password: ${{ secrets.DOCKERHUB_TOKEN }}
+          path: ./scripts/docker
+          file: ./scripts/docker/Dockerfile.release
+          repository: ${{ secrets.DOCKERHUB_USERNAME }}/polkadot
+          tags: latest
           push: true
-          file: scripts/docker/Dockerfile.release
-          tags: martinparity/polkadot:latest
       - name: Image digest
         run: echo ${{ steps.docker_build.outputs.digest }}
--- a/.github/workflows/publish-docker-release.yml
+++ b/.github/workflows/publish-docker-release.yml
@@ -10,12 +10,17 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
+      - name: Login to Dockerhub
+        uses: docker/login-action@v1
+        with:
+          username: ${{ secrets.DOCKERHUB_USERNAME }}
+          password: ${{ secrets.DOCKERHUB_TOKEN }}
       - name: Build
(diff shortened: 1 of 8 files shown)


.github/workflows/publish-docker-release.yml
```diff
@@ -0,0 +1,30 @@
...
+        uses: docker/login-action@v1
+        with:
+          username: ${{ secrets.DOCKERHUB_USERNAME }}
+          password: ${{ secrets.DOCKERHUB_TOKEN }}
+      - name: Build
+        id:   docker_build
+        uses: docker/build-push-action@v2
+        with:
+          push: true
+          file: scripts/docker/Dockerfile.release
```
kirushik, line 25: You can actually cache layers in docker builds with Buildx, which gives a significant performance boost here.

First you'd need to set up caching, with a separate step (prior `build-push-action`):
```yml
      - name: Cache Docker layers
        uses: actions/cache@v2
        with:
          path: /tmp/.buildx-cache
          key: ${{ runner.os }}-buildx-${{ github.sha }}
          restore-keys: |
            ${{ runner.os }}-buildx-
```

And then in your `build-push-action` itself you just add 
```yml
          cache-from: type=local,src=/tmp/.buildx-cache
          cache-to: type=local,dest=/tmp/.buildx-cache
```
properties, and everything seems to work. (At least on that silly side-project of mine.)

kirushik (approved):

scripts/docker/Dockerfile.release
```diff
@@ -0,0 +1,46 @@
+FROM debian:buster-slim
```
TriplEight, line 1: it's better to use "Dockerfile" as an extension: `release.Dockerfile`. Other way many linters do not recognize it.

scripts/docker/Dockerfile.release
```diff
@@ -0,0 +1,46 @@
...
+
+# metadata
+ARG VCS_REF
+ARG BUILD_DATE
+
+LABEL io.parity.image.authors="devops-team@parity.io" \
+	io.parity.image.vendor="Parity Technologies" \
+	io.parity.image.title="parity/polkadot" \
+	io.parity.image.description="polkadot: a platform for web3" \
+	io.parity.image.source="https://github.com/paritytech/polkadot/blob/${VCS_REF}/scripts/docker/Dockerfile" \
```
TriplEight, line 11:
```suggestion
	io.parity.image.source="https://github.com/paritytech/polkadot/blob/${VCS_REF}/scripts/docker/release.Dockerfile" \
```

docker/Dockerfile
```diff
@@ -1,33 +1,21 @@
...
-        scripts/init.sh && \
-        cargo build --$PROFILE
+RUN cargo build --$PROFILE
 
 # ===== SECOND STAGE ======
 
-FROM phusion/baseimage:0.11
-LABEL maintainer "chevdor@gmail.com"
+FROM debian:buster-slim
 LABEL description="This is the 2nd stage: a very small image where we copy the Polkadot binary."
```
TriplEight, line 14: could use the same Label as in `release.Dockerfile` below

scripts/docker/Dockerfile.release
```diff
@@ -0,0 +1,46 @@
...
+USER polkadot
+
+# check if executable works in this container
+RUN /usr/bin/polkadot --version
+
+EXPOSE 30333 9933 9944
+VOLUME ["/polkadot"]
+
+ENTRYPOINT ["/usr/bin/polkadot"]
+
```
TriplEight, line 46: also I have some questions on `diff scripts/docker/Dockerfile scripts/docker/Dockerfile.release`
- why different base?
- why different endpoint for the binary?
- why did you omit creating a user?

scripts/docker/Dockerfile.release
```diff
@@ -0,0 +1,46 @@
...
+# install tools and dependencies
+RUN apt-get update && \
+	DEBIAN_FRONTEND=noninteractive apt-get upgrade -y && \
+	DEBIAN_FRONTEND=noninteractive apt-get install -y \
+		libssl1.1 \
+		ca-certificates \
+		curl \
+    gnupg && \
+    gpg --recv-keys --keyserver hkps://keys.mailvelope.com 9D4B2B6EB8F97156D19669A9FF0812D491B96798 && \
```
TriplEight, line 27: indentation here and below

TriplEight (changes requested):

scripts/docker/Dockerfile.release
s3krit: >why different base?

good spot, was an oversight on my part

>why different endpoint for the binary?

This dockerfile installs polkadot from our apt repository, so it gets placed in `/usr/bin/` rather than `/usr/local/bin`.

>why did you omit creating a user?

The `polkadot` user is created as part of installing the debian package.

docker/Dockerfile
s3krit: The intention of this Dockerfile is to be used by users - i.e., for dev purposes etc., so it doesn't make sense to have vendor info and references to variables (like `$VCS_REF`) that won't be present

gabreal (approved):
overall looks very good to me. what are the arguments to build and release docker images from github actions rather than gitlab ci (it's actually already there, could also use the deb package, runs on self-run infra)?

s3krit:
> overall looks very good to me. what are the arguments to build and release docker images from github actions rather than gitlab ci (it's actually already there, could also use the deb package, runs on self-run infra)?

The main advantage is that we trigger it when the release is *published* rather than *tagged*. Since we perform a bunch of tests after the tag is pushed but before the release is published, it makes more sense to only push our docker images once the release is actually published.

docker/Dockerfile
TriplEight: OCI recommends filling labels as a best practice. At least it can be handy to know where is the code of this Dockerfile, this could be in label.

scripts/docker/Dockerfile.release
TriplEight: > which is based on buster also

oh, then we should change the base for the other image.

TriplEight: > The polkadot user is created as part of installing the debian package.

It's worth mentioning in the comment there.

TriplEight (approved):
Looks good, thanks for updating both Dockerfiles to a new base.

PR #1811, opened 2020-10-13
Add ParasSudoWrapper to test-runtime and bring back register_parachain to test-service

bkchr:
Adds the ParasSudoWrapper to the test-runtime to support registration of parachains.

Besides that it brings back the functionality of the test node to
register a parachain.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5592,6 +5592,7 @@ dependencies = [
  "polkadot-primitives",
  "polkadot-rpc",
  "polkadot-runtime-common",
+ "polkadot-runtime-parachains",
  "polkadot-service",
  "polkadot-test-runtime",
  "rand 0.7.3",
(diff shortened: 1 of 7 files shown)


bkchr:
> why you would do that instead of using `paras_sudo_wrapper` which does essentially the same (minus checking for his para being already registered) and is intended for testing environments?

I was not aware of this. What is than the reason for Registrar to exist?

montekki:
>   What is than the reason for Registrar to exist?
AFAIK the idea is that for test environments and testnets we just use the `paras_sudo_wrapper` and sudo register/unregister stuff since it is a test environment completely controlled by us.
The `registrar` exists to provide its functionality for upper layers of the non-test permissionless environment such as slot auctions.

bkchr:
@montekki done

montekki (approved):

PR #1821, opened 2020-10-15
Finish XCM Executor Assets w/ Tests

shawntabrizi:
This implements the missing functionality of the Assets implementation in XCM Executor.

Some fixes and tests along the way.

### DIFF
--- a/xcm/xcm-executor/src/assets.rs
+++ b/xcm/xcm-executor/src/assets.rs
@@ -18,6 +18,7 @@ use sp_std::{prelude::*, mem::swap, collections::{btree_map::BTreeMap, btree_set
 use xcm::v0::{MultiAsset, MultiLocation, AssetInstance};
 use sp_runtime::RuntimeDebug;
 
+/// Classification of an asset being concrete or abstract.
 #[derive(Clone, Eq, PartialEq, Ord, PartialOrd, RuntimeDebug)]
 pub enum AssetId {
 	Concrete(MultiLocation),
@@ -25,6 +26,7 @@ pub enum AssetId {
 }
 
 impl AssetId {
+	/// Prepend a MultiLocation to a concrete asset, giving it a new root location.
 	pub fn reanchor(&mut self, prepend: &MultiLocation) -> Result<(), ()> {
 		if let AssetId::Concrete(ref mut l) = self {
 			l.prepend_with(prepend.clone()).map_err(|_| ())?;
@@ -33,6 +35,7 @@ impl AssetId {
 	}
 }
 
+/// List of concretely identified fungible and non-fungible assets.
 #[derive(Default, Clone, RuntimeDebug)]
 pub struct Assets {
 	pub fungible: BTreeMap<AssetId, u128>,
@@ -56,6 +59,7 @@ impl From<Assets> for Vec<MultiAsset> {
 }
 
 impl Assets {
+	/// An iterator over the fungible assets.
 	pub fn fungible_assets_iter<'a>(&'a self) -> impl Iterator<Item=MultiAsset> + 'a {
 		self.fungible.iter()
 			.map(|(id, &amount)| match id.clone() {
@@ -64,6 +68,7 @@ impl Assets {
 			})
 	}
 
+	/// An iterator over the non-fungible assets.
 	pub fn non_fungible_assets_iter<'a>(&'a self) -> impl Iterator<Item=MultiAsset> + 'a {
 		self.non_fungible.iter()
 			.map(|&(ref class, ref instance)| match class.clone() {
@@ -72,6 +77,7 @@ impl Assets {
 			})
 	}
 
+	/// An iterator over all assets taking ownership of the values.
 	pub fn into_assets_iter(self) -> impl Iterator<Item=MultiAsset> {
 		let fungible = self.fungible.into_iter()
 			.map(|(id, amount)| match id {
@@ -86,37 +92,35 @@ impl Assets {
 		fungible.chain(non_fungible)
 	}
 
+	/// An iterator over all assets borrowing the values.
 	pub fn assets_iter<'a>(&'a self) -> impl Iterator<Item=MultiAsset> + 'a {
 		let fungible = self.fungible_assets_iter();
 		let non_fungible = self.non_fungible_assets_iter();
 		fungible.chain(non_fungible)
 	}
 
-	/// Modify `self` to include `MultiAsset`, saturating if necessary.
+	/// Modify `self` to include a `MultiAsset`, saturating if necessary.
+	/// Only works on concretely identified assets; wildcards will be swallowed without error.
 	pub fn saturating_subsume(&mut self, asset: MultiAsset) {
 		match asset {
 			MultiAsset::ConcreteFungible { id, amount } => {
-				self.fungible
-					.entry(AssetId::Concrete(id))
-					.and_modify(|e| *e = e.saturating_add(amount))
-					.or_insert(amount);
+				self.saturating_subsume_fungible(AssetId::Concrete(id), amount);
 			}
 			MultiAsset::AbstractFungible { id, amount } => {
-				self.fungible
-					.entry(AssetId::Abstract(id))
-					.and_modify(|e| *e = e.saturating_add(amount))
-					.or_insert(amount);
+				self.saturating_subsume_fungible(AssetId::Abstract(id), amount);
 			}
 			MultiAsset::ConcreteNonFungible { class, instance} => {
-				self.non_fungible.insert((AssetId::Concrete(class), instance));
+				self.saturating_subsume_non_fungible(AssetId::Concrete(class), instance);
 			}
 			MultiAsset::AbstractNonFungible { class, instance} => {
-				self.non_fungible.insert((AssetId::Abstract(class), instance));
+				self.saturating_subsume_non_fungible(AssetId::Abstract(class), instance);
 			}
 			_ => (),
 		}
 	}
 
+	/// Modify `self` to include a new fungible asset by `id` and `amount`,
+	/// saturating if necessary.
 	pub fn saturating_subsume_fungible(&mut self, id: AssetId, amount: u128) {
 		self.fungible
 			.entry(id)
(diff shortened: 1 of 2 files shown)


xcm/xcm-executor/src/assets.rs
```diff
@@ -190,6 +250,98 @@ impl Assets {
...
+						if let Some(&e) = maybe_value {
+							if e > amount {
+								self.fungible.insert(id.clone(), e - amount);
+								result.saturating_subsume_fungible(id, amount);
+							} else {
+								self.fungible.remove(&id);
+								result.saturating_subsume_fungible(id, e.clone());
+							}
+						}
+					}
```
shawntabrizi, line 294: I am not confident that this is the optimal way to do this. Would love a close review (as this logic is duplicated 4 times below)

xcm/xcm-executor/src/assets.rs
```diff
@@ -198,14 +350,16 @@ impl Assets {
...
 							result.saturating_subsume_fungible(id, amount);
-							*e = *e - amount;
 						} else {
-							result.saturating_subsume_fungible(id, *e);
-							*e = 0
-						});
+							self.fungible.remove(&id);
+							result.saturating_subsume_fungible(id, e.clone());
+						}
+					}
```
shawntabrizi, line 362: The old behavior would leave behind an asset with 0 amount.
This new logic removes an asset from the map if the value is zero.

xcm/xcm-executor/src/assets.rs
```diff
@@ -70,52 +92,43 @@ impl Assets {
 		fungible.chain(non_fungible)
 	}
 
+	/// An iterator over all assets borrowing the values.
```
pepyakin, line 95: Does it borrow the values though?

xcm/xcm-executor/src/assets.rs
```diff
@@ -140,22 +153,78 @@ impl Assets {
 	/// Return the assets in `self`, but (asset-wise) of no greater value than `assets`.
 	///
 	/// Result is undefined if `assets` includes elements which match to the same asset more than once.
-	pub fn min<'a, I: Iterator<Item=&'a MultiAsset>>(&self, assets: I) -> Self {
+	pub fn min<'a, M: 'a + std::borrow::Borrow<MultiAsset>, I: Iterator<Item=M>>(&self, assets: I) -> Self {
```
pepyakin, line 156: I don't think it's a good idea to use `std` here. A more worrying issue is that the CI passes, therefore we don't have testing for non std environment

xcm/xcm-executor/src/assets.rs
```diff
@@ -140,22 +153,78 @@ impl Assets {
 	/// Return the assets in `self`, but (asset-wise) of no greater value than `assets`.
 	///
 	/// Result is undefined if `assets` includes elements which match to the same asset more than once.
```
pepyakin, line 155: Can I suggest you to move or add a doc test here? I think it would be a great aid for understanding what this fn supposed to do.

xcm/xcm-executor/src/assets.rs
```diff
@@ -190,6 +251,92 @@ impl Assets {
 			match asset {
 				MultiAsset::None => (),
 				MultiAsset::All => return self.swapped(Assets::default()),
+				MultiAsset::AllFungible => {
+					// Copy all fungible assets into result.
+					for (id, amount) in self.fungible.clone().into_iter() {
```
pepyakin, line 256: You here take a reference to a vector, then create a new vector, then copy all the elements into the new vector, and then destruct the vector using `.into_iter` to iterate on every element.

I think a more idiomatic and efficient way to do the same is to create an iterator from a reference to a vector using `.iter` and then apply the `.cloned` combinator. 

```suggestion
					for (id, amount) in self.fungible.iter().cloned() {
```

xcm/xcm-executor/src/assets.rs
```diff
@@ -190,6 +251,92 @@ impl Assets {
...
+					// Clear all non-fungible assets.
+					self.non_fungible = Default::default();
+				},
+				MultiAsset::AllAbstractFungible { id } => {
+					let all_abstract_fungible: Vec<(AssetId, u128)> = self.fungible.clone()
+						.into_iter()
+						.filter(|(iden, _)| match iden {
+							AssetId::Abstract(iden) => id == *iden,
+							_ => false,
+						}).collect();
```
pepyakin, line 276: Here we have the same as above, but with additional `.collect` which would allocate a new vector. I believe that is not necessary since you destruct that vector by iterating on it just a line below (such a `for` loop calls `into_iter` under the hood).

xcm/xcm-executor/src/assets.rs
```diff
@@ -216,26 +365,273 @@ impl Assets {
...
-				// | MultiAsset::AllConcreteFungible { id }
-				// | MultiAsset::AllConcreteNonFungible { class } => (),
 			}
 		}
 		result
 	}
 
+	/// Swaps two mutable Assets, without deinitializing either one.
 	pub fn swapped(&mut self, mut with: Assets) -> Self {
 		swap(&mut *self, &mut with);
```
pepyakin, line 378: In cases like this, it is not entirely clear (although a seasoned rustecean would have a clue) where this function comes from. In cases like this, people usually import `mem` instead of `mem::swap` and here they'd call `mem::swap(..)`.

xcm/xcm-executor/src/assets.rs
```diff
@@ -216,26 +365,273 @@ impl Assets {
...
+	#[test]
+	fn into_assets_iter_works() {
+		let assets = test_assets();
+		let mut iter = assets.into_assets_iter();
+		// Order defined by implementation: CF, AF, CNF, ANF
+		assert_eq!(Some(CF(300)), iter.next());
+		assert_eq!(Some(AF(1, 100)), iter.next());
+		assert_eq!(Some(CNF(400)), iter.next());
+		assert_eq!(Some(ANF(2, 200)), iter.next());
+		assert_eq!(None, iter.next());
```
pepyakin, line 421: nit: could you `.collect::<Vec<_>>()` and then write it like 

```rust
assert_eq!(assets_vec, vec![CF(300), AF(1, 100), CNF(400), ANF(2, 200)]);
```

looks more legible to me

xcm/xcm-executor/src/assets.rs
```diff
@@ -140,22 +153,99 @@ impl Assets {
...
+	/// 	MultiAsset::AbstractFungible { id: vec![0], amount: 50 },
+	/// ].into();
+	///
+	/// let assets_we_can_trade: Assets = assets_i_have.min(assets_they_want.assets_iter());
+	/// assert_eq!(assets_we_can_trade.into_assets_iter().collect::<Vec<_>>(), vec![
+	/// 	MultiAsset::ConcreteFungible { id: MultiLocation::Null, amount: 100 },
+	/// 	MultiAsset::AbstractFungible { id: vec![0], amount: 50 },
+	/// ]);
+	/// ```
+	pub fn min<'a, M: 'a + sp_std::borrow::Borrow<MultiAsset>, I: Iterator<Item=M>>(&self, assets: I) -> Self {
```
shawntabrizi, line 177: The api for `saturating_take` expects a `Vec<MultiAssets>` while this API expects an Assets iterator.

Is there a reason these APIs are not the same?

pepyakin: To be fair, I am not aware of any. How about this?

```patch
From 91e5ba56079000bbdbe75142d38c35aee166df1c Mon Sep 17 00:00:00 2001
From: Sergei Shulepov <s.pepyakin@gmail.com>
Date: Tue, 20 Oct 2020 13:01:23 +0200
Subject: [PATCH] How about this?

---
 xcm/xcm-executor/src/assets.rs | 15 +++++++++++----
 1 file changed, 11 insertions(+), 4 deletions(-)

diff --git a/xcm/xcm-executor/src/assets.rs b/xcm/xcm-executor/src/assets.rs
index 7df21b4c9..8616b7d8b 100644
--- a/xcm/xcm-executor/src/assets.rs
+++ b/xcm/xcm-executor/src/assets.rs
@@ -174,7 +174,11 @@ impl Assets {
 	/// 	MultiAsset::AbstractFungible { id: vec![0], amount: 50 },
 	/// ]);
 	/// ```
-	pub fn min<'a, M: 'a + sp_std::borrow::Borrow<MultiAsset>, I: Iterator<Item=M>>(&self, assets: I) -> Self {
+	pub fn min<'a, M, I>(&self, assets: I) -> Self
+	where
+		M: 'a + sp_std::borrow::Borrow<MultiAsset>,
+		I: IntoIterator<Item = M>,
+	{
 		let mut result = Assets::default();
 		for asset in assets.into_iter() {
 			match asset.borrow() {
@@ -288,7 +292,10 @@ impl Assets {
 	/// 	MultiAsset::ConcreteFungible { id: MultiLocation::Null, amount: 100 },
 	/// ]);
 	/// ```
-	pub fn saturating_take(&mut self, assets: Vec<MultiAsset>) -> Assets {
+	pub fn saturating_take<I>(&mut self, assets: I) -> Assets
+	where
+		I: IntoIterator<Item = MultiAsset>,
+	{
 		let mut result = Assets::default();
 		for asset in assets.into_iter() {
 			match asset {
@@ -538,7 +545,7 @@ mod tests {
 
 		let taken_none = assets.saturating_take(none);
 		assert_eq!(None, taken_none.assets_iter().next());
-		let taken_all = assets.saturating_take(all.into());
+		let taken_all = assets.saturating_take(all);
 		// Everything taken
 		assert_eq!(None, assets.assets_iter().next());
 		let all_iter = taken_all.assets_iter();
@@ -610,7 +617,7 @@ mod tests {
 		assets2_vec.push(CNF(400));
 		let assets2: Assets = assets2_vec.into();
 
-		let taken = assets1.saturating_take(assets2.into());
+		let taken = assets1.saturating_take(Vec::from(assets2));
 		let taken = taken.into_assets_iter().collect::<Vec<_>>();
 		assert_eq!(taken, vec![CF(300), AF(1, 50), CNF(400)]);
 
-- 
2.28.0
```

pepyakin (approved):
Looks good

PR #1832, opened 2020-10-20
Fix locked ksm in council votes.

kianenigma:
This will fix the issue reported by https://github.com/paritytech/substrate/issues/7223. 

The core of the issue is that the council `VotingBond` has always been 5 *CENTS, but at some point when releasing the 2008 runtime, `DOLLARS` was reduced: 

```diff
- pub const DOLLARS: Balance = DOTS;
+ pub const DOLLARS: Balance = DOTS / 6;
```

which means that the a portion of the bond (namely `old_bond - new_bond`) became inaccessible. 

The amount of money being touched here is not much, but still needs careful reviews. The process to obtain the account ids is explained in the source issues.


runtime/kusama/src/lib.rs
```diff
@@ -889,6 +889,222 @@ impl pallet_proxy::Trait for Runtime {
 	type AnnouncementDepositFactor = AnnouncementDepositFactor;
 }
 
+pub struct FixCouncilHistoricalVotes;
+impl frame_support::traits::OnRuntimeUpgrade for FixCouncilHistoricalVotes {
+	fn on_runtime_upgrade() -> frame_support::weights::Weight {
+		use frame_support::traits::ReservableCurrency;
+		let mut failure: Balance = 0;
+		// https://github.com/paritytech/polkadot/pull/1252/files#diff-cba4e599a9fdd88fe8d33b5ed913958d63f844186b53c5cbe9bc73a2e2944857R22
+		let old_bond = 5 * CENTS * 6;
```
xlc, line 898: will we have integer rounding errors here? e.g. `100 / 6 * 6 != 100`

kianenigma: Indeed, I think it is better to hardcode the exact value. These few lines are mostly for demonstration of what should happen.

shawntabrizi:
There needs to also be a PR which fixes the underlying issue and track the real deposit of users in voting.

kianenigma:
> There needs to also be a PR which fixes the underlying issue and track the real deposit of users in voting.

There is: https://github.com/paritytech/substrate/pull/7040. And basically it is blocked on this migration to happen first.

runtime/kusama/src/lib.rs
```diff
@@ -1019,7 +1235,8 @@ pub type Executive = frame_executive::Executive<
 	frame_system::ChainContext<Runtime>,
 	Runtime,
 	AllModules,
-	CustomOnRuntimeUpgrade
+	// TODO: CustomOnRuntimeUpgrade is done, lets just remove it here?
```
shawntabrizi, line 1238: I think we leave it there just to make it easier to hack such logic in the future.

shawntabrizi (approved):

runtime/kusama/src/lib.rs
shawntabrizi:
```
	#[test]
	fn what_is_dollar() {
		use crate::currency::{DOTS, DOLLARS};
		println!("{:?} / 6 = {:?}", DOTS, DOLLARS);
		println!("Deposit Before: {:?}", 5 * DOTS);
		println!("Deposit After: {:?}", 5 * DOLLARS);
		assert!(false);
	}
```

```
---- constants::tests::what_is_dollar stdout ----
1000000000000 / 6 = 166666666666
Deposit Before: 5000000000000
Deposit After: 833333333330
thread 'constants::tests::what_is_dollar' panicked at 'assertion failed: false', runtime/kusama/src/constants.rs:128:9
```

runtime/kusama/src/lib.rs
```diff
@@ -889,6 +889,222 @@ impl pallet_proxy::Trait for Runtime {
...
 }
 
+pub struct FixCouncilHistoricalVotes;
+impl frame_support::traits::OnRuntimeUpgrade for FixCouncilHistoricalVotes {
+	fn on_runtime_upgrade() -> frame_support::weights::Weight {
+		use frame_support::traits::ReservableCurrency;
+		let mut failure: Balance = 0;
+		// https://github.com/paritytech/polkadot/pull/1252/files#diff-cba4e599a9fdd88fe8d33b5ed913958d63f844186b53c5cbe9bc73a2e2944857R22
+		let old_bond = 5 * CENTS * 6;
+		let current_bond = 5 * CENTS;
```
shawntabrizi, line 899:
```suggestion
		let old_bond = 50000000000;
		let current_bond = 8333333330;
```

```
	#[test]
	fn what_is_cent() {
		use crate::constants::currency::{DOTS, DOLLARS};
		println!("{:?} / 6 = {:?}", DOTS / 100, CENTS);
		println!("Deposit Before: {:?}", 5 * (DOTS / 100));
		println!("Deposit After: {:?}", 5 * CENTS);
		assert!(false);
	}
```

```
---- constants::tests::what_is_cent stdout ----
10000000000 / 6 = 1666666666
Deposit Before: 50000000000
Deposit After: 8333333330
thread 'constants::tests::what_is_cent' panicked at 'assertion failed: false', runtime/kusama/src/constants.rs:128:9
```

kianenigma:
This is stuck in a bit of a deadlock. We realized in a debug session with @apopiak somehow we have corrupt state in both polkadot and kusama. Some accounts are voters while they seem to have zero reserved. This must be a runtime bug and needs to be resolved first.

kianenigma:
This will be stale for a while until we fix some other issues related.

kianenigma:
This is ready for review and merge into the next upgrade again. 

https://github.com/paritytech/substrate/pull/7040 Will add recorded deposits to the pallet. 
https://kusama.polkassembly.io/post/356 will be executed soon which fixes the issue of some users having their reserved balance drained. 

so all is good here.

rrtti:
A note to add: motions reimbursing slashed candidates and runners up are also ongoing in Kusama and Polkadot: 

https://kusama.polkassembly.io/motion/239
https://polkadot.polkassembly.io/motion/43

runtime/kusama/src/lib.rs
```diff
@@ -889,6 +889,222 @@ impl pallet_proxy::Trait for Runtime {
...
+			[166, 148, 132, 242, 177, 14, 194, 241, 222, 161, 147, 148, 66, 61, 87, 111, 145, 198, 181, 171, 35, 21, 179, 137, 244, 225, 8, 188, 240, 170, 40, 64],
+		]
+			.into_iter()
+			.map(|acc|
+				AccountId::from(acc)
+			).for_each(|acc| {
+				failure += Balances::unreserve(&acc, to_unreserve);
+			});
+
+		frame_support::debug::info!("Migration to fix voters happened with errors: {}", failure);
```
thiolliere, line 1103: I guess this can happen if a pallet slash_reserved too much.
So this can happen due to the candidate being slash_reserved too much. I haven't follow the motion https://kusama.polkassembly.io/motion/239 , does it bring back some reserved balance ? if so and if there is some overlap with those account maybe the motion must go through first.

thiolliere (commented):
apart from maybe overlap with the motion and nitpicks, looks good

runtime/kusama/src/lib.rs
kianenigma: No while it was possible to try and compensate some of the reserved balances, I concluded that it is not possible to recover the exact reserved balance within a reasonable amount of time, so for now we refunded everyone to free balance, and it is expected for some errors to happen here.

runtime/kusama/src/lib.rs
```diff
@@ -992,7 +1208,7 @@ pub type Executive = frame_executive::Executive<
 	frame_system::ChainContext<Runtime>,
 	Runtime,
 	AllModules,
-	CustomOnRuntimeUpgrade
+	FixCouncilHistoricalVotes,
```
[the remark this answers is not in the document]
kianenigma: discussed here: https://github.com/paritytech/polkadot/pull/1832#discussion_r509948547

kianenigma:
@thiolliere @thiolliere @shawntabrizi unless if any objections pop up, this will be merged by EOD.

runtime/kusama/src/lib.rs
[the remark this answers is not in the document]
thiolliere: ok then, anyway not too important

runtime/kusama/src/lib.rs
thiolliere: ah I see, good then

thiolliere (approved):

PR #1835, opened 2020-10-21
node/service/src/lib: Do not spawn authority discovery on sentries

mxinden:
The notion of sentry nodes has been deprecated (see [1] for details).
Support for sentry nodes in the `client/authority-discovery` module has
been removed.

This commit adjusts the instantiation of the authority discovery worker
accordingly, only spawning the module on authority nodes.

Companion for https://github.com/paritytech/substrate/pull/7368.

[1] https://github.com/paritytech/substrate/issues/6845


ordian (approved):
Looks good, but this will conflict with #1795 (cc @coriolinus). I'm fine with merging this PR first as it is smaller.

PR #1836, opened 2020-10-21
Update Tips to go 100% to Author

shawntabrizi:
This PR updates the split of transaction tips so that 100% goes to the block author as originally outlined in the W3F research papers:

https://research.web3.foundation/en/latest/polkadot/Token%20Economics.html#-2.-slow-adjusting-mechanism

### DIFF
--- a/runtime/kusama/src/lib.rs
+++ b/runtime/kusama/src/lib.rs
@@ -30,8 +30,8 @@ use primitives::v1::{
 };
 use runtime_common::{
 	claims, SlowAdjustingFeeUpdate, CurrencyToVote,
-	impls::ToAuthor,
-	NegativeImbalance, BlockHashCount, MaximumBlockWeight, AvailableBlockRatio,
+	impls::DealWithFees,
+	BlockHashCount, MaximumBlockWeight, AvailableBlockRatio,
 	MaximumBlockLength, BlockExecutionWeight, ExtrinsicBaseWeight, RocksDbWeight,
 	MaximumExtrinsicWeight, ParachainSessionKeyPlaceholder,
 };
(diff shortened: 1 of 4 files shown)


runtime/kusama/src/lib.rs
```diff
@@ -214,13 +217,21 @@ parameter_types! {
...
->;
+pub struct DealWithFees;
+impl OnUnbalanced<NegativeImbalance<Runtime>> for DealWithFees {
+	fn on_unbalanceds<B>(mut fees_then_tips: impl Iterator<Item=NegativeImbalance<Runtime>>) {
+		if let Some(fees) = fees_then_tips.next() {
+			// for fees, 80% to treasury, 20% to author
+			let mut split = fees.ration(80, 20);
+			if let Some(tips) = fees_then_tips.next() {
+				// for tips, if any, 100% to author
+				tips.ration_merge_into(0, 100, &mut split);
```
bkchr, line 228:
```suggestion
				tips.merge_into(&mut split.1);
```

bkchr (commented):
Would nice to get a test :D 

And the type should be moved to the common crate.

runtime/kusama/src/lib.rs
shawntabrizi: okie dokie

bkchr (approved):

kianenigma (approved):

shawntabrizi:
bot merge

burdges:
We discussed several mitigations in case if we ever run into block producers stealing tips by building forks.  We'll explore them again if and when we notice sketchy looking forks in future.

PR #1842, opened 2020-10-23
Runtime API for checking validation outputs

### LINKED ISSUE #1594: Introduce a Runtime API for checking results against TransientValidationData
This will allow logic to be shared on the node-side and runtime code for checking candidate commitments.

After #1539

pepyakin:
Closes #1594

This PR features structured commits. Please see each commit individually. 

The difference between this PR and is that I had to promote the struct `ValidationOutputs` to top-level primitives. 

This is because CandidateCommitments host not only the validation outputs, but also `erasure_root`. Getting erasure root happens in the backing subsystem, (i.e. later) and I don't think it's a good idea to move it earlier. Moreover, the runtime API in question should check the acceptance criteria and `erasure_root` is not required for that. `ValidationOutputs` is the best thing we have so far that. Unfortunately, it lives in node primitives. So I just promoted it to top-level primitives.
As a potential refactoring we could express `CandidateCommitments` in terms of `ValidationOutputs`.

I used fully-qualified names instead of imports deliberately to minimize conflicts. Before merging I prefer to switch over to imports.

Split off from https://github.com/paritytech/polkadot/pull/1679

### DIFF
--- a/node/core/candidate-validation/src/lib.rs
+++ b/node/core/candidate-validation/src/lib.rs
@@ -34,8 +34,8 @@ use polkadot_node_subsystem_util::{
 use polkadot_subsystem::errors::RuntimeApiError;
 use polkadot_node_primitives::{ValidationResult, InvalidCandidate};
 use polkadot_primitives::v1::{
-	ValidationCode, PoV, CandidateDescriptor, ValidationData, PersistedValidationData,
-	TransientValidationData, OccupiedCoreAssumption, Hash, ValidationOutputs,
+	ValidationCode, PoV, CandidateDescriptor, PersistedValidationData,
+	OccupiedCoreAssumption, Hash, ValidationOutputs,
 };
 use polkadot_parachain::wasm_executor::{
 	self, ValidationPool, ExecutionMode, ValidationError,
@@ -161,7 +161,6 @@ async fn run(
 				}
 				CandidateValidationMessage::ValidateFromExhaustive(
 					persisted_validation_data,
-					transient_validation_data,
 					validation_code,
 					descriptor,
 					pov,
@@ -171,7 +170,6 @@ async fn run(
 						&mut ctx,
 						execution_mode.clone(),
 						persisted_validation_data,
-						transient_validation_data,
 						validation_code,
 						descriptor,
 						pov,
@@ -214,7 +212,7 @@ async fn runtime_api_request<T>(
 
 #[derive(Debug)]
 enum AssumptionCheckOutcome {
-	Matches(ValidationData, ValidationCode),
+	Matches(PersistedValidationData, ValidationCode),
 	DoesNotMatch,
 	BadRequest,
 }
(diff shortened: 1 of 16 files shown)


node/core/candidate-validation/src/lib.rs
```diff
@@ -269,70 +267,100 @@ async fn check_assumption_validation_data(
...
-				validation_code,
-				descriptor,
-				pov,
-				spawn,
-			).await;
-		}
-		AssumptionCheckOutcome::DoesNotMatch => {},
-		AssumptionCheckOutcome::BadRequest => return Ok(Err(ValidationFailed("Bad request".into()))),
+		OccupiedCoreAssumption::TimedOut,
+		// TODO: Why don't we check `Free`? The guide assumes there are only two possible assumptions.
```
rphmeier, line 282: `Included` and `TimedOut` on a free core are both no-ops, so any candidate which assumes the core is free would already pass one of those. Adding the third check is a no-op.

`Free` is kind of an odd one out because there are very few situations where you actually want to check to see that a core is fully free. Off the top of my head, I can't think of any.

primitives/src/v1.rs
```diff
@@ -666,6 +680,10 @@ sp_api::decl_runtime_apis! {
 		fn persisted_validation_data(para_id: Id, assumption: OccupiedCoreAssumption)
 			-> Option<PersistedValidationData<N>>;
 
+		// TODO: Adding a Runtime API should be backwards compatible... right?
```
rphmeier, line 683: Yeah, it is fine. We don't need to worry about backcompat until we've deployed on Rococo. This TODO needs resolving before merge

runtime/parachains/src/inclusion.rs
```diff
@@ -654,6 +658,57 @@ const fn availability_threshold(n_validators: usize) -> usize {
...
+							>= self.config.validation_upgrade_frequency
+				});
+			ensure!(valid_upgrade_attempt, Error::<T>::PrematureCodeUpgrade);
+			ensure!(
+				new_validation_code.0.len() <= self.config.max_code_size as _,
+				Error::<T>::NewCodeTooLarge
+			);
+		}
+
+		// TODO: messaging acceptance criteria rules will go here.
```
rphmeier, line 706: needs resolving

rphmeier (approved):
A couple stray TODOs but looks good!

pepyakin:
Those aren't stray actually : ) I placed two of them to source the comments on review. So they fulfilled their purpose and now can go. 

Then last one is actually a marker (so I will have less fun time rebasing / resolving conflicts) and is supposed to be resolved by within #1679. Now I realized that I can put a PR number there to be inline with our TODO guidelines.

Another thing I realized is that I probably need to bump the runtime spec versions.

rphmeier:
Ok, I know I need to review #1679 :) happy to leave those TODOs be until then

PR #1857, opened 2020-10-27
test-runtime: use sp-std instead of rstd

pepyakin:
At some point we had aliases for all the substrate crates but at some point we've switched to the full names to avoid confusion. It seems that we missed this instance.

Was split off #1679


thiolliere (approved):

coriolinus (approved):

pepyakin:
bot merge

PR #1859, opened 2020-10-27
Downward Message Processing implementation

### LINKED ISSUE #1678: Implement DMP

pepyakin:
Split off from #1679 
Closes #1678
Related to #1806

Note that this change:

- includes updates of XCM, but doesn't include the changes #1679. The current plan is to land the guide changes together. 
- the last commit from #1679 was split from the HRMP part and properly implemented.
- I didn't address [the suggestion](https://github.com/paritytech/polkadot/pull/1679#discussion_r512037298) of @coriolinus since I don't feel the discussion is over. I think it is better to continue it here if my arguments are not conviencing or if there are better suggestions.


rphmeier (approved):
I'll leave the bikeshed up to you and @coriolinus

coriolinus:
> if I am coding some stuff that works with the downward messages, what should I take: downward message or a downward message contents?

I guess it depends what you want to do with it. To me, if you have `DownwardMessage` and `DownwardMessageContents`, then the implication is that the former wraps and possibly extends the latter. 

That said, I don't want to hold anything up with bikeshedding; if you feel that the current naming scheme is better, then just go ahead with it as it is.

pepyakin:
ok, then I am going to go ahead and merge. Although I'll try to keep this issue in mind when writing the client code.

pepyakin:
bot merge

PR #1862, opened 2020-10-27
Fix proposer factory prometheus registration

bkchr:
The proposer wasn't registered in prometheus and thus, we could not see
its metrics.


ordian (approved):
Good catch!

Issue #1846, opened 2020-10-26
Subsystem results should be propagated or logged


coriolinus:
Why does this function even returns a `Result` when we ignore it anyway?

_Originally posted by @bkchr in https://github.com/paritytech/polkadot/pull/1795#r511870919_

PR #1876, opened 2020-10-28
Log whenever a subsystem's fn run returns an Err

coriolinus:
Closes https://github.com/paritytech/polkadot/issues/1846.

### DIFF
--- a/node/subsystem/src/lib.rs
+++ b/node/subsystem/src/lib.rs
@@ -219,9 +219,33 @@ pub trait SubsystemContext: Send + 'static {
 ///
 /// [`Overseer`]: struct.Overseer.html
 /// [`Subsystem`]: trait.Subsystem.html
+#[async_trait]
 pub trait Subsystem<C: SubsystemContext> {
+	/// This subsystem's name
+	const NAME: &'static str;
+
+	/// Run this subsystem's main loop.
+	async fn run(self, ctx: C) -> SubsystemResult<()>;
+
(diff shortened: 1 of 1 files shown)


node/subsystem/src/lib.rs
```diff
@@ -219,9 +219,33 @@ pub trait SubsystemContext: Send + 'static {
...
+	/// This subsystem's name
+	const NAME: &'static str;
+
+	/// Run this subsystem's main loop.
+	async fn run(self, ctx: C) -> SubsystemResult<()>;
+
 	/// Start this `Subsystem` and return `SpawnedSubsystem`.
-	fn start(self, ctx: C) -> SpawnedSubsystem;
+	///
+	/// This is implemented by default in terms of `run`.
```
pepyakin, line 232: I think it would be useful to orient the client on which method they need to call under which circumstances and the implementer which method to implement under which circumstances.

node/subsystem/src/lib.rs
```diff
@@ -219,9 +219,33 @@ pub trait SubsystemContext: Send + 'static {
 ///
 /// [`Overseer`]: struct.Overseer.html
 /// [`Subsystem`]: trait.Subsystem.html
+#[async_trait]
 pub trait Subsystem<C: SubsystemContext> {
+	/// This subsystem's name
```
pepyakin, line 224: Worth noting that it only makes difference if the `start` method is implemented.

bkchr:
Done by https://github.com/paritytech/polkadot/pull/1878

PR #1878, opened 2020-10-28
Change SpawnedSubsystem type to log subsystem errors

### LINKED ISSUE #1846: Subsystem results should be propagated or logged

montekki:
Closes #1846


montekki:
A bit simpler than #1876

bkchr (approved):
Nice! That is what I expected :)

PR #1883, opened 2020-10-29
Don't bring node down when a runtime api call fails

bkchr:


ordian (approved):
FYI: I'm working on an overhaul of subsystem error handling now.

montekki (approved):
Approving as a firefighting move; but overall this is not an intended way of handling things and should be fixed as soon as @ordian improves error handling

PR #1885, opened 2020-10-29
Upward Message Passing implementation

### LINKED ISSUE #1677: Implement UMP

### LINKED ISSUE #1702: Implementer's Guide: Dekindification / XCM

pepyakin:
Closes #1677 
The UMP part of and Closes #1702
The UMP part of #1869
Split from #1679


roadmap/implementers-guide/src/runtime/router.md
```diff
@@ -15,20 +15,36 @@ OutgoingParas: Vec<ParaId>;
...
 /// is the total length (in bytes) of the message payloads.
 ///
-/// Note that this is an auxilary mapping: it's possible to tell the byte size  and the number of
+/// Note that this is an auxilary mapping: it's possible to tell the byte size and the number of
 /// messages only looking at `RelayDispatchQueues`. This mapping is separate to avoid the cost of
 /// loading the whole message queue if only the total size and count are required.
+///
+/// Invariant:
+/// - The set of keys should exactly match the set of keys of `RelayDispatchQueues`.
 RelayDispatchQueueSize: map ParaId => (u32, u32);
```
coriolinus, line 36:
```suggestion
///
/// Output: `(byte_size, num_messages)`.
RelayDispatchQueueSize: map ParaId => (u32, u32);
```

When returning tuples of the same type, I like to explicitly document what each output field represents, even if it's implied by other text in the docstring.

pepyakin: agreed, although I wouldn't call it `Output`: this is a data scheme declaration after all

roadmap/implementers-guide/src/runtime/router.md
```diff
@@ -222,58 +211,85 @@ Candidate Enactment:
...
+
+The following entry-points are meant to be used for HRMP channel management.
+
+Those entry-points are meant to be called from a parachain. `origin` is defined as the `ParaId` of
+the parachain executed the message.
+
+* `hrmp_init_open_channel(recipient, max_places, max_message_size)`:
+    1. Check that the `origin` is not `recipient`.
+    1. Check that `max_places` is less or equal to `config.hrmp_channel_max_places` and greater than zero.
+    1. Check that `max_message_size` is less or equal to `config.hrmp_channel_max_message_size` and greater than zero.
```
coriolinus, line 257: Can we use unsigned types to avoid checks that values are greater than 0?

roadmap/implementers-guide/src/types/runtime.md
```diff
@@ -35,7 +35,7 @@ struct HostConfiguration {
 	/// The amount of blocks ahead to schedule parathreads.
 	pub scheduling_lookahead: u32,
 	/// Total number of individual messages allowed in the parachain -> relay-chain message queue.
-	pub max_upward_queue_count: u32,
+	pub max_upward_queue_capacity: u32,
```
coriolinus, line 38: Nit: `max_upward_queue_capacity` suggests that queue capacity is adjustable, and this is the max legal value. Either `max_upward_queue_count` or `upward_queue_capacity` would be appropriate IMO, but I think `max_upward_queue_capacity` changes the meaning a bit.

roadmap/implementers-guide/src/runtime/router.md
pepyakin: actually, that it's unsigned and the problem still exists. Here we avoid 0 sized messages.

coriolinus (approved):
Looks good to me! I haven't done a detailed review of `ump.rs`, but at a quick read-through I didn't see any problems there.

runtime/parachains/src/configuration.rs
```diff
@@ -285,7 +354,12 @@ mod tests {
 				chain_availability_period: 10,
 				thread_availability_period: 8,
 				scheduling_lookahead: 3,
+				max_upward_queue_count: 1337,
+			    max_upward_queue_size: 228,
```
montekki, line 358: formatting?

runtime/parachains/src/router/ump.rs
```diff
@@ -0,0 +1,717 @@
...
+		});
+	}
+
+	/// Check that all the upward messages sent by a candidate pass the acceptance criteria. Returns
+	/// false, if any of the messages doesn't pass.
+	pub(crate) fn check_upward_messages(
+		config: &HostConfiguration<T::BlockNumber>,
+		para: ParaId,
+		upward_messages: &[UpwardMessage],
+	) -> bool {
```
montekki, line 80: The UX of this concerns me a bit: so if sending 100 messages fail at 42-nd message the operators of the parachain would have to manually try to understand which one caused the failure and why and so on

montekki (approved):

runtime/parachains/src/configuration.rs
drahnr, line 358:
```suggestion
			max_upward_queue_size: 228,
```

runtime/parachains/src/router/ump.rs
```diff
@@ -0,0 +1,717 @@
...
+			Some(para) => match needs_dispatch.binary_search(&para) {
+				Ok(found_idx) => found_idx,
+				Err(_supposed_idx) => {
+					// well that's weird because we maintain an invariant that
+					// `NextDispatchRoundStartWith` must point into one of the items in
+					// `NeedsDispatch`.
+					//
+					// let's select 0 as the starting index as a safe bet.
+					debug_assert!(false);
+					0
```
drahnr, line 294: Is there a case this could be called with zero dispatchables and this hence being oob?

runtime/parachains/src/router/ump.rs
```diff
@@ -0,0 +1,717 @@
...
+
+	/// Flushes the dispatcher state into the persistent storage.
+	fn flush<T: Trait>(self) {
+		let next_one = self.peek();
+		<Module<T> as Store>::NextDispatchRoundStartWith::set(next_one);
+		<Module<T> as Store>::NeedsDispatch::put(self.needs_dispatch);
+	}
+}
+
+#[cfg(test)]
```
drahnr, line 341: Probably a good idea to split into two files here.

runtime/parachains/src/router/ump.rs
pepyakin: I don't think we can do with the `bool` anything: down the line it will have to fit thru the `decl_error`'s bottleneck.

But at least we can log?

runtime/parachains/src/router/ump.rs
```diff
@@ -0,0 +1,717 @@
...
+
+		// make sure that the queue is not overfilled.
+		// we do it here only once since returning false invalidates the whole relay-chain block.
+		if para_queue_count > config.max_upward_queue_count
+			|| para_queue_size > config.max_upward_queue_size
+		{
+			return false;
+		}
+
+		true
```
drahnr, line 105:
```suggestion
		para_queue_count <= config.max_upward_queue_count
			&& para_queue_size <= config.max_upward_queue_size
```

pepyakin: for the record I prefer the former version, but an independent pollee agreed with this version lol )

runtime/parachains/src/router/ump.rs
pepyakin: nah, cur_idx doesn't have to point on an existing element, we always access the underlying vector with `.get`

runtime/parachains/src/router/ump.rs
pepyakin: while I usually prefer doing splitting, I don't think this is a good case for that. First, we would have either `ump_tests.rs` (since the `/router` hosts a bunch of different modules already) or another level of hierarchy like `router/ump.rs` and `router/ump/tests.rs`, and both options seem awkward. Secondly, the tests are still manageable.

runtime/parachains/src/router/ump.rs
pepyakin: I've created an issue https://github.com/paritytech/polkadot/issues/1898

pepyakin:
bot merge

runtime/parachains/src/router/ump.rs
drahnr: Just saying, using an `Option<usize>` might not be entirely silly in the long run.

pepyakin: Try to imagine how would it look like in the current version. We would have to add a couple of maps (or god forbid expects). I am not sure that it worth it presently. That said, I don't disagree that it might be more appropriate in the future

PR #1890, opened 2020-10-30
Update Proxy Filters

shawntabrizi:
This is a small PR which cleans up and normalizes the proxy filters.

Specifically, since: https://github.com/paritytech/substrate/pull/6503

All Utility calls have become safe for all proxy types to call.

- [x] ^^ The above statement needs confirmation ^^

Thus specific filters like:

```
Call::Utility(pallet_utility::Call::batch(..))
```

Are not needed. Instead:

```
Call::Utility(..)
```

can be used.

This also adds `Session` to the Staking Proxy Filter. cc @joepetrowski @wpank

### DIFF
--- a/runtime/kusama/src/lib.rs
+++ b/runtime/kusama/src/lib.rs
@@ -852,6 +852,7 @@ impl InstanceFilter<Call> for ProxyType {
 			),
 			ProxyType::Staking => matches!(c,
 				Call::Staking(..) |
+				Call::Session(..) |
 				Call::Utility(..)
 			),
 			ProxyType::IdentityJudgement => matches!(c,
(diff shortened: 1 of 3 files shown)


shawntabrizi:
Note I also added `Session` to the staking filters.

gavofyork (approved):

gavofyork:
Would be nice to have a filter for governance calls that cannot result in a slash.

shawntabrizi:
@gavofyork ack. Will open an issue and address in a future PR

shawntabrizi:
bot merge

redzsina:
We are done with our review for this PR and don't see any issues with it.

PR #1892, opened 2020-10-30
A real overseer feature

montekki:
To test:
```
cargo run --release --features="real-overseer"  -- --tmp --chain rococo-local --validator --alice --force-authoring
```

### DIFF
--- a/node/service/src/lib.rs
+++ b/node/service/src/lib.rs
@@ -23,7 +23,7 @@ mod grandpa_support;
 mod client;
 
 use grandpa::{self, FinalityProofProvider as GrandpaFinalityProofProvider};
-#[cfg(all(feature = "full-node", feature = "real-overseer"))]
+#[cfg(all(feature = "full-node"))]
 use {
 	std::convert::TryInto,
 	std::time::Duration,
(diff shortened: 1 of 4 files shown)


ordian (approved):

node/service/src/lib.rs
```diff
@@ -23,7 +23,7 @@ mod grandpa_support;
 mod client;
 
 use grandpa::{self, FinalityProofProvider as GrandpaFinalityProofProvider};
-#[cfg(feature = "full-node")]
+#[cfg(all(feature = "full-node"))]
```
bkchr, line 26:
```suggestion
#[cfg(feature = "full-node")]
```

bkchr (approved):

rphmeier:
bot merge

joelamouche:
Hi, so I'm trying to learn more about Overseer and why I need to pass `--features=real-overseer` when testing my parachain (moonbeam).
What happens if I **don't** pass it?
Thanks in advance

rphmeier:
If you don't pass `--features=real-overseer` then all parachains consensus components are replaced with dummy versions that don't do anything.

joelamouche:
So block won't be produced on the parachains?

PR #1894, opened 2020-10-31
fix service build: enable notifications protocol only under real overseer

rphmeier:
#1864 and #1892 conflicted, and this resolves that


rphmeier:
bot merge

ordian (approved):

PR #1897, opened 2020-11-02
Backing fixes

rphmeier:
* Add `attested_candidate` function to statement table
* fix: Ensure that imported statements always are fed to the statement table
* Issue backed candidates to the provisioner
* fix: construct backed candidates with the correct para_id for the group
* fix: tests

### DIFF
--- a/node/core/backing/src/lib.rs
+++ b/node/core/backing/src/lib.rs
@@ -337,13 +337,19 @@ impl CandidateBackingJob {
 		let issued_statement = statement.is_some();
 
 		if let Some(statement) = statement {
-			if let Some(signed_statement) = self.sign_statement(statement).await {
-				self.import_statement(&signed_statement).await?;
-				self.distribute_signed_statement(signed_statement).await?;
-			}
+			self.sign_import_and_distribute_statement(statement).await?
 		}
 
 		Ok(issued_statement)
+	}
+
+	async fn sign_import_and_distribute_statement(&mut self, statement: Statement) -> Result<(), Error> {
+		if let Some(signed_statement) = self.sign_statement(statement).await {
+			self.import_statement(&signed_statement).await?;
+			self.distribute_signed_statement(signed_statement).await?;
+		}
+
+		Ok(())
 	}
 
 	fn get_backed(&self) -> Vec<NewBackedCandidate> {
@@ -354,9 +360,9 @@ impl CandidateBackingJob {
 			let TableAttestedCandidate { candidate, validity_votes, .. } = p;
 
 			let (ids, validity_votes): (Vec<_>, Vec<_>) = validity_votes
-						.into_iter()
-						.map(|(id, vote)| (id, vote.into()))
-						.unzip();
+				.into_iter()
+				.map(|(id, vote)| (id, vote.into()))
+				.unzip();
 
 			let group = match self.table_context.groups.get(&self.assignment) {
 				Some(group) => group,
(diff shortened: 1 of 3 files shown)


node/core/backing/src/lib.rs
```diff
@@ -428,9 +450,27 @@ impl CandidateBackingJob {
 
 		let summary = self.table.import_statement(&self.table_context, stmt);
 
+		if let Some(ref summary) = summary {
+			if let Some(attested) = self.table.attested_candidate(
+				&summary.candidate,
+				&self.table_context,
+			) {
+				if self.backed.insert(summary.candidate) {
```
bkchr, line 458: If we do it this way we will miss some validity votes or even include a candidate that we got an invalid message for at a later point.

montekki: If there is >= threshold votes and an `Invalid` message is received someone is lying so in any way this case has to be handled by misbehavior reporting?

montekki (approved):

node/core/backing/src/lib.rs
bkchr: As far as I understand this function here: https://github.com/paritytech/polkadot/blob/bkchr-start-backing-fix/statement-table/src/generic.rs#L521

Its not a misbehavior to send an `Invalid` message and it would also be weird to make this a misbehavior.

montekki: I always thought that the idea is this: issuing any statement that is not true (`Invalid` on a valid andidate and vice versa) is a misbehavior. For instance this can happen when candidates are re-checked by randomly assigned validators and in this case a dispute(?) has to happen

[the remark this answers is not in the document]
rphmeier: i always forget this as well : )

rphmeier: > If we do it this way we will miss some validity votes or even include a candidate that we got an invalid message for at a later point.

I just removed the `Invalid` check for includability. But yeah, we might miss some votes. Maybe it makes more sense to have this structure:
  * Backing subsystem notifies `Provisioner` of the `CandidateReceipt` that is includable.
  * Provisioner then can fetch full `BackedCandidate` for any of those, which will include all votes on the candidate.

This gives us the benefits that the `Provisioner` (high-level code) gets to decide which backed candidates to include and we don't drop votes. To avoid round-trips on the overseer, the request to the `CandidateBacking` subsystem should be a `Vec<Hash>` with a `Vec<BackedCandidate>` response.

node/core/backing/src/lib.rs
```diff
@@ -242,6 +245,44 @@ fn primitive_statement_to_table(s: &SignedFullStatement) -> TableSignedStatement
...
+) -> Option<BackedCandidate> {
+	let TableAttestedCandidate { candidate, validity_votes, group_id: para_id } = attested;
+
+	let (ids, validity_votes): (Vec<_>, Vec<_>) = validity_votes
+		.into_iter()
+		.map(|(id, vote)| (id, vote.into()))
+		.unzip();
+
+	let group = match table_context.groups.get(&para_id) {
+		Some(group) => group,
```
bkchr, line 265:
```suggestion
```

node/core/backing/src/lib.rs
```diff
@@ -242,6 +245,44 @@ fn primitive_statement_to_table(s: &SignedFullStatement) -> TableSignedStatement
...
+		None => return None,
```
bkchr, line 266:
```suggestion
```

node/core/backing/src/lib.rs
```diff
@@ -242,6 +245,44 @@ fn primitive_statement_to_table(s: &SignedFullStatement) -> TableSignedStatement
...
+	};
```
bkchr, line 267:
```suggestion
```

bkchr (approved):

node/core/backing/src/lib.rs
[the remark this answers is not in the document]
rphmeier: added a comment

rphmeier:
bot merge

node/core/backing/src/lib.rs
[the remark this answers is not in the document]
burdges: I suppose `HasMap::insert` must return `Some` when filled, so `HashSet::insert` returns morally the opposite.

Issue #1898, opened 2020-11-02
Diagnostics for the ValidationOutputs checker


pepyakin:
The router module adds quite a few acceptance criteria checks. However, it's hard to tell which check hasn't passed. Moreover, e.g. for upward messages, any of the messages can fail the check and there is no indication of which one.

We need to introduce better informing for why a certain candidate didn't pass the checks.

pepyakin:
Because it would be also nice to see which candidate failed the acceptance criteria, in case there are multiple (i.e. during the inclusion).

Perhaps, this time we need to introduce an error enum and handle it in higher level, i.e. either within `CandidateCheckContext::check_validation_outputs` or at its call sites. This would also allow us to add a prefix message, since a log about some upward message doesn't immediately makes it clear that the acceptance criteria didn't pass.

There is also similar thing to #1900

pepyakin:
I guess, while we are at it, we can do the inclusion module error handling.

PR #1900, opened 2020-11-02
Implement HRMP

### LINKED ISSUE #1400: Implement HRMP for parachains

### LINKED ISSUE #1642: Guide: Tighten HRMP
Currently, we defer some structural changes to the session boundary, to make contextual execution easier and not invalidate a candidate suddenly while it was inflight.

However, that creates a problem with mass channel closure. If a parachain that has maximum number of inbound channels and outbound channels decides to close all channels that it has all of them will be handled at the session boundary. That can also happen if a parachain is offboarded. Potentially if we offboard a number of parachains we can end up closing a very large number of channels at the same time.

For XCMP so far we came up with a proper mitigation, i.e. we don't close the channels right away, but let leave them on the balance of the recipient and we add additional rule for acceptance criteria that disallows sending the messages in that channel.

For HRMP we decided to just put aggressive limits in place. The goal of this issue is to make sure we put limits that won't allow to exploit the aforementioned problem. We can limit the total number of channels by, say, 20 inbound channels and 20 outbound channels.

### LINKED ISSUE #1663: Implementer's Guide: HRMP: Reliable way to notify PVF about incoming channels

### LINKED ISSUE #1704: Implementer's Guide: Demote HRMP channel management to regular dispatchables

### LINKED ISSUE #1806: HRMP: Provide a collator access to incoming messages

### LINKED ISSUE #1869: UMP, HRMP: CandidateCommitments size limits

pepyakin:
This PR subsumes the Router PR: #1679 

Closes #1400
Closes #1642 
Closes #1663 
Closes #1704
Closes #1806
Closes #1869

- [x] #1806
- [x] #1869
- [x] the router module HRMP tests

Regarding #1663 I decided to go with a simple implementation: just to send a DM to notify paras. For now, it makes sense to start with a simpler approach to see how it works and then extend it as needed. My concern about this solution was that these notifications should be processed in a timely manner but a downward message queue can be clogged. Well, that's a bit too forward looking since most likely the queues won't be filled in the foreseeable future (assuming we can upgrade later). Apart from that, if that becomes an issue, it is possible for the PVF to enact messages out-of-order therefore getting some sort of QoS. But again I doubt that this will be required for now.

### DIFF
--- a/parachain/src/primitives.rs
+++ b/parachain/src/primitives.rs
@@ -186,7 +186,14 @@ impl<T: Encode + Decode + Default> AccountIdConversion<T> for Id {
 	}
 }
 
-/// A type that uniquely identifies an HRMP channel. An HRMP channel is unidirectional
+/// A type that uniquely identifies an HRMP channel. An HRMP channel is established between two paras.
+/// In text, we use the notation `(A, B)` to specify a channel between A and B. The channels are
+/// unidirectional, meaning that `(A, B)` and `(B, A)` refer to different channels. The convention is
+/// that we use the first item tuple for the sender and the second for the recipient. Only one channel
+/// is allowed between two participants in one direction, i.e. there cannot be 2 different channels
+/// identified by `(A, B)`.
+///
+/// `HrmpChannelId` has a defined ordering: first `sender` and tie is resolved by `recipient`.
 #[derive(Clone, PartialEq, Eq, Encode, Decode, RuntimeDebug)]
 #[cfg_attr(feature = "std", derive(Hash))]
 pub struct HrmpChannelId {
--- a/parachain/src/primitives.rs
+++ b/parachain/src/primitives.rs
@@ -28,7 +28,7 @@ use serde::{Serialize, Deserialize};
 #[cfg(feature = "std")]
 use sp_core::bytes;
 
-use polkadot_core_primitives::Hash;
+use polkadot_core_primitives::{Hash, OutboundHrmpMessage};
 
 /// Block number type used by the relay chain.
 pub use polkadot_core_primitives::BlockNumber as RelayChainBlockNumber;
@@ -186,6 +186,16 @@ impl<T: Encode + Decode + Default> AccountIdConversion<T> for Id {
 	}
 }
 
+/// A type that uniquely identifies an HRMP channel. An HRMP channel is unidirectional
+#[derive(Clone, PartialEq, Eq, Encode, Decode, RuntimeDebug)]
+#[cfg_attr(feature = "std", derive(Hash))]
+pub struct HrmpChannelId {
+	/// The para that acts as the sender in this channel.
+	pub sender: Id,
+	/// The para that acts as the recipient in this channel.
+	pub recipient: Id,
+}
+
 /// A message from a parachain to its Relay Chain.
 pub type UpwardMessage = Vec<u8>;
 
@@ -212,7 +222,7 @@ pub struct ValidationParams {
 }
 
 /// The result of parachain validation.
-// TODO: egress and balance uploads (https://github.com/paritytech/polkadot/issues/220)
+// TODO: balance uploads (https://github.com/paritytech/polkadot/issues/220)
 #[derive(PartialEq, Eq, Encode)]
 #[cfg_attr(feature = "std", derive(Debug, Decode))]
 pub struct ValidationResult {
@@ -222,8 +232,12 @@ pub struct ValidationResult {
 	pub new_validation_code: Option<ValidationCode>,
 	/// Upward messages send by the Parachain.
 	pub upward_messages: Vec<UpwardMessage>,
+	/// Outbound horizontal messages sent by the parachain.
+	pub horizontal_messages: Vec<OutboundHrmpMessage<Id>>,
 	/// Number of downward messages that were processed by the Parachain.
 	///
 	/// It is expected that the Parachain processes them from first to last.
 	pub processed_downward_messages: u32,
+	/// The mark which specifies the block number up to which all inbound HRMP messages are processed.
+	pub hrmp_watermark: RelayChainBlockNumber,
 }
--- a/parachain/src/primitives.rs
+++ b/parachain/src/primitives.rs
@@ -192,8 +192,6 @@ impl<T: Encode + Decode + Default> AccountIdConversion<T> for Id {
 /// that we use the first item tuple for the sender and the second for the recipient. Only one channel
 /// is allowed between two participants in one direction, i.e. there cannot be 2 different channels
 /// identified by `(A, B)`.
-///
-/// `HrmpChannelId` has a defined ordering: first `sender` and tie is resolved by `recipient`.
 #[derive(Clone, PartialEq, Eq, Encode, Decode, RuntimeDebug)]
 #[cfg_attr(feature = "std", derive(Hash))]
 pub struct HrmpChannelId {
--- a/parachain/test-parachains/adder/src/wasm_validation.rs
+++ b/parachain/test-parachains/adder/src/wasm_validation.rs
@@ -17,12 +17,13 @@
 //! WASM validation for adder parachain.
 
 use crate::{HeadData, BlockData};
-use core::{intrinsics, panic};
+use core::panic;
+use sp_std::vec::Vec;
 use parachain::primitives::{ValidationResult, HeadData as GenericHeadData};
 use codec::{Encode, Decode};
 
 #[no_mangle]
-pub extern fn validate_block(params: *const u8, len: usize) -> u64 {
+pub extern "C" fn validate_block(params: *const u8, len: usize) -> u64 {
 	let params = unsafe { parachain::load_params(params, len) };
 	let parent_head = HeadData::decode(&mut &params.parent_head.0[..])
 		.expect("invalid parent head format.");
@@ -38,7 +39,9 @@ pub extern fn validate_block(params: *const u8, len: usize) -> u64 {
 			head_data: GenericHeadData(new_head.encode()),
 			new_validation_code: None,
 			upward_messages: sp_std::vec::Vec::new(),
+			horizontal_messages: sp_std::vec::Vec::new(),
 			processed_downward_messages: 0,
+			hrmp_watermark: params.relay_chain_height,
 		}
 	)
 }
--- a/roadmap/implementers-guide/src/runtime/inclusion.md
+++ b/roadmap/implementers-guide/src/runtime/inclusion.md
@@ -70,8 +70,7 @@ All failed checks should lead to an unrecoverable error making the block invalid
   1. call `Router::check_upward_messages(para, commitments.upward_messages)` to check that the upward messages are valid.
   1. call `Router::check_processed_downward_messages(para, commitments.processed_downward_messages)` to check that the DMQ is properly drained.
   1. call `Router::check_hrmp_watermark(para, commitments.hrmp_watermark)` for each candidate to check rules of processing the HRMP watermark.
-  1. check that in the commitments of each candidate the horizontal messages are sorted by ascending recipient ParaId and there is no two horizontal messages have the same recipient.
-  1. using `Router::verify_outbound_hrmp(sender, commitments.horizontal_messages)` ensure that the each candidate send a valid set of horizontal messages
+  1. using `Router::check_outbound_hrmp(sender, commitments.horizontal_messages)` ensure that the each candidate send a valid set of horizontal messages
   1. create an entry in the `PendingAvailability` map for each backed candidate with a blank `availability_votes` bitfield.
   1. create a corresponding entry in the `PendingAvailabilityCommitments` with the commitments.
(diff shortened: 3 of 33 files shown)


rphmeier (commented):
some comments, but looks OK

roadmap/implementers-guide/src/runtime/router.md
```diff
@@ -144,22 +148,34 @@ HrmpCloseChannelRequests: map HrmpChannelId => Option<()>;
...
 HrmpIngressChannelsIndex: map ParaId => Vec<ParaId>;
 HrmpEgressChannelsIndex: map ParaId => Vec<ParaId>;
 /// Storage for the messages for each channel.
 /// Invariant: cannot be non-empty if the corresponding channel in `HrmpChannels` is `None`.
 HrmpChannelContents: map HrmpChannelId => Vec<InboundHrmpMessage>;
 /// Maintains a mapping that can be used to answer the question:
 /// What paras sent a message at the given block number for a given reciever.
-/// Invariant: The para ids vector is never empty.
+/// Invariants:
+/// - The para ids vector is never empty.
```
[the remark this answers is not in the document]
pepyakin: Oh, I meant the inner vector, that is parameterized by `ParaId`

runtime/parachains/src/inclusion.rs
```diff
@@ -739,6 +757,23 @@ impl<T: Trait> CandidateCheckContext<T> {
 			),
 			Error::<T>::InvalidUpwardMessages,
 		);
+		ensure!(
+			<router::Module<T>>::check_hrmp_watermark(
+				para_id,
+				self.relay_parent_number,
+				// TODO: Hmm, we should settle on a single represenation of T::BlockNumber
+				T::BlockNumber::from(hrmp_watermark),
```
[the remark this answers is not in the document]
pepyakin: Makes sense although a bit annoying, I wish we used a concrete type. I am going to introduce an error type in #1898 and it would require parameterizing that enum with `BlockNumber` and that seems excessive to me.

parachain/src/primitives.rs
```diff
@@ -186,6 +186,23 @@ impl<T: Encode + Decode + Default> AccountIdConversion<T> for Id {
...
+/// In text, we use the notation `(A, B)` to specify a channel between A and B. The channels are
+/// unidirectional, meaning that `(A, B)` and `(B, A)` refer to different channels. The convention is
+/// that we use the first item tuple for the sender and the second for the recipient. Only one channel
+/// is allowed between two participants in one direction, i.e. there cannot be 2 different channels
+/// identified by `(A, B)`.
+///
+/// `HrmpChannelId` has a defined ordering: first `sender` and tie is resolved by `recipient`.
+#[derive(Clone, PartialEq, Eq, Encode, Decode, RuntimeDebug)]
+#[cfg_attr(feature = "std", derive(Hash))]
+pub struct HrmpChannelId {
```
coriolinus, line 199: If there is a defined ordering worth mentioning in the documentation, why not derive `PartialOrd, Ord`?

roadmap/implementers-guide/src/runtime/inclusion.md
```diff
@@ -70,18 +70,17 @@ All failed checks should lead to an unrecoverable error making the block invalid
   1. call `Router::check_upward_messages(para, commitments.upward_messages)` to check that the upward messages are valid.
   1. call `Router::check_processed_downward_messages(para, commitments.processed_downward_messages)` to check that the DMQ is properly drained.
   1. call `Router::check_hrmp_watermark(para, commitments.hrmp_watermark)` for each candidate to check rules of processing the HRMP watermark.
-  1. check that in the commitments of each candidate the horizontal messages are sorted by ascending recipient ParaId and there is no two horizontal messages have the same recipient.
-  1. using `Router::verify_outbound_hrmp(sender, commitments.horizontal_messages)` ensure that the each candidate send a valid set of horizontal messages
+  1. using `Router::check_outbound_hrmp(sender, commitments.horizontal_messages)` ensure that the each candidate send a valid set of horizontal messages
```
coriolinus, line 73:
```suggestion
  1. using `Router::check_outbound_hrmp(sender, commitments.horizontal_messages)` ensure that the each candidate sent a valid set of horizontal messages
```

roadmap/implementers-guide/src/runtime/router.md
```diff
@@ -90,26 +92,28 @@ struct HrmpChannel {
...
+    limit_message_size: u32,
     /// The maximum number of messages that can be pending in the channel at once.
-    limit_used_places: u32,
+    max_capacity: u32,
     /// The maximum total size of the messages that can be pending in the channel at once.
-    limit_used_bytes: u32,
+    max_total_size: u32,
     /// The maximum message size that could be put into the channel.
-    limit_message_size: u32,
+    max_message_size: u32,
```
coriolinus, line 102: What's the difference between `max_message_size` and `limit_message_size`?

roadmap/implementers-guide/src/runtime/router.md
```diff
@@ -90,26 +92,28 @@ struct HrmpChannel {
     sender_deposit: Balance,
     /// The amount that the recipient supplied as a deposit when accepting opening this channel.
     recipient_deposit: Balance,
+    /// The maximum message size that could be put into the channel.
+    limit_message_size: u32,
```
coriolinus, line 96:
```suggestion
```

If, as I suspect, this is a redundant bit of data, then we can just remove it.

roadmap/implementers-guide/src/runtime/router.md
```diff
@@ -144,22 +148,34 @@ HrmpCloseChannelRequests: map HrmpChannelId => Option<()>;
 HrmpCloseChannelRequestsList: Vec<HrmpChannelId>;
 
 /// The HRMP watermark associated with each para.
+/// Invariant: every para should be onboarded.
```
coriolinus, line 151: It would be nice to have a little more description here, or a link to the guide if appropriate. It's not obvious what onboarding a para entails.

roadmap/implementers-guide/src/runtime/router.md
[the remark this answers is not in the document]
coriolinus:
```suggestion
/// - The inner `Vec<ParaId>` is never empty.
```

roadmap/implementers-guide/src/runtime/router.md
```diff
@@ -144,22 +148,34 @@ HrmpCloseChannelRequests: map HrmpChannelId => Option<()>;
...
 /// Storage for the messages for each channel.
 /// Invariant: cannot be non-empty if the corresponding channel in `HrmpChannels` is `None`.
 HrmpChannelContents: map HrmpChannelId => Vec<InboundHrmpMessage>;
 /// Maintains a mapping that can be used to answer the question:
 /// What paras sent a message at the given block number for a given reciever.
-/// Invariant: The para ids vector is never empty.
+/// Invariants:
+/// - The para ids vector is never empty.
+/// - The para ids vector doesn't contain duplicates.
+/// - The outer vector is sorted ascending by block number and there are no duplicates.
```
coriolinus, line 178: Ambiguous: for a given block number, would it be legal for the outer vector to contain both `(n, vec1)` and `(n, vec2)` where `vec1 != vec2`? That's the property we get from a naive deduplication. If there cannot be two digests with the same block number, regardless of the associated vectors, then that's a stricter property, and we should mention it specifically.

runtime/parachains/src/router/hrmp.rs
```diff
@@ -0,0 +1,1244 @@
...
+		// a -> [x, y]     x -> [a, b]
+		// b -> [x, z]     y -> [a]
+		//                 z -> [b]
+		//
+		// we derive a list of channels they represent.
+		//
+		//   (a, x)         (a, x)
+		//   (a, y)         (a, y)
+		//   (b, x)         (b, x)
+		//   (b, y)         (b, y)
```
coriolinus, line 976:
```suggestion
		// ingress         egress
		//
		// a -> [x, y]     x -> [a, b]
		// b -> [x, z]     y -> [a]
		//                 z -> [b]
		//
		// we derive a list of channels they represent.
		//
		//   (a, x)         (a, x)
		//   (a, y)         (a, y)
		//   (b, x)         (b, x)
		//   (b, z)         (b, z)
```

Looks like a typo in the conversion table: `b` maps to `z` in two cases, not `y`.

parachain/src/primitives.rs
pepyakin: That's a great catch! Apparently, this property was specified because we needed sorting somewhere but now the impl revealed that the spec didn't actually rely on this.

I'll remove this note

roadmap/implementers-guide/src/runtime/router.md
pepyakin: Your suspicion is correct! This PR undergone quite a few of rebases and [git] fixups and I guess this is one artifact of that.

roadmap/implementers-guide/src/runtime/router.md
pepyakin: Oh, this is such a valuable catch!

coriolinus (approved):

pepyakin:
bot merge

parachain/test-parachains/adder/src/wasm_validation.rs
```diff
@@ -17,12 +17,13 @@
...
 use crate::{HeadData, BlockData};
-use core::{intrinsics, panic};
+use core::panic;
+use sp_std::vec::Vec;
 use parachain::primitives::{ValidationResult, HeadData as GenericHeadData};
 use codec::{Encode, Decode};
 
 #[no_mangle]
-pub extern fn validate_block(params: *const u8, len: usize) -> u64 {
+pub extern "C" fn validate_block(params: *const u8, len: usize) -> u64 {
```
drahnr, line 26: It would be preferred to have platform independent integer width length for `len`, since this is exported.

pepyakin: I guess, this won't harm, but note that this code is for wasm only and we assume that it's 32 bits only.

drahnr: It's just a beartrap in the woods, soon to be forgotten of its existence, and somebody will step into it eventually - it might not be a :bear:

pepyakin: https://github.com/paritytech/polkadot/pull/1944

PR #1905, opened 2020-11-02
Parachain improvements

bkchr:
- Set the parachains configuration in Rococo genesis
- Don't stop the overseer when a subsystem job is stopped
- Several small code changes


node/subsystem-util/src/lib.rs
```diff
@@ -867,7 +820,7 @@ where
 					.forward(drain())
 					.await
 				{
-					log::error!("failed to stop all jobs on conclude signal: {:?}", e);
+					log::error!("failed to stop all jobs({}) on conclude signal: {:?}", Job::NAME, e);
```
coriolinus, line 823:
```suggestion
					log::error!("failed to stop all jobs ({}) on conclude signal: {:?}", Job::NAME, e);
```

node/subsystem-util/src/lib.rs
```diff
@@ -877,30 +830,18 @@ where
...
-							if let Err(err) = Job::handle_unanchored_msg(to_job) {
-								log::error!("Failed to handle unhashed message: {:?}", err);
-								let e = JobsError::Job(err);
-								Self::fwd_err(None, e, err_tx).await;
-								return true;
-							}
-						}
+						Some(hash) => jobs.send_msg(hash, to_job).await,
+						None => log::debug!(
+							"Trying to send a message to a job({}) without specifying a relay parent.",
```
coriolinus, line 835:
```suggestion
							"Trying to send a message to a job ({}) without specifying a relay parent.",
```

node/subsystem-util/src/lib.rs
```diff
@@ -877,30 +830,18 @@ where
...
+						),
 					}
 				}
 			}
 			Ok(Signal(BlockFinalized(_))) => {}
 			Err(err) => {
-				log::error!("error receiving message from subsystem context: {:?}", err);
-				let e = JobsError::Utility(Error::from(err));
-				Self::fwd_err(None, e, err_tx).await;
+				log::error!("error receiving message from subsystem context for job({}): {:?}", Job::NAME, err);
```
coriolinus, line 843:
```suggestion
				log::error!("error receiving message from subsystem context for job ({}): {:?}", Job::NAME, err);
```

node/subsystem-util/src/lib.rs
```diff
@@ -637,54 +607,33 @@ impl<Spawner: SpawnNamed, Job: 'static + JobTrait> Jobs<Spawner, Job> {
...
-// we just abort them all. Still better than letting them dangle.
-#[pinned_drop]
-impl<Spawner, Job: JobTrait> PinnedDrop for Jobs<Spawner, Job> {
-	fn drop(self: Pin<&mut Self>) {
-		for job_handle in self.running.values() {
-			job_handle.abort_handle.abort();
+	async fn send_msg(&mut self, parent_hash: Hash, msg: Job::ToJob) {
+		if let Entry::Occupied(mut job) = self.running.entry(parent_hash) {
+			if job.get_mut().send_msg(msg).await.is_err() {
+				log::debug!("failed to send message to job({}), will remove it", Job::NAME);
```
coriolinus, line 634:
```suggestion
				log::debug!("failed to send message to job ({}), will remove it", Job::NAME);
```

coriolinus (approved):
Some error formatting nits, but nothing blocking.

node/core/backing/src/lib.rs
```diff
@@ -838,7 +838,7 @@ impl util::JobTrait for CandidateBackingJob {
 
 			let (assignment, required_collator) = match assignment {
 				None => return Ok(()), // no need to work.
-				Some((a, r)) => (a, r),
+				Some(r) => r,
```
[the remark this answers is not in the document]
bkchr: Why does it makes the readability worse? You see that this match returns a tuple by checking the let statement.

montekki (approved):

primitives/src/v1.rs
```diff
@@ -687,7 +703,7 @@ sp_api::decl_runtime_apis! {
 			-> Option<PersistedValidationData<N>>;
 
 		/// Checks if the given validation outputs pass the acceptance criteria.
-		fn check_validation_outputs(para_id: Id, outputs: ValidationOutputs) -> bool;
+		fn check_validation_outputs(para_id: Id, outputs: ValidationOutputs) -> Result<(), CheckValidationOutputsError>;
```
bkchr, line 706: @pepyakin okay?

node/core/backing/src/lib.rs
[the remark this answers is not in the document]
coriolinus: It's because to get to the original tuple, you need to chase it back two layers instead of one. `let` => `Some(r)` => `assignment`. 

I don't want to block this PR on this, but you did change this in a way that is semantically identical, so I'll ask again: why make the change in the first place?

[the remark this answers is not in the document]
bkchr: Because it is less code and easier to read. There is no reason to write more text for this. It makes no sense to write out the entire tuple when you can all of this with one character. 

> It's because to get to the original tuple, you need to chase it back two layers instead of one. let => Some(r) => assignment.

No one reading code in this way, you start with the let and directly see that it expects a tuple. You don't start in the tuple.

primitives/src/v1.rs
```diff
@@ -655,6 +655,22 @@ pub enum CandidateEvent<H = Hash> {
 	CandidateTimedOut(CandidateReceipt<H>, HeadData),
 }
 
+/// Errors that can happen when checking the validation outputs of a parachain candidate.
+#[derive(Clone, Encode, Decode, Copy)]
+#[cfg_attr(feature = "std", derive(PartialEq, Debug, displaydoc::Display))]
+pub enum CheckValidationOutputsError {
+	/// The head data is too large.
+	HeadDataTooLarge,
+	/// The parachain tried to upgrade its code to early.
```
coriolinus, line 664:
```suggestion
	/// The parachain tried to upgrade its code too early.
```

primitives/src/v1.rs
```diff
@@ -655,6 +655,22 @@ pub enum CandidateEvent<H = Hash> {
...
+	PrematureCodeUpgrade,
+	/// The new parachain code is too large.
+	NewCodeTooLarge,
+	/// The parachain didn't handled the downward messages correctly.
```
coriolinus, line 668:
```suggestion
	/// The parachain didn't handle the downward messages correctly.
```

primitives/src/v1.rs
pepyakin: The rationale behind this bool is to minimize potential effects on the runtime API: changing this struct is backward incompatible. My logic was that the detailed programmatic information about the source of the failure is not really important since it doesn't allow for recovery. It is important for diagnostics but that's better handled by logging, I guess.

bkchr: I first had logging from the runtime. However, I was not really happy with it. In the end we also have stuff like transactionvalidityerror as enum that is send over this bound. But I can also change it back.

I think it is quite important to log this stuff, otherwise it is impossible to debug all this. I just found the bug by manually adding printlns etc. And this shouldn't be the solution here ;)

runtime/parachains/src/runtime_api_impl/v1.rs
```diff
@@ -221,8 +221,22 @@ pub fn persisted_validation_data<T: initializer::Trait>(
...
-	<inclusion::Module<T>>::check_validation_outputs(para_id, outputs)
+) -> bool {
+	match <inclusion::Module<T>>::check_validation_outputs(para_id, outputs) {
+		Ok(()) => true,
+		Err(e) => {
+			frame_support::debug::RuntimeLogger::init();
+			let err: &'static str = e.into();
+			log::debug!(
+				target: "candidate_validation",
+				"Validation output checking for parachain `{}` failed: {}",
```
pepyakin, line 232: nit: we use plural

```suggestion
				"Validation outputs checking for parachain `{}` failed: {}",
```

bkchr:
bot merge

PR #1906, opened 2020-11-03
add native logging to check_upward_messages

### LINKED ISSUE #1898: Diagnostics for the ValidationOutputs checker

coriolinus:
This doesn't affect the WASM builds, but ensures that native versions
(such as are used for testing) emit diagnostics anytime the
check_upward_messages function exits, which should reduce the pain
of debugging when something goes wrong.

Closes #1898

### DIFF
--- a/runtime/parachains/src/router/ump.rs
+++ b/runtime/parachains/src/router/ump.rs
@@ -18,7 +18,7 @@ use super::{Trait, Module, Store};
 use crate::configuration::{self, HostConfiguration};
 use sp_std::prelude::*;
 use sp_std::collections::{btree_map::BTreeMap, vec_deque::VecDeque};
-use frame_support::{StorageMap, StorageValue, weights::Weight, traits::Get};
+use frame_support::{StorageMap, StorageValue, weights::Weight, traits::Get, debug::native as log};
 use primitives::v1::{Id as ParaId, UpwardMessage};
 
 /// All upward messages coming from parachains will be funneled into an implementation of this trait.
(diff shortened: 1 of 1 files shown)


runtime/parachains/src/router/ump.rs
```diff
@@ -96,6 +111,20 @@ impl<T: Trait> Module<T> {
 
 		// make sure that the queue is not overfilled.
 		// we do it here only once since returning false invalidates the whole relay-chain block.
+		if para_queue_count > config.max_upward_queue_count {
+			log::warn!(
+				target: LOG_TARGET,
+				"para_queue has more items than permitted by config ({} > {})",
```
pepyakin, line 117:
```suggestion
				"the ump queue has more items than permitted by config ({} > {})",
```

runtime/parachains/src/router/ump.rs
```diff
@@ -96,6 +111,20 @@ impl<T: Trait> Module<T> {
...
+				para_queue_count, config.max_upward_queue_count,
+			);
+		}
+		if para_queue_size > config.max_upward_queue_size {
+			log::warn!(
+				target: LOG_TARGET,
+				"para_queue has grown past the max size permitted by config ({} > {})",
```
pepyakin, line 124:
```suggestion
				"the ump queue has grown past the max size permitted by config ({} > {})",
```

runtime/parachains/src/router/ump.rs
pepyakin: also, it's not yet happened, so I guess it should be "would have" instead of "hash".

pepyakin (approved):
great, thanks!

coriolinus:
bot merge

PR #1911, opened 2020-11-03
Moare fixes for parachains

bkchr:
- Sending data to a job should always contain a relay parent. Done this
for the provisioner
- Fixed the `select_availability_bitfields` function. It was assuming we
have one core per validator, while we only have one core per parachain.
- Drive by async "rewrite" in proposer

### DIFF
--- a/node/core/provisioner/src/lib.rs
+++ b/node/core/provisioner/src/lib.rs
@@ -41,7 +41,7 @@ use polkadot_primitives::v1::{
 	BackedCandidate, BlockNumber, CoreState, Hash, OccupiedCoreAssumption,
 	SignedAvailabilityBitfield,
 };
-use std::{collections::HashMap, convert::TryFrom, pin::Pin};
+use std::{collections::HashSet, convert::TryFrom, pin::Pin};
 use thiserror::Error;
 
 struct ProvisioningJob {
@@ -211,7 +211,7 @@ impl ProvisioningJob {
 				ToJob::Provisioner(RequestBlockAuthorshipData(_, sender)) => {
 					self.provisionable_data_channels.push(sender)
 				}
-				ToJob::Provisioner(ProvisionableData(data)) => {
+				ToJob::Provisioner(ProvisionableData(_, data)) => {
 					let mut bad_indices = Vec::new();
 					for (idx, channel) in self.provisionable_data_channels.iter_mut().enumerate() {
 						match channel.send(data.clone()).await {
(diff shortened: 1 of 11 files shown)


drahnr (approved):

node/core/provisioner/src/lib.rs
```diff
@@ -211,7 +211,7 @@ impl ProvisioningJob {
 				ToJob::Provisioner(RequestBlockAuthorshipData(_, sender)) => {
 					self.provisionable_data_channels.push(sender)
 				}
-				ToJob::Provisioner(ProvisionableData(data)) => {
+				ToJob::Provisioner(ProvisionableData(_, data)) => {
```
coriolinus, line 214: Why add the relay parent to `ProvisionableData` if we just ignore it?

primitives/src/v1.rs
```diff
@@ -623,6 +623,11 @@ impl<N> CoreState<N> {
 			Self::Free => None,
 		}
 	}
+
+	/// Does this core state is `Self::Occupied`?
```
coriolinus, line 627:
```suggestion
	/// Is this core state `Self::Occupied`?
```

node/core/provisioner/src/lib.rs
bkchr: Why is that not obvious? How do you want to know to which job the message needs to be send?

coriolinus: Nvm, you're right; missed the change in `messages.rs`.

node/subsystem/src/messages.rs
```diff
@@ -536,7 +536,7 @@ impl ProvisionerMessage {
 		match self {
 			Self::RequestBlockAuthorshipData(hash, _) => Some(*hash),
 			Self::RequestInherentData(hash, _) => Some(*hash),
-			Self::ProvisionableData(_) => None,
+			Self::ProvisionableData(hash, _) => Some(*hash),
```
coriolinus, line 539: Given that most variants of `ProvisionableData` include a `Hash`, might it make more sense to just delegate to that hash instead of adding a new one to this enum?

https://github.com/paritytech/polkadot/blob/23d3de11a63b5365b7dc88920086bb5211e7306f/node/subsystem/src/messages.rs#L499-L508

bkchr: Yeah I thought the same, but I wanted to make it more explicit here. I can also change it

PR #1912, opened 2020-11-03
Do not validate a candidate in candidate selection

bkchr:
The candidate selection subsystem should not validate a candidate, as
this is done by the backing subsystem on a `Second` request. Otherwise
we validate one candidate twice.


rphmeier (changes requested):
missing corresponding guide changes

rphmeier (approved):

PR #1913, opened 2020-11-03
fix bitfield selection

rphmeier:
It was doing some nonsensical 'bitfield per core' thing which was definitely not written after reading the implementer's guide

### DIFF
--- a/node/core/provisioner/src/lib.rs
+++ b/node/core/provisioner/src/lib.rs
@@ -39,9 +39,10 @@ use polkadot_node_subsystem_util::{
 };
 use polkadot_primitives::v1::{
 	BackedCandidate, BlockNumber, CoreState, Hash, OccupiedCoreAssumption,
-	SignedAvailabilityBitfield,
+	SignedAvailabilityBitfield, ValidatorIndex,
 };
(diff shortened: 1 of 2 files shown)


node/core/provisioner/src/lib.rs
```diff
@@ -324,32 +325,30 @@ fn select_availability_bitfields(
...
-		if !seen_validators.insert(bitfield.validator_index()) {
-			continue;
+	'a:
+	for bitfield in bitfields.iter().cloned() {
+		if bitfield.payload().0.len() != cores.len() {
+			continue
 		}
 
-		for (idx, _) in cores.iter().enumerate().filter(|v| v.1.is_occupied()) {
+		let is_better = selected.get(&bitfield.validator_index())
```
bkchr, line 336: If we accepted the other bitfield and the new one has more ones it must mean that it is invalid? Because it would have a bit set for a free core

rphmeier: why? we don't require that all occupied cores have the bit set.

bkchr: 🤦🤦🤦

bkchr (approved):

PR #1692, opened 2020-09-09
Companion: Handle construct_runtime breaking change.

thiolliere:
companion for: https://github.com/paritytech/substrate/pull/6969

construct_runtime changes the way it encodes Origin, Call, Event, this PR make use of newly introduced indices to make Call unchanged, and migrate scheduler usage of old pallet origin.

This PR does break Event but as far as I know this is not an issue, event should be decoded using metadata.

EDIT: If it needs more test, looking at code, and comparing expansion everything looks good. I can also do some call and origin encoding and decoding between master and this PR if wanted.

## TODO

* [x] double check no module make use of ModuleToIndex in way that could break them.

## Note for reviwer

the breaking change is only about the encoding of Event, Call, Origin, and implementation of ModuleToIndex. But there is no breaking usage of ModuleToIndex (as there is no such usage on polkadot (and substrate only usage is for Errors which should be read from metadata)). Call encoding is not breaking as index are used in a way not to break them, Event are breaking but similarly to Error they should be read from metadata as well.
Only Origin is a breaking change, and its only usage on chain is through scheduler, thus migration code has been written.

PR #1914, opened 2020-11-03
Remove stale upgrades from polkadot + kusama

s3krit:
Copied the same approach as Westend for having an empty `on_runtime_upgrade` function - I hope that's correct


thiolliere (approved):
those runtime upgrade were introduced in https://github.com/paritytech/polkadot/pull/1692/files
Which is part of polkadot runtime version 25 (and same for kusama)

bkchr (approved):

PR #1916, opened 2020-11-04
Make `CandidateHash` a real type

bkchr:
This pr adds a new type `CandidateHash` that is used instead of the
opaque `Hash` type. This helps to ensure on the type system level that
we are passing the correct types.

This pr also fixes wrong usage of `relay_parent` as `candidate_hash`
when communicating with the av storage.

### DIFF
--- a/core-primitives/src/lib.rs
+++ b/core-primitives/src/lib.rs
@@ -52,6 +52,8 @@ pub type Hash = sp_core::H256;
 
 /// Unit type wrapper around [`Hash`] that represents a candidate hash.
 ///
+/// This type is produced by [`CandidateReceipt::hash`].
+///
 /// This type makes it easy to enforce that a hash is a candidate hash on the type level.
 #[derive(Clone, Copy, codec::Encode, codec::Decode, Hash, Eq, PartialEq, Debug, Default)]
 pub struct CandidateHash(pub Hash);
--- a/core-primitives/src/lib.rs
+++ b/core-primitives/src/lib.rs
@@ -50,6 +50,12 @@ pub type ChainId = u32;
 /// A hash of some data used by the relay chain.
 pub type Hash = sp_core::H256;
 
+/// Unit type wrapper around [`Hash`] that represents a candidate hash.
+///
+/// This type makes it easy to enforce that a hash is a candidate hash on the type level.
+#[derive(Clone, Copy, codec::Encode, codec::Decode, Hash, Eq, PartialEq, Debug, Default)]
+pub struct CandidateHash(pub Hash);
+
 /// Index of a transaction in the relay chain. 32-bit should be plenty.
 pub type Nonce = u32;
 
--- a/node/core/av-store/src/lib.rs
+++ b/node/core/av-store/src/lib.rs
@@ -729,7 +729,10 @@ where
 	Ok(())
 }
 
-fn available_data(db: &Arc<dyn KeyValueDB>, candidate_hash: &CandidateHash) -> Option<StoredAvailableData> {
+fn available_data(
+	db: &Arc<dyn KeyValueDB>,
+	candidate_hash: &CandidateHash,
+) -> Option<StoredAvailableData> {
 	query_inner(db, columns::DATA, &available_data_key(candidate_hash))
 }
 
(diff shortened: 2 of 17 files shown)


node/network/availability-distribution/src/lib.rs
```diff
@@ -296,8 +296,9 @@ impl ProtocolState {
 				// remove from the ancestry index
 				self.ancestry.remove(relay_parent);
 				// and also remove the actual receipt
-				self.receipts.remove(relay_parent);
-				self.per_candidate.remove(relay_parent);
+				if let Some(candidates) = self.receipts.remove(relay_parent) {
+					candidates.into_iter().for_each(|c| { self.per_candidate.remove(&c.0); });
```
bkchr, line 300: I think this is actually a bug, as a pending candidate could exists for multiple relay parents? Maybe I'm wrong, I didn't looked closer for now. I just tried to replicate the old behavior.

node/network/availability-distribution/src/lib.rs
```diff
@@ -313,8 +314,9 @@ impl ProtocolState {
 						// remove from the ancestry index
 						self.ancestry.remove(&ancestor);
 						// and also remove the actual receipt
-						self.receipts.remove(&ancestor);
-						self.per_candidate.remove(&ancestor);
+						if let Some(candidates) = self.receipts.remove(&ancestor) {
+							candidates.into_iter().for_each(|c| { self.per_candidate.remove(&c.0); });
```
bkchr, line 318: Same

core-primitives/src/lib.rs
```diff
@@ -50,6 +50,12 @@ pub type ChainId = u32;
 /// A hash of some data used by the relay chain.
 pub type Hash = sp_core::H256;
 
+/// Unit type wrapper around [`Hash`] that represents a candidate hash.
+///
+/// This type makes it easy to enforce that a hash is a candidate hash on the type level.
```
coriolinus, line 55:
```suggestion
/// Unit type wrapper around [`Hash`] that represents a candidate hash.
///
/// This type is produced by [`CandidateReceipt::hash`].
///
/// This type makes it easy to enforce that a hash is a candidate hash on the type level.
```

node/network/availability-distribution/src/lib.rs
montekki: `CandidateReceipt` includes the `CandidateDescriptor` field that in turn references a `relay_parent`. So two candidates with multiple relay parents will not hash to the same value

montekki (approved):

node/network/availability-distribution/src/lib.rs
drahnr: `if descendants.is_empty() {` should prevent this, but given your changes have to look into it again.

coriolinus (approved):
Looks great! I have not audited the rest of the codebase to ensure that there are no further instances of `Hash` which might correctly become `CandidateHash`. I did look at all instances within this PR of that replacement, and agree that they seem appropriate.

node/core/av-store/src/lib.rs
```diff
@@ -729,7 +729,7 @@ where
 	Ok(())
 }
 
-fn available_data(db: &Arc<dyn KeyValueDB>, candidate_hash: &Hash) -> Option<StoredAvailableData> {
+fn available_data(db: &Arc<dyn KeyValueDB>, candidate_hash: &CandidateHash) -> Option<StoredAvailableData> {
```
pepyakin, line 732: Wrap this line?

node/network/availability-distribution/src/lib.rs
```diff
@@ -969,7 +970,7 @@ where
 }
 
 /// Query the proof of validity for a particular candidate hash.
-async fn query_data_availability<Context>(ctx: &mut Context, candidate_hash: Hash) -> Result<bool>
+async fn query_data_availability<Context>(ctx: &mut Context, candidate_hash: CandidateHash) -> Result<bool>
```
pepyakin, line 973: Wrap this line?

node/core/av-store/src/lib.rs
```diff
@@ -564,7 +564,7 @@ where
 				log::trace!(
 					target: LOG_TARGET,
 					"Updating pruning record for finalized block {}",
-					record.candidate_hash,
+					record.block_number,
```
drahnr, line 567: I guess  this eases reading the output?

pepyakin (commented):
some pesky line width nits.

👍 for the newtype

drahnr (approved):

node/core/av-store/src/lib.rs
bkchr: It says "finalized_block" and the candidate hash was never and will never be a finalized block :P

statement-table/src/lib.rs
```diff
@@ -18,77 +18,35 @@ pub mod generic;
 
 pub use generic::{Table, Context};
 
-/// Concrete instantiations suitable for v0 primitives.
-pub mod v0 {
```
rphmeier, line 22: 🙇

rphmeier (approved):

PR #1917, opened 2020-11-04
Introduce runtime upgrade testing crate

seunlanlege:
Using [substrate-test-runner](https://github.com/paritytech/substrate-test-runner)


bkchr:
Does this really needs to be part of the Polkadot repo? Could we not store this in a different repo?

seunlanlege:
Yeah now that i think about it, you're right this should go into another repo where it always depends on the latest polkadot tag.

PR #1918, opened 2020-11-04
Burnin for substrate#7464

tomaka:
https://github.com/paritytech/substrate/pull/7464
Not intended to be merged.

Since `error!` logs aren't detected, I put a `debug-assertions = true` flag in the root `Cargo.toml` in order to trigger all the `debug_assert!(false)` in the networking code in case of an issue.

Also had to do `cargo update -p jsonrpc-core-client -p jsonrpc-derive` for some reason in order for this thing to build.


tomaka:
Burn-in finished a long time ago

PR #1919, opened 2020-11-04
[burn-in]  grandpa: store the authority id that is used for voting per round

andresilva:
https://github.com/paritytech/substrate/pull/7454


andresilva:
Burn-in didn't show any issues.

PR #1920, opened 2020-11-04
Simplify the bitfield signing job

bkchr:
Besides that the pr also adds a simple test.

### DIFF
--- a/node/core/bitfield-signing/src/lib.rs
+++ b/node/core/bitfield-signing/src/lib.rs
@@ -16,20 +16,16 @@
 
 //! The bitfield signing subsystem produces `SignedAvailabilityBitfield`s once per block.
 
-#![deny(unused_crate_dependencies, unused_results)]
+#![deny(unused_crate_dependencies)]
 #![warn(missing_docs)]
+#![recursion_limit="256"]
 
-use bitvec::bitvec;
-use futures::{
-	channel::{mpsc, oneshot},
-	prelude::*,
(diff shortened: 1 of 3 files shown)


node/core/bitfield-signing/src/lib.rs
```diff
@@ -181,85 +169,55 @@ async fn get_core_availability(
...
-				Err(err) => {
-					errs_ref.lock().await.push(err);
-					return;
-				}
-			};
-			out_ref.lock().await.set(idx, availability);
-		})
-		.await;
+	// Wrap the sender in a Mutex to share it between the futures.
+	let sender = Mutex::new(sender);
```
montekki, line 212: Why not `clone()`?

montekki (approved):

coriolinus (approved):

node/core/bitfield-signing/src/lib.rs
drahnr: That would increase the capacity by 1 for each clone.

bkchr: Exactly, this would mean that you can always send at least one item.

montekki:
bot merge

node/core/bitfield-signing/src/lib.rs
pepyakin: Could one of you guys please send a follow-up to leave a comment for this? I think it is important.

PR #1921, opened 2020-11-05
Make sure we clear the freed the availability cores

bkchr:
If a core is freed, we need to make sure it is inserted as free into
availability cores, because we can not be sure that the core is occupied
directly again.


runtime/parachains/src/scheduler.rs
```diff
@@ -495,6 +495,7 @@ impl<T: Trait> Module<T> {
 
 		Scheduled::set(scheduled);
 		ParathreadQueue::set(parathread_queue);
+		AvailabilityCores::set(cores);
```
rphmeier, line 498: 😄

rphmeier (approved):

runtime/parachains/src/scheduler.rs
bkchr: Yeah, in the end everything was there :D

PR #1926, opened 2020-11-06
Improve diagnostics for the ValidationOutputs checker / inclusion

### LINKED ISSUE #1898: Diagnostics for the ValidationOutputs checker

pepyakin:
Closes #1898

This moves from logging statements in modules like HRMP, UMP to a dedicated enums that are printed centrally in the inclusion module. If an error happens during processing inclusion inherent, the number of the candidate will be reported as well.

### DIFF
--- a/runtime/parachains/src/inclusion.rs
+++ b/runtime/parachains/src/inclusion.rs
@@ -773,7 +773,7 @@ impl<T: Trait> CandidateCheckContext<T> {
 	) -> Result<(), AcceptanceCheckErr<T::BlockNumber>> {
 		ensure!(
 			head_data.0.len() <= self.config.max_head_data_size as _,
-			AcceptanceCheckErr::HeadDataTooLarge
+			AcceptanceCheckErr::HeadDataTooLarge,
 		);
 
 		// if any, the code upgrade attempt is allowed.
@@ -786,11 +786,11 @@ impl<T: Trait> CandidateCheckContext<T> {
 				});
 			ensure!(
 				valid_upgrade_attempt,
-				AcceptanceCheckErr::PrematureCodeUpgrade
+				AcceptanceCheckErr::PrematureCodeUpgrade,
 			);
 			ensure!(
 				new_validation_code.0.len() <= self.config.max_code_size as _,
-				AcceptanceCheckErr::NewCodeTooLarge
+				AcceptanceCheckErr::NewCodeTooLarge,
 			);
 		}
 
--- a/runtime/parachains/src/inclusion.rs
+++ b/runtime/parachains/src/inclusion.rs
@@ -186,8 +186,9 @@ decl_module! {
 	}
 }
 
-impl<T: Trait> Module<T> {
+const LOG_TARGET: &str = "parachains_runtime_inclusion";
 
+impl<T: Trait> Module<T> {
 	/// Block initialization logic, called by initializer.
 	pub(crate) fn initializer_initialize(_now: T::BlockNumber) -> Weight { 0 }
 
@@ -400,7 +401,7 @@ impl<T: Trait> Module<T> {
 			// In the meantime, we do certain sanity checks on the candidates and on the scheduled
 			// list.
 			'a:
-			for candidate in &candidates {
+			for (candidate_idx, candidate) in candidates.iter().enumerate() {
 				let para_id = candidate.descriptor().para_id;
 
 				// we require that the candidate is in the context of the parent block.
@@ -413,15 +414,26 @@ impl<T: Trait> Module<T> {
 					Error::<T>::NotCollatorSigned,
 				);
 
-				check_cx.check_validation_outputs(
-					para_id,
-					&candidate.candidate.commitments.head_data,
-					&candidate.candidate.commitments.new_validation_code,
-					candidate.candidate.commitments.processed_downward_messages,
-					&candidate.candidate.commitments.upward_messages,
-					T::BlockNumber::from(candidate.candidate.commitments.hrmp_watermark),
-					&candidate.candidate.commitments.horizontal_messages,
-				)?;
+				if let Err(err) = check_cx
+					.check_validation_outputs(
+						para_id,
+						&candidate.candidate.commitments.head_data,
+						&candidate.candidate.commitments.new_validation_code,
+						candidate.candidate.commitments.processed_downward_messages,
+						&candidate.candidate.commitments.upward_messages,
+						T::BlockNumber::from(candidate.candidate.commitments.hrmp_watermark),
+						&candidate.candidate.commitments.horizontal_messages,
+					)
+				{
+					log::debug!(
+						target: LOG_TARGET,
+						"Validation outputs checking during inclusion of a candidate {} for parachain `{}` failed: {:?}",
+						candidate_idx,
+						u32::from(para_id),
+						err,
+					);
+					Err(err.strip_into_dispatch_err::<T>())?;
+				};
 
 				for (i, assignment) in scheduled[skip..].iter().enumerate() {
 					check_assignment_in_order(assignment)?;
@@ -542,13 +554,11 @@ impl<T: Trait> Module<T> {
 	}
 
 	/// Run the acceptance criteria checks on the given candidate commitments.
-	///
-	/// Returns an 'Err` if any of the checks doesn't pass.
 	pub(crate) fn check_validation_outputs(
 		para_id: ParaId,
 		validation_outputs: primitives::v1::ValidationOutputs,
-	) -> Result<(), DispatchError> {
-		CandidateCheckContext::<T>::new().check_validation_outputs(
+	) -> bool {
+		if let Err(err) = CandidateCheckContext::<T>::new().check_validation_outputs(
 			para_id,
 			&validation_outputs.head_data,
 			&validation_outputs.new_validation_code,
@@ -556,7 +566,18 @@ impl<T: Trait> Module<T> {
 			&validation_outputs.upward_messages,
 			T::BlockNumber::from(validation_outputs.hrmp_watermark),
 			&validation_outputs.horizontal_messages,
-		)
+		) {
+			frame_support::debug::RuntimeLogger::init();
+			log::debug!(
+				target: LOG_TARGET,
+				"Validation outputs checking for parachain `{}` failed: {:?}",
+				u32::from(para_id),
+				err,
+			);
+			false
+		} else {
+			true
+		}
 	}
 
 	fn enact_candidate(
@@ -692,6 +713,34 @@ const fn availability_threshold(n_validators: usize) -> usize {
 	threshold
 }
 
+#[derive(derive_more::From, Debug)]
+enum AcceptanceCheckErr<BlockNumber> {
+	HeadDataTooLarge,
+	PrematureCodeUpgrade,
+	NewCodeTooLarge,
+	ProcessedDownwardMessages(router::ProcessedDownwardMessagesAcceptanceErr),
+	UpwardMessages(router::UpwardMessagesAcceptanceCheckErr),
+	HrmpWatermark(router::HrmpWatermarkAcceptanceErr<BlockNumber>),
+	OutboundHrmp(router::OutboundHrmpAcceptanceErr),
+}
+
+impl<BlockNumber> AcceptanceCheckErr<BlockNumber> {
+	/// Returns the same error so that it can be threaded through a needle of `DispatchError` and
+	/// ultimately returned from a `Dispatchable`.
+	fn strip_into_dispatch_err<T: Trait>(self) -> Error<T> {
+		use AcceptanceCheckErr::*;
+		match self {
+			HeadDataTooLarge => Error::<T>::HeadDataTooLarge,
+			PrematureCodeUpgrade => Error::<T>::PrematureCodeUpgrade,
+			NewCodeTooLarge => Error::<T>::NewCodeTooLarge,
+			ProcessedDownwardMessages(_) => Error::<T>::IncorrectDownwardMessageHandling,
+			UpwardMessages(_) => Error::<T>::InvalidUpwardMessages,
+			HrmpWatermark(_) => Error::<T>::HrmpWatermarkMishandling,
+			OutboundHrmp(_) => Error::<T>::InvalidOutboundHrmp,
+		}
+	}
+}
+
 /// A collection of data required for checking a candidate.
 struct CandidateCheckContext<T: Trait> {
 	config: configuration::HostConfiguration<T::BlockNumber>,
@@ -720,10 +769,10 @@ impl<T: Trait> CandidateCheckContext<T> {
 		upward_messages: &[primitives::v1::UpwardMessage],
 		hrmp_watermark: T::BlockNumber,
 		horizontal_messages: &[primitives::v1::OutboundHrmpMessage<ParaId>],
-	) -> Result<(), DispatchError> {
+	) -> Result<(), AcceptanceCheckErr<T::BlockNumber>> {
 		ensure!(
 			head_data.0.len() <= self.config.max_head_data_size as _,
-			Error::<T>::HeadDataTooLarge
+			AcceptanceCheckErr::HeadDataTooLarge
 		);
 
 		// if any, the code upgrade attempt is allowed.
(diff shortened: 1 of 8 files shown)


runtime/parachains/src/inclusion.rs
```diff
@@ -413,15 +414,26 @@ impl<T: Trait> Module<T> {
...
+					)
+				{
+					log::debug!(
+						target: LOG_TARGET,
+						"Validation outputs checking during inclusion of a candidate {} for parachain `{}` failed: {:?}",
+						candidate_idx,
+						u32::from(para_id),
+						err,
+					);
+					Err(err.strip_into_dispatch_err::<T>())?;
```
coriolinus, line 435: Is it desirable to continue checking candidate to determine all candidate validation failures, which can be collected into a single item, or is it fine to terminate early at the first candidate validation failure (current behavior)?

pepyakin: Good idea for checking multiple candidates at once!

But I don't believe this is the current behavior though.

pepyakin: I guess in this case, we would have to return the first or the last failure as a dispatch err.

runtime/parachains/src/inclusion.rs
```diff
@@ -692,6 +713,34 @@ const fn availability_threshold(n_validators: usize) -> usize {
...
+	ProcessedDownwardMessages(router::ProcessedDownwardMessagesAcceptanceErr),
+	UpwardMessages(router::UpwardMessagesAcceptanceCheckErr),
+	HrmpWatermark(router::HrmpWatermarkAcceptanceErr<BlockNumber>),
+	OutboundHrmp(router::OutboundHrmpAcceptanceErr),
+}
+
+impl<BlockNumber> AcceptanceCheckErr<BlockNumber> {
+	/// Returns the same error so that it can be threaded through a needle of `DispatchError` and
+	/// ultimately returned from a `Dispatchable`.
+	fn strip_into_dispatch_err<T: Trait>(self) -> Error<T> {
```
coriolinus, line 730: 👍 I like this strategy.

runtime/parachains/src/inclusion.rs
coriolinus: Yeah, I had in my head that we could return like a `Err(Vec<u32>)` with the indices, but if we have to strip that out as a dispatch error anyway, then there's not much point.

runtime/parachains/src/inclusion.rs
```diff
@@ -413,15 +414,26 @@ impl<T: Trait> Module<T> {
...
+						para_id,
+						&candidate.candidate.commitments.head_data,
+						&candidate.candidate.commitments.new_validation_code,
+						candidate.candidate.commitments.processed_downward_messages,
+						&candidate.candidate.commitments.upward_messages,
+						T::BlockNumber::from(candidate.candidate.commitments.hrmp_watermark),
+						&candidate.candidate.commitments.horizontal_messages,
+					)
+				{
+					log::debug!(
```
bkchr, line 428: This would only be printed on native but also compiled for wasm :P

runtime/parachains/src/inclusion.rs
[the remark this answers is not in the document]
pepyakin: I deliberately went with a inherent method rather than a `From` impl. This way we have an opportunity to leave a comment and IMO `. strip_into_dispatch_err` gives a little bit better sense of what would happen compared to a `.into()`.

runtime/parachains/src/inclusion.rs
pepyakin: oh, why is that? Is it because the logger is not initialized?

bkchr: Yeah

runtime/parachains/src/inclusion.rs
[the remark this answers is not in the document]
bkchr: You can also put a comment above a `From` implementation :shrug: 

Plus, you will need to do `Error::<T>::from` anyway.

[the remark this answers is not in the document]
pepyakin: True, but I still prefer a specific `.strip_into_dispatch_err()` to generic `.into` (or `from`) at the call site and the comment would show up in the rustdoc (if you are using a proper IDE anyway lol).

runtime/parachains/src/inclusion.rs
pepyakin: That's a good catch, thanks. Is there a way to make it work on wasm? 

Or do you recommend to switch to only native logging? I was shooting to get logging in both native and wasm and the cost of the whole PR doesn't seem too big: the binary size of this PR compared to master is ~1000 bytes.

runtime/parachains/src/inclusion.rs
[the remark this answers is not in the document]
bkchr: I mean you would need to write:
```
Error::<T>::from(err)
```

Just calling `err.into()` would not work.

runtime/parachains/src/inclusion.rs
bkchr: I'm fine with logging from wasm.

We would need to initialize the `RuntimeLogger`. Like I had done it in the runtime api

runtime/parachains/src/inclusion.rs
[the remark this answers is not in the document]
pepyakin: Ah, yeah, gotcha.

runtime/parachains/src/inclusion.rs
pepyakin: Ok, I can slap a call to `RuntimeLogger::init` just before logging.

But that feels such an ad-hoc place to initialize a logger. Ideally we do that only once, close to the entrypoint into the runtime. I know that initializing the logger twice won't harm, but that nonetheless feels non elegant. Something that we should think about

runtime/parachains/src/inclusion.rs
```diff
@@ -734,45 +784,28 @@ impl<T: Trait> CandidateCheckContext<T> {
 						&& self.relay_parent_number.saturating_sub(last)
 							>= self.config.validation_upgrade_frequency
 				});
-			ensure!(valid_upgrade_attempt, Error::<T>::PrematureCodeUpgrade);
+			ensure!(
+				valid_upgrade_attempt,
+				AcceptanceCheckErr::PrematureCodeUpgrade
```
bkchr, line 789:
```suggestion
				AcceptanceCheckErr::PrematureCodeUpgrade,
```

runtime/parachains/src/inclusion.rs
```diff
@@ -720,10 +770,10 @@ impl<T: Trait> CandidateCheckContext<T> {
 		upward_messages: &[primitives::v1::UpwardMessage],
 		hrmp_watermark: T::BlockNumber,
 		horizontal_messages: &[primitives::v1::OutboundHrmpMessage<ParaId>],
-	) -> Result<(), DispatchError> {
+	) -> Result<(), AcceptanceCheckErr<T::BlockNumber>> {
 		ensure!(
 			head_data.0.len() <= self.config.max_head_data_size as _,
-			Error::<T>::HeadDataTooLarge
+			AcceptanceCheckErr::HeadDataTooLarge
```
bkchr, line 776:
```suggestion
			AcceptanceCheckErr::HeadDataTooLarge,
```

runtime/parachains/src/inclusion.rs
```diff
@@ -734,45 +784,28 @@ impl<T: Trait> CandidateCheckContext<T> {
...
 				});
-			ensure!(valid_upgrade_attempt, Error::<T>::PrematureCodeUpgrade);
+			ensure!(
+				valid_upgrade_attempt,
+				AcceptanceCheckErr::PrematureCodeUpgrade
+			);
 			ensure!(
 				new_validation_code.0.len() <= self.config.max_code_size as _,
-				Error::<T>::NewCodeTooLarge
+				AcceptanceCheckErr::NewCodeTooLarge
```
bkchr, line 793:
```suggestion
				AcceptanceCheckErr::NewCodeTooLarge,
```

runtime/parachains/src/router/dmp.rs
```diff
@@ -18,6 +18,7 @@ use super::{Trait, Module, Store};
 use crate::configuration::HostConfiguration;
 use frame_support::{StorageMap, weights::Weight, traits::Get};
 use sp_std::prelude::*;
+use sp_std::fmt;
```
bkchr, line 21:
```suggestion
use sp_std::{fmt, prelude::*};
```

runtime/parachains/src/router/dmp.rs
```diff
@@ -28,6 +29,38 @@ pub enum QueueDownwardMessageError {
 	ExceedsMaxMessageSize,
 }
 
+/// An error returned by `check_processed_downward_messages` that indicates an acceptance check
```
bkchr, line 32:
```suggestion
/// An error returned by [`check_processed_downward_messages`] that indicates an acceptance check
```

runtime/parachains/src/router/ump.rs
```diff
@@ -17,8 +17,9 @@
 use super::{Trait, Module, Store};
 use crate::configuration::{self, HostConfiguration};
 use sp_std::prelude::*;
+use sp_std::fmt;
```
bkchr, line 20:
```suggestion
use sp_std::{fmt, prelude::*};
```

runtime/parachains/src/router/ump.rs
```diff
@@ -50,7 +51,58 @@ impl UmpSink for () {
 	}
 }
 
-const LOG_TARGET: &str = "runtime-parachains::upward-messages";
+/// An error returned by `check_upward_messages` that indicates a violation of one of acceptance
```
bkchr, line 54:
```suggestion
/// An error returned by [`check_upward_messages`] that indicates a violation of one of acceptance
```

runtime/parachains/src/router/ump.rs
```diff
@@ -50,7 +51,58 @@ impl UmpSink for () {
...
+			),
+			AcceptanceCheckErr::CapacityExceeded { count, limit } => write!(
+				fmt,
+				"the ump queue would have more items than permitted by config ({} > {})",
+				count, limit,
+			),
+			AcceptanceCheckErr::TotalSizeExceeded { total_size, limit } => write!(
+				fmt,
+				"the ump queue would have grown past the max size permitted by config ({} > {})",
+				total_size, limit,
```
bkchr, line 101:
```suggestion
				sent,
				permitted,
			),
			AcceptanceCheckErr::MessageSize {
				idx,
				msg_size,
				max_size,
			} => write!(
				fmt,
				"upward message idx {} larger than permitted by config ({} > {})",
				idx,
				msg_size,
				max_size,
			),
			AcceptanceCheckErr::CapacityExceeded { count, limit } => write!(
				fmt,
				"the ump queue would have more items than permitted by config ({} > {})",
				count,
				limit,
			),
			AcceptanceCheckErr::TotalSizeExceeded { total_size, limit } => write!(
				fmt,
				"the ump queue would have grown past the max size permitted by config ({} > {})",
				total_size, 
				limit,
```

runtime/parachains/src/router/hrmp.rs
```diff
@@ -14,22 +14,20 @@
...
 use primitives::v1::{
 	Balance, Hash, HrmpChannelId, Id as ParaId, InboundHrmpMessage, OutboundHrmpMessage,
 	SessionIndex,
 };
 use sp_runtime::traits::{BlakeTwo256, Hash as HashT};
-use sp_std::collections::{btree_set::BTreeSet, btree_map::BTreeMap};
-use sp_std::mem;
+use sp_std::collections::{btree_map::BTreeMap, btree_set::BTreeSet};
+use sp_std::{mem, fmt};
 use sp_std::prelude::*;
```
bkchr, line 31:
```suggestion
use sp_std::{mem, fmt, collections::{btree_map::BTreeMap, btree_set::BTreeSet}, prelude::*};
```

runtime/parachains/src/router/hrmp.rs
```diff
@@ -79,7 +77,130 @@ pub struct HrmpChannel {
 	pub mqc_head: Option<Hash>,
 }
 
-const LOG_TARGET: &str = "runtime-parachains::hrmp";
+/// An error returned by `check_hrmp_watermark` that indicates an acceptance criteria check
```
bkchr, line 80:
```suggestion
/// An error returned by [`check_hrmp_watermark`] that indicates an acceptance criteria check
```

runtime/parachains/src/router/hrmp.rs
```diff
@@ -79,7 +77,130 @@ pub struct HrmpChannel {
...
+	AheadRelayParent {
+		new_watermark: BlockNumber,
+		relay_chain_parent_number: BlockNumber,
+	},
+	LandsOnBlockWithNoMessages {
+		new_watermark: BlockNumber,
+	},
+}
+
+/// An error returned by `check_outbound_hrmp` that indicates an acceptance criteria check
```
bkchr, line 96:
```suggestion
/// An error returned by [`check_outbound_hrmp`] that indicates an acceptance criteria check
```

runtime/parachains/src/router/hrmp.rs
```diff
@@ -79,7 +77,130 @@ pub struct HrmpChannel {
...
+				relay_chain_parent_number,
+			} => write!(
+				fmt,
+				"the HRMP watermark is ahead the relay-parent ({:?} > {:?})",
+				new_watermark, relay_chain_parent_number
+			),
+			LandsOnBlockWithNoMessages { new_watermark } => write!(
+				fmt,
+				"the HRMP watermark ({:?}) doesn't land on a block with messages received",
+				new_watermark
```
bkchr, line 153:
```suggestion
				new_watermark,
				last_watermark,
			),
			AheadRelayParent {
				new_watermark,
				relay_chain_parent_number,
			} => write!(
				fmt,
				"the HRMP watermark is ahead the relay-parent ({:?} > {:?})",
				new_watermark,
				relay_chain_parent_number,
			),
			LandsOnBlockWithNoMessages { new_watermark } => write!(
				fmt,
				"the HRMP watermark ({:?}) doesn't land on a block with messages received",
				new_watermark,
```

runtime/parachains/src/router/hrmp.rs
```diff
@@ -79,7 +77,130 @@ pub struct HrmpChannel {
...
+			),
+			NotSorted { idx } => write!(
+				fmt,
+				"the HRMP messages are not sorted (first unsorted is at index {})",
+				idx,
+			),
+			NoSuchChannel { idx, channel_id } => write!(
+				fmt,
+				"the HRMP message at index {} is sent to a non existent channel {:?}->{:?}",
+				idx, channel_id.sender, channel_id.recipient,
```
bkchr, line 176:
```suggestion
				idx,
				channel_id.sender,
				channel_id.recipient,
```

runtime/parachains/src/router/hrmp.rs
```diff
@@ -79,7 +77,130 @@ pub struct HrmpChannel {
...
+}
+
+impl fmt::Debug for OutboundHrmpAcceptanceErr {
+	fn fmt(&self, fmt: &mut fmt::Formatter) -> fmt::Result {
+		use OutboundHrmpAcceptanceErr::*;
+		match self {
+			MoreMessagesThanPermitted { sent, permitted } => write!(
+				fmt,
+				"more HRMP messages than permitted by config ({} > {})",
+				sent, permitted,
```
bkchr, line 166:
```suggestion
				sent,
				permitted,
```

runtime/parachains/src/router/hrmp.rs
```diff
@@ -79,7 +77,130 @@ pub struct HrmpChannel {
...
+				idx, channel_id.sender, channel_id.recipient,
+			),
+			MaxMessageSizeExceeded {
+				idx,
+				msg_size,
+				max_size,
+			} => write!(
+				fmt,
+				"the HRMP message at index {} exceeds the negotiated channel maximum message size ({} > {})",
+				idx, msg_size, max_size,
```
bkchr, line 185:
```suggestion
				idx,
				msg_size,
				max_size,
```

runtime/parachains/src/router/hrmp.rs
```diff
@@ -79,7 +77,130 @@ pub struct HrmpChannel {
...
+			),
+			TotalSizeExceeded {
+				idx,
+				total_size,
+				limit,
+			} => write!(
+				fmt,
+				"sending the HRMP message at index {} would exceed the neogitiated channel total size  ({} > {})",
+				idx, total_size, limit,
```
bkchr, line 194:
```suggestion
				idx,
				total_size,
				limit,
```

runtime/parachains/src/router/hrmp.rs
```diff
@@ -79,7 +77,130 @@ pub struct HrmpChannel {
...
+			),
+			CapacityExceeded { idx, count, limit } => write!(
+				fmt,
+				"sending the HRMP message at index {} would exceed the neogitiated channel capacity  ({} > {})",
+				idx, count, limit,
```
bkchr, line 199:
```suggestion
				idx,
				count,
				limit,
```

bkchr (approved):

pepyakin:
bot merge

PR #1928, opened 2020-11-07
Adds integration test based on adder collator

bkchr:
This adds an integration test for parachains that uses the adder
collator. The test will start two relay chain nodes and one collator and
waits until 4 blocks are build and enacted by the parachain.

### DIFF
--- a/cli/src/command.rs
+++ b/cli/src/command.rs
@@ -219,11 +219,11 @@ pub fn run() -> Result<()> {
 		Some(Subcommand::ValidationWorker(cmd)) => {
 			let _ = sc_cli::init_logger("", sc_tracing::TracingReceiver::Log, None);
 
-			if cfg!(feature = "browser") {
+			if cfg!(feature = "browser") || cfg!(target_os = "android") {
 				Err(sc_cli::Error::Input("Cannot run validation worker in browser".into()))
 			} else {
-				#[cfg(all(not(feature = "browser"), not(feature = "service-rewr")))]
-				service::run_validation_worker(&cmd.mem_id)?;
+				#[cfg(not(all(target_os = "android", feature = "browser")))]
+				polkadot_parachain::wasm_executor::run_worker(&cmd.mem_id)?;
 				Ok(())
 			}
 		},
--- a/cli/src/command.rs
+++ b/cli/src/command.rs
@@ -222,7 +222,7 @@ pub fn run() -> Result<()> {
 			if cfg!(feature = "browser") || cfg!(target_os = "android") {
 				Err(sc_cli::Error::Input("Cannot run validation worker in browser".into()))
 			} else {
-				#[cfg(not(all(target_os = "android", feature = "browser")))]
+				#[cfg(not(any(target_os = "android", feature = "browser")))]
 				polkadot_parachain::wasm_executor::run_worker(&cmd.mem_id)?;
 				Ok(())
 			}
--- a/node/test/service/src/lib.rs
+++ b/node/test/service/src/lib.rs
@@ -23,17 +23,22 @@ pub mod chain_spec;
 pub use chain_spec::*;
 use futures::future::Future;
 use polkadot_overseer::OverseerHandler;
-use polkadot_primitives::v1::{Id as ParaId, HeadData, ValidationCode, Balance};
+use polkadot_primitives::v1::{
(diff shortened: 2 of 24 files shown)


runtime/test-runtime/src/lib.rs
```diff
@@ -24,8 +24,9 @@ use pallet_transaction_payment::CurrencyAdapter;
 use sp_std::prelude::*;
 use codec::Encode;
 use polkadot_runtime_parachains::{
-	configuration,
+	configuration as parachains_configuration,
```
pepyakin, line 27: oh, that's unfortunate, does this have anything to do with added `Config`?

parachain/test-parachains/adder/collator/src/lib.rs
```diff
@@ -123,6 +142,20 @@ impl Collator {
 			async move { Some(collation) }.boxed()
 		})
 	}
+
+	/// Wait that `blocks` are being build and enacted.
```
pepyakin, line 146:
```suggestion
	/// Wait until `blocks` are built and enacted.
```

node/test/service/src/lib.rs
```diff
@@ -255,26 +300,49 @@ impl PolkadotTestNode {
...
 	pub fn wait_for_blocks(&self, count: usize) -> impl Future<Output = ()> {
 		self.client.wait_for_blocks(count)
 	}
+
+	/// Register the collator functionality in the overseer of this node.
+	pub async fn register_collator(
+		&mut self,
+		collator_key: CollatorPair,
+		para_id: ParaId,
+		collator: Box<dyn Fn(Hash, &ValidationData) -> Pin<Box<dyn Future<Output = Option<Collation>> + Send>> + Send + Sync>,
```
pepyakin, line 329: Would be nice to wrap this line

pepyakin:
Looks good, some nits.

runtime/test-runtime/src/lib.rs
bkchr: No, just wanted similar naming as in rococo.

pepyakin: ah, ok, 
we should go over all of them then.

It's ok for the later PR though.

cli/src/command.rs
```diff
@@ -219,11 +219,11 @@ pub fn run() -> Result<()> {
...
 			let _ = sc_cli::init_logger("", sc_tracing::TracingReceiver::Log, None);
 
-			if cfg!(feature = "browser") {
+			if cfg!(feature = "browser") || cfg!(target_os = "android") {
 				Err(sc_cli::Error::Input("Cannot run validation worker in browser".into()))
 			} else {
-				#[cfg(all(not(feature = "browser"), not(feature = "service-rewr")))]
-				service::run_validation_worker(&cmd.mem_id)?;
+				#[cfg(not(all(target_os = "android", feature = "browser")))]
+				polkadot_parachain::wasm_executor::run_worker(&cmd.mem_id)?;
```
coriolinus, line 226: Doesn't matter a ton because we're in the `else` clause, but the `cfg` statement should use `not(any(` instead of `not(all(`, in order to comply with the `||` in the `if` statement.

cli/src/command.rs
```diff
@@ -219,11 +219,11 @@ pub fn run() -> Result<()> {
 		Some(Subcommand::ValidationWorker(cmd)) => {
 			let _ = sc_cli::init_logger("", sc_tracing::TracingReceiver::Log, None);
 
-			if cfg!(feature = "browser") {
+			if cfg!(feature = "browser") || cfg!(target_os = "android") {
 				Err(sc_cli::Error::Input("Cannot run validation worker in browser".into()))
 			} else {
-				#[cfg(all(not(feature = "browser"), not(feature = "service-rewr")))]
-				service::run_validation_worker(&cmd.mem_id)?;
+				#[cfg(not(all(target_os = "android", feature = "browser")))]
```
bkchr, line 225:
```suggestion
				#[cfg(not(any(target_os = "android", feature = "browser")))]
```

coriolinus (approved):
I like the test implementation; it came together more cleanly than I'd realized was possible.

pepyakin (approved):

pepyakin:
bot merge

PR #1930, opened 2020-11-08
Implementer's guide: Assignments counting procedure

burdges:
Alters the assignments counting procedure from #1691 to take at least one tranche per no-show.  This changes a few things:
- We're now forced to keep expected assignees per tranche small, like 1-2, except for tranche zero.  We planned this anyways because if prevents an adversary from bruit forcing their bad assignees into having lower tranches than many honest assignees.  
- We consume more assignees resources but only slightly more and only under niche scenarios. 
- We're much less vulnerable to an adversary with as nearly unlimited capability to DoS honest assignees:  Imagine they wait until they've many bad assignees into low tranches and then just DoS all the honest ones, making them no-shows.  We previously counted these bad assignees towards replacing the honest DoSed no-shows, but now we add an expected 1-2 honest checkers per tranche aka per no-show.

I needed to write `check_approval` and `tranches_to_approve` as basically one method [`advance_assignee_status`](https://github.com/paritytech/polkadot/pull/1558/files#diff-6c8c2073cb884813e271487a2873c2849331ad0e63a23409df3d65e6b62e38c5R281) in the pseduo-code #1558 so they'll turn out similar here I think, but right now this suffices I guess.

---

@rphmeier  There is another "iterating VRFs for no-shows" scheme that's stronger under attack by a powerful DoS adversary:  We could cover no-shows, not by using additional tranches, but by running fresh VRFs whose story is the no-show, meaning this second (third) order VRF input is the VRF output of the no-show, and it returns a tranche.  

In this, adversary knows literally nothing about the no-show checks until they happen, which makes DoS attacks almost worthless, excepting in conjunction with relay chain equivocation of course.  Yet, there are several serious down sides for this iterating VRFs scheme: 
 1. We'd replace honest no-shows very slowly because we've no analog of the fat zeroth tranche here, so like maybe a single no-show delays finality for minutes!  
 2. We'd complicate the code with yet more gossip messages, etc.  I know `tranches_to_approve` looks complex, but it's a single loop for which we can write extensive unit tests, while this iterating VRFs requires more complex tests that cover the gossip messages, etc. 
 3. We might address 1 with a VRF bidding scheme where anyone with low results sends a message, but this makes 2 much much worse.  And it worsens our relay chain equivocation situation somewhat.

I'm therefore proposing the "simpler" scheme in this PR instead of iterating VRFs for no-shows.


roadmap/implementers-guide/src/node/approval/approval-voting.md
```diff
@@ -207,16 +207,16 @@ On receiving an `ApprovedAncestor(Hash, BlockNumber, response_channel)`:
...
-    * Count no-shows in tranches `k..l` and for each of those, take tranches until all no-shows are covered. Repeat so on until either
-      * We run out of tranches to take, having not received any assignments past a certain point. In this case we set `n_tranches` to a special value `ALL` which indicates that new assignments are needed.
-      * All no-shows are covered. Set `n_tranches` to the number of tranches taken
-    * return `n_tranches`
+    1. First, set `t := session_info.needed_approvals`.  Set the base tranche `l=0`.
+    2. Take assignments from tranches `l..` until we have at least `t` assignments.  Let `k` denote the highest tranche taken.  Count the number `assigned` of assignments taken in tranches `0..k`.  If `assigned < t` then return a special value `ALL` which indicates we wait for more assignments. 
+    3. If `assigned > 128` then return the candidate as unfinalizable and advise block production to build another fork from the inclusion relay parent.  This condition can be reverted if some no-shows turnning up eventually.
```
burdges, line 212: We could discuss omitting this check for now if anyone is worried about it adding complexity.  We're approaching breaks in underlying security assumptions if `assigned` gets so large anyways.

roadmap/implementers-guide/src/node/approval/approval-voting.md
```diff
@@ -180,6 +180,8 @@ On receiving a `CheckAndImportApproval(indirect_approval_vote, response_channel)
 
 #### `ApprovalVotingMessage::ApprovedAncestor`
 
+We record final approvals on-chain both so that we can pay the approval voters, and to map GRANDPA votes to actual finalized blocks.
```
rphmeier, line 183: Omit this line

roadmap/implementers-guide/src/node/approval/approval-voting.md
```diff
@@ -192,7 +194,7 @@ On receiving an `ApprovedAncestor(Hash, BlockNumber, response_channel)`:
 
 #### `import_checked_assignment`
   * Load the candidate in question and access the `approval_entry` for the block hash the cert references.
-  * Ensure the validator index is not part of the backing group for the candidate.
+  * Ensure the validator index is not part of the backing group for the candidate.  We count late backing votes via the backing system.
```
rphmeier, line 197: We don't, at the moment

roadmap/implementers-guide/src/node/approval/approval-voting.md
```diff
@@ -205,16 +207,16 @@ On receiving an `ApprovedAncestor(Hash, BlockNumber, response_channel)`:
...
-    * Count no-shows in tranches `k..l` and for each of those, take tranches until all no-shows are covered. Repeat so on until either
-      * We run out of tranches to take, having not received any assignments past a certain point. In this case we set `n_tranches` to a special value `ALL` which indicates that new assignments are needed.
-      * All no-shows are covered. Set `n_tranches` to the number of tranches taken
-    * return `n_tranches`
+    1. First, set `t := session_info.needed_approvals`.  Set the base tranche `l=0`.
+    2. Take assignments from tranches `l..` until we have at least `t` assignments.  Let `k` denote the highest tranche taken.  Count the number `assigned` of assignments taken in tranches `0..k`.  If `assigned < t` then return a special value `ALL` which indicates we wait for more assignments. 
+    3. If `assigned > 128` then return the candidate as unfinalizable and advise block production to build another fork from the inclusion relay parent.  This condition can be reverted if some no-shows turnning up eventually.
```
rphmeier, line 212: > If `assigned > 128` then return the candidate as unfinalizable

This should say to return `ALL`.

> advise block production to build another fork from the inclusion relay parent

Omit this - it's underspecified at the moment. Maybe add a TODO and a GitHub issue for blacklisting those forks.

roadmap/implementers-guide/src/node/approval/approval-voting.md
```diff
@@ -205,16 +207,16 @@ On receiving an `ApprovedAncestor(Hash, BlockNumber, response_channel)`:
...
+    4. Count the number `noshows` of no-shows in tranches `l..k`.  If `noshows` is zero then return success with `n_tranches := k`.  Of course this happens early and does not indicate final termination, as we may later return `ALL` after more no-shows, but if all these assigned checkers vote valid then we are done.
```
rphmeier, line 213: > Of course this happens early and does not indicate final termination, as we may later return `ALL` after more no-shows, but if all these assigned checkers vote valid then we are done.

This sounds like logic beyond the scope of the function - `check_approval` is the part that checks approval.

roadmap/implementers-guide/src/node/approval/approval-voting.md
```diff
@@ -205,16 +207,16 @@ On receiving an `ApprovedAncestor(Hash, BlockNumber, response_channel)`:
...
+    5. For each no-show in `noshows`, we require both another checker and another tranche, which ever means more tranches.  Take assignments from at least `no_shows` subsequent tranches and then if we have not yet covered all noshows then continue taking tranches until we do cover all no-shows.  e.g. if there are 2 no-shows, we might only need to take 1 additional tranche with >= 2 assignments. Or we might need to take 3 tranches, where one is empty and the other two have 1 assignment each.
```
rphmeier, line 214: We should address the "we run out of tranches to take, having not received any assignments past a certain point" case.

roadmap/implementers-guide/src/node/approval/approval-voting.md
```diff
@@ -205,16 +207,16 @@ On receiving an `ApprovedAncestor(Hash, BlockNumber, response_channel)`:
...
+    6. At this point, set `l := k` and set `k` to be the number of tranches taken so far.  Also set `t := assigned - noshows` so that all uncovered assignments must eventually be covered.  Repeat from step 2.  
 
 #### `check_approval(block_entry, approval_entry, n_tranches) -> bool`
   * If `n_tranches` is ALL, return false
-  * Otherwise, if all validators in `n_tranches` have approved, return `true`. If any validator in these tranches has not yet approved but is not yet considered a no-show, return `false`.
+  * Otherwise, if all validators in `n_tranches` have either approved or been replaced as a no-show, then return `true`.  If any validator in these tranches has not yet approved but is not yet considered a no-show, return `false`.
```
rphmeier, line 219: This function doesn't have information to understand replacement. This function is invoked after the replacement procedure has selected extra tranches. I could see it gaining an extra parameter for `tolerated_missing_approvals`. We would also need to alter `tranches_to_approve` to return a `(n_tranches, n_replaced)` tuple, and I suppose supply `tolerated_missing_approvals = n_replaced` in `check_approval`. Then we would not approve if there are any missing approval votes in those tranches beyond the no-shows we've already accounted for.

burdges: As I mentioned elsewhere, I needed `tranches_to_approve` and `check_approval` to be the same function to have everything in one place.

roadmap/implementers-guide/src/node/approval/approval-voting.md
burdges: Ok sure.  I just felt this looked like it needed clarification that it serves two roles, not one.

roadmap/implementers-guide/src/node/approval/approval-voting.md
burdges: Remove so this represents the current goal, or use future/conditional tense to represent later plans?

roadmap/implementers-guide/src/node/approval/approval-voting.md
burdges: We should remove this entirely I think and just create an issue.

roadmap/implementers-guide/src/node/approval/approval-voting.md
burdges: It's a remark that this function is non-monotonic.  It can increase, and then decrease (and return approved once it gets merged with `check_approval` again).

roadmap/implementers-guide/src/node/approval/approval-voting.md
burdges: That was 3 which I just removed.  If we run out of tranches without 3 then every validator is assigned to check the block.  Yes I suppose we should say so.  If we run out of tranches with 3 then we've decided this block sucks and we're going to fork the chain or something.

roadmap/implementers-guide/src/node/approval/approval-voting.md
burdges: https://github.com/w3f/research-security-issues/issues/47

roadmap/implementers-guide/src/node/approval/approval-voting.md
rphmeier: They aren't the same function in this PR, though. So they would need to be unified

roadmap/implementers-guide/src/node/approval/approval-voting.md
rphmeier: Can you change the language a bit? Right now it sounds sort of like the function can return more than once, whereas what we actually want to get across is that it is non-monotonic.

roadmap/implementers-guide/src/node/approval/approval-voting.md
rphmeier: We need to know how many tranches to take, but remember that this is from the perspective of a single node executing these algorithms, and this node may simply have not received those assignments yet.

roadmap/implementers-guide/src/node/approval/approval-voting.md
```diff
@@ -203,22 +203,22 @@ On receiving an `ApprovedAncestor(Hash, BlockNumber, response_channel)`:
...
-    2. Take assignments from tranches `l..` until we have at least `t` assignments.  Let `k` denote the highest tranche taken.  If we run out of tranches to take then we have de facto escalated to every validator checking the block, so return this fact.
-    3. Count the number `assigned` of assignments taken in tranches `0..k`.  If `assigned < t` then return a special value `ALL` which indicates we wait for more assignments. 
-    4. Count the number `noshows` of no-shows in tranches `l..k`.  If `noshows` is zero then return success with `n_tranches := k`.  Of course this happens early and does not indicate final termination, as we may later return `ALL` after more no-shows, but if all these assigned checkers vote valid then we are done.
+#### ` assignees_status(approval_entry) -> AssigneeStatus`
+  * Summarise our view of this approval entry's run by iterating over assignment and approval vote records 
+    1. First, set `needed := session_info.needed_approvals`.  Set the base tranche `l=0`.
+    2. Take assignments from tranches `l..` until we have at least `needed` assignments or hit our timeouts.  Let `tranches` denote the highest tranche taken (plus one).
```
rphmeier, line 209: What does "hit our timeouts" mean?

roadmap/implementers-guide/src/node/approval/approval-voting.md
```diff
@@ -203,22 +203,22 @@ On receiving an `ApprovedAncestor(Hash, BlockNumber, response_channel)`:
...
+  * Summarise our view of this approval entry's run by iterating over assignment and approval vote records 
+    1. First, set `needed := session_info.needed_approvals`.  Set the base tranche `l=0`.
+    2. Take assignments from tranches `l..` until we have at least `needed` assignments or hit our timeouts.  Let `tranches` denote the highest tranche taken (plus one).  
+    3. Count the number `assigned` of assignments taken in tranches `0..tranches`.  If `assigned < needed` then return a special value `PENDING` which indicates we wait for more assignments. 
+    4. Count the number `approvals` in tranches `0..tranches`.  Also, count the number of `noshows` of no-shows in tranches `l..tranches`.  If `noshows` is zero then return `DONE(approvals,assigned,needed,tranches)`.  Of course, this indicates potential approval only if `approvals == assigned` and we hide the assignment timeout for `tranches`.  In fact, this is an over simplification since we care about arrival times, so counting tranches does not suffice here.  If `approvals < assigned` then more no-shows could occur on future invokations, returning us to `PEMNDING`.
```
rphmeier, line 211: I can't merge over-simplification into the implementer's guide!

roadmap/implementers-guide/src/node/approval/approval-voting.md
```diff
@@ -203,22 +203,22 @@ On receiving an `ApprovedAncestor(Hash, BlockNumber, response_channel)`:
...
 
-#### `tranches_to_approve(approval_entry) -> tranches`
-  * Determine the amount of tranches `n_tranches` our view of the protocol requires of this approval entry
-    1. First, set `t := session_info.needed_approvals`.  Set the base tranche `l=0`.
-    2. Take assignments from tranches `l..` until we have at least `t` assignments.  Let `k` denote the highest tranche taken.  If we run out of tranches to take then we have de facto escalated to every validator checking the block, so return this fact.
-    3. Count the number `assigned` of assignments taken in tranches `0..k`.  If `assigned < t` then return a special value `ALL` which indicates we wait for more assignments. 
-    4. Count the number `noshows` of no-shows in tranches `l..k`.  If `noshows` is zero then return success with `n_tranches := k`.  Of course this happens early and does not indicate final termination, as we may later return `ALL` after more no-shows, but if all these assigned checkers vote valid then we are done.
+#### ` assignees_status(approval_entry) -> AssigneeStatus`
```
rphmeier, line 206: `AssigneeStatus` is not defined anywhere, is it?

roadmap/implementers-guide/src/node/approval/approval-voting.md
```diff
@@ -203,22 +203,22 @@ On receiving an `ApprovedAncestor(Hash, BlockNumber, response_channel)`:
...
-    6. At this point, set `l := k` and set `k` to be the number of tranches taken so far.  Also set `t := assigned - noshows` so that all uncovered assignments must eventually be covered.  Repeat from step 2.  
+    6. At this point, set `l := tranches` and again set `tranches` to be the number of tranches taken so far.  Also set `needed := assigned - noshows` so that all uncovered assignments must eventually be covered.  Repeat from step 2.  
+    7. We return `DONE(..)` if we run out of tranches.    
 
 #### `check_approval(block_entry, approval_entry, n_tranches) -> bool`
-  * If `n_tranches` is ALL, return false
-  * Otherwise, if all validators in `n_tranches` have either approved or been replaced as a no-show, then return `true`.  If any validator in these tranches has not yet approved but is not yet considered a no-show, return `false`.
+  * Invoke `assignees_status` and then check if `approvals == assigned == needed` or if `needed >= num_validators` then we ask that `3 approvals > 2 num_validators`.
```
rphmeier, line 217: The `3 * approvals > 2 * num_validators` thing is new? Is `num_validators` the total number of validators?

coriolinus: I think that's just an integer-safe expression of the inequality `approvals/num_validators > 2/3`.

rphmeier: Yes, it's just not something that was described in the previous writeup

burdges: We'll do something better here later, but some 2/3 condition is simple and suffices for an MVP I think.  I also like this "worse" condition for initial testnets because if we've some bug that causes a checker explosion then this make it show up more in testing.

roadmap/implementers-guide/src/node/approval/approval-voting.md
rphmeier: I suggest removal

burdges:
We should split tracker.rs in https://github.com/paritytech/polkadot/pull/1558/files into an independent implementation of the counting algorithm, and a separate implementation of its database.  We can then adapt the stories.rs, criteria.rs, and the counting algorithm for actual use, and someone can reimplement all the data storage logic twice, once for on-chain and once for off-chain.

burdges:
I've updated the description to far more closely resemble tracker.rs in https://github.com/paritytech/polkadot/pull/1558/files which I also updated for the new no show scheme.  

I'm worried this new description is a "bit too efficient" and while the text I removed was vague it maybe communicated important aspects of the goal.  The gap is discussed in https://github.com/paritytech/polkadot/pull/1930/commits/db69ad9977f05f8f8b71b1fc02fa3415da17fca6#diff-588e17953264507ee591a633edc92d8dd229f9da717eef8a65245396ab1e12d4R249

burdges:
#1972 is better

PR #1934, opened 2020-11-09
Add Basic XCM

shawntabrizi:
This PR adds a basic XCM Sink and enables the Relay Chain to pass downward XCM messages to Parachains.


runtime/parachains/src/ump.rs
```diff
@@ -52,6 +52,32 @@ impl UmpSink for () {
...
+pub struct XcmSink<Config>(sp_std::marker::PhantomData<Config>);
+impl<Config: xcm_executor::Config> UmpSink for XcmSink<Config> {
+	fn process_upward_message(origin: ParaId, msg: Vec<u8>) -> Weight {
+		use parity_scale_codec::Decode;
+		use xcm::v0::{Xcm, Junction, MultiLocation, ExecuteXcm};
+		use xcm_executor::XcmExecutor;
+
+		let mut weight: Weight = 0;
+
+		// TODO: Use versioned XCM Format?
```
pepyakin, line 66: Yeah, I'd imagine so, otherwise, what's the reason of having it.

runtime/parachains/src/ump.rs
```diff
@@ -52,6 +52,32 @@ impl UmpSink for () {
...
+			let xcm_junction: Junction = Junction::Parachain { id: origin.into() };
+			let xcm_location: MultiLocation = xcm_junction.into();
+
+			// TODO: Do something with result.
+			let _result = XcmExecutor::<Config>::execute_xcm(xcm_location, xcm_message);
+
+			weight = weight.saturating_add(0)
+		}
+
+		weight
```
pepyakin, line 77: Let's leave a reminder that this implementation should ensure that it doesn't consume more weight than some pre-configured value.

pepyakin (commented):
A couple of changes requested. 

The problem I see here is that the implementation of xcm sink doesn't account for any weight. Also, there is no guarantee on how much weight execution of a message can take and that's important.

I am not going to block on this, since XCM is a huge experiment anyway, but we should keep that in mind.

parachain/src/primitives.rs
```diff
@@ -285,3 +285,34 @@ pub struct ValidationResult {
 	/// The mark which specifies the block number up to which all inbound HRMP messages are processed.
 	pub hrmp_watermark: RelayChainBlockNumber,
 }
+
+pub type Weight = u64;
```
shawntabrizi, line 289: Weight is a frame primitive, but we dont necessarily want all parachains to depend on frame, so redefining this here.

pepyakin (approved):

runtime/parachains/src/ump.rs
```diff
@@ -52,6 +52,37 @@ impl UmpSink for () {
...
+impl<Config: xcm_executor::Config> UmpSink for XcmSink<Config> {
+	fn process_upward_message(origin: ParaId, msg: Vec<u8>) -> Weight {
+		use parity_scale_codec::Decode;
+		use xcm::VersionedXcm;
+		use xcm::v0::{Junction, MultiLocation, ExecuteXcm};
+		use xcm_executor::XcmExecutor;
+
+		let weight: Weight = 0;
+
+		if let Some(versioned_xcm_message) = VersionedXcm::decode(&mut &msg[..]).ok() {
```
bkchr, line 68:
```suggestion
		if let Ok(versioned_xcm_message) = VersionedXcm::decode(&mut &msg[..]) {
```

And maybe some logging when it fails? :see_no_evil:

bkchr (approved):

shawntabrizi:
bot merge

PR #1939, opened 2020-11-10
Breakdown the Router module on Dmp, Ump, Hrmp modules

pepyakin:
Even though this PR looks big it actually isn't. What we do here is we split the Router module on three separate independent modules: Dmp, Ump and Hrmp, and update tests and plumbing. I tried to collate a proper commit history so I'd recommend going over all commits.

As I said in our parachains team call, there is a not so nice duplication of `OutgoingParas`. On the bonus side, the external modules (such as registrar) now can call into a single entrypoint rather than calling both Router and Paras `schedule_para_cleanup`. In future, we would need to tackle this.


pepyakin:
borked rebase, converting into draft

roadmap/implementers-guide/src/glossary.md
```diff
@@ -24,6 +24,7 @@ Here you can find definitions of a bunch of jargon, usually specific to the Polk
 - Parathread: A parachain which is scheduled on a pay-as-you-go basis.
 - Proof-of-Validity (PoV): A stateless-client proof that a parachain candidate is valid, with respect to some validation function.
 - Relay Parent: A block in the relay chain, referred to in a context where work is being done in the context of the state at this block.
+- Router: The router module used to be a runtime module responsible for routing messages between paras and the relay chain. At some point it was split up into separate runtime modules: Dmp, Ump, Hrmp, each responsible for the respective part of message routing.
```
drahnr, line 27:
```suggestion
- Router: The router module is a meta module that consists of three runtime module responsible for routing messages between paras and the relay chain. The three separate separate runtime modules: Dmp, Ump, Hrmp, each responsible for the respective part of message routing.
```

roadmap/implementers-guide/src/glossary.md
```diff
@@ -24,6 +24,7 @@ Here you can find definitions of a bunch of jargon, usually specific to the Polk
 - Parathread: A parachain which is scheduled on a pay-as-you-go basis.
 - Proof-of-Validity (PoV): A stateless-client proof that a parachain candidate is valid, with respect to some validation function.
 - Relay Parent: A block in the relay chain, referred to in a context where work is being done in the context of the state at this block.
+- Router: The router module is a meta module that consists of three runtime module responsible for routing messages between paras and the relay chain. The three separate separate runtime modules: Dmp, Ump, Hrmp, each responsible for the respective part of message routing.
```
drahnr, line 27:
```suggestion
- Router: The router module is a meta module that consists of three runtime modules responsible for routing messages between paras and the relay chain. The three separate separate runtime modules are: Dmp, Ump, Hrmp, each responsible for the respective part of message routing.
```

roadmap/implementers-guide/src/runtime/hrmp.md
```diff
@@ -1,6 +1,6 @@
-# Router Module
+# HRMP Module
 
-The Router module is responsible for all messaging mechanisms supported between paras and the relay chain, specifically: UMP, DMP, HRMP and later XCMP.
+A module responsible for HRMP. See [Messaging Overview](../messaging.md) for more details.
```
drahnr, line 3: Do links work like this? Asking for a friend..

drahnr (approved):
Looks good!

roadmap/implementers-guide/src/runtime/hrmp.md
pepyakin: What's wrong with it? Genuine question

drahnr: Wasn't sure if the link targets have to end in `.html`, but they don't, that's only relevant for rust struct targets i.e. `[foo is great](struct.Foo.html)`

pepyakin: Actually, mdbook supports both notations! But I prefer and encourage `*.md`, because it also works when you are viewing the docs on github

roadmap/implementers-guide/src/runtime/dmp.md
```diff
@@ -0,0 +1,57 @@
...
+
+## Initialization
+
+No initialization routine runs for this module.
+
+## Routines
+
+Candidate Acceptance Function:
+
+* `check_processed_downward_messages(P: ParaId, processed_downward_messages)`:
```
coriolinus, line 38: `processed_downward_messages` should have a type. On first read, it wasn't clear if it was a slice or integer.

roadmap/implementers-guide/src/runtime/ump.md
```diff
@@ -0,0 +1,98 @@
+# UMP Module
```
coriolinus, line 1: Please expand the acronym here also.

roadmap/implementers-guide/src/runtime/ump.md
```diff
@@ -0,0 +1,98 @@
...
+
+## Initialization
+
+No initialization routine runs for this module.
+
+## Routines
+
+Candidate Acceptance Function:
+
+* `check_upward_messages(P: ParaId, Vec<UpwardMessage>`):
```
coriolinus, line 60: `messages` should be named.

roadmap/implementers-guide/src/runtime/ump.md
```diff
@@ -0,0 +1,98 @@
...
+
+* `enact_upward_messages(P: ParaId, Vec<UpwardMessage>)`:
+    1. Process each upward message `M` in order:
+        1. Append the message to `RelayDispatchQueues` for `P`
+        1. Increment the size and the count in `RelayDispatchQueueSize` for `P`.
+        1. Ensure that `P` is present in `NeedsDispatch`.
+
+The following routine is intended to be called in the same time when `Paras::schedule_para_cleanup` is called.
+
+`schedule_para_cleanup(ParaId)`:
```
coriolinus, line 75: `para` should be named.

roadmap/implementers-guide/src/runtime/hrmp.md
```diff
@@ -1,6 +1,4 @@
-# Router Module
-
-The Router module is responsible for all messaging mechanisms supported between paras and the relay chain, specifically: UMP, DMP, HRMP and later XCMP.
+# HRMP Module
```
coriolinus, line 1: Please expand this acronym also.

roadmap/implementers-guide/src/glossary.md
```diff
@@ -24,7 +24,7 @@ Here you can find definitions of a bunch of jargon, usually specific to the Polk
 - Parathread: A parachain which is scheduled on a pay-as-you-go basis.
 - Proof-of-Validity (PoV): A stateless-client proof that a parachain candidate is valid, with respect to some validation function.
 - Relay Parent: A block in the relay chain, referred to in a context where work is being done in the context of the state at this block.
-- Router: The router module is a meta module that consists of three runtime module responsible for routing messages between paras and the relay chain. The three separate separate runtime modules: Dmp, Ump, Hrmp, each responsible for the respective part of message routing.
+- Router: The router module is a meta module that consists of three runtime modules responsible for routing messages between paras and the relay chain. The three separate separate runtime modules are: Dmp, Ump, Hrmp, each responsible for the respective part of message routing.
```
coriolinus, line 27: Repetition: "The three separate separate runtime"...

coriolinus (approved):
Overall seems fine with just minor nits. The only real question I have: from the commit messages, it seems like you extracted the DMP, UMP, HRMP modules from existing implementations. However, I don't see where the previous implementation has been removed. If there is dead code remaining, it would be good to get rid of it.

roadmap/implementers-guide/src/runtime/ump.md
pepyakin: should it though? Isn't it already clear that when we say "the messages" it refers to those?

I am not entirely against that, but in that case we should probably address it throughout the whole doc

roadmap/implementers-guide/src/runtime/dmp.md
```diff
@@ -0,0 +1,57 @@
+# DMP Module
```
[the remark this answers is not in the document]
pepyakin: My logic the users of this document should use accronyms extensibly. To the point that we never expand HRMP, we just use it to name everything related to it.

Secondly, I am really not a fan of duplicating information, because in the definition is given already in the glossary and perhaps the messaging overview.

roadmap/implementers-guide/src/runtime/ump.md
pepyakin: again, the same logic as [here](https://github.com/paritytech/polkadot/pull/1939#discussion_r523009204)

pepyakin:
I've addressed some of the uncontroversial comments. The router module implementation was removed in this commit

https://github.com/paritytech/polkadot/pull/1939/commits/a870f718c367f7dbe041376b420e61f24216081c

pepyakin:
Yeah, @coriolinus you were right! That commit did not include removal of the (sub-)modules themselves.

roadmap/implementers-guide/src/runtime/dmp.md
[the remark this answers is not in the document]
coriolinus: I guess it's a philosophical difference: I like to document things everywhere, because it's hard to predict the directions from which users will come. HRMP is a great example of that, actually: internally, I always expand it to "higher-ranked message passing", which I am fairly confident is incorrect, but we're just using the acronym everywhere. Most of the time that's fine, in usage, but it seems like a fairly tame redundancy to also define it in its section header in the guide.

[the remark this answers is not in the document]
pepyakin: Ok fair, I am gonna expand these acronyms however only in the text below. The reasoning here is that the title is the module name, and that is the abbreviation.

pepyakin:
bot merge

PR #1941, opened 2020-11-10
sudo_queue_downward_message

pepyakin:
This PR adds a nifty entrypoint for sending a downward message from root/sudo.

### DIFF
--- a/runtime/common/src/paras_sudo_wrapper.rs
+++ b/runtime/common/src/paras_sudo_wrapper.rs
@@ -29,7 +29,9 @@ use runtime_parachains::{
 use primitives::v1::Id as ParaId;
 
 /// The module's configuration trait.
-pub trait Trait: configuration::Trait + paras::Trait + dmp::Trait + ump::Trait + hrmp::Trait {}
+pub trait Trait: 
+	configuration::Trait + paras::Trait + dmp::Trait + ump::Trait + hrmp::Trait 
+{}
 
 decl_error! {
 	pub enum Error for Module<T: Trait> {
@@ -76,10 +78,10 @@ decl_module! {
 			ensure!(<paras::Module<T>>::is_valid_para(id), Error::<T>::ParaDoesntExist);
 			let config = <configuration::Module<T>>::config();
 			<dmp::Module<T>>::queue_downward_message(&config, id, payload)
-				.map_err(|e| {match e {
+				.map_err(|e| match e {
 					dmp::QueueDownwardMessageError::ExceedsMaxMessageSize =>
 						Error::<T>::ExceedsMaxMessageSize.into(),
-				}})
+				})
 		}
 	}
 }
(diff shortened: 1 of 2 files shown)


runtime/common/src/paras_sudo_wrapper.rs
```diff
@@ -58,5 +65,21 @@ decl_module! {
 			runtime_parachains::schedule_para_cleanup::<T>(id);
 			Ok(())
 		}
+
+		/// Send a downward message to the given para.
+		///
+		/// The given parachain should exist and the payload should not exceed the preconfigured size
+		/// `config.max_downward_message_size`.
+		#[weight = (1_000, DispatchClass::Operational)]
+		pub fn sudo_queue_downward_message(origin, id: ParaId, payload: Vec<u8>) -> DispatchResult {
```
shawntabrizi, line 74: probably isnt right to use the term "sudo" here. A root call can also come from democracy or other systems which generate a root origin. `root_queue_downward_message` would make more sense but i think `queue_downward_message` would be enough?

shawntabrizi (approved):
I would not have directly used the term sudo here since you really mean `root` origin which could come from multiple places like democracy, but ya thats just a nitpick

pepyakin:
Yeah, I know. That's just what the module is already called.

pepyakin:
Rebased against the current master. The sole commit of this PR remains unchanged.

runtime/common/src/paras_sudo_wrapper.rs
```diff
@@ -16,22 +16,29 @@
...
 
 /// The module's configuration trait.
-pub trait Trait: paras::Trait + dmp::Trait + ump::Trait + hrmp::Trait { }
+pub trait Trait: configuration::Trait + paras::Trait + dmp::Trait + ump::Trait + hrmp::Trait {}
 
 decl_error! {
-	pub enum Error for Module<T: Trait> { }
+	pub enum Error for Module<T: Trait> {
+		/// The specified parachain or parathread is not registered.
+		ParaDoesnExist,
```
coriolinus, line 37:
```suggestion
		ParaDoesntExist,
```

runtime/common/src/paras_sudo_wrapper.rs
```diff
@@ -58,5 +65,21 @@ decl_module! {
...
 		}
+
+		/// Send a downward message to the given para.
+		///
+		/// The given parachain should exist and the payload should not exceed the preconfigured size
+		/// `config.max_downward_message_size`.
+		#[weight = (1_000, DispatchClass::Operational)]
+		pub fn sudo_queue_downward_message(origin, id: ParaId, payload: Vec<u8>) -> DispatchResult {
+			ensure_root(origin)?;
+			ensure!(<paras::Module<T>>::is_valid_para(id), Error::<T>::ParaDoesnExist);
```
coriolinus, line 76:
```suggestion
			ensure!(<paras::Module<T>>::is_valid_para(id), Error::<T>::ParaDoesntExist);
```

coriolinus (approved):

bkchr (approved):

PR #1942, opened 2020-11-10
[CI] Fix release Docker github action

s3krit:
* Add some package metadata
* Fix the errors in the Github Action that were causing [this](https://github.com/paritytech/polkadot/actions/runs/352786916) issue
* Fix a few [hadolint](https://github.com/hadolint/hadolint) nags
* Now pins the version installed from the Docker image - will fail if the specific version of Polkadot is not present in our Deb repository

### DIFF
--- a/.github/workflows/publish-docker-release.yml
+++ b/.github/workflows/publish-docker-release.yml
@@ -14,12 +14,12 @@ jobs:
         uses: docker/setup-buildx-action@v1
       - name: Cache Docker layers
         uses: actions/cache@v2
-      - name: Login to Dockerhub
         with:
          path: /tmp/.buildx-cache
          key: ${{ runner.os }}-buildx-${{ github.sha }}
          restore-keys: |
            ${{ runner.os }}-buildx-
+      - name: Login to Dockerhub
         uses: docker/login-action@v1
         with:
           username: ${{ secrets.DOCKERHUB_USERNAME }}
@@ -29,7 +29,7 @@ jobs:
         uses: docker/build-push-action@v2
         with:
           push: true
-          file: scripts/docker/Dockerfile.release
+          file: scripts/docker/release.Dockerfile
           tags: |
             parity/polkadot:latest
             parity/polkadot:${{ github.event.release.tag_name }}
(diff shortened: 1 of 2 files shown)


General-Beck (approved):

.github/workflows/publish-docker-release.yml
```diff
@@ -29,10 +29,14 @@ jobs:
         uses: docker/build-push-action@v2
         with:
           push: true
-          file: scripts/docker/Dockerfile.release
+          file: scripts/docker/release.Dockerfile
           tags: |
-            parity/polkadot:latest
-            parity/polkadot:${{ github.event.release.tag_name }}
+            martinparity/polkadot:latest
```
gabreal, line 34: is this docker repository change intentional?

scripts/docker/release.Dockerfile
```diff
@@ -18,21 +19,21 @@ ENV RUST_BACKTRACE 1
...
 # install tools and dependencies
 RUN apt-get update && \
-		DEBIAN_FRONTEND=noninteractive apt-get upgrade -y && \
-		DEBIAN_FRONTEND=noninteractive apt-get install -y \
+		DEBIAN_FRONTEND=noninteractive apt-get install -y --no-install-recommends \
 			libssl1.1 \
 			ca-certificates \
 			curl \
 			gnupg && \
+		useradd -m -u 1000 -U -s /bin/sh -d /polkadot polkadot && \
```
gabreal, line 27: doesn't the debian package create this user later?

TriplEight (changes requested):
I assume this dockerfile is built with GHA because it's easier to use it with github labels?

.github/workflows/publish-docker-release.yml
s3krit: Yep sorry - committed so that I could test it without it smashing the `parity/polkadot` image. change reverted.

s3krit:
> I assume this dockerfile is built with GHA because it's easier to use it with github labels?

No, it's actually so that it is triggered on the *publishing* of a release. Since the publishing of the draft release is the final step in creating a release, it is the best indicator of 'this version is now released'.

>AFAIR polkadot bin should create this user, right?

It is, but we had a user run into issues with the fact the user installed by the debian package had a different UID to the one created by the 'old' release docker image.

TriplEight (approved):

PR #1944, opened 2020-11-11
Change `len` from `usize` to `u32`.

pepyakin:
I don't think this is particularly important, but shouldn't harm either.

### DIFF
--- a/parachain/test-parachains/adder/src/wasm_validation.rs
+++ b/parachain/test-parachains/adder/src/wasm_validation.rs
@@ -24,7 +24,7 @@ use codec::{Encode, Decode};
 
 #[no_mangle]
 pub extern "C" fn validate_block(params: *const u8, len: u32) -> u64 {
-	let params = unsafe { parachain::load_params(params, len) };
+	let params = unsafe { parachain::load_params(params, len as u32) };
 	let parent_head = HeadData::decode(&mut &params.parent_head.0[..])
 		.expect("invalid parent head format.");
 
--- a/parachain/test-parachains/adder/src/wasm_validation.rs
+++ b/parachain/test-parachains/adder/src/wasm_validation.rs
@@ -24,7 +24,7 @@ use codec::{Encode, Decode};
 
 #[no_mangle]
 pub extern "C" fn validate_block(params: *const u8, len: u32) -> u64 {
-	let params = unsafe { parachain::load_params(params, len as u32) };
+	let params = unsafe { parachain::load_params(params, len as usize) };
 	let parent_head = HeadData::decode(&mut &params.parent_head.0[..])
 		.expect("invalid parent head format.");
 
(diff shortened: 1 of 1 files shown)


coriolinus (approved):

drahnr (approved):

drahnr (approved):

parachain/test-parachains/adder/src/wasm_validation.rs
```diff
@@ -23,8 +23,8 @@ use parachain::primitives::{ValidationResult, HeadData as GenericHeadData};
 use codec::{Encode, Decode};
 
 #[no_mangle]
-pub extern "C" fn validate_block(params: *const u8, len: usize) -> u64 {
-	let params = unsafe { parachain::load_params(params, len) };
+pub extern "C" fn validate_block(params: *const u8, len: u32) -> u64 {
+	let params = unsafe { parachain::load_params(params, len as u32) };
```
niklasad1, line 27: redundant cast.

```suggestion
	let params = unsafe { parachain::load_params(params, len) };
```

parachain/test-parachains/adder/src/wasm_validation.rs
```diff
@@ -24,7 +24,7 @@ use codec::{Encode, Decode};
 
 #[no_mangle]
 pub extern "C" fn validate_block(params: *const u8, len: u32) -> u64 {
-	let params = unsafe { parachain::load_params(params, len) };
+	let params = unsafe { parachain::load_params(params, len as u32) };
```
coriolinus, line 27: I don't understand the motivation here if `len` is already `u32` per the function definition.

pepyakin: oh, I meant to cast it by `usize` 🙈 , I need to devote more attention to this PR.

pepyakin:
Hold on guys, the third time is the charm

bkchr (changes requested):
If you want to do this correctly, please change the `load_params' function ;)

Otherwise this pr doesn't bring any value.

pepyakin:
Ok, I see your concern, had the same one. I chose the presented solution because the body of `load_params` requires `usize` anyway. So this would just move the cast down. 

Also note, that this is not a destructive cast, since we never assume that the word size is below 32 bits. But if it were, then moving the cast won't help either, we would need to panic to somehow signal this destruction. 

And finally, I **would have** argued that this PR has some value, something along the lines that the risk of UB coming from a broken signature is worse than potential trimming problems in 16 bits platforms yadyadyadya (🙈), but I am not myself completely sure that the root problem is a sufficiently realistic scenario. 

So I am not sure if I am going to persue this PR. Feel free to close it.

pepyakin:
Oh, and look, the CI has passed xD

pepyakin:
Ok, Basti showed me [another example](https://github.com/paritytech/cumulus/blob/3ef4b34a75739a26bddb89cc4b2751d288812f61/runtime/src/validate_block/mod.rs#L64) where `usize` is used and I also found one in substrate and perhaps there are more.

I am closing this, but if anybody feels motivated enough to pick up, feel free to reopen or open another PR

PR #1947, opened 2020-11-12
Restrict collating_using_adder_collator test on feature real-overseer

coriolinus:
This test can only pass given that feature, and runs forever otherwise.

This change ensure that a careless `cargo test --all` doesn't run forever.

### DIFF
--- a/parachain/test-parachains/adder/collator/tests/integration.rs
+++ b/parachain/test-parachains/adder/collator/tests/integration.rs
@@ -19,6 +19,7 @@
 
 // If this test is failing, make sure to run all tests with the `real-overseer` feature being enabled.
 #[substrate_test_utils::test]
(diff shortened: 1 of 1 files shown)


ordian (approved):
Looks good, although I prefer
`required-features` annotation on a test
https://doc.rust-lang.org/cargo/reference/cargo-targets.html?highlight=required-features#configuring-a-target

[example](https://github.com/paritytech/parity-common/blob/ac082b501416e0a7f09fedeb91fe1fe9167bb733/uint/Cargo.toml#L29-L31)

coriolinus:
If we used `required-features`, then no future tests in the same crate would run unless that feature were specified, which I think is not the behavior we'd expect and desire. I like the granularity of disabling only this single test in the event that it can't pass.

PR #1949, opened 2020-11-12
Make sure adder collator test is executed in CI

bkchr:
…overseer (#1947)"

This reverts commit 355558ae880b6f2d256550f02fc41f680b0f752b.

### DIFF
--- a/parachain/test-parachains/adder/collator/tests/integration.rs
+++ b/parachain/test-parachains/adder/collator/tests/integration.rs
@@ -19,6 +19,7 @@
 
 // If this test is failing, make sure to run all tests with the `real-overseer` feature being enabled.
 #[substrate_test_utils::test]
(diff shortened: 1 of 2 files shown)


ordian (approved):
Is there a way to accomplish what #1947 was trying to solve?
Namely, not run this test with `cargo test --workspace --release`?

bkchr:
We could change to the directory and run it directly. However this will probably require a huge rebuild and this will increase the already not so small CI time. 

However, I don't expect this feature to last for that long and I added a comment to the test if it is failing to make the user aware that the feature needs to be activated.

bkchr:
And the test is also timing out after 10 minutes.

bkchr:
Okay, CI difference is below 2 minutes. Fine by me, we can keep this for now.

PR #1950, opened 2020-11-13
Add CI job to verify extrinsic ordering

s3krit:
**What?**: This PR adds a job that uses @jacogr 's `polkadot-js-metadata-cmp` tool to verify that if the ordering/indexing of the extrinsics of a runtime have changed since the last release, `transaction_version` for that runtime has been bumped.

**How?**: We grab the polkadot binary produced in the `test-linux-stable` job, and for each runtime, we run the client in the background with `--chain=$RUNTIME-local`. Then we use `polkadot-js-metadata-cmp` to compare the metadata with the live chain.

This essentially automates [this](https://github.com/paritytech/polkadot/blob/master/.github/ISSUE_TEMPLATE/release.md#extrinsic-ordering) check that is completed as part of our release process - but on a per-PR basis. In design and implementation it's somewhat similar to our [`check_runtime.sh`](https://github.com/paritytech/polkadot/blob/master/scripts/gitlab/check_runtime.sh) CI job.

After merging, we should add this check as a must-pass to our branch protection rules for `master`

An example of a passing job can be found [here](https://gitlab.parity.io/parity/polkadot/-/jobs/714696)
An example of a failing job can be found [here](https://gitlab.parity.io/parity/polkadot/-/jobs/714605) (to cause it to fail, I had it compare the metadata between kusama and polkadot)

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -147,8 +147,8 @@ check-runtime-benchmarks:          &test
 check-transaction-versions:
   image:                          node:15
   stage:                          build
-  dependencies:
-    - test-linux-stable
+  needs:
+    - job:                        test-linux-stable
   before_script:
     - npm install -g @polkadot/metadata-cmp
     - git fetch origin release
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -198,7 +198,7 @@ generate-impl-guide:
     - mdbook build roadmap/implementers-guide
 
 check-transaction-versions:
-  stage:                          build
+  stage:                          publish
   dependencies:
     - build-linux-release
   script: "scripts/gitlab/check_extrinsics_ordering.sh"
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -200,9 +200,13 @@ generate-impl-guide:
     - mdbook build roadmap/implementers-guide
 
 check-transaction-versions:
-  stage:                          test
-    # dependencies:
-    # - build-linux-release
+  image:                          node:15
+  stage:                          publish
+  dependencies:
+    - build-linux-release
+  before_script:
+    - npm install -g @polkadot/metadata-cmp
+    - git fetch origin release
   script: "scripts/gitlab/check_extrinsics_ordering.sh"
 
 .publish-build:                    &publish-build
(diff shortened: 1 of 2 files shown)


.gitlab-ci.yml
```diff
@@ -141,6 +144,16 @@ check-runtime-benchmarks:          &test
     - ./scripts/gitlab/check_runtime_benchmarks.sh
     - sccache -s
 
+check-transaction-versions:
+  image:                          node:15
+  stage:                          build
+  dependencies:
```
TriplEight, line 150:
```suggestion
  needs:
```

.gitlab-ci.yml
```diff
@@ -141,6 +144,16 @@ check-runtime-benchmarks:          &test
...
+    - test-linux-stable
```
TriplEight, line 151:
```suggestion
    - job:                          test-linux-stable
  stage:                          build
```

TriplEight: `dependencies` is deprecated and doesn't work well with modern syntax.

scripts/gitlab/check_extrinsics_ordering.sh
```diff
@@ -0,0 +1,59 @@
...
+  "kusama"
+  "polkadot"
+)
+
+for RUNTIME in "${runtimes[@]}"; do
+  echo "[+] Checking runtime: ${RUNTIME}"
+
+  release_transaction_version=$(
+    git show "origin/release:runtime/${RUNTIME}/src/lib.rs" | \
+    grep 'transaction_version'
```
TriplEight, line 20: maybe 1 more tab here

scripts/gitlab/check_extrinsics_ordering.sh
```diff
@@ -0,0 +1,59 @@
...
+  $BIN --chain="$RUNTIME-local"  &
+  jobs
+
+  changed_extrinsics=$(
+    polkadot-js-metadata-cmp "$LIVE_WS" "$LOCAL_WS" \
+    | sed 's/^ \+//g' | grep -e 'idx: [0-9]\+ -> [0-9]\+'
+  )
+
+  if [ -n "$changed_extrinsics" ]; then
+    echo "[!] Extrinsics indexing/ordering has changed in the ${RUNTIME} runtime! If this change is intentional, please bump transaction_version in lib.rs. Changed extrinsics:"
```
TriplEight, line 51: ain't it too long?

gabreal (approved):
lgtm

PR #1951, opened 2020-11-13
Approval Distribution Subsystem

### LINKED ISSUE #1603: Guide: Approval Voting Distribution Subsystem
This subsystem is responsible for distributing assignments and approval votes generated by the Approval Voting Subsystem (#1601) to and from other validators.

These votes need to make their way to the provisioner, where they can be included on-chain. Gossip is suitable for this subsystem.

We might want this subsystem to be responsible for distributing disputes as well.

rphmeier:
Closes #1603


roadmap/implementers-guide/src/node/approval/approval-distribution.md
```diff
@@ -0,0 +1,63 @@
...
+
+We implement this protocol as a gossip protocol, and like other parachain-related gossip protocols our primary concerns are about ensuring fast message propagation while maintaining an upper bound on the number of messages any given node must store at any time.
+
+Approval messages should always follow assignments, so we need to be able to discern two pieces of information based on our [View](../../types/network.md#universal-types):
+  1. Is a particular assignment relevant under a given `View`?
+  2. Is a particular approval relevant to any assignment in a set?
+
+These two queries need not be perfect, but they must never yield false positives. For our own local view, they should not yield false negatives. When applied to our peers' views, it is acceptable for them to yield false negatives. The reason for that is that our peers' views may be beyond ours, and we are not capable of fully evaluating them. Once we have caught up, we can check again for false negatives to continue distributing.
```
coriolinus, line 17:
```suggestion
These two queries need not be perfect, but they must never yield false positives. For our own local view, they must not yield false negatives. When applied to our peers' views, it is acceptable for them to yield false negatives. The reason for that is that our peers' views may be beyond ours, and we are not capable of fully evaluating them. Once we have caught up, we can check again for false negatives to continue distributing.
```

"Should" has the connotation that misbehavior is allowable.

roadmap/implementers-guide/src/node/approval/approval-distribution.md
```diff
@@ -0,0 +1,63 @@
...
+
+For assignments, what we need to be checking is whether we are aware of the (block, candidate) pair that the assignment references. For approvals, we need to be aware of an assignment by the same validator which references the candidate being approved.
+
+However, awareness on its own of a (block, candidate) pair would imply that even ancient candidates all the way back to the genesis are relevant. We are actually not interested in anything before finality. did
```
coriolinus, line 21: Trailing "did".

```suggestion
However, awareness on its own of a (block, candidate) pair would imply that even ancient candidates all the way back to the genesis are relevant. We are actually not interested in anything before finality. 
```

roadmap/implementers-guide/src/types/overseer-protocol.md
```diff
@@ -40,9 +40,9 @@ struct ActiveLeavesUpdate {
 Messages received by the approval voting subsystem.
 
 ```rust
-enum VoteCheckResult {
+enum VoteCheckResult<T> {
 	// The vote was accepted and should be propagated onwards.
-	Accepted,
+	Accepted(T),
```
coriolinus, line 45: There's some documentation in `ApprovalVotingMessage` about what `T` will likely contain, but that's not where people are likely to look first when trying to figure out this enum. Copying that note here would help; expanding it would help more.

rphmeier: What kind of documentation do you expect? It's a generic

coriolinus: Yes, it's generic, with no locally observable constraints. That lack of information is hard to work with. I'd go with documentation like:

> In typical use, the payload of `Accepted` is either `()` when routed via `CheckAndImportApproval`, or `(Hash, u32)` when routed via `CheckAndImportAssignment`. In the latter case, those values are the block hash and candidate index of the assignment.

The goal is just to provide a local jumping-off point, so the reader can have some idea of what's going on there without needing to correlate it with the entire rest of the codebase.

rphmeier: I find that to be too fragile for the moment. It would be easy to update one of those types in a small iteration without updating that doc-string, leading the reader even more confused. Whereas the `VoteImportResult` itself is easily searchable.

roadmap/implementers-guide/src/types/network.md
```diff
@@ -8,7 +8,12 @@ These types are those that are actually sent over the network to subsystems.
 type RequestId = u64;
 type ProtocolVersion = u32;
 struct PeerId(...); // opaque, unique identifier of a peer.
-struct View(Vec<Hash>); // Up to `N` (5?) chain heads.
+struct View {
+	// Up to `N` (5?) chain heads.
+	heads: Vec<Hash>,
```
coriolinus, line 13: It feels like we should encode the size restriction into the typesystem somehow, just as self-protection. Possibly like

```rust
struct LimitedSizeVec<T> {
  items: SmallVec<T>,
  capacity: usize,
}
```

That's probably out of scope of this PR, though.

roadmap/implementers-guide/src/node/approval/approval-voting.md
```diff
@@ -166,6 +166,7 @@ On receiving a `ApprovalVotingMessage::CheckAndImportAssignment` message, we che
...
     * If the cert kind is `RelayVRFDelay`, then we check if the VRF is valid for the validator's key with the input `block_entry.relay_vrf_story ++ cert.core_index.encode()` as described in [the approvals protocol section](../../protocol-approval.md#assignment-criteria). The cert can be ignored if the block did not cause inclusion of a candidate on that core index. Otherwise, this is a valid assignment for the included candidate. The delay tranche for the assignment is determined by reducing `(vrf.make_bytes().to_u64() % (session_info.n_delay_tranches + session_info.zeroth_delay_tranche_width)).saturating_sub(session_info.zeroth_delay_tranche_width)`.
+    * If the delay tranche is too far in the future, return `VoteCheckResult::Ignore`.
```
coriolinus, line 169: What counts as "too far"?

roadmap/implementers-guide/src/node/approval/approval-distribution.md
```diff
@@ -40,14 +40,25 @@ struct State {
   finalized_number: BlockNumber,
 }
 
+enum MessageFingerprint {
+  Assigment(Hash, u32, ValidatorIndex),
+  Approval(Hash, u32, ValidatorIndex),
```
coriolinus, line 45: What do the integers mean here?

roadmap/implementers-guide/src/types/overseer-protocol.md
```diff
@@ -40,36 +40,38 @@ struct ActiveLeavesUpdate {
...
 
 ```rust
-enum VoteCheckResult<T> {
+enum AssignmentCheckResult {
 	// The vote was accepted and should be propagated onwards.
-	Accepted(T),
+	Accepted(u32),
 	// The vote was valid but duplicate and should not be propagated onwards.
-	AcceptedDuplicate(T),
+	AcceptedDuplicate(u32),
```
coriolinus, line 47: What does the integer signify?

roadmap/implementers-guide/src/types/overseer-protocol.md
```diff
@@ -40,36 +40,38 @@ struct ActiveLeavesUpdate {
...
+		ResponseChannel<AssignmentCheckResult>,
 	),
 	/// Check if the approval vote is valid and can be accepted by our view of the
 	/// protocol.
 	///
 	/// Should not be sent unless the block hash within the indirect vote is known.
 	CheckAndImportApproval(
 		IndirectSignedApprovalVote,
-		ResponseChannel<VoteCheckResult<()>>,
+		ResponseChannel<ApprovalCehckResult>,
```
coriolinus, line 74: Typo: "Cehck" -> "Check"

roadmap/implementers-guide/src/node/approval/approval-distribution.md
```diff
@@ -182,4 +179,15 @@ Imports an approval signature referenced by block hash and candidate index.
...
     * We've checked that this specific approval has a corresponding assignment within the `BlockEntry`.
-    * Thus, all peers are aware of the assignment or have a message to them in-flight which will make them so. 
\ No newline at end of file
+    * Thus, all peers are aware of the assignment or have a message to them in-flight which will make them so. 
+
+
+#### `unify_with_peer(peer: PeerId, view)`:
+
+For each block in the view:
+  1. Initialize `fresh_blocks = {}`
```
coriolinus, line 188: We should be explicit that `fresh_blocks` is a set, not a map.

roadmap/implementers-guide/src/runtime/approvals_inherent.md
```diff
@@ -0,0 +1 @@
+# ApprovalsInherent Module
```
coriolinus, line 1: This feels like an accidentally included file of the sort which is generated when there's an error in the TOC.

roadmap/implementers-guide/src/types/network.md
```diff
@@ -58,6 +77,14 @@ enum StatementDistributionV1Message {
 }
 ```
 
+### Approval Distribution V1
+
+```rust
+enum ApprovalDistributionV1Message {
+	// TODO [now]
```
coriolinus, line 84: Hanging TODO.

roadmap/implementers-guide/src/node/approval/approval-distribution.md
```diff
@@ -25,31 +27,75 @@ However, awareness on its own of a (block, candidate) pair would imply that even
...
 ## Functionality
 
+```rust
+type BlockScopedCandidate = (Hash, CandidateHash);
+
+/// The `State` struct is responsible for tracking the overall state of the subsystem.
+///
+/// It tracks metadata about our view of the chain, which assignments and approvals we have seen, and our peers' views.
+struct State {
+  block_view: HashMap<BlockNumber, Vec<Hash>>,
```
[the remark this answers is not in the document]
rphmeier: For `n <= finalized_number`, there is no entry in this map

roadmap/implementers-guide/src/node/approval/approval-distribution.md
```diff
@@ -25,31 +27,75 @@ However, awareness on its own of a (block, candidate) pair would imply that even
...
+}
+
+/// Information about blocks in our current view as well as whether peers know of them.
+struct BlockEntry {
+  // Peers who we know are aware of this block and thus, the candidates within it.
+  known_by: HashSet<PeerId>,
+  // The parent hash of the block.
+  parent_hash: Hash,
+  // A votes entry for each candidate.
+  candidates: IndexMap<CandidateHash, CandidateEntry>,
```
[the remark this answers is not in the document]
rphmeier: because we refer to them by index constantly

roadmap/implementers-guide/src/types/network.md
rphmeier: yes, out of scope

roadmap/implementers-guide/src/types/overseer-protocol.md
rphmeier: candidate index

roadmap/implementers-guide/src/node/approval/approval-distribution.md
rphmeier: why?

roadmap/implementers-guide/src/types/overseer-protocol.md
```diff
@@ -86,7 +86,17 @@ enum ApprovalVotingMessage {
 Messages received by the approval Distribution subsystem.
 
 ```rust
+struct BlockApprovalMeta {
+	hash: Hash,
```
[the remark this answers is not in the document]
rphmeier: yep

roadmap/implementers-guide/src/node/approval/approval-distribution.md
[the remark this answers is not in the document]
coriolinus: That's an even better constraint! It would be useful to have it written.

roadmap/implementers-guide/src/node/approval/approval-distribution.md
coriolinus: Because we're writing this for the future, and the `{}` notation is ambiguous. It's possible to figure it out from later context, but why make it harder on the reader?

roadmap/implementers-guide/src/types/overseer-protocol.md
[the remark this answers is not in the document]
coriolinus: Cool. In that case, what do you think of

```suggestion
	/// Hash of the block number under consideration.
	hash: Hash,
```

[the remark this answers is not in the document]
rphmeier: It's not the hash of the block number, but the hash of the block

coriolinus (approved):

roadmap/implementers-guide/src/node/approval/approval-distribution.md
```diff
@@ -0,0 +1,197 @@
...
+
+We implement this protocol as a gossip protocol, and like other parachain-related gossip protocols our primary concerns are about ensuring fast message propagation while maintaining an upper bound on the number of messages any given node must store at any time.
+
+Approval messages should always follow assignments, so we need to be able to discern two pieces of information based on our [View](../../types/network.md#universal-types):
+  1. Is a particular assignment relevant under a given `View`?
+  2. Is a particular approval relevant to any assignment in a set?
+
+These two queries need not be perfect, but they must never yield false positives. For our own local view, they must not yield false negatives. When applied to our peers' views, it is acceptable for them to yield false negatives. The reason for that is that our peers' views may be beyond ours, and we are not capable of fully evaluating them. Once we have caught up, we can check again for false negatives to continue distributing.
```
burdges, line 19: "it is acceptable for our conclusions to yield false negatives"

roadmap/implementers-guide/src/node/approval/approval-distribution.md
```diff
@@ -0,0 +1,197 @@
...
+
+For assignments, what we need to be checking is whether we are aware of the (block, candidate) pair that the assignment references. For approvals, we need to be aware of an assignment by the same validator which references the candidate being approved.
```
burdges, line 21: "..references, or the candidate equivocation in the case of an equivocation in relay chain block production".

Yes, I know we're not doing those yet, but I wanted to make a note.

roadmap/implementers-guide/src/node/approval/approval-distribution.md
burdges: Unclear.  We add a received timestame to assignments when verifying their VRF signature.  Approval votes might benefit form a received timestame too.

roadmap/implementers-guide/src/node/approval/approval-distribution.md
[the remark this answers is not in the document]
burdges: We'll need to refer to them by availability core index at their inclusion time too, which is not this index.

roadmap/implementers-guide/src/node/approval/approval-distribution.md
```diff
@@ -0,0 +1,197 @@
...
+    * check if `peer` appears under `known_by` and whether the fingerprint is in the `known_messages` of the peer. If the peer does not know the block, report for providing data out-of-view and proceed. If the peer does know the block and the knowledge contains the fingerprint, report for providing replicate data and return.
+    * If the message fingerprint appears under the `BlockEntry`'s `Knowledge`, give the peer a small positive reputation boost and return. Note that we must do this after checking for out-of-view to avoid being spammed. If we did this check earlier, a peer could provide data out-of-view repeatedly and be rewarded for it.
+    * Dispatch `ApprovalVotingMessage::CheckAndImportAssignment(assignment)` and wait for the response.
+    * If the result is `AssignmentCheckResult::Accepted(candidate_index)` or `AssignmentCheckResult::AcceptedDuplicate(candidate_index)` 
+      * If `candidate_index` does not match `claimed_candidate_index`, punish the peer's reputation, recompute the fingerprint, and re-do our knowledge checks. The goal here is to accept messages which peers send us that are labeled wrongly, but punish them for it as they've made us do extra work.
```
burdges, line 154: "labeled wrongly"?  Are these indexes the vacated cores number?  If so, we could always include the index inside the assignment, even when doing a modulo VRF criteria for tranche zero, so it could be read before proceeding, which makes the assignment outright invalid if the index is wrong.  Is that helpful?

roadmap/implementers-guide/src/types/overseer-protocol.md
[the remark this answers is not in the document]
burdges: "relay chain block"

roadmap/implementers-guide/src/node/approval/approval-distribution.md
burdges: Actually I'm unsure where `claimed_candidate_index` originates.  I suppose the peer sends us data it believes important, but why does it need to tell us why the data matters?  We either agree or not, yes?

roadmap/implementers-guide/src/node/approval/approval-distribution.md
```diff
@@ -0,0 +1,197 @@
...
+      * Give the peer a positive reputation boost and add the fingerprint to both our and the peer's knowledge.
+    * If the result is `VoteCheckResult::Bad`:
+      * Report the peer and return.
+  * Load the candidate entry for the given candidate index. It should exist unless there is a logic error in the approval voting subsystem.
+  * Set the approval state for the validator index to `ApprovalState::Approved`. It should already be in the `Assigned` state as our `BlockEntry` knowledge contains a fingerprint for the assignment.
+  * Dispatch a `ApprovalDistributionV1Message::Approval(approval)` to all peers in the `BlockEntry`'s `known_by` set, excluding the peer in the `source`, if `source` has kind `MessageSource::Peer`. Add the fingerprint of the assignment to the knowledge of each peer. Note that this obeys the politeness conditions:
+    * We guarantee elsewhere that all peers within `known_by` are aware of all assignments relative to the block.
+    * We've checked that this specific approval has a corresponding assignment within the `BlockEntry`.
+    * Thus, all peers are aware of the assignment or have a message to them in-flight which will make them so.
```
burdges, line 186: All this sounds fine, but it's quadratic so another reason candidates should be rejected long before "escalation by DoS". 

It breaks gossip assumptions if one compares fingerprints between peers using publicly keyed bloom filters, but if each pair of peers computes their own shared secret bloom filter key, then they could compare using bloom filters.  Yet another optimization I guess.

roadmap/implementers-guide/src/node/approval/approval-voting.md
burdges: We think two relay chain blocks works.  Assuming tranches are half seconds and relay chain blocks are 6 seconds, then 24 tranches but maybe that'll change.  Increasing this somewhat sounds fine too.

roadmap/implementers-guide/src/node/approval/approval-voting.md
```diff
@@ -221,8 +223,8 @@ On receiving an `ApprovedAncestor(Hash, BlockNumber, response_channel)`:
   * Set `n_tranches = tranches_to_approve(approval_entry)`
   * If `OurAssignment` has tranche `<= n_tranches`, the tranche is live according to our local clock (based against block slot), and we have not triggered the assignment already
     * Import to `ApprovalEntry`
-    * Broadcast on network with an `ApprovalNetworkingMessage::DistributeAssignment`.
-    * Kick off approval work with `launch_approval`
+    * Broadcast on network with an `ApprovalDistributionMessage::DistributeAssignment`.
+    * Kick off approval work with `launch_approval`. Note that if the candidate appears in multiple current blocks, we will launch approval for each block it appears in. It may make sense to shortcut around this with caching either at this level or on the level of the other subsystems invoked by that function.
```
burdges, line 227: Yes, these approval checkers being different is an unfortunate consequence of using the relay VRF modulo criteria for tranche zero.  

If the relay chain largely avoids forks then the relay VRF modulo criteria ensures all validators get the same number of tranche zero assignments, which improves tranche zero assignments in several respects:
- We avoid a choice between either overworking nodes so much or else giving them more arbitrary choices about what they check.  
- We estimate the number of checkers with a binimial distribution, which gives a variance n p q which looks less than the variance obtained from the more complex distributions resulting from other strategies.
- We could support validators of different computing capacities if we force validators to take assignments with sample numbers 1,..,k where k cannot exceed by too much their previous average number of samples completed on the same inclusion batch.  In this way, we pay more for validators with more CPU but permit others to participate.  This might not be a good thing, but nice option.

roadmap/implementers-guide/src/node/approval/approval-distribution.md
rphmeier: All of the text you wrote about approvals is in a document called `protocol-approvals` in the guide, and it mentions this. It's less confusing to mention only stuff that's implemented here, IMO.

roadmap/implementers-guide/src/node/approval/approval-distribution.md
rphmeier: Yeah, I think including that data in the VRF would make this point redundant. At the moment, the vacated core index isn't part of the VRF, and we have peers send us the vacated core index alongside the assignment which makes it easier for us to do some politeness checks. 

So the conundrum this is solving is what we should do if a peer sends us something labeled with the wrong core index but that is still a valid assignment for some other core. And my answer was to re-label it, punish the peer, and distribute.

Your general advice so far has been to keep as much data as possible out of the VRF, so it didn't cross my mind that was an option.

roadmap/implementers-guide/src/node/approval/approval-voting.md
rphmeier: All that sounds fine, but the point here is actually about the fact that we literally re-execute the candidate even if we've just validated it and we should have some LRU cache somewhere so that being "re-assigned" to the same candidate across multiple forks is ultra cheap.

burdges: I missed interpreted and read this in terms of across validators, thanks!

roadmap/implementers-guide/src/node/approval/approval-distribution.md
burdges: Our VRF signatures have an input that corresponds to the output, which yes we minimize, but the same VRF also signs an extra auxiliary message, that influences only it proof, not its output.  

We can pack anything we like into this extra auxiliary message, like whatever claims simplify the code, or even the entire gossip message containing the VRF (and doing so saves at least 45ms and 64 bytes per assignment message).

roadmap/implementers-guide/src/types/overseer-protocol.md
burdges: If this is a core index then it should maybe get a type alias or wrapper type for clarity.

burdges:
LGTM :)

roadmap/implementers-guide/src/node/approval/approval-distribution.md
rphmeier: I see. So it is possible to include the core index under the proof, even if that core index was unknown prior to the output being generated? What's the `schnorrkel` API for that?

roadmap/implementers-guide/src/node/approval/approval-voting.md
```diff
@@ -153,7 +153,7 @@ On receiving an `OverseerSignal::ActiveLeavesUpdate(update)`:
   * Ensure that the `CandidateEntry` contains a `block_assignments` entry for the block, with the correct backing group set.
   * If a validator in this session, compute and assign `our_assignment` for the `block_assignments`
     * Only if not a member of the backing group.
-    * Run `RelayVRFModulo` and `RelayVRFDelay` according to the [the approvals protocol section](../../protocol-approval.md#assignment-criteria)
+    * Run `RelayVRFModulo` and `RelayVRFDelay` according to the [the approvals protocol section](../../protocol-approval.md#assignment-criteria). Ensure that the assigned core derived from the output is covered by the auxiliary signature aggregated in the `VRFPRoof`.
```
burdges, line 165: Seems fine.  The core is always either and input or an output already.  If you want it to always be an input then we can just include it in the message and check it from the output.

roadmap/implementers-guide/src/node/approval/approval-voting.md
```diff
@@ -166,6 +166,7 @@ On receiving a `ApprovalVotingMessage::CheckAndImportAssignment` message, we che
...
     * If the cert kind is `RelayVRFDelay`, then we check if the VRF is valid for the validator's key with the input `block_entry.relay_vrf_story ++ cert.core_index.encode()` as described in [the approvals protocol section](../../protocol-approval.md#assignment-criteria). The cert can be ignored if the block did not cause inclusion of a candidate on that core index. Otherwise, this is a valid assignment for the included candidate. The delay tranche for the assignment is determined by reducing `(vrf.make_bytes().to_u64() % (session_info.n_delay_tranches + session_info.zeroth_delay_tranche_width)).saturating_sub(session_info.zeroth_delay_tranche_width)`.
+    * We also check that the core index derived by the output is covered by the `VRFProof` by means of an auxiliary signature.
```
burdges, line 178: The auxiliary message checked by the proof is just like a signature, so it gives us the ability to include stuff that is not in the VRF input.  The core is either an input or output of the given criteria already, but we can include it in the auxiliary message and check that it is correct if that simplifies some code somewhere.

roadmap/implementers-guide/src/node/approval/approval-voting.md
rphmeier: Can we feasibly include the core as an input to the VRF, given that it's derived from the output? That seems cyclical, but maybe through the power of cryptography...

roadmap/implementers-guide/src/node/approval/approval-voting.md
rphmeier: @coriolinus Yeah, I left this vague, but it'll be somehow parameterizable.

PR #1969, opened 2020-11-17
[CI] Only increase priority of a release based on client changes

s3krit:
Currently, the priority assigned to a release is based on the max priority of *all* changes. However, we only need to increase the priority of users upgrading their client for *client*-related changes (i.e., high-priority runtime changes have no bearing on how urgently a user should update their client).


TriplEight (approved):
LGTM

PR #1972, opened 2020-11-18
change approval voting counting procedure

rphmeier:
cc @burdges 

Supersedes #1930 

Change description:
  * Take one new non-empty tranche per no-show, as opposed to the minimal number of tranches which cover the no-show
  * If the required number of checkers (assigned + no-shows) equals or surpasses the total number of validators, we require a simple supermajority to approve.
  * We no longer broadcast our assignment if we already perceive the candidate as having been approved by others.


roadmap/implementers-guide/src/node/approval/approval-voting.md
```diff
@@ -203,23 +203,38 @@ On receiving an `ApprovedAncestor(Hash, BlockNumber, response_channel)`:
...
-      * All no-shows are covered. Set `n_tranches` to the number of tranches taken
+      * We run out of tranches to take, having not received any assignments past a certain point. In this case we set `n_tranches` to a special value `RequiredTranches::Pending` which indicates that new assignments are needed.
+      * All no-shows are covered by at least one non-empty tranche. Set `n_tranches` to the number of tranches taken
+      * The amount of assignments in non-empty & taken tranches plus the amount of needed extras equals or exceeds the total number of validators for the approval entry, which can be obtained by measuring the bitfield. In this case we return a special value `RequiredTranches::All` indicating that all validators have effectively been assigned to check.
     * return `n_tranches`
 
 #### `check_approval(block_entry, approval_entry, n_tranches) -> bool`
-  * If `n_tranches` is ALL, return false
+  * If `n_tranches` is `RequiredTranches::Pending`, return false
+  * If `n_tranches` is `RequiredTranches::All`,  then we return true if `3 * n_approvals > 2 n_validators` and false otherwise.
```
coriolinus, line 231: Explicit multiplication, implicit boolean

```suggestion
  * If `n_tranches` is `RequiredTranches::All`,  then we return `3 * n_approvals > 2 * n_validators`.
```

coriolinus (approved):

burdges:
We could fix the no-show handling in two different ways:  Your new "Take one new non-empty tranche per no-show" vs my old one "Take one new tranche per no-show or the minimal number of tranches which cover the no-show, which ever is more."

I think either provides similar security, but I guess your new text yields simpler code.  :)

roadmap/implementers-guide/src/node/approval/approval-voting.md
```diff
@@ -203,23 +203,38 @@ On receiving an `ApprovedAncestor(Hash, BlockNumber, response_channel)`:
...
+  Exact(DelayTranche),
+}
+```
+
+
   * Determine the amount of tranches `n_tranches` our view of the protocol requires of this approval entry
     * First, take tranches until we have at least `session_info.needed_approvals`. Call the number of tranches taken `k`
-    * Then, count no-shows in tranches `0..k`. For each no-show, we require another checker. Take new tranches until each no-show is covered, so now we've taken `l` tranches. e.g. if there are 2 no-shows, we might only need to take 1 additional tranche with >= 2 assignments. Or we might need to take 3 tranches, where one is empty and the other two have 1 assignment each.
+    * Then, count no-shows in tranches `0..k`. For each no-show, we require another non-empty tranche. Take new tranches until each no-show is covered, so now we've taken `l = k + j` tranches, where `j` is at least the number of no-shows within tranches `0..k`.
     * Count no-shows in tranches `k..l` and for each of those, take tranches until all no-shows are covered. Repeat so on until either
```
burdges, line 223: You want "take another non-empty tranche for each no-show" here too, not just "covered", which is ambiguous.

burdges: I see it's explained below.. well still..

burdges:
There is a time shifting effect that I implemented and discussed in #1930 that does not appear here.  Arguably it's implicit but not really I think.

rphmeier:
@burdges wasn't the outcome of our calls that we didn't need to do the time-shifting thing?

burdges:
We've count no shows from when we received the assignment, which avoids misscount other's late assignments as no show.  We do however require some mechanism that prevents us sending announcements too early when a no show happens.  If a no show takes 1.5 relay chain blocks, so 36 tranches, then we should avoid everyone from tranche 1-35 announcing when one person from tranche zero no shows.

rphmeier:
>  If a no show takes 1.5 relay chain blocks, so 36 tranches, then we should avoid everyone from tranche 1-35 announcing when one person from tranche zero no shows. It's straightforward to set a stoping condition in the assignment counter loop to achieve this I think.

Yeah, I'll add something like this today which should prevent everyone from broadcasting at once.

roadmap/implementers-guide/src/node/approval/approval-voting.md
```diff
@@ -190,42 +190,68 @@ On receiving an `ApprovedAncestor(Hash, BlockNumber, response_channel)`:
...
+    * Count no-shows in tranches `k..l` and for each of those, take another non-empty tranche for each no-show. Repeat so on until either
+      * We run out of tranches to take, having not received any assignments past a certain point. In this case we set `n_tranches` to a special value `RequiredTranches::Pending(last_taken_tranche + uncovered_no_shows)` which indicates that new assignments are needed. `uncovered_no_shows` is the number of no-shows we have not yet covered with `last_taken_tranche`.
+      * All no-shows are covered by at least one non-empty tranche. Set `n_tranches` to the number of tranches taken and return `RequiredTranches::Exact(n_tranches)`.
+      * The amount of assignments in non-empty & taken tranches plus the amount of needed extras equals or exceeds the total number of validators for the approval entry, which can be obtained by measuring the bitfield. In this case we return a special value `RequiredTranches::All` indicating that all validators have effectively been assigned to check.
```
burdges, line 231: Yes, I think using the bitfield is reasonable in the short term.  We'll want something smaller eventually, but asked discussed previously exactly what can be discussed later.  In the short term, `All` could trigger a dispute style 2/3 check, or it could simply trash the candidate and fork the chain, but if we become happy with some lower threshold then maybe trashing the candidate actually becomes more efficient.  We'd slash eventually either way.

burdges (approved):
It's fine.  It does not address the concern I poorly described as "time shifting effect" because..

This does not address when we broadcast our own assignments.   I'd think that's another PR probably.

burdges:
I prefer this version over my own suggestions:   We've many tranche zero assignments so we'll have multipe tranche zero no-shows occasionally.  We delay slightly for broadcasting the second obligatory tranche in my version, while we broadcast both tranches immediately in this version.  

It's also simpler due to completely purging the previous counting login in favor of this notion of non-empty tranche.

roadmap/implementers-guide/src/node/approval/approval-voting.md
rphmeier: I bring up the bitfield here just because it's the easiest way to get `n_validators` with the data we have available in this scope. A `usize` would be smaller, but since the ~1000 bits are already referenced by the `ApprovalEntry`, it's harmless to count the length here.

rphmeier:
bot merge

PR #1984, opened 2020-11-19
add max_pov_size to runtime config and PersistedValidationData

### LINKED ISSUE #1572: Max block size should be a limit in the runtime configuration
The `CandidateValidation` subsystem expects to have a local `max_block_data_size` which the underlying Wasm validation logic should use to do checks. This value should instead be drawn from the chain itself.

This should either be added  as a separate runtime API or added to the `ValidationData` (see #1539), probably as `Persisted`

coriolinus:
Closes #1572.

### DIFF
--- a/node/core/candidate-validation/src/lib.rs
+++ b/node/core/candidate-validation/src/lib.rs
@@ -347,16 +347,16 @@ async fn spawn_validate_exhaustive(
 /// are passed, `Err` otherwise.
 fn perform_basic_checks(
 	candidate: &CandidateDescriptor,
-	max_block_data_size: Option<u64>,
+	max_pov_size: u32,
 	pov: &PoV,
 ) -> Result<(), InvalidCandidate> {
+	use std::convert::TryInto;
+
 	let encoded_pov = pov.encode();
 	let hash = pov.hash();
 
-	if let Some(max_size) = max_block_data_size {
-		if encoded_pov.len() as u64 > max_size {
-			return Err(InvalidCandidate::ParamsTooLarge(encoded_pov.len() as u64));
-		}
+	if encoded_pov.len().try_into().unwrap_or(!0) > max_pov_size {
+		return Err(InvalidCandidate::ParamsTooLarge(encoded_pov.len() as u64));
 	}
 
 	if hash != candidate.pov_hash {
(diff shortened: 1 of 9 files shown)


runtime/parachains/src/configuration.rs
```diff
@@ -43,6 +43,8 @@ pub struct HostConfiguration<BlockNumber> {
 	pub max_code_size: u32,
 	/// The maximum head-data size, in bytes.
 	pub max_head_data_size: u32,
+	/// THe maximum POV block size, in bytes.
+	pub max_pov_size: u32,
```
coriolinus, line 47: Will we need a runtime upgrade to initialize this to a non-0 value? If yes, what should we use as a default? I haven't really played with those yet.

node/core/candidate-validation/src/lib.rs
```diff
@@ -859,15 +859,15 @@ mod tests {
 
 	#[test]
 	fn candidate_validation_bad_return_is_invalid() {
-		let validation_data: PersistedValidationData = Default::default();
+		let validation_data = PersistedValidationData::default().with(|vd| vd.max_pov_size=1024);
```
bkchr, line 862:
```suggestion
		let validation_data = PersistedValidationdata { max_pov_size = 1024, ..Default::default());
```

For this you don't need any special function.

node/core/candidate-validation/src/lib.rs
```diff
@@ -347,16 +347,16 @@ async fn spawn_validate_exhaustive(
...
+	use std::convert::TryInto;
+
 	let encoded_pov = pov.encode();
 	let hash = pov.hash();
 
-	if let Some(max_size) = max_block_data_size {
-		if encoded_pov.len() as u64 > max_size {
-			return Err(InvalidCandidate::ParamsTooLarge(encoded_pov.len() as u64));
-		}
+	if encoded_pov.len().try_into().unwrap_or(!0) > max_pov_size {
```
bkchr, line 358: Either switch the sides or just cast the max pov size to usize as this fits into usize. Than you don't need TryInto

runtime/parachains/src/configuration.rs
coriolinus: I think that just setting it in the three genesis locations will do it for us. Note: as of now, this is only set for rococo and the test function. If it should also be set for polkadot/kusama, then more changes are needed.

rphmeier: No runtime migration needed!

rphmeier (approved):
Thanks!

PR #1994, opened 2020-11-20
Sane messaging defaults for rococo

### LINKED ISSUE #1986: Sane messaging configuration values for rococo
Right now, the whole parachains configuration is initialized with default values (at least mostly).

Luckily, we are just not doing a great job at checking those, that allows us to conduct basic testing with parachains. I.e. there is no check that validation code is within limits for `schedule_para_initialize`.

But a default configuration would be nonetheless nice.

pepyakin:
Closes #1986


shawntabrizi (approved):

pepyakin:
bot merge

PR #1997, opened 2020-11-23
Companion: macro attribute: rename frame_system Trait -> Config

thiolliere:
companion for https://github.com/paritytech/substrate/pull/6877

Fix the only breaking change: rename frame_system Trait to Config


thiolliere:
closed as system trait is already renamed config

PR #2005, opened 2020-11-24
Do not send messages twice in bitfield distribution

bkchr:
This removes a bug which resulted in sending bitfield messages multiple
times by not checking if we already relayed them. Besides that it also
adds an optimization to not relay a message to a peer that send us
this message.

### DIFF
--- a/node/network/bitfield-distribution/src/lib.rs
+++ b/node/network/bitfield-distribution/src/lib.rs
@@ -330,12 +330,16 @@ where
 			// check interest in the peer in this message's relay parent
 			if view.contains(&message.relay_parent) {
 				// track the message as sent for this peer
-				message_sent_to_peer
+				let is_new = message_sent_to_peer
 					.entry(peer.clone())
 					.or_default()
 					.insert(validator.clone());
 
-				Some(peer.clone())
+				if is_new {
+					Some(peer.clone())
+				} else {
+					None
+				}
 			} else {
 				None
 			}
@@ -443,6 +447,9 @@ where
 		}
 		one_per_validator.insert(validator.clone(), message.clone());
 
+		// If the peer has sent us a message, we don't need to send him the same.
+		job_data.message_sent_to_peer.entry(origin.clone()).or_default().insert(validator.clone());
+
 		relay_message(ctx, job_data, &mut state.peer_views, validator, message).await;
 
 		modify_reputation(ctx, origin, BENEFIT_VALID_MESSAGE_FIRST).await
@@ -529,11 +536,7 @@ async fn handle_peer_view_change<Context>(
 where
 	Context: SubsystemContext<Message = BitfieldDistributionMessage>,
 {
-	let current = state.peer_views.entry(origin.clone()).or_default();
-
-	let added: Vec<Hash> = view.difference(&*current).cloned().collect();
-
-	*current = view;
+	let added = state.peer_views.entry(origin.clone()).or_default().replace_difference(view).cloned().collect::<Vec<_>>();
 
 	// Send all messages we've seen before and the peer is now interested
 	// in to that peer.
@@ -755,7 +758,7 @@ mod test {
 	use polkadot_node_subsystem_util::TimeoutExt;
 	use sp_keystore::{SyncCryptoStorePtr, SyncCryptoStore};
 	use sp_application_crypto::AppKey;
-	use sc_keystore::LocalKeystore;
+	use sp_keystore::testing::KeyStore;
 	use std::sync::Arc;
 	use std::time::Duration;
 	use assert_matches::assert_matches;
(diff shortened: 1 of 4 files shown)


node/network/bitfield-distribution/src/lib.rs
```diff
@@ -443,6 +447,9 @@ where
 		}
 		one_per_validator.insert(validator.clone(), message.clone());
 
+		// If the peer has sent us a message, we don't need to send him the same.
+		job_data.message_sent_to_peer.entry(origin.clone()).or_default().insert(validator.clone());
```
rphmeier, line 451: Isn't this already done above here? https://github.com/paritytech/polkadot/pull/2005/files#diff-6e99314d667011a73a36844509b05e6bfbf87e4cdcfa04d43d758d9ecf65fefbR424

rphmeier (commented):
LGTM except for my one comment about duplicate check

node/network/bitfield-distribution/src/lib.rs
bkchr: No, this is the set of received messages.

rphmeier: Oh, good point. But why should we update the `message_sent_to_peer` if we've only received the message? It seems like we should only update the `message_received_from_peer`

rphmeier: We should also check `message_received_from_peer` in the `interested_peers` computation in `relay_peers`, I guess.

bkchr: > Oh, good point. But why should we update the message_sent_to_peer if we've only received the message? It seems like we should only update the message_received_from_peer

If I have received the message from *you*, what is the benefit of sending you the same message again? This would happen below in `relay_message'. While you will not decrease my reputation for this, as I did not yet send you this message, I just waste my bandwidth for a message that is already known by you.

bkchr: > We should also check message_received_from_peer in the interested_peers computation in relay_peers, I guess.

We could also do it that way. I can change it that way if you like.

rphmeier: > While you will not decrease my reputation for this, as I did not yet send you this message, I just waste my bandwidth for a message that is already known by you.

Who says they're the same message? We only know that they're from the same validator. The bitfield distribution section of the guide goes deeper into the rationale.

So I prefer the other change, if you don't mind doing it

bkchr: > Who says they're the same message? We only know that they're from the same validator. The bitfield distribution section of the guide goes deeper into the rationale.

That is a good point, but we actually don't account for this in the code? As far as I have seen, we only check if we have send/received a message based on the validator is message is coming from.

You are saying that we actually should look into the message and check if they are different, is that correct?

rphmeier: > You are saying that we actually should look into the message and check if they are different, is that correct?

No, the change I suggest is just to check if we've received a message from the peer before sending a message. I think the thing this is trying to work around is where both peers send each other the message at the same time and then report each other.

node/network/bitfield-distribution/src/lib.rs
```diff
@@ -1253,4 +1320,84 @@ mod test {
...
+		let peer_b = PeerId::random();
+		assert_ne!(peer_a, peer_b);
+
+		// validator 0 key pair
+		let (mut state, signing_context, keystore, validator) = state_with_view(view![hash], hash);
+
+		// create a signed message by validator 0
+		let payload = AvailabilityBitfield(bitvec![bitvec::order::Lsb0, u8; 1u8; 32]);
+		let signed_bitfield =
+			executor::block_on(Signed::<AvailabilityBitfield>::sign(&keystore, payload, &signing_context, 0, &validator))
```
pepyakin, line 1343: Would you mind wrapping this and other lines introduced by this PR that grew unwieldly?

node/network/bitfield-distribution/src/lib.rs
bkchr: No, I don't try to work around this and this did not happen. We also just report a peer if it sends us the same message twice. We don't check if we send the message to this peer already for reporting the peer. Because as you said, this is very likely to end in a race condition.

However, I will do your requested changes.

bkchr: The other thing that I spoke about is that the code currently assumes that each validator only sends one bitfield per relay chain. Will this always be true or will we may change this in the future?

rphmeier (approved):

node/network/bitfield-distribution/src/lib.rs
rphmeier: I think this will be true for the foreseeable future

PR #2006, opened 2020-11-24
backing: reorder votes to match bitfield in backed candidate

rphmeier:
fixes a bug raised by @pepyakin and adds a test case that failed before and passes now.


pepyakin:
I've checked this fix (although with my local implementation of the fix) and it worked. Thus, can you also remove [this](https://github.com/paritytech/polkadot/blob/ab5dd449f854c5dcffea225f3e4a9854e57786a3/runtime/parachains/src/inclusion_inherent.rs#L141) one?

coriolinus (approved):

montekki (approved):

rphmeier:
> I've checked this fix (although with my local implementation of the fix) and it worked without any inherent problems. Thus, can you also remove this one?

That is a separate thing, which I don't feel comfortable removing at this time. If there is a bug in candidate backing or bitfield selection which causes the validators to always attempt to author blocks with invalid inherent, I would rather halt all parachains than halt the relay chain.

PR #2010, opened 2020-11-25
Move erasure root out of candidate commitments and into descriptor

rphmeier:
Only descriptors are available at the point where we do availability recovery, so we needed this to be in the descriptor.

It also made it easier to reason about commitments as stuff that the chain needs to be notified of. In fact, it made the `ValidationOutputs` struct completely redundant, so I removed it and used `CandidateCommitments` everywhere it used to be.

Candidate backing has the largest changeset. I made sure we are still checking:
  1. That the validated commitments match the ones sent to us by a collator / another validator
  1. That the erasure root matches the ones sent to us by a collator / another validator.

Previously, these checks were the same, but now they are separate.

### DIFF
--- a/node/core/backing/src/lib.rs
+++ b/node/core/backing/src/lib.rs
@@ -31,8 +31,7 @@ use polkadot_primitives::v1::{
 	CommittedCandidateReceipt, BackedCandidate, Id as ParaId, ValidatorId,
 	ValidatorIndex, SigningContext, PoV, CandidateHash,
 	CandidateDescriptor, AvailableData, ValidatorSignature, Hash, CandidateReceipt,
-	CandidateCommitments, CoreState, CoreIndex, CollatorId, ValidationOutputs,
-	ValidityAttestation,
+	CoreState, CoreIndex, CollatorId, ValidityAttestation,
 };
 use polkadot_node_primitives::{
 	FromTableMisbehavior, Statement, SignedFullStatement, MisbehaviorReport, ValidationResult,
@@ -229,6 +228,8 @@ impl TryFrom<AllMessages> for FromJob {
 	}
 }
 
+struct InvalidErasureRoot;
+
 // It looks like it's not possible to do an `impl From` given the current state of
 // the code. So this does the necessary conversion.
 fn primitive_statement_to_table(s: &SignedFullStatement) -> TableSignedStatement {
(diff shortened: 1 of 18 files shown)


node/core/backing/src/lib.rs
```diff
@@ -733,18 +736,16 @@ impl CandidateBackingJob {
...
 	//
-	// This calls an inspection function before making the PoV available for any last checks
-	// that need to be done. If the inspection function returns an error, this function returns
-	// early without making the PoV available.
-	#[tracing::instrument(level = "trace", skip(self, pov, with_commitments), fields(subsystem = LOG_TARGET))]
-	async fn make_pov_available<T, E>(
+	// This will compute the erasure root internally and compare it to the expected erasure root.
+	// This returns `Err()` iff there is an internal error. Otherwise, it returns either `Ok(Ok(()))` or `Ok(Err(_))`.
+	#[tracing::instrument(level = "trace", skip(self, pov), fields(subsystem = LOG_TARGET))]
+	async fn make_pov_available(
```
coriolinus, line 742: It might be worth factoring this function into two: one which checks the erasure root, and one which stores the available data. The goal is not really to reduce the amount of work done; it's more to avoid non-sequitur lines like `let erasure_valid = self.make_pov_available()`.

node/core/backing/src/lib.rs
```diff
@@ -347,36 +348,38 @@ impl CandidateBackingJob {
...
+							let candidate = CommittedCandidateReceipt {
+								descriptor: candidate.descriptor().clone(),
+								commitments,
+							};
+
+							self.issued_statements.insert(candidate_hash);
+							Some(Statement::Seconded(candidate))
+						}
+						Err(InvalidErasureRoot) => {
+							self.issue_candidate_invalid_message(candidate.clone()).await?;
```
coriolinus, line 380: It looks like we don't use `candidate` past this point, so we shouldn't need to `clone` it.

coriolinus (approved):

node/core/backing/src/lib.rs
pepyakin: It may help remove this pesky nested `Result` as well

pepyakin (approved):

node/core/backing/src/lib.rs
rphmeier: beyond the scope of this PR

node/core/backing/src/lib.rs
rphmeier:
```suggestion
							self.issue_candidate_invalid_message(candidate).await?;
```

rphmeier:
bot merge

pepyakin:
apparently we do need clone, because what we have at hand is a reference, but `issue_candidate_invalid_message` expects an owned copy.

pepyakin:
bot merge

PR #2011, opened 2020-11-25
guide: Availability Recovery

### LINKED ISSUE #1597: Guide: Availability Recovery Subsystem
This subsystem is the inverse of the availability distribution subsystem: validators will serve the availability chunks kept in the availability store to nodes who connect to them. And the subsystem will also implement the other side: the logic for nodes to connect to validators, request availability pieces, and reconstruct the `AvailableData`.

One main difficulty will be addressing the fact that the pieces may have been made available during session N, while we are currently in session N+K, where session keys may have changed. Thus, we will expect some data that lives on-chain to allow us to determine the session in which some receipt was included, who the validators were during that session, and what their authority-discovery key was. This will enable us to do the necessary DHT lookup to find the node to connect to, as long as their actual address hasn't changed.

This data can live partially on-chain and mostly off-chain, by using the technique introduced in https://github.com/paritytech/substrate/pull/6220, where the chain stores cryptographic commitments in the state and triggers a write to an off-chain database which is maintained by the full node.

rphmeier:
Closes #1597

### DIFF
--- a/roadmap/implementers-guide/src/node/availability/availability-recovery.md
+++ b/roadmap/implementers-guide/src/node/availability/availability-recovery.md
@@ -147,7 +147,7 @@ On `Conclude`, shut down the subsystem.
 
 1. Compute the threshold from the session info. It should be `f + 1`, where `n = 3f + k`, where `k in {1, 2, 3}`, and `n` is the number of validators.
 1. Set the various fields of `Interaction` based on the validator lists in `session_info`. Compute a random shuffling of the validator indices.
-1. Set the `to_state` sender to be equal to 
+1. Set the `to_state` sender to be equal to a clone of `state.from_interaction_tx`.
 1. Initialize `received_chunks` to an empty set, as well as `requesting_chunks`.
 
 Launch the interaction as a background task running `interaction_loop(interaction)`.
--- a/roadmap/implementers-guide/src/node/availability/availability-recovery.md
+++ b/roadmap/implementers-guide/src/node/availability/availability-recovery.md
@@ -4,6 +4,8 @@
 
 This subsystem is the inverse of the [Availability Distribution](availability-distribution.md) subsystem: validators will serve the availability chunks kept in the availability store to nodes who connect to them. And the subsystem will also implement the other side: the logic for nodes to connect to validators, request availability pieces, and reconstruct the `AvailableData`.
 
+This version of the availability recovery subsystem is based off of direct connections to validators. In order to recover any given `AvailableData`, we must recover at least `f + 1` pieces from validators of the session. Thus, we will connect to and query randomly chosen validators until we have received `f + 1` pieces.
+
 ## Protocol
 
 `PeerSet`: `Validation`
@@ -11,7 +13,7 @@ This subsystem is the inverse of the [Availability Distribution](availability-di
 Input:
 
 - NetworkBridgeUpdateV1(update)
-- TODO: input message to request a fetch.
+- AvailabilityRecoveryMessage::RecoverAvailableData(candidate, session, response)
 
 Output:
 
--- a/roadmap/implementers-guide/src/node/availability/availability-recovery.md
+++ b/roadmap/implementers-guide/src/node/availability/availability-recovery.md
@@ -22,3 +22,34 @@ Output:
 - AvailabilityStore::QueryChunk
 
 ## Functionality
+
+We hold a state which tracks the current recovery interactions we have live, as well as which request IDs correspond to which interactions. An interaction is a structure encapsulating all interaction with the network necessary to recover the available data.
+
+```rust
+struct State {
+    /// The interaction handle may just be the interaction itself, but it might be useful to implement each interaction
+    /// as its own `Future`, running in its own task, in which case the handle will be a channel to communicate with
+    /// the interaction.
+    interactions: Map<CandidateHash, InteractionHandle>,
+    live_requests: Map<RequestId, (PeerId, CandidateHash)>,
+    next_request_id: RequestId,
+
+    // An LRU cache of recently recovered data.
+    availability_lru: LruCache<CandidateHash, AvailableData>,
+}
+```
+
+```rust
+struct Interaction {
+    responses: ResponseChannel<Option<AvailableData>>,
+    validator_authority_keys: Vec<AuthorityId>,
+    validators: Vec<ValidatorId>,
+    // a random shuffling of the validators which indicates the order in which we connect to the validators and
+    // request the chunk from them.
+    shuffling: Vec<ValidatorIndex>, 
+    // The number of pieces needed.
+    threshold: usize, 
+    received_chunks: Map<ValidatorIndex, ErasureChunk>,
+    requesting_chunks: Set<ValidatorIndex>,
+}
+```
--- a/roadmap/implementers-guide/src/node/availability/availability-recovery.md
+++ b/roadmap/implementers-guide/src/node/availability/availability-recovery.md
@@ -26,22 +26,58 @@ Output:
 We hold a state which tracks the current recovery interactions we have live, as well as which request IDs correspond to which interactions. An interaction is a structure encapsulating all interaction with the network necessary to recover the available data.
 
 ```rust
+type ChunkResponse = Result<(PeerId, ErasureChunk), Unavailable>;
+
+struct AwaitedChunk {
+    issued_at: Instant,
+    validator_index: ValidatorIndex,
+    candidate_hash: CandidateHash,
+    response: ResponseChannel<ChunkResponse>,
+}
+
 struct State {
-    /// The interaction handle may just be the interaction itself, but it might be useful to implement each interaction
-    /// as its own `Future`, running in its own task, in which case the handle will be a channel to communicate with
-    /// the interaction.
+    /// Each interaction is implemented as its own async task, and these handles are for communicating with them.
     interactions: Map<CandidateHash, InteractionHandle>,
-    live_requests: Map<RequestId, (PeerId, CandidateHash)>,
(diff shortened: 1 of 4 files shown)


burdges:
As an optimization, approval assignments could implicitly requests pieces:  

Alice received assignment from Bob to check candidate/core C.  Alice computes a permutation P of the set {0,..,_num_validators} using PRNG seeded by Bob's VRF output, so that everyone gets the same permutation.  Alice iterates over the availability votes covering C in the order P.  Alice halts if she finds 1/3 + epsilon "yes available" votes before reaching her own index P[i].  If Alice reaching her own index as P[i] then she send Bob her piece after a courtesy delay computed from i. 

Let k denote the least index such that P[i] for i=0,..,k contains num_validators/3 "yes available" votes.  If Bob detects any pieces running late as computed using courtesy delays, then Bob starts requesting from the validator with index P[i] for i=k,..,num_validators, and eventually cycles back into i=0,..,k using roughly `AvailabilityRecoveryV1Message`. 

All this complexity could be omitted for now since we'll always need explicit requests.

roadmap/implementers-guide/src/node/availability/availability-recovery.md
```diff
@@ -26,30 +26,144 @@ Output:
...
\ No newline at end of file
+```
+
+### Signal Handling
+
+On `ActiveLeavesUpdate`, if `activated` is non-emtpy, set `state.live_block_hash` to the first block in `Activated. Also, remove any items that have been in 
+
+Ignore `BlockFinalized` signals.
+
+On `Conclude`, shut down the subsystem.
```
burdges, line 101: Is the node shutting down here? We'd prefer if recovery finished for whatever assignments we annouced, but doing that can be future work.

roadmap/implementers-guide/src/node/availability/availability-recovery.md
```diff
@@ -26,30 +26,144 @@ Output:
...
+1. If there exists an entry under `r_id`, remove it. If there doesn't exist one, report the peer and return. If the peer in the entry doesn't match the sending peer, reinstate the entry, report the peer, and return.
+1. Send the chunk response on the `awaited_chunk` for the interaction to handle.
+
+### Interaction logic
+
+#### `launch_interaction(session_index, session_info, candidate_receipt, candidate_hash)`
+
+1. Compute the threshold from the session info. It should be `f + 1`, where `n = 3f + k`, where `k in {1, 2, 3}`, and `n` is the number of validators.
+1. Set the various fields of `Interaction` based on the validator lists in `session_info`. Compute a random shuffling of the validator indices.
+1. Set the `to_state` sender to be equal to
```
burdges, line 150: missing text

roadmap/implementers-guide/src/node/availability/availability-recovery.md
rphmeier: Yeah, this is the node shutting down. I think we'll file this as 'future work': https://github.com/paritytech/polkadot/issues/2029

rphmeier: I think we'd best handle this on the level of other subsystems - it's really the requesting subsystem that knows best if a particular availability recovery interaction can be safely cancelled or not. I think this subsystem should be ready to conclude at any time, whereas the approval voting subsystem will have to be aware of the recovery requests it has in-flight.

roadmap/implementers-guide/src/node/availability/availability-recovery.md
```diff
@@ -20,3 +22,148 @@ Output:
...
+#### `interaction_loop(interaction)`
+
+```rust
+// How many parallel requests to have going at once.
+const N_PARALLEL: usize = 50;
+```
+
+Loop:
+  * Poll for new updates from `requesting_chunks`. Check merkle proofs of any received chunks, and any failures should lead to issuance of a `FromInteraction::ReportPeer` message.
+  * If `received_chunks` has more than `threshold` entries, attempt to recover the data. If that fails, or a re-encoding of it doesn't match the expected erasure root, break and issue a `FromInteraction::Concluded(RecoveryError::Invalid)`. Otherwise, issue a `FromInteraction::Concluded(Ok(()))`.
```
montekki, line 164: Recovery errors are a sign that some validators have issued faulty chunks right? If that is so shouldn't this issue be lifted to disputes? If that should what is the mechanism for proving to everyone that these validators issued these wrong chunks?

roadmap/implementers-guide/src/node/availability/availability-recovery.md
```diff
@@ -20,3 +22,148 @@ Output:
...
+#### On receiving `AvailabilityRecoveryV1::Chunk(r_id, chunk)`
+
+1. If there exists an entry under `r_id`, remove it. If there doesn't exist one, report the peer and return. If the peer in the entry doesn't match the sending peer, reinstate the entry, report the peer, and return.
+1. Send the chunk response on the `awaited_chunk` for the interaction to handle.
+
+### Interaction logic
+
+#### `launch_interaction(session_index, session_info, candidate_receipt, candidate_hash)`
+
+1. Compute the threshold from the session info. It should be `f + 1`, where `n = 3f + k`, where `k in {1, 2, 3}`, and `n` is the number of validators.
```
montekki, line 148: DW: I always believed that this threshold in the AV protocol simply put says this: "as long as 2/3+1 validators are honest and will send us the piece the recovery would work. So if we request a piece from every validator at least 2/3+1 would send us a chunk and then this will be enough to recover data". Also `inclusion.rs` defines  the threshold as

```rust
const fn availability_threshold(n_validators: usize) -> usize {
	let mut threshold = (n_validators * 2) / 3;
	threshold += (n_validators * 2) % 3;
	threshold
}
```

So two questions here:
 1. Which is the correct formula for the threshold?
 1. We should probably be querying all validators and waiting for at least `threshold` correct pieces shouldn't we?

roadmap/implementers-guide/src/node/availability/availability-recovery.md
```diff
@@ -20,3 +22,148 @@ Output:
...
+    candidate_hash: Hash,
+    erasure_root: Hash,
+    received_chunks: Map<ValidatorIndex, ErasureChunk>,
+    requesting_chunks: FuturesUnordered<Receiver<ChunkResponse>>,
+}
+```
+
+### Signal Handling
+
+On `ActiveLeavesUpdate`, if `activated` is non-emtpy, set `state.live_block_hash` to the first block in `Activated. Also, remove any items that have been in
```
montekki, line 97:
```suggestion
On `ActiveLeavesUpdate`, if `activated` is non-emtpy, set `state.live_block_hash` to the first block in `Activated`. Also, remove any items that have been in 
```

The sentence looks unfinished.

roadmap/implementers-guide/src/node/availability/availability-recovery.md
```diff
@@ -20,3 +22,148 @@ Output:
...
+    validator_index: ValidatorIndex,
+    candidate_hash: CandidateHash,
+    response: ResponseChannel<ChunkResponse>,
+}
+
+struct State {
+    /// Each interaction is implemented as its own async task, and these handles are for communicating with them.
+    interactions: Map<CandidateHash, InteractionHandle>,
+    /// A recent block hash for which state should be available.
+    live_block_hash: Hash,
```
montekki, line 42: Probably this should be a `BlockNumber` or at least we should check that when we update this field the new block is higher than the old one?

roadmap/implementers-guide/src/node/availability/availability-recovery.md
rphmeier: Yeah, it should be escalated to disputes. My intention was to have approval voting handle that. The dispute participation subsystem will also make availability recovery requests and we'd expect it should encounter the same error. I elaborate on why that is below.

One of the things we are trying to avoid is the attack vector where some of the pieces successfully recover the data and some do not. Events might unfold in this order:
 * Validator A is given a set of pieces containing a bad piece and returns `RecoveryError::Invalid`. I am assuming that the adversary, by controlling the network and some bad validators, can ensure that only validator A can recover the bad piece.
 * Validators B-Z observe the dispute and each recovers a set of pieces which correctly recovery the `AvailableData`.
  * Validators B-Z vote to slash A, who has behaved according to protocol.

The way that we mitigate this attack is with by checking that "a re-encoding of it doesn't match the expected erasure root". Because we have this check, validators B-Z will all notice that the `erasure_root` in the candidate descriptor doesn't match, and thus return a `RecoveryError::Invalid`. Thus, the dispute will resolve against the candidate's backers.

cc @drahnr

roadmap/implementers-guide/src/node/availability/availability-recovery.md
rphmeier: 1. We only require `f + 1` (1/3 + 1) pieces to recover from the erasure coding because that is a subset which is guaranteed to contain at least one honest node's piece. However, we consider something available (by the bitfield system) only when `2f + 1` validators claim that they have their piece, because this means that there is at least a set of `f + 1` honest nodes who all have their piece, and thus the data will still be available even if all `f` adversarial nodes refuse to respond to requests.

2. That's effectively what the `Interaction` does. It queries random batches of validators for the session, a few at a time, and asks them each for their piece. Until we have f+1 pieces, and then we attempt to recover.

roadmap/implementers-guide/src/node/availability/availability-recovery.md
rphmeier: What we really care about is `SlotNumber`, not block number. But yeah, there is some risk of moving backwards.

montekki (approved):

montekki:
bot merge

PR #2026, opened 2020-11-27
proposer: guard all provisioner data work with timeout

andresilva:
From `wait_for_activation` documentation:

```
	/// Note that due the fact the overseer doesn't store the whole active-leaves set, only deltas,
	/// the response channel may never return if the hash was deactivated before this call.
	/// In this case, it's the caller's responsibility to ensure a timeout is set.
```

I have replicated the following error: when syncing the chain from scratch as a validator we might start authoring when we are close to the tip (since the major sync heuristic will start returning false), but in the meantime we still import more blocks. If we start authoring at block N and in the meantime it's children N+1 and N+2 are imported, then the active leaf for block N will have already been pruned, but we'll re-add the listener on the call to `wait_for_activation` and we'll be waiting on an activation that never happens.

This PR makes the whole work block in `get_provisioner_data` be guarded by the timeout.

### DIFF
--- a/node/core/proposer/src/lib.rs
+++ b/node/core/proposer/src/lib.rs
@@ -141,20 +141,23 @@ where
 		let mut overseer = self.overseer.clone();
 		let parent_header_hash = self.parent_header_hash.clone();
 
-		let (sender, receiver) = futures::channel::oneshot::channel();
+		let pid = async {
+			let (sender, receiver) = futures::channel::oneshot::channel();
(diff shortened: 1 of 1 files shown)


ordian (approved):
Thanks, I somehow assumed this was already the case that we handle timeouts here:
https://github.com/paritytech/polkadot/pull/1616#issuecomment-677672651.

PR #2027, opened 2020-11-27
Proof of concept: Jaeger integration to debug parachains

tomaka:
This PR isn't intended to be merged but to demonstrate the idea of reporting timelines to Jaeger.

Because I'm not very familiar with the codebase, I've only put spans at a few selected areas and not everywhere, but here is for example the result I obtain for each relay chain block:

image

Each relay chain block has its own individual trace (i.e. the timeline shown on the screenshot) that collects the information reported by every single node regarding this block.

The way you use it is to run:

```sh
docker run -d --name jaeger \
  -e COLLECTOR_ZIPKIN_HTTP_PORT=9411 \
  -p 5775:5775/udp \
  -p 6831:6831/udp \
  -p 6832:6832/udp \
  -p 5778:5778 \
  -p 16686:16686 \
  -p 14268:14268 \
  -p 14250:14250 \
  -p 9411:9411 \
  jaegertracing/all-in-one:1.21
```

Then run the nodes. They will automatically send their spans through UDP to the agent.
The UI (where to see the traces) can be accessed through <https://localhost:16686>.

### DIFF
--- a/node/subsystem/src/lib.rs
+++ b/node/subsystem/src/lib.rs
@@ -28,7 +28,7 @@ use futures::prelude::*;
 use futures::channel::{mpsc, oneshot};
 use futures::future::BoxFuture;
 
-use polkadot_primitives::v1::Hash;
+use polkadot_primitives::v1::{Hash, PoV};
 use async_trait::async_trait;
 use smallvec::SmallVec;
 use thiserror::Error;
@@ -77,7 +77,7 @@ impl PartialEq for ActiveLeavesUpdate {
 	///
 	/// Instead, it means equality when `activated` and `deactivated` are considered as sets.
 	fn eq(&self, other: &Self) -> bool {
-		self.activated.len() == other.activated.len() && self.deactivated.len() == other.deactivated.len() 
+		self.activated.len() == other.activated.len() && self.deactivated.len() == other.deactivated.len()
 			&& self.activated.iter().all(|a| other.activated.contains(a))
 			&& self.deactivated.iter().all(|a| other.deactivated.contains(a))
 	}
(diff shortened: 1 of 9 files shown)


node/subsystem/src/lib.rs
```diff
@@ -261,3 +261,76 @@ where
...
+		buf.copy_from_slice(&hash.as_ref()[0..16]);
+		std::num::NonZeroU128::new(u128::from_be_bytes(buf)).unwrap()
+	};
+
+	TRACES_IN.lock().unwrap().as_ref().unwrap().span(trace_id, span_name)
+}
+
+/// Must be called otherwise the other jaeger-related functions will panic.
+///
+/// THIS API IS REALLY BAD AND SHOULDN'T BE FOUND IN THE MASTER BRANCH
```
ordian, line 310: what do you think the way to move forward would be?

can we use these libs?
https://github.com/tokio-rs/tracing/tree/HEAD/tracing-opentelemetry
https://github.com/open-telemetry/opentelemetry-rust/tree/HEAD/opentelemetry-jaeger

tomaka: > what do you think the way to move forward would be?

Properly integrate it in Substrate, and pass an `Arc<TracesIn>` as parameter to the various subsystems.

> can we use these libs?

I'm going to paste this conversation here, for posterity:

image

node/subsystem/src/lib.rs
```diff
@@ -261,3 +261,76 @@ where
...
+//
+
+/// Shortcut for [`hash_span`] with the hash of the `PoV`.
+pub fn pov_span(pov: &PoV, span_name: &str) -> mick_jaeger::Span {
+	hash_span(&pov.hash(), span_name)
+}
+
+/// Creates a `Span` referring to the given hash. All spans created with [`hash_span`] with the
+/// same hash (even from multiple different nodes) will be visible in the same view on Jaeger.
+pub fn hash_span(hash: &Hash, span_name: &str) -> mick_jaeger::Span {
```
rphmeier, line 298: we should add a `candidate_span` similar to `pov_span` but for candidate hashes

tomaka:
Closing in favour of #2047, as there's no reason this leave this PR open.

PR #2036, opened 2020-11-30
rename storage Config to ActiveConfig

thiolliere:
This PR rename the storage `Config` in Configuration pallet to `ActiveConfig`.

This is to prevent name collision with incoming renaming of `Trait` into `Config`.

I think this pallet is used in some tests chains, thus I implemented the migration. Note that I don't think the migration needs any version check and bump because it only needs to migrate a value whenever it exist.

related: https://github.com/paritytech/polkadot/pull/2014

### DIFF
--- a/runtime/parachains/src/configuration.rs
+++ b/runtime/parachains/src/configuration.rs
@@ -23,6 +23,7 @@ use primitives::v1::{Balance, ValidatorId, SessionIndex};
 use frame_support::{
 	decl_storage, decl_module, decl_error,
 	ensure,
+	storage::migration,
 	dispatch::DispatchResult,
 	weights::{DispatchClass, Weight},
 };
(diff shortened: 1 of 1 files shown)


runtime/parachains/src/configuration.rs
```diff
@@ -150,6 +151,20 @@ decl_module! {
...
+				b"",
+			);
+
+			if let Some(value) = value {
+				ActiveConfig::<T>::put(value)
+			}
+
+			0
+		}
+
```
bkchr, line 167:
```suggestion
```

bkchr (approved):
This is not yet used anywhere, so we don't need any migrations! :)

thiolliere:
then I'll close this one and do the rename directly in the Trait to Config PR

PR #2037, opened 2020-11-30
Simplify subsystem jobs

bkchr:
This pr simplifies the subsystem jobs interface. Instead of requiring an
extra message that is used to signal that a job should be ended, a job
now ends when the receiver returns `None`. Besides that it changes the
interface to enforce that messages to a job provide a relay parent.

### DIFF
--- a/node/subsystem-util/src/lib.rs
+++ b/node/subsystem-util/src/lib.rs
@@ -56,6 +56,7 @@ use std::{
 	pin::Pin,
 	task::{Poll, Context},
 	time::Duration,
+	fmt,
 };
 use streamunordered::{StreamUnordered, StreamYield};
 use thiserror::Error;
@@ -117,18 +118,11 @@ pub async fn request_from_runtime<RequestBuilder, Response, FromJob>(
 ) -> Result<RuntimeApiReceiver<Response>, Error>
 where
 	RequestBuilder: FnOnce(RuntimeApiSender<Response>) -> RuntimeApiRequest,
-	FromJob: TryFrom<AllMessages>,
-	<FromJob as TryFrom<AllMessages>>::Error: std::fmt::Debug,
+	FromJob: From<AllMessages>,
 {
 	let (tx, rx) = oneshot::channel();
 
-	sender
-		.send(
-			AllMessages::RuntimeApi(RuntimeApiMessage::Request(parent, request_builder(tx)))
-				.try_into()
-				.map_err(|err| Error::SenderConversion(format!("{:?}", err)))?,
-		)
-		.await?;
+	sender.send(AllMessages::RuntimeApi(RuntimeApiMessage::Request(parent, request_builder(tx))).into()).await?;
 
 	Ok(rx)
 }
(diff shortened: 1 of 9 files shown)


node/subsystem-util/src/lib.rs
```diff
@@ -394,13 +394,10 @@ impl Validator {
...
-/// Normally, this will be some subset of `Allmessages`, and a `Stop` variant.
+/// Normally, this will be some subset of `Allmessages`.
 pub trait ToJobTrait: TryFrom<AllMessages> {
-	/// The `Stop` variant of the ToJob enum.
-	const STOP: Self;
-
-	/// If the message variant contains its relay parent, return it here
-	fn relay_parent(&self) -> Option<Hash>;
+	/// Get the relay parent this message is assigned to.
+	fn relay_parent(&self) -> Hash;
```
coriolinus, line 400: This makes `ToJobTrait` a bit more rigid: it ensures that we'll never write a subsystem using this util which can have jobs not tied to relay parents.

I think this is a reasonable simplification at this point.

node/subsystem-util/src/lib.rs
```diff
@@ -394,13 +394,10 @@ impl Validator {
 
 /// ToJob is expected to be an enum declaring the set of messages of interest to a particular job.
 ///
-/// Normally, this will be some subset of `Allmessages`, and a `Stop` variant.
+/// Normally, this will be some subset of `Allmessages`.
 pub trait ToJobTrait: TryFrom<AllMessages> {
```
coriolinus, line 398: Given the changes already in this PR, it's not clear that we actually need `ToJobTrait` anymore. It could simplify subsystem definitions if their inbound channel accepted `AllMessages`: their matches on inbound messages would need to explicitly discard irrelevant ones instead of implicitly in the `TryFrom<AllMessages> for ToJob` definitions, but the total amount of code would still be reduced.

node/subsystem/src/messages.rs
```diff
@@ -526,12 +526,12 @@ pub enum ProvisionerMessage {
 }
 
 impl ProvisionerMessage {
-	/// If the current variant contains the relay parent hash, return it.
-	pub fn relay_parent(&self) -> Option<Hash> {
+	/// Returns the relay parent this message is assigned to.
+	pub fn relay_parent(&self) -> Hash {
```
coriolinus, line 530: Except where you have changed it in this PR, all `*Message::relay_parent()` returns `Option<Hash>`. It would be nice to maintain consistency across the interface, even if we don't abstract it into a trait.

coriolinus (commented):
Agree this does what it says. It could go farther: getting rid of `ToJob` would enable us to remove even more code.

node/subsystem/src/messages.rs
bkchr: We could also remove the other ones if they are not required.

I see this as any message to a subsystem that spawns jobs needs to provide a relay parent. While messages to long running subsystems don't need to provide a relay parent.

bkchr: This means, we can either remove the other functions or just keep it the way it is done currently.

coriolinus: Right, I see what you mean. I don't have any strong objections to just removing `fn relay_parent` from those messages which don't need it.

coriolinus (approved):

ordian (approved):
Nice :+1:

PR #2042, opened 2020-11-30
Cadidate selection check assignment

montekki:
> This subsystem is only ever interested in parablocks assigned to the particular parachain which this validator is currently handling.

### DIFF
--- a/node/core/candidate-selection/src/lib.rs
+++ b/node/core/candidate-selection/src/lib.rs
@@ -132,6 +132,26 @@ enum Error {
 	ChainApi(#[from] ChainApiError),
 }
 
+macro_rules! try_runtime_api {
+	($x: expr) => {
+		match $x {
+			Ok(x) => x,
+			Err(e) => {
+				tracing::warn!(
+					target: LOG_TARGET,
+					err = ?e,
+					"Failed to fetch runtime API data for job",
+				);
+
+				// We can't do candidate selection work if we don't have the
+				// requisite runtime API data. But these errors should not take
+				// down the node.
+				return Ok(());
+			}
+		}
+	}
+}
+
 impl JobTrait for CandidateSelectionJob {
 	type ToJob = ToJob;
 	type FromJob = FromJob;
@@ -153,33 +173,13 @@ impl JobTrait for CandidateSelectionJob {
 		mut sender: mpsc::Sender<FromJob>,
 	) -> Pin<Box<dyn Future<Output = Result<(), Self::Error>> + Send>> {
 		Box::pin(async move {
-			macro_rules! try_runtime_api {
-				($x: expr) => {
-					match $x {
-						Ok(x) => x,
-						Err(e) => {
-							tracing::warn!(
-								target: LOG_TARGET,
-								err = ?e,
-								"Failed to fetch runtime API data for job",
-							);
-
-							// We can't do candidate validation work if we don't have the
-							// requisite runtime API data. But these errors should not take
-							// down the node.
-							return Ok(());
-						}
-					}
-				}
-			}
-
 			let (groups, cores) = futures::try_join!(
-				request_validator_groups(relay_parent, &mut sender).await?,
-				request_from_runtime(
+				try_runtime_api!(request_validator_groups(relay_parent, &mut sender).await),
+				try_runtime_api!(request_from_runtime(
 					relay_parent,
 					&mut sender,
 					|tx| RuntimeApiRequest::AvailabilityCores(tx),
-				).await?,
+				).await),
 			)?;
 
 			let (validator_groups, group_rotation_info) = try_runtime_api!(groups);
@@ -193,9 +193,8 @@ impl JobTrait for CandidateSelectionJob {
 				Err(err) => return Err(Error::Util(err)),
 			};
 
-			let mut groups = HashMap::new();
-
 			let mut assignment = None;
+
 			for (idx, core) in cores.into_iter().enumerate() {
 				// Ignore prospective assignments on occupied cores for the time being.
 				if let CoreState::Scheduled(scheduled) = core {
@@ -206,7 +205,6 @@ impl JobTrait for CandidateSelectionJob {
 							assignment = Some(scheduled.para_id);
 							break;
 						}
-						let _ = groups.insert(scheduled.para_id, g.clone());
 					}
 				}
 			}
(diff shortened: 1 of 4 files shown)


node/core/candidate-selection/src/lib.rs
```diff
@@ -206,6 +281,16 @@ impl CandidateSelectionJob {
...
 		let _timer = self.metrics.time_handle_collation();
 
+		if self.assignment != para_id {
+			tracing::info!(
+				target: LOG_TARGET,
+				"Collator {:?} sent a collation outside of our assignment {:?}",
+				collator_id,
+				para_id,
+			);
+			return;
```
montekki, line 291: probably we should report or disconnect from this collator

node/core/candidate-selection/src/lib.rs
```diff
@@ -19,27 +19,38 @@
 
 #![deny(missing_docs, unused_crate_dependencies, unused_results)]
 
+use std::collections::HashMap;
```
bkchr, line 22:
```suggestion
```

node/core/candidate-selection/src/lib.rs
```diff
@@ -19,27 +19,38 @@
...
+	request_from_runtime,
+	Validator,
+	JobTrait, ToJobTrait, FromJobCommand,
 	metrics::{self, prometheus},
 };
-use polkadot_primitives::v1::{CandidateReceipt, CollatorId, Hash, Id as ParaId, PoV};
+use polkadot_primitives::v1::{
+	CandidateReceipt, CollatorId, CoreState, CoreIndex, Hash, Id as ParaId, PoV,
+};
 use std::{convert::TryFrom, pin::Pin};
```
bkchr, line 47:
```suggestion
use std::{convert::TryFrom, pin::Pin, collections::HashMap};
```

node/core/candidate-selection/src/lib.rs
```diff
@@ -126,24 +137,86 @@ impl JobTrait for CandidateSelectionJob {
...
+		relay_parent: Hash,
+		keystore: Self::RunArgs,
 		metrics: Self::Metrics,
 		receiver: mpsc::Receiver<ToJob>,
-		sender: mpsc::Sender<FromJob>,
+		mut sender: mpsc::Sender<FromJob>,
 	) -> Pin<Box<dyn Future<Output = Result<(), Self::Error>> + Send>> {
 		Box::pin(async move {
-			let job = CandidateSelectionJob::new(metrics, sender, receiver);
+			macro_rules! try_runtime_api {
```
bkchr, line 157: Does this macro really needs to be declared in this function?

node/core/candidate-selection/src/lib.rs
```diff
@@ -126,24 +137,86 @@ impl JobTrait for CandidateSelectionJob {
...
+							// requisite runtime API data. But these errors should not take
+							// down the node.
+							return Ok(());
+						}
+					}
+				}
+			}
+
+			let (groups, cores) = futures::try_join!(
+				request_validator_groups(relay_parent, &mut sender).await?,
```
bkchr, line 178: Below we just print a warning, but here we return the error?

node/core/candidate-selection/src/lib.rs
```diff
@@ -126,24 +137,86 @@ impl JobTrait for CandidateSelectionJob {
...
+
+			let mut assignment = None;
+			for (idx, core) in cores.into_iter().enumerate() {
+				// Ignore prospective assignments on occupied cores for the time being.
+				if let CoreState::Scheduled(scheduled) = core {
+					let core_index = CoreIndex(idx as _);
+					let group_index = group_rotation_info.group_for_core(core_index, n_cores);
+					if let Some(g) = validator_groups.get(group_index.0 as usize) {
+						if g.contains(&validator.index()) {
+							assignment = Some(scheduled.para_id);
```
bkchr, line 207:
```suggestion
							assignment = Some(scheduled.para_id);
							break;
```

?

node/core/candidate-selection/src/lib.rs
```diff
@@ -126,24 +137,86 @@ impl JobTrait for CandidateSelectionJob {
...
+						}
+						let _ = groups.insert(scheduled.para_id, g.clone());
```
bkchr, line 209: What is the purpose of this?

node/core/candidate-selection/src/lib.rs
rphmeier: Yeah. Collators shouldn't be advertising to validators who aren't in their group.

node/core/candidate-selection/src/lib.rs
coriolinus: Looks like it's related to `deny(unused_results)` at the top of the file.

bkchr: That was not my question. `groups` is not used for anything.

And this lint is useless anyway...

node/core/candidate-selection/src/lib.rs
bkchr: Why? This blows up the function and makes it unreadable. Putting it above will not put it into any broader scope, besides the scope of this module.

If we follow this reasoning, every free standing function that is called in here should be declared in the function that uses it.

node/core/candidate-selection/src/lib.rs
montekki: removed

node/core/candidate-selection/src/lib.rs
montekki: moved outside

node/core/candidate-selection/src/lib.rs
```diff
@@ -57,30 +63,92 @@ enum Error {
...
 	type ToJob = CandidateSelectionMessage;
 	type Error = Error;
-	type RunArgs = ();
+	type RunArgs = SyncCryptoStorePtr;
 	type Metrics = Metrics;
 
 	const NAME: &'static str = "CandidateSelectionJob";
 
-	#[tracing::instrument(skip(_relay_parent, _run_args, metrics, receiver, sender), fields(subsystem = LOG_TARGET))]
+	#[tracing::instrument(skip(relay_parent, keystore, metrics, receiver, sender), fields(subsystem = LOG_TARGET))]
```
montekki, line 94:
```suggestion
	#[tracing::instrument(skip(keystore, metrics, receiver, sender), fields(subsystem = LOG_TARGET))]
```

rphmeier (approved):

PR #2043, opened 2020-11-30
parachains runtime: configurable maximum validators per core

rphmeier:
This will make it easier to deploy parachains onto networks that have large numbers of validators. Previously, if we had 1000 validators and 1 parachain, all 1000 validators would be assigned onto that parachain. Now governance can configure a maximum number of validators per core (parachain / parathread)


ordian (approved):
Would be nice to update the guide as well:
https://w3f.github.io/parachain-implementers-guide/types/runtime.html#host-configuration
and
https://w3f.github.io/parachain-implementers-guide/runtime/scheduler.html#session-change

rphmeier:
Thanks for the reminder to update the guide!

rphmeier:
bot merge

burdges:
Yes, we only want multiple backers for redundancy, so we've no reason for the backer group size to grow beyond 3 or 5.

PR #2049, opened 2020-12-01
execute_with_client: add 'static bound for Backend

pepyakin:
This is not a big change since the user definition AbstractClient already is 'static.


bkchr (approved):

bkchr:
bot merge

Issue #1980, opened 2020-11-18
guide: GRANDPA voting rule


rphmeier:
We want a GRANDPA voting rule that:
  * does not vote on blocks which haven't been approved. we can determine this with a call to `ApprovalVoting::ApprovedAncestor`
  * does not vote on blocks where there are any live disputes which would trigger a rollback to before that point. It's likely easiest to just abstract over that in this point and do the implementation of that message in the dispute participation subsystem (#1977)

burdges:
We de facto lock disputed parachains because we lock backed parachain cores until they become available and get included.  We might ensure core rotation leaves the disputed parachain locked though.  We do not afaik actually lock the parachain until its dispute concludes, although mabye we should do so, not sure..

rphmeier:
> We de facto lock disputed parachains because we lock backed parachain cores until they become available and get included. 

I was not aware of this - neither of the guide and code mention this. My feeling was that we'd accept disputes on included blocks only, anyway, as there isn't much use disputing something pre-availability.

burdges:
A priori, we'd accept remote disputes on anything I'd think.  Yes, if it never gets included anyplace then sure maybe it's less risky, but actually understanding remote inclusion kinda sucks.  I'd envisioned accepting local or remote disputes without inclusions evidence, but if the disputed candidates were not yet availabile then we continue its availability run off-chain.  We time out eventually but could be restarted manually I guess.

Imagine we've a parachain P and relay chain forks X and Y with X1 backing U on P and Y1 backing V on P.  If we then dispute V on Y then we'll dispute V on X but should we lock P on X?  I think no..

rphmeier:
Closed by #2112

PR #2053, opened 2020-12-01
add a brief description of goals of the GRANDPA voting rule

rphmeier:
Related to #1980 . I won't close that yet, because we haven't figured out the disputes portion of it yet


burdges:
I forget if it's said explicitly elsewhere, but here it's only implicit that we implement the voting rule off-chain, although we do put approval votes on-chain for payments.  It'll a bit take more work to make those align this way I guess.

PR #2056, opened 2020-12-02
Make it possible for the adder collator to calculate any state

bkchr:
This is very useful for when wanting to have multiple running or when
wanting to restart the collator.


parachain/test-parachains/adder/collator/src/lib.rs
```diff
@@ -33,6 +33,25 @@ use test_parachain_adder::{execute, hash_state, BlockData, HeadData};
 /// This is a constant to make tests easily reproducible.
 const ADD: u64 = 2;
 
+/// Calculates the head and state for the given `number`.
```
pepyakin, line 36:
```suggestion
/// Calculates the head and state for the block with the given `number`.
```

To emphasize that it's the block number, not the number found in the state.

pepyakin (approved):

pepyakin:
bot merge

PR #2060, opened 2020-12-02
Reduce the inclusion inherent's actual weight if the block is already heavy

### LINKED ISSUE #1555: Inclusion and heavy blocks
In Substrate there are some heavy operational extrinsics or other stuff that consumes the whole block weight.

For example, if governance enacts a new runtime upgrade, then it consumes the whole block worth of weight. At the same time, the inclusion inherent currently is required thus all the parachain processing is mandatory.

I wonder if it can lead to problems and if we should allow not skip inclusion if there is a lot of work scheduled in the block already?

coriolinus:
Closes #1555.

This should give us these behaviors:

- the `inclusion` inherent is prioritized over standard transactions
- but if it's too heavy, i.e. in case of runtime upgrade, it'll be dropped.

It is my belief that allowing the proposer to just not include
this data won't have any adverse effects: it's equivalent to replacing
them with empty versions of themselves, which the `ProvideInherent`
impl already does.


node/core/proposer/src/lib.rs
```diff
@@ -193,12 +193,56 @@ where
...
+			//
+			// ```rust
+			// let expect_heavy_block = sum_of_tx_weights > maximum_block_weight - MARGIN;
+			// ```
+			//
+			// This is complicated by the fact that transactions are code, not data: in principle,
+			// it would be possible for an attacker to craft a transaction which is heavy and looks
+			// valid to the transaction pool, but which aborts cheaply when it is executed,
+			// preventing its costs from being deducted from the attacker. Spamming the relay chain
+			// with sufficient of these transactions would prevent all parachain progress.
```
coriolinus, line 233: @pepyakin I'd like your opinion: is this attack something we should actually worry about / mitigate, or can we somehow ensure that there are no relay chain transactions which would be suitable for this attack?

pepyakin: yeah, it's theoretically possible but that's already accounted as a DoS vector. That's one of the reasons why we are so careful with panics within the Runtime.

pepyakin:
So the problem I find in the outlined description is that you still don't know what extrinsics will be included. The proposer doesn't take all extrinsics from the pool. Also there is no way to tell what extrinsics will be proposed, at least not until the proposer does it job. 

Instead, there is a long line of extrinsics waiting to be included in the block and then the proposer adds one by one keeping an eye on the cumulative consumed weight. After it reaches a certain threshold it tries to shove some more extrinsics. So all in all this boils down to the first approach.

There are two observations that may expand possibly ways to explore on how to tackle this:

1. After the substrate proposer initializes the block, it should already know the amount of weight consumed by the `on_initialize` and `on_finalize`. Assuming that this is the source of most heavy (i.e. most weight) work - we could already know the amount of weight the on_initialize+on_finalize bracket will perform.
1. In case, we need to handle some extrinsics (I guess some heavy operational extrinsics) then another observation is that inherents could be moved in the extrinsics list. If so, we could have a custom logic that moves the inclusion inherent after on_initialize and heavy operational/mandatory extrinsics. At this point there is enough information to decide whether we want to skip inclusion or not.

As a side note, this can be also performed on the runtime side, btw. In fact, there is already [this](https://github.com/paritytech/polkadot/blob/ab5dd449f854c5dcffea225f3e4a9854e57786a3/runtime/parachains/src/inclusion_inherent.rs#L141) ([a relevant comment](https://github.com/paritytech/polkadot/pull/2006#issuecomment-733044682))

bkchr:
> There are two observations that may expand possibly ways to explore on how to tackle this:
> 
>     1. After the substrate proposer initializes the block, it should already know the amount of weight consumed by the `on_initialize` and `on_finalize`. Assuming that this is the source of most heavy (i.e. most weight) work - we could already know the amount of weight the on_initialize+on_finalize bracket will perform.

We can not know the weight of `on_finalize` before running it and this is executed after all transaction are applied. However, currently most of the important heavy stuff is done in `on_initialize`. This means, if we see that `on_initialize` already eat up most of the block weight, we could just skip the inherent when it is applied. But this should be done from inside the runtime.

> 
>     2. In case, we need to handle some extrinsics (I guess some heavy operational extrinsics) then another observation is that inherents could be moved in the extrinsics list. If so, we could have a custom logic that moves the inclusion inherent after on_initialize and heavy operational/mandatory extrinsics. At this point there is enough information to decide whether we want to skip inclusion or not.

How should the node know what each extrinsic is doing? If you have some important operational extrinsic, you could just schedule it, so that it is executed before the inherents in the next block. However opertional extrinsics also have more weight budget. 

As already said in Riot, I don't see this as anything that should be handled from the runtime. Apply the inherent and the inclusion logic will skip the data in the inherent if we already used too much weight.

rphmeier:
> As already said in Riot, I don't see this as anything that should be handled from the runtime. Apply the inherent and the inclusion logic will skip the data in the inherent if we already used too much weight.

Do you mean that we should alter `inclusion_inherent` in the runtime to perform this purpose? Not a bad idea, although we will waste block space, potentially without necessity.

pepyakin:
> We can not know the weight of on_finalize before running it and this is executed after all transaction are applied. 

Isn't the `on_initialize` returns weight for both `on_initialize` and `on_finalize`? Otherwise, you already get into the problem when a block packed tightly and then there is `on_finalize` on top.

> How should the node know what each extrinsic is doing?

I was thinking about special casing this in the node, if needed.

> As already said in Riot, I don't see this as anything that should be handled from the runtime. Apply the inherent and the inclusion logic will skip the data in the inherent if we already used too much weight.

Could you point to that discussion?

bkchr:
> > As already said in Riot, I don't see this as anything that should be handled from the runtime. Apply the inherent and the inclusion logic will skip the data in the inherent if we already used too much weight.
> 
> Do you mean that we should alter `inclusion_inherent` in the runtime to perform this purpose? 

Exactly.

> Not a bad idea, although we will waste block space, potentially without necessity.

Ahh yeah, good point.

bkchr:
> > We can not know the weight of on_finalize before running it and this is executed after all transaction are applied.
> 
> Isn't the `on_initialize` returns weight for both `on_initialize` and `on_finalize`? Otherwise, you already get into the problem when a block packed tightly and then there is `on_finalize` on top.

Ohh yeah you are right, somehow I thought this wasn't the case. :facepalm: 

> 
> > How should the node know what each extrinsic is doing?
> 
> I was thinking about special casing this in the node, if needed.

Special casing it in which way? Call into the runtime and get back the weights? Extrinsics are anyway negligible, because the inclusion inherent is in 99% of the cases more important.

bkchr:
> > Not a bad idea, although we will waste block space, potentially without necessity.
> 
> Ahh yeah, good point.

If we want to prevent this, we could just simply add a runtime api that returns the current and max weight and let you decide on this. While this still doesn't give you the weight that the inherent will consume.

pepyakin:
> Special casing it in which way? Call into the runtime and get back the weights? Extrinsics are anyway negligible, because the inclusion inherent is in 99% of the cases more important.

Perhaps, or just straight hardcoding in the node. But yeah, I agree that this perhaps is not required.

bkchr:
Hardcode what? Sorry, I can not follow ^^

runtime/parachains/src/inclusion_inherent.rs
```diff
@@ -127,6 +129,30 @@ decl_module! {
 	}
 }
 
+// At the point that this is run, the provisioner has already chosen a set of extrinsics,
+// so we can rely on calculations like block weight.
+fn inclusion_inherents_would_overload_block<T: 'static + frame_system::Config>() -> bool {
+	frame_system::Module::<T>::block_weight().total() > <T as frame_system::Config>::MaximumBlockWeight::get() - BLOCK_WEIGHT_MARGIN
```
coriolinus, line 135: Instead of calculating a fixed margin, we could pretty easily redo this calculation as a fraction of total block weight, if desired.

runtime/parachains/src/inclusion_inherent.rs
```diff
@@ -127,6 +129,30 @@ decl_module! {
 	}
 }
 
+// At the point that this is run, the provisioner has already chosen a set of extrinsics,
+// so we can rely on calculations like block weight.
```
coriolinus, line 133: I _think_ this assertion is true, but would appreciate someone more knowledgable confirming or denying it.

runtime/parachains/src/inclusion_inherent.rs
```diff
@@ -39,6 +39,8 @@ use crate::{
 };
 use inherents::{InherentIdentifier, InherentData, MakeFatalError, ProvideInherent};
 
+const BLOCK_WEIGHT_MARGIN: u64 = 1000;
```
coriolinus, line 42: This value is entirely arbitrary; suggestions for better values would be welcomed.

rphmeier: @coriolinus can you work with some of the folks more familiar with the benchmarking / weights in practice on this?

runtime/parachains/src/inclusion_inherent.rs
rphmeier: I don't think this is necessarily true.

node/core/proposer/src/lib.rs
thiolliere: I think such an extrinsic can happen due to weight refund.
Basically extrinsic can specify what was the actual weight of the extrinsic after execution.

An extrinsic can have a heavy weight due to compute-heavy worst-case while finally being very cheap. And this is not just about invalid extrinsic. I think some valid extrinsic can expect big weight refund in principle

runtime/parachains/src/inclusion_inherent.rs
thiolliere: Polkadot adjust fee with `SlowAdjustingFeeUpdate`, thus fees slowly decrease or increase depending on if the weight of the normal-class extrinsics of the block fill more or less than 25% of what is available to them (See `TargetBlockFullness`. So this can be a first indicator.

Then this value is configured when defining the runtime. inside pallets I guess best is to make it configurable in `Config` or to `system::Config::BlockWeights::get().max_block` maybe.

thiolliere:
> Do you mean that we should alter inclusion_inherent in the runtime to perform this purpose? Not a bad idea, although we will waste block space, potentially without necessity.

The waste is if we exclude the InclusionInherent due to a heuristic, when it actually would have fit, isn't it ?

As far as I can see what we want ultimately is to have this inherent prioritised with other extrinsics in the transaction pool. This inherent would be of class normal or operational and priority high, thus in case of runtime upgrade and when block is full then it won't be included. But I don't know how difficult it is.

So yes if we execute this inherent before all other non-inherent extrinsics then we can update check if there is enough weight available, and maybe update the used weight of the block. But then this extrinsic has more priority than operational signed/unsigned extrinsics. But considering this is probably what we want most of the time and that Operationals have some reserved weight amount (weight amount available only for operationals extrinsics), then I think it is fine.

runtime/parachains/src/inclusion_inherent.rs
thiolliere: maybe this inherent could be simply benchmarked, then we would just check that block weight can handle this inherent weight.

runtime/parachains/src/inclusion_inherent.rs
thiolliere: as far as I know, for now all inherents are run before non-inherent extrinsics. thus the block weight here only includes on_runtime_upgrade, on_initialize weights and previous inherent weights.

node/core/proposer/src/lib.rs
pepyakin: It seems that I misread the text, and agree with @thiolliere on this one. I also now have realized why I this question was addressed to me.

bkchr:
> The waste is if we exclude the InclusionInherent due to a heuristic, when it actually would have fit, isn't it ?

He means waste of block space. Because the inherent would still be part of the block extrinsics. However, the size of this probably not that much.

thiolliere:
I'm confused now, what if we just make the call `inclusion` operational or normal instead of mandatory.

If we do `inclusion` as operational, then the weight of the extrinsic will be checked, and the transaction validity will be `Err(ExhaustRessource)`. Such an extrinsic should not be included in the block. and this is what we want.

I'm very not familiar about how the client organize the inherent extrinsics. It doesn't use the `priority` information but it does check the validity no ?

coriolinus:
So, if it's operational, then the weight stuff gets much easier, because the normal provisioner takes care of including it or not as space is available; we don't need to mess with it here. However, if it's not included, then we need to get appropriate defaults when we read it.

... I can't think of a reason that isn't a good idea. I'm going to try implementing it.

runtime/parachains/src/inclusion_inherent.rs
```diff
@@ -70,13 +68,11 @@ decl_module! {
 		}
 
 		fn on_finalize() {
-			if Included::take().is_none() {
-				panic!("Bitfields and heads must be included every block");
-			}
+			Included::take();
```
coriolinus, line 71: Is there a useful way to emit a log line here? Even if we don't want to panic anymore, it might be worth mentioning to the node operator.

rphmeier:
I still feel like the inclusion inherent should be mandatory and we should simply scale down the amount of parachains represented in it based on weight.

I don't like the idea of having to trust the transaction queue to include parachain blocks at all.

bkchr:
> I don't like the idea of having to trust the transaction queue to include parachain blocks at all.

The inherent would still be pushed directly to the block, it does not go through the transaction queue at all.

runtime/parachains/src/inclusion_inherent.rs
```diff
@@ -47,8 +47,6 @@ decl_storage! {
 		///
 		/// The `Option<()>` is effectively a bool, but it never hits storage in the `None` variant
 		/// due to the guarantees of FRAME's storage APIs.
-		///
-		/// If this is `None` at the end of the block, we panic and render the block invalid.
 		Included: Option<()>;
```
bkchr, line 50: This makes no sense anymore.

coriolinus: We still use it to ensure it's not included twice.

bkchr:
FWIW: I'm still in favor of letting the parachain inclusion logic decide on what it should do. 

I don't expect that the waste of block resources is that *much*. Has someone checked the size of the inherent at all?

The inclusion could for example just process one candidate or do some other processing that isn't that heavy.. There are a lot of options on what could still be done.

thiolliere:
I agree with you both.

Last thought: considering the position of inherents is not enforced, we have to keep in mind that a block producer could put this inherent as last extrinsic of the block.
This is probably not a problem if a the economic incentive of inclusion is high.

Otherwise maybe we could enforce some inherent to be the firsts extrinsics of the block. The runtime could check for it. (this would require of frame-executive, `execute_block` would need to retrieve the inherents from `InherentData` and check for it. Or other implementation).

bkchr:
> Otherwise maybe we could enforce some inherent to be the firsts extrinsics of the block. The runtime could check for it. (this would require of frame-executive, `execute_block` would need to retrieve the inherents from `InherentData` and check for it. Or other implementation).

This task is somewhere on my never ending task list :P And we should not assume anything else here.

coriolinus:
Promoting some comments out of the commit message (66834f5):

> This approach replaces making the inclusion inherent non-mandatory. It's still not ideal in that we have to configure a heuristic for how much each backed candidate 'weighs', instead of directly measuring it somehow.
> 
> This approach also never truncates the signed bitfields. The rationale for that depends on some assumptions:
> 
> - processing the signed bitfields is cheap compared to the   backed candidates
> - it is beneficial to the progress of the relay chain to update the signed bitfields even if not all backed candidates are updated

runtime/parachains/src/inclusion_inherent.rs
```diff
@@ -106,6 +107,14 @@ decl_module! {
 
 			<scheduler::Module<T>>::schedule(freed);
 
+			// Temporarily mutably borrow this list so it can be truncated, then restore immutability.
+			// This only works because we own the vector anyway.
+			let backed_candidates = {
+				let mut backed_candidates = backed_candidates;
+				limit_backed_candidates::<T>(&mut backed_candidates);
+				backed_candidates
+			};
```
bkchr, line 116:
```suggestion
			let backed_candidates = limit_backed_candidates::<T>(backed_candidates);
```

The comment doesn't bring any value. If you make the function take the value by value, it will be much cleaner code.

runtime/parachains/src/configuration.rs
```diff
@@ -130,6 +130,8 @@ pub struct HostConfiguration<BlockNumber> {
 	///
 	/// This parameter affects the upper bound of size of `CandidateCommitments`.
 	pub hrmp_max_message_num_per_candidate: u32,
+	/// The amount of block weight which we assume is consumed to process a single backed candidate.
+	pub backed_candidate_block_weight: u32,
```
bkchr, line 134: Why is that a configuration value?

This value is calculated by benchmarking and not by the council or whatever.

runtime/parachains/src/inclusion_inherent.rs
```diff
@@ -127,6 +136,43 @@ decl_module! {
...
 }
 
+/// Limit the number of backed candidates processed in order to stay within block weight limits.
+///
+/// Use a configured assumption about the weight required to process a backed candidate and the
+/// current block weight as of the execution of this function to ensure that we don't overload
+/// the block with candidate processing.
+///
+/// Retains the first N candidates which fit, with no attempt at sorting by priority.
+fn limit_backed_candidates<T: Config>(backed_candidates: &mut Vec<BackedCandidate<T::Hash>>) {
```
bkchr, line 146: Why don't we drop all or nothing?

If the backed candidates are sorted, this could lead to unfair situations.

coriolinus: I assumed it was desirable to process some backed candidates, not all or nothing. Agree that if we don't trust the provisioner, then all or nothing is more fair. Can we assume we trust the provisioner?

runtime/parachains/src/configuration.rs
coriolinus: Will change it.

runtime/parachains/src/inclusion_inherent.rs
bkchr: I would say no.

runtime/parachains/src/configuration.rs
coriolinus: 174f9064

runtime/parachains/src/inclusion_inherent.rs
coriolinus: 174f9064

runtime/parachains/src/inclusion_inherent.rs
coriolinus: 63b4869a

runtime/parachains/src/inclusion_inherent.rs
```diff
@@ -127,6 +129,47 @@ decl_module! {
...
 
+/// Limit the number of backed candidates processed in order to stay within block weight limits.
+///
+/// Use a configured assumption about the weight required to process a backed candidate and the
+/// current block weight as of the execution of this function to ensure that we don't overload
+/// the block with candidate processing.
+///
+/// If the backed candidates exceed the available block weight remaining, then skips all of them.
+/// This is somewhat less desirable than attempting to fit some of them, but is more fair in the
+/// even that we can't trust the provisioner to provide a fair / random ordering of candidates.
```
drahnr, line 140: I guess we don't want any logic, i.e. deadline + binpacking combo algo come to mind.

runtime/parachains/src/inclusion_inherent.rs
```diff
@@ -127,6 +129,47 @@ decl_module! {
...
+fn limit_backed_candidates<T: Config>(
+	backed_candidates: Vec<BackedCandidate<T::Hash>>,
+) -> Vec<BackedCandidate<T::Hash>> {
+	const BACKED_CANDIDATE_WEIGHT_ASSUMPTION: usize = 10_000;
+
+	let block_weight_remaining = <T as frame_system::Config>::MaximumBlockWeight::get()
+		.saturating_sub(frame_system::Module::<T>::block_weight().total());
+
+	if backed_candidates.len() * BACKED_CANDIDATE_WEIGHT_ASSUMPTION > block_weight_remaining as usize {
```
drahnr, line 149: With my limited knowledge, what happens to the inherents if they don't make it here? That would imply the candidates never make?
The threshold also seems to be very coarse. I think most of these points were already discussed in the comments.

runtime/parachains/src/inclusion_inherent.rs
```diff
@@ -127,6 +129,47 @@ decl_module! {
...
+/// If the backed candidates exceed the available block weight remaining, then skips all of them.
+/// This is somewhat less desirable than attempting to fit some of them, but is more fair in the
+/// even that we can't trust the provisioner to provide a fair / random ordering of candidates.
+fn limit_backed_candidates<T: Config>(
+	backed_candidates: Vec<BackedCandidate<T::Hash>>,
+) -> Vec<BackedCandidate<T::Hash>> {
+	const BACKED_CANDIDATE_WEIGHT_ASSUMPTION: usize = 10_000;
+
+	let block_weight_remaining = <T as frame_system::Config>::MaximumBlockWeight::get()
+		.saturating_sub(frame_system::Module::<T>::block_weight().total());
```
thiolliere, line 147: the definition of weight has slightly changed since https://github.com/paritytech/polkadot/pull/1420 can you merge polkadot master ?

runtime/parachains/src/inclusion_inherent.rs
thiolliere, line 149: It seems weight is added 2 times here. I mean at the moment this is executed (i.e. the extrinsic is applied) frame_system block weight already include the extrinsic weight (it is added by check_weight extension). So actually if you want to check if `inclusion` has reached block weight limit you just need to check if here block weight is more than block weight limit.

But I see that `inclusion` weight is `1_000_000_000` and so doesn't include the weight part factor of `backed_candidates.len()`. But then you never write this amount into the used weight amount.

So either it is documented that `inclusion` weight doesn't include the full weight as it doesn't include the variable part factor of `backed_candidates.len()`. And we write this weight somewhere.

Or (and I think it is the best) we use refund here:
Basically we can have precise and full weight info for `inclusion` (so either benchmark or manual estimation but this will end up being a formula `K1 + backed_candidates.len() * K2`). and here we just check if we went beyond the full block weight.
In case we reached the block weight limit then we don't include any candidates.
At the end of inclusion we can still do a weight refund if no candidate were included (i.e. `K1`).

runtime/parachains/src/inclusion_inherent.rs
coriolinus: I don't know what a fair algorithm would be for the candidates, given the information we have available. Adding one would be a reasonable new task, but I don't think it fits within the scope of this particular issue.

runtime/parachains/src/inclusion_inherent.rs
coriolinus: If they don't get included, then they'll be retried later. This is analogous to discarding the candidate as we do [here](https://github.com/paritytech/polkadot/blob/7fb1e57d7a90badbbe0b8098443b9da85d3f4da0/runtime/parachains/src/inclusion_inherent.rs#L141-L145) in `master` already.

runtime/parachains/src/inclusion_inherent.rs
coriolinus: Good catch, thanks! 5bc1f4f5.

runtime/parachains/src/inclusion_inherent.rs
coriolinus: 3dbed6d9

runtime/parachains/src/inclusion_inherent.rs
```diff
@@ -39,6 +39,12 @@ use crate::{
 };
 use inherents::{InherentIdentifier, InherentData, MakeFatalError, ProvideInherent};
 
+// In the future, we should benchmark these consts; these are all untested assumptions for now.
```
thiolliere, line 42: why not adding benchmarks now ? it shouldn't be too hard with `benchmark!` macro, you can look at other pallet how they write benchmark.

But I'm not familiar with the inclusion code so I might underestimate the task.

runtime/parachains/src/inclusion_inherent.rs
```diff
@@ -122,11 +131,53 @@ decl_module! {
...
+/// If the backed candidates exceed the available block weight remaining, then skips all of them.
+/// This is somewhat less desirable than attempting to fit some of them, but is more fair in the
+/// even that we can't trust the provisioner to provide a fair / random ordering of candidates.
+fn limit_backed_candidates<T: Config>(
+	backed_candidates: Vec<BackedCandidate<T::Hash>>,
+) -> Vec<BackedCandidate<T::Hash>> {
+	let block_weight_remaining = <T as frame_system::Config>::BlockWeights::get().max_block
+		.saturating_sub(frame_system::Module::<T>::block_weight().total());
+
+	if backed_candidates.len() as Weight * BACKED_CANDIDATE_WEIGHT > block_weight_remaining  {
```
thiolliere, line 157: block_weight_remaining already include `INCLUSION_INHERENT_CLAIMED_WEIGHT` here. it has been added to block weight with the extension check_weight in frame_system.

So you just need to check if `<T as frame_system::Config>::BlockWeights::get().max_block < frame_system::Module::<T>::block_weight().total()`

runtime/parachains/src/inclusion_inherent.rs
```diff
@@ -76,12 +82,12 @@ decl_module! {
 		}
 
 		/// Include backed candidates and bitfields.
-		#[weight = (1_000_000_000, DispatchClass::Mandatory)]
+		#[weight = (INCLUSION_INHERENT_CLAIMED_WEIGHT, DispatchClass::Mandatory)]
```
thiolliere, line 85: At some point why not just this ? you have access to calls argument in this expression.
```suggestion
		#[weight = (MINIMAL_INCLUSION_INHERENT_WEIGHT + backed_candidates.len() * BACKED_CANDIDATE_WEIGHT, DispatchClass::Mandatory)]
```

coriolinus: That doesn't appear to work, unfortunately. It would be a nice solution, but doesn't compile locally.

coriolinus: No, I'm wrong: I mis-pasted, which caused the compilation failures. This is a good idea!

coriolinus: 622f267

runtime/parachains/src/inclusion_inherent.rs
coriolinus: Yup: 84afca72.

runtime/parachains/src/inclusion_inherent.rs
coriolinus: Promoted this to a new issue: https://github.com/paritytech/polkadot/issues/2124. It's kind of out of scope of this PR because the old numbers were essentially guesses as well.

bkchr (commented):
Please add tests.

rphmeier:
Processing availability bitfields is probably heavier than you give it credit for. It's possibly even heavier than processing backing. But we can tweak ratios later on.

rphmeier:
Needs unit tests and to be tested in practice as well.

coriolinus:
More tests on the way; I just haven't yet figured out all the details of how to implement tests that the inherent weight changes, or not, according to the block weight before execution.

coriolinus:
@bkchr Have added tests; please let me know if you would like more.

runtime/parachains/src/inclusion_inherent.rs
```diff
@@ -80,19 +85,21 @@ decl_module! {
...
 			parent_header: Header,
-		) -> DispatchResult {
+		) -> DispatchResultWithPostInfo {
 			ensure_none(origin)?;
 			ensure!(!<Included>::exists(), Error::<T>::TooManyInclusionInherents);
 
 			// Check that the submitted parent header indeed corresponds to the previous block hash.
-			let now = <frame_system::Module<T>>::block_number();
-			let parent_hash = <frame_system::Module<T>>::block_hash(now - One::one());
+			let parent_hash = <frame_system::Module<T>>::parent_hash();
```
coriolinus, line 102: @pepyakin I'm pretty sure this change is functionally equivalent, but it makes testing possible, which it isn't when looking up the parent hash by block number. Do you agree?

pepyakin: ah, yeah, for sure

drahnr (approved):
The changes are in line with the previous comments :+1:

rphmeier (approved):

PR #2069, opened 2020-12-04
Add an upper number of maximum parallel runtime api requests

bkchr:
Instead of spawning all runtime api requests in the background and using
all wasm instances. This pr adds a maximum number of parallel requests.

### DIFF
--- a/node/core/runtime-api/src/lib.rs
+++ b/node/core/runtime-api/src/lib.rs
@@ -42,7 +42,7 @@ use std::{sync::Arc, collections::VecDeque, pin::Pin};
 const LOG_TARGET: &str = "runtime_api";
 
 /// The number of maximum runtime api requests can be executed in parallel. Further requests will be buffered.
-const MAX_PARALLEL_REQUESTS: usize = 2;
+const MAX_PARALLEL_REQUESTS: usize = 4;
 
 /// The name of the blocking task that executes a runtime api request.
 const API_REQUEST_TASK_NAME: &str = "polkadot-runtime-api-request";
--- a/node/core/runtime-api/src/lib.rs
+++ b/node/core/runtime-api/src/lib.rs
@@ -126,7 +126,7 @@ impl<Client> RuntimeApiSubsystem<Client> where
 	async fn poll_requests(&mut self) {
 		// If there are no active requests, this future should be pending forever.
 		if self.active_requests.len() == 0 {
-			futures::pending!()
+			return futures::pending!()
 		}
 
 		// If there are active requests, this will always resolve to `Some(_)` when a request is finished.
--- a/node/core/runtime-api/src/lib.rs
+++ b/node/core/runtime-api/src/lib.rs
@@ -36,22 +36,38 @@ use polkadot_primitives::v1::{Block, BlockId, Hash, ParachainHost};
 use sp_api::ProvideRuntimeApi;
 use sp_core::traits::SpawnNamed;
 
-use futures::prelude::*;
-use std::sync::Arc;
+use futures::{prelude::*, stream::FuturesUnordered, channel::oneshot, select};
+use std::{sync::Arc, collections::VecDeque, pin::Pin};
 
 const LOG_TARGET: &str = "runtime_api";
+
+/// The number of maximum parallel runtime api requests.
+const MAX_PARALLEL_REQUESTS: usize = 2;
+
+/// The name of the blocking task that executes a runtime api request.
+const API_REQUEST_TASK_NAME: &str = "polkadot-runtime-api-request";
 
 /// The `RuntimeApiSubsystem`. See module docs for more details.
 pub struct RuntimeApiSubsystem<Client> {
 	client: Arc<Client>,
 	metrics: Metrics,
 	spawn_handle: Box<dyn SpawnNamed>,
+	/// If there are [`MAX_PARALLEL_REQUESTS`] requests being executed, we buffer them in here until they can be executed.
+	waiting_requests: VecDeque<(Pin<Box<dyn Future<Output = ()> + Send>>, oneshot::Receiver<()>)>,
+	/// All the active runtime api requests that are currently being executed.
+	active_requests: FuturesUnordered<oneshot::Receiver<()>>,
 }
 
 impl<Client> RuntimeApiSubsystem<Client> {
 	/// Create a new Runtime API subsystem wrapping the given client and metrics.
 	pub fn new(client: Arc<Client>, metrics: Metrics, spawn_handle: impl SpawnNamed + 'static) -> Self {
-		RuntimeApiSubsystem { client, metrics, spawn_handle: Box::new(spawn_handle) }
+		RuntimeApiSubsystem {
+			client,
+			metrics,
+			spawn_handle: Box::new(spawn_handle),
+			waiting_requests: Default::default(),
+			active_requests: Default::default(),
+		}
 	}
 }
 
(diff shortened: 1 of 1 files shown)


node/core/runtime-api/src/lib.rs
```diff
@@ -36,22 +36,38 @@ use polkadot_primitives::v1::{Block, BlockId, Hash, ParachainHost};
...
 
-use futures::prelude::*;
-use std::sync::Arc;
+use futures::{prelude::*, stream::FuturesUnordered, channel::oneshot, select};
+use std::{sync::Arc, collections::VecDeque, pin::Pin};
 
 const LOG_TARGET: &str = "runtime_api";
 
+/// The number of maximum parallel runtime api requests.
+const MAX_PARALLEL_REQUESTS: usize = 2;
```
drahnr, line 45: Is this intentionally very low? There are at least two locations I know from the top of my head that require to distinct chunks of information which could be perallalized (or already are) and would deplete this right away.

node/core/runtime-api/src/lib.rs
```diff
@@ -36,22 +36,38 @@ use polkadot_primitives::v1::{Block, BlockId, Hash, ParachainHost};
...
 use sp_core::traits::SpawnNamed;
 
-use futures::prelude::*;
-use std::sync::Arc;
+use futures::{prelude::*, stream::FuturesUnordered, channel::oneshot, select};
+use std::{sync::Arc, collections::VecDeque, pin::Pin};
 
 const LOG_TARGET: &str = "runtime_api";
 
+/// The number of maximum parallel runtime api requests.
```
pepyakin, line 44:
```suggestion
/// The number of maximum runtime api requests can be executed in parallel. Further requests will be buffered.
```

node/core/runtime-api/src/lib.rs
```diff
@@ -68,34 +84,73 @@ impl<Client, Context> Subsystem<Context> for RuntimeApiSubsystem<Client> where
...
+				client,
+				metrics,
+				relay_parent,
+				request,
+			);
+			let _ = sender.send(());
+		}.boxed();
+
+		if self.active_requests.len() >= MAX_PARALLEL_REQUESTS {
+			self.waiting_requests.push_back((request, receiver));
```
drahnr, line 110: Should we warn or drop requests if we reach a defined upper bound of `N` waiting requests?

node/core/runtime-api/src/lib.rs
```diff
@@ -68,34 +84,73 @@ impl<Client, Context> Subsystem<Context> for RuntimeApiSubsystem<Client> where
...
+			self.active_requests.push(receiver);
+		}
+	}
+
+	/// Poll the active runtime api requests.
+	async fn poll_requests(&mut self) {
+		// If there are no active requests, this future should be pending forever.
+		if self.active_requests.len() == 0 {
+			futures::pending!()
+		} if self.active_requests.next().await.is_none() {
```
drahnr, line 122: this a bit of an odd read

node/core/runtime-api/src/lib.rs
```diff
@@ -68,34 +84,73 @@ impl<Client, Context> Subsystem<Context> for RuntimeApiSubsystem<Client> where
...
-							client,
-							metrics,
-							relay_parent,
-							request,
-						)
-					}.boxed())
-				},
-			}
+		select! {
+			req = ctx.recv().fuse() => match req? {
```
coriolinus, line 143: There was some work recently taking `.fuse()` out of various `select!` macros; should we do that here also?

node/core/runtime-api/src/lib.rs
```diff
@@ -261,6 +317,7 @@ mod tests {
 			}
 
 			fn availability_cores(&self) -> Vec<CoreState> {
+				let _ = self.availability_cores_wait.lock().unwrap();
```
coriolinus, line 320: What is the point of this mutex lock?

bkchr: Please read the test

node/core/runtime-api/src/lib.rs
bkchr: No. This doesn't change anything.

node/core/runtime-api/src/lib.rs
coriolinus: Ah, misunderstood it, sorry.

node/core/runtime-api/src/lib.rs
```diff
@@ -68,34 +84,73 @@ impl<Client, Context> Subsystem<Context> for RuntimeApiSubsystem<Client> where
...
+				FromOverseer::Signal(OverseerSignal::Conclude) => return Ok(()),
+				FromOverseer::Signal(OverseerSignal::ActiveLeaves(_)) => {},
+				FromOverseer::Signal(OverseerSignal::BlockFinalized(_)) => {},
+				FromOverseer::Communication { msg } => match msg {
+					RuntimeApiMessage::Request(relay_parent, request) => {
+						subsystem.spawn_request(relay_parent, request);
+					},
+				}
+			},
+			_ = subsystem.poll_requests().fuse() => {},
```
pepyakin, line 153: Here it seems that in case there were no message from the overseer we will end up in `poll_requests`. In case there is a next `active_requests` we will await on it and as soon as the request finishes we return control here.

During this time span, we don't serve the `ctx.recv()`? So for example, if a conclude signal comes we will wait until the requests finishes. Is that right? 

Not that bad I guess, just testing my understanding.

node/core/runtime-api/src/lib.rs
```diff
@@ -916,4 +973,41 @@ mod tests {
...
+		let (ctx, mut ctx_handle) = test_helpers::make_subsystem_context(TaskExecutor::new());
+		let runtime_api = Arc::new(MockRuntimeApi::default());
+		let relay_parent = [1; 32].into();
+		let spawner = sp_core::testing::TaskExecutor::new();
+		let mutex = runtime_api.availability_cores_wait.clone();
+
+		let subsystem = RuntimeApiSubsystem::new(runtime_api.clone(), Metrics(None), spawner);
+		let subsystem_task = run(ctx, subsystem).map(|x| x.unwrap());
+		let test_task = async move {
+			let lock = mutex.lock().unwrap();
```
pepyakin, line 988: This whole mutex thing seems a bit unclear IMO. I would appreciate leaving a comment describing why we do this.

node/core/runtime-api/src/lib.rs
```diff
@@ -916,4 +973,41 @@ mod tests {
...
+				}).await;
+
+				receivers.push(rx);
+			}
+
+			let join = future::join_all(receivers);
+
+			drop(lock);
+
+			join.await.into_iter().for_each(|r| assert_eq!(r.unwrap().unwrap(), runtime_api.availability_cores));
```
pepyakin, line 1006: Could we wrap this expression?

node/core/runtime-api/src/lib.rs
pepyakin, line 45: I think we can do more and need more. For example, during backing there is a place where we start 4 runtime API requests at the same time. Maybe we should do 4?

pepyakin (commented):
This looks good and useful!

node/core/runtime-api/src/lib.rs
```diff
@@ -68,34 +84,82 @@ impl<Client, Context> Subsystem<Context> for RuntimeApiSubsystem<Client> where
...
+			self.spawn_handle.spawn_blocking(API_REQUEST_TASK_NAME, request);
+			self.active_requests.push(receiver);
+		}
+	}
+
+	/// Poll the active runtime api requests.
+	async fn poll_requests(&mut self) {
+		// If there are no active requests, this future should be pending forever.
+		if self.active_requests.len() == 0 {
+			futures::pending!()
```
pepyakin, line 129:
```suggestion
			return futures::pending!()
```

🤔 ?

pepyakin: wait, I mean `.await`

bkchr: Should not make a difference, but okay :)

pepyakin (approved):

pepyakin:
bot merge

PR #2081, opened 2020-12-07
Rework `ConnectionsRequests`

bkchr:
Instead of implementing the `Stream` trait, this struct now provides a
function `next()`. This enables us to encode into the type system that
it will always return a value or block indefinitely.


ordian (approved):
Looks sane.

rphmeier (approved):

PR #2082, opened 2020-12-07
Bump substrate, and bump version to v0.8.27

s3krit:


bkchr (approved):

Cargo.lock
```diff
@@ -1462,15 +1462,15 @@ checksum = "3f9eec918d3f24069decb9af1554cad7c880e2da24a9afd88aca000531ab82c1"
 [[package]]
 name = "fork-tree"
 version = "2.0.0"
-source = "git+https://github.com/paritytech/substrate#0dfdc328fa2b9514847b354e6968553a9e29d61e"
+source = "git+https://github.com/paritytech/substrate?branch=master#eaab146094360ca02538f5287abe859b2d815c3c"
```
tomaka, line 1465: I have the impression that this `?branch=master` is constantly being added and removed

tomaka (approved):

Cargo.lock
bkchr: Probably depends on the Cargo version. I'm not sure if the new syntax already hit stable or not. (with master included)

s3krit: Ah yeah - I should have specified +stable when bumping sp-io... Is it a problem? Should I *not* apply the `?branch=master`?

bkchr: That should be no problem. CI is green, so it was happy :)

PR #2089, opened 2020-12-08
Reward validators for participating in parachains

rphmeier:
We want to give reward points according to the schedule here: https://w3f-research.readthedocs.io/en/latest/polkadot/Token%20Economics.html 

That document doesn't yet account for bitfields, but I imagine we want to record every '1' bit set in candidates that actually get included. Likewise, I only rewarded validators backing candidates that became available later on.

TODO:
  - [x] Fix tests
  - [x] Test backing rewards
  - [x] Add bitfield rewards
  - [x] Test bitfield rewards
  - [x] Add an implementation of `RewardValidators` that uses `pallet-staking` and real values under the hood

### DIFF
--- a/roadmap/implementers-guide/src/runtime/inclusion.md
+++ b/roadmap/implementers-guide/src/runtime/inclusion.md
@@ -17,6 +17,7 @@ struct CandidatePendingAvailability {
   descriptor: CandidateDescriptor,
   availability_votes: Bitfield, // one bit per validator.
   relay_parent_number: BlockNumber, // number of the relay-parent.
+  backers: Bitfield, // one bit per validator, set for those who backed the candidate.
   backed_in_number: BlockNumber,
 }
 ```
@@ -77,6 +78,7 @@ All failed checks should lead to an unrecoverable error making the block invalid
 * `enact_candidate(relay_parent_number: BlockNumber, CommittedCandidateReceipt)`:
   1. If the receipt contains a code upgrade, Call `Paras::schedule_code_upgrade(para_id, code, relay_parent_number + config.validationl_upgrade_delay)`.
     > TODO: Note that this is safe as long as we never enact candidates where the relay parent is across a session boundary. In that case, which we should be careful to avoid with contextual execution, the configuration might have changed and the para may de-sync from the host's understanding of it.
+  1. Reward all backing validators of each candidate, contained within the `backers` field.
   1. call `Ump::enact_upward_messages` for each backed candidate, using the [`UpwardMessage`s](../types/messages.md#upward-message) from the [`CandidateCommitments`](../types/candidate.md#candidate-commitments).
   1. call `Dmp::prune_dmq` with the para id of the candidate and the candidate's `processed_downward_messages`.
   1. call `Hrmp::prune_hrmp` with the para id of the candiate and the candidate's `hrmp_watermark`.
--- a/runtime/common/src/paras_registrar.rs
+++ b/runtime/common/src/paras_registrar.rs
@@ -262,7 +262,7 @@ mod tests {
 		}, testing::{UintAuthorityId, TestXt}, Perbill, curve::PiecewiseLinear,
 	};
 	use primitives::v1::{
-		Balance, BlockNumber, Header, Signature, AuthorityDiscoveryId,
+		Balance, BlockNumber, Header, Signature, AuthorityDiscoveryId, ValidatorIndex,
 	};
 	use frame_support::{
 		traits::{Randomness, OnInitialize, OnFinalize},
@@ -473,8 +473,16 @@ mod tests {
 
 	impl configuration::Config for Test { }
 
+	pub struct TestRewardValidators;
+
+	impl inclusion::RewardValidators for TestRewardValidators {
+		fn reward_backing(_: impl IntoIterator<Item = ValidatorIndex>) { }
+		fn reward_bitfields(_: impl IntoIterator<Item = ValidatorIndex>) { }
+	}
+
 	impl inclusion::Config for Test {
 		type Event = ();
+		type RewardValidators = TestRewardValidators;
 	}
 
 	impl session_info::AuthorityDiscoveryConfig for Test {
--- a/runtime/parachains/src/inclusion.rs
+++ b/runtime/parachains/src/inclusion.rs
@@ -1394,7 +1394,7 @@ mod tests {
 				availability_votes: default_availability_votes(),
 				relay_parent_number: 0,
 				backed_in_number: 0,
-				backers: default_backing_bitfield(),
+				backers: backing_bitfield(&[3, 4]),
 			});
 			PendingAvailabilityCommitments::insert(chain_a, candidate_a.commitments);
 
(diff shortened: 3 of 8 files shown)


burdges:
Any reward for availability bitfield votes should stay quite small since it only represent downloading about 3/num_validators of the candidate and storing it for a day.  We'd just pay them for delivering the pieces ideally, which sounds like future work.  

Approval checkers could select from whom they fetch pieces randomly and claim from whom they obtained pieces in their approval vote, called a provider bitfield.  We now encounter a problem that validators could (a) obtain their pieces preferentially from their friends or (b) obtain their pieces from anyone but then claim they obtained their pieces from their friends.

We might address (a) by paying validators more based upon the aggregate distribution of their provider bitfield.  We'd implement and want uniform but honest behavior creates a Gaussian distribution by the central limit theorem.  Is there a nice statistical test for "anywhere between uniform and Gaussian" which which we can reward for the actual behavior without rewarding for deviations from the implementation?

We might address (b) using tit-for-tat:  If you think someone who downloaded your piece stiffed you in its provider bitfield then you demote them somewhat in the otherwise random order in which you fetch pieces.  As a result, you continue behaving honestly yourself, not stiffing anyone, but you do punish the guy who stiffed you.  In fact, you've punished them enough after only one time when you skipped fetching from them and your approval vote got counted.  Isn't this some easy Nash equilibrium result @jonasW3F ?

rphmeier:
Thanks for the write-up Jeff. I'll put in a small reward for each bit of the bitfield. We'll need subsystems for watching the chain and doing that tit-for-tat calculation, but that's future work.

jonasW3F:
> We might address (b) using tit-for-tat: If you think someone who downloaded your piece stiffed you in its provider bitfield then you demote them somewhat in the otherwise random order in which you fetch pieces. As a result, you continue behaving honestly yourself, not stiffing anyone, but you do punish the guy who stiffed you. In fact, you've punished them enough after only one time when you skipped fetching from them and your approval vote got counted, so maybe even punishing twice buys nothing. Also they'll never know if your order uses local randomness, which makes retaliation impossible. Isn't this second argument some easy Nash equilibrium result @jonasW3F or @SamuelHaefner ? Is there anything we should worry about like extortion?

Sorry, I was off yesterday. Yes, this is exactly what I was referring to in our previous discussion (I think it was in elements?, can't find an issue for this). A local reputation, system where you basically be nice to someone providing you the piece in the hope that they'll be nice to you and provide their piece. Otherwise you punish them by avoiding requesting pieces from them in the future (tit-for-tat or some other trigger strategy). I try to understand this issue more and have some questions on that: 

1) In our previous discussion rewarding for providing pieces was excluded due to the load of transactions, has that changed?
2) How sure can you be that somebody downloaded your piece but not did incorporate you in their provider bitfield?
3) Not being sure that you are getting punished might also be a downside, because noisy signals complicate the incentives. It would be fine if you'd knew that you are getting punished, because you would know for sure what happens before you act maliciously and then it might be easier to think about a NE.
4) How available are the pieces in general? How many backers can an approval checker expect to have the piece they need? If the set is not too small, I could also imagine a situation where cartels emerge and only request pieces from each other. This would not be good? There is no speed bonus?
 
I guess more questions than answers...sorry :)

burdges:
> How sure can you be that somebody downloaded your piece but not did incorporate you in their provider bitfield?

The punisher cannot know this but the punisher always behave honestly, never stiffs the punishee, and never does anything detectable, so this should remain a Nash equilibrium I think.

> Not being sure that you are getting punished might also be a downside, because noisy signals complicate the incentives. It would be fine if you'd knew that you are getting punished, because you would know for sure what happens before you act maliciously and then it might be easier to think about a NE.

Anyone stiffing a piece provider ala (a) knows they'll be punished if the piece provider is honest.

> How available are the pieces in general?  How many backers can an approval checker expect to have the piece they need? If the set is not too small, I could also imagine a situation where cartels emerge and only request pieces from each other. This would not be good? There is no speed bonus?

If there are n = 3f+1 validators then there are n unique pieces of each candidate.  We need 2f+1 different validators to each obtain their unique pieces from the 3ish backers before approvals start, or else we trash the candidate.  Approval checkers need any f+1 pieces before they can reconstruct the candidate.  I suppose cartels could make their own nodes faster or save networking costs, but a cartel with f+1 validators breaks polkedot anyways.  

I discussed speed bonuses in https://github.com/w3f/research-security-issues/issues/50 because they avoided O(n^2) per piece payments, making it an optimization over this scheme.  If we're doing bitfields based payments then provider bitfields remain O(n^2) but have 1/2 to 1/3 the fullness of availability bitfields. 

In between speed bonuses and the naive accounding, one could do logarithmic bitfield accumulation, meaning we "add" provider bitfields by producing a "carry" bitfield.  I implemented this in-gossip aggregation of BLS signatures in https://github.com/w3f/bls/blob/master/src/bit.rs but it's maybe even more useful here.  I think someone like @drskalman could extract out the logarithmic bitfield accumulation code into some generally usable utility crate for runtime code that handles bitfields on chain.

rphmeier:
@burdges By provider bitfields, do you mean the bitfields that the validators sent to the chain, where availability bitfields are the 'aggregated' versions of specific bits in those pertaining to a particular core?

burdges:
No.  An approval checker would compose, and attach to their approval vote, a "provider bitfield" that identifies whose pieces they used in reconstruction, so a provider bitfield must have exactly f+1 bits set since reconstruction works with exactly f+1 pieces.

If you "sum" the provider bitfields from all approval checkers across an entire epoch then you obtain an aggregate votes saying which validators more readily provided their pieces for approval checkers.  This is a nicer reward metric for availability providers than availability bitfields.

We've zero validation of provider bitfields however, so validators could bias the bitfield towards their friends either by semi-honestly biasing our reconstruction (a) or dishonestly by stiffing (b).  We can easily punish stiffing with tit-for-tat, without even being dishonest and stiffing back, but semi-honest bias requires a more subtle tit-for-tat.

rphmeier:
Thanks for elaborating @burdges . I don't think we'll do that in this PR, but it's something to have an eye on for Polkadot economics

burdges:
Yes, there is some work involved in adding the provider bitfields, especially with an aggregation optimization, and the tit-for-tat protocols, so not really MVP.  They look quite efficient and avoid the tragedy of the commons that comes from either not paying availability voters or else paying them only for voting yes (as proposed here).

burdges:
We could deal with (a) using the assignment VRF instead of looking at distributions or some tit-for-tat. 

At the beginning of the epoch, you seed ChaChaRng with a VRF output based on the epoch randomness and then use the block number and core for the chacha nonce.  This gives you an Rng that tells you the order in which you fetch pieces from different validators, assuming they vote that they have the piece.

We then do the (b) first tit-for-tat with three "count fields":  You also have a black mark "count field" from the previous epoch based upon which you occasionally skip nodes that pop up in your fetch priority.  If you skip one then you remove one black mark against him.   You track you own cumulative provider count field, as well as a providee count field of what contributions you expect others to credit you for.  

After the the epoch, you all post your provider count field accumulated over the epoch, and then you figure out  who credited you for less than you think you deserved.  You add a black mark count fields for them.  You also attach your VRF to your provider count field.  

You are paid according to how many times nodes named you a provider and how well your naming of people as providers stuck to your VRF's predictions.  

In this, you give up some earnings to punish people, but that's probably okay.

montekki (approved):

montekki:
bot merge

PR #2104, opened 2020-12-09
Fix bug and further optimizations in availability distribution

bkchr:
- There was a bug that resulted in only getting one candidate per block
as the candidates were put into the hashmap with the relay block hash as
key. The solution for this is to use the candidate hash and the relay
block hash as key.
- We stored received/sent messages with the candidate hash and chunk
index as key. The candidate hash wasn't required in this case, as the
messages are already stored per candidate.

Further refactoring by @rphmeier (who edited this post to write this :) ):
- Add a `HashSet<Hash>` to `PerCandidate` which is used to ref-count the `PerCandidate` structs against the relay-blocks which consider them live.
- Change interface and API of `receipts` and `query_live_candidates` to be easier to follow and avoid unnecessary data
- Storing `live_candidates` in `PerRelayParent`. This includes the candidates needed by the ancestors and is the dual of the `PerCandidate` map of relay-parents.
- Remove `ancestry` mapping, as its functionality was made obsolete by `live_candidates`.


node/network/availability-distribution/src/lib.rs
```diff
@@ -384,27 +378,16 @@ where
...
-			.await?;
+		state.add_relay_parent(ctx, *added, validators, validator_index).await?;
 	}
 
 	// handle all candidates
 	for (candidate_hash, _receipt) in state.cached_live_candidates_unioned(view.difference(&old_view)) {
-		let per_candidate = state.per_candidate.entry(candidate_hash).or_default();
-
-		// assure the node has the validator role
-		if per_candidate.validator_index.is_none() {
```
rphmeier, line 397: This check is no longer done, that I can see. Rationale for removal?

node/core/bitfield-signing/src/lib.rs
```diff
@@ -265,7 +284,7 @@ impl JobTrait for BitfieldSigningJob {
 			sender
 				.send(
 					AllMessages::from(
-						BitfieldDistributionMessage::DistributeBitfield(relay_parent, signed_bitfield),
+						BitfieldDistributionMessage::DistributeBitfield(relay_parent, dbg!(signed_bitfield)),
```
rphmeier, line 287:
```suggestion
						BitfieldDistributionMessage::DistributeBitfield(relay_parent, signed_bitfield),
```

node/network/availability-distribution/src/lib.rs
```diff
@@ -836,7 +817,7 @@ async fn query_live_candidates<Context>(
 	ctx: &mut Context,
 	state: &mut ProtocolState,
 	relay_parents: impl IntoIterator<Item = Hash>,
-) -> Result<HashMap<Hash, (CandidateHash, CommittedCandidateReceipt)>>
+) -> Result<HashMap<(Hash, CandidateHash), CommittedCandidateReceipt>>
```
rphmeier, line 820: I would expect this to be either `HashMap<Hash, Vec<(CandidateHash, CommittedCandidateReceipt)>>` or `HashMap<CandidateHash, CommittedCandidateRecipt>`. Note that with the first variant, as well as what is currently committed in this PR, the same candidate is likely to appear more than once, if it's been pending availability for more than one block.

I'm not sure where the relay-parent is supposed to enter into this.

To be honest, I don't really understand where the relay-parent needs to fit in here.

node/network/availability-distribution/src/lib.rs
```diff
@@ -156,11 +156,11 @@ struct PerCandidate {
 	/// candidate hash + erasure chunk index -> gossip message
 	message_vault: HashMap<u32, AvailabilityGossipMessage>,
 
-	/// Track received candidate hashes and validator indices from peers.
-	received_messages: HashMap<PeerId, HashSet<(CandidateHash, ValidatorIndex)>>,
+	/// Track received erasure chunk indices per peer.
+	received_messages: HashMap<PeerId, HashSet<ValidatorIndex>>,
```
drahnr, line 160: :+1:

node/network/availability-distribution/src/lib.rs
```diff
@@ -170,10 +170,10 @@ struct PerCandidate {
 }
 
 impl PerCandidate {
-	/// Returns `true` iff the given `message` is required by the given `peer`.
-	fn message_required_by_peer(&self, peer: &PeerId, message: &(CandidateHash, ValidatorIndex)) -> bool {
-		self.received_messages.get(peer).map(|v| !v.contains(message)).unwrap_or(true)
-			&& self.sent_messages.get(peer).map(|v| !v.contains(message)).unwrap_or(true)
+	/// Returns `true` iff the given `validator_index` is required by the given `peer`.
```
drahnr, line 173: We should make this clearer that the peer wants the associated erasure chunk, not the validator index in itself.

node/network/availability-distribution/src/lib.rs
bkchr: I removed this by intention. 

I think that is wrong, however that means I also need to change more in here.

Let me explain what I think:

I removed it because below we are also checking the message vault. Non validator nodes also hold this message vault and will be able to use this to propagate chunks. This was the reason I removed this check.

However I think, I'm wrong here. If we switch to a new view, we can only send new chunks to peers for that we have the availability data in the store and that just got "activated" by this view. Every other candidate that was already known by us, should already be exchanges with the peers as far as we know their view. So, continuing on this, we don't need to check the message vault, as there shouldn't be any messages in it yet, but I will bring this check back.

rphmeier: I don't think we really need this check. Ultimately, we don't distribute pieces unless we have them in our availability store. As a non-validator, or indeed as a non backing validator for this particular candidate, we won't have those pieces.

drahnr (dismissed):
The complexity is way lower than before :+1:

node/network/availability-distribution/src/tests.rs
```diff
@@ -1081,3 +1051,222 @@ fn k_ancestors_in_session() {
...
+
+	let per_relay_parent = state.per_relay_parent.get(&relay_parent).unwrap();
+
+	assert!(per_relay_parent.live_candidates.contains(&candidate_hash_a));
+	assert!(per_relay_parent.live_candidates.contains(&candidate_hash_b));
+}
+
+#[test]
+fn query_pending_availability_at_pulls_from_and_updates_receipts() {
+	let hash_a = [0u8; 32].into();
```
drahnr, line 1160: There is `Hash::repeat_byte` which is used elsewhere.

node/network/availability-distribution/src/lib.rs
bkchr: I have brought it back and added some further optimizations.

node/network/availability-distribution/src/lib.rs
```diff
@@ -360,18 +365,28 @@ where
 			.map(|(peer, _view)| peer.clone())
 			.collect();
 
+		// If we have no peers that are interested, skip the rest.
+		if peers.is_empty() {
```
rphmeier, line 369: I don't think this optimization will work. AFAIK we have no other place where we would load the chunks other than the code below.

`send_tracked_gossip_message_to_peers` should place the messages in the message vault even though `peers.is_empty()`, but now that I look closer, it sadly doesn't.

So this change preserves current behavior, but the current behavior is slightly broken.

node/network/availability-distribution/src/lib.rs
```diff
@@ -360,18 +365,28 @@ where
...
 
 		let validator_count = per_candidate.validators.len();
 
 		// distribute all erasure messages to interested peers
 		for chunk_index in 0u32..(validator_count as u32) {
-			let erasure_chunk = if let Some(message) = per_candidate.message_vault.get(&chunk_index) {
-				message.erasure_chunk.clone()
-			} else if let Some(erasure_chunk) = query_chunk(ctx, candidate_hash, chunk_index as ValidatorIndex).await? {
+			let erasure_chunk = if let Some(erasure_chunk) =
+				query_chunk(ctx, candidate_hash, chunk_index as ValidatorIndex).await? {
```
rphmeier, line 380: Why go to the availability store unconditionally, if the message may exist in the message vault?

drahnr: This change makes the vault obsolete, and these two changes are co-dependant.

bkchr: At this point, the message can not yet exist in the message vault.

If we are switching to a new view, it means we either fetched old candidates which would be handled by `on_peer_view_changed`. Or we fetched new candidates. If we fetched new candidates, there can also be no message in the message vault, because we would have rejected any message up to now for this candidate, because our view wasn't yet aware of the relay block that introduced this candidate.

node/network/availability-distribution/src/lib.rs
bkchr: I see what you mean! Let me write a test and fix this!

node/network/availability-distribution/src/lib.rs
rphmeier: I don't believe this is true.

Note that this is doing `state.cached_live_candidates_unioned`, not `query_pending_availability`. If the parablock was pending availability in a recent ancestor or a sibling block that we've already imported, then we'll have already queried the data from disk.

node/network/availability-distribution/src/lib.rs
```diff
@@ -503,12 +456,11 @@ where
 	}
```
drahnr, line 456: This `peers.is_empty()` has the same issue a the newly introduced one.

The following must be called at the very least for each item provided by `messages`.
```rust
		per_candidate
			.message_vault
			.insert(message.erasure_chunk.index, message.clone());
```

rphmeier: Thanks. This is what I was referencing with my other comment.

node/network/availability-distribution/src/lib.rs
drahnr: Just removing this one is not enough, see line `:454`

node/network/availability-distribution/src/lib.rs
bkchr: Yeah sorry. I see what you mean. When peer updates it view before us, we would not send it the required data.

rphmeier: Actually that is what I meant with the other comment but not this one.

Let's say our initial view is `[A]`. `live_candidates(A) = [X, Y]`. Now we get a new block B and our view is `[A, B]`. `live_candidates(B) = [Y, Z]`. So `cached_live_candidates_unioned = [X, Y, Z]`. We already have a `per_candidate` with filled out message vault for `[X, Y]`, so we don't need to go to the av-store again. We only need to go to disk for the new `Z` candidate.

node/network/availability-distribution/src/tests.rs
rphmeier: `Hash::repeat_byte(0)` is longer than
`[0u8; 32].into()`

FWIW

drahnr: But invariant over the hash length :)

node/network/availability-distribution/src/lib.rs
```diff
@@ -718,10 +718,10 @@ impl AvailabilityDistributionSubsystem {
 	{
 		// work: process incoming messages from the overseer.
 		loop {
-			let message = ctx
+			let message = dbg!(ctx
```
drahnr, line 721: spurious `dbg!`

drahnr:
We should merge https://github.com/paritytech/polkadot/pull/2115 into this one to get a test into this branch that verifies the fixes that make sure any sent message also makes it into the message vault (also in that pr)

drahnr:
Just merged this PR into https://github.com/paritytech/polkadot/pull/2115 so it should merge just fine.

node/network/availability-distribution/src/lib.rs
```diff
@@ -417,77 +369,64 @@ where
...
+				}
 			} else {
-				continue;
+				tracing::error!(
+					target: LOG_TARGET,
+					%chunk_index,
+					?candidate_hash,
+					"Availability store reported that we have the availability data, but we could not retrieve a chunk of it!",
+				);
+				break;
```
drahnr, line 410: This became `break` that was a `continue`. Should we not try to retrieve as many chunks as possible?

node/network/availability-distribution/src/lib.rs
```diff
@@ -501,37 +440,27 @@ async fn send_tracked_gossip_messages_to_peers<Context>(
 where
 	Context: SubsystemContext<Message = AvailabilityDistributionMessage>,
 {
-	if peers.is_empty() {
-		return;
-	}
```
drahnr, line 506: :heavy_check_mark:

drahnr (approved):
LGTM

node/network/availability-distribution/src/lib.rs
bkchr: Reverted.

node/network/availability-distribution/src/lib.rs
```diff
@@ -807,96 +756,97 @@ where
...
 }
 
-/// Obtain all live candidates based on an iterator of relay heads.
-#[tracing::instrument(level = "trace", skip(ctx, relay_parents), fields(subsystem = LOG_TARGET))]
-async fn query_live_candidates_without_ancestors<Context>(
+#[derive(Debug)]
+enum FetchedLiveCandidate {
+	Cached,
+	Fresh(CandidateDescriptor),
+}
```
pepyakin, line 763: A comment would be welcome here

node/network/availability-distribution/src/lib.rs
```diff
@@ -807,96 +756,97 @@ where
...
-) -> Result<HashSet<CommittedCandidateReceipt>>
+	relay_blocks: impl IntoIterator<Item = Hash>,
+	receipts: &mut HashMap<Hash, HashSet<CandidateHash>>,
+) -> Result<HashMap<CandidateHash, FetchedLiveCandidate>>
 where
 	Context: SubsystemContext<Message = AvailabilityDistributionMessage>,
 {
-	let iter = relay_parents.into_iter();
-	let hint = iter.size_hint();
+	let mut live_candidates = HashMap::new();
```
pepyakin, line 781: Changing `live_candidates` requires a guide update. [There](https://w3f.github.io/parachain-implementers-guide/node/availability/availability-distribution.html) it assumes `live_candidates(relay_heads) -> Set<CommittedCandidateReceipt>`

node/network/availability-distribution/src/lib.rs
```diff
@@ -417,77 +369,64 @@ where
...
-			// obtain the chunks from the cache, if not fallback
-			// and query the availability store
-			let message_id = (candidate_hash, chunk_index);
-			let erasure_chunk = if let Some(message) = per_candidate.message_vault.get(&chunk_index) {
-				message.erasure_chunk.clone()
+			let message = if let Some(message) = per_candidate.message_vault.get(&chunk_index) {
+				tracing::trace!(
+					target: LOG_TARGET,
+					%chunk_index,
+					?candidate_hash,
```
pepyakin, line 387: TIL this syntax

node/core/av-store/src/lib.rs
```diff
@@ -748,7 +774,11 @@ where
 			}
 		}
 		StoreAvailableData(hash, id, n_validators, av_data, tx) => {
-			match store_available_data(subsystem, &hash, id, n_validators, av_data) {
+			let res = store_available_data(subsystem, &hash, id, n_validators, av_data);
```
pepyakin, line 777: nit: inconsistency, `let res` vs `let result` above

node/core/av-store/src/lib.rs
```diff
@@ -719,25 +719,51 @@ where
...
+			tracing::trace!(
+				target: LOG_TARGET,
+				%chunk_index,
+				?candidate_hash,
+				%block_number,
+				result = ?res,
+				"Stored chunk",
+			);
+
+			match res {
```
bkchr, line 766:
```suggestion
			let result = store_chunk(subsystem, &candidate_hash, validator_index, chunk, block_number);

			tracing::trace!(
				target: LOG_TARGET,
				%chunk_index,
				?candidate_hash,
				%block_number,
				?result,
				"Stored chunk",
			);

			match result {
```

node/core/av-store/src/lib.rs
```diff
@@ -748,7 +774,11 @@ where
 			}
 		}
 		StoreAvailableData(hash, id, n_validators, av_data, tx) => {
-			match store_available_data(subsystem, &hash, id, n_validators, av_data) {
+			let res = store_available_data(subsystem, &hash, id, n_validators, av_data);
+
+			tracing::trace!(target: LOG_TARGET, candidate_hash = ?hash, result = ?res, "Stored available data");
+
+			match res {
```
bkchr, line 781:
```suggestion
			let result = store_available_data(subsystem, &hash, id, n_validators, av_data);

			tracing::trace!(target: LOG_TARGET, candidate_hash = ?hash, ?result, "Stored available data");

			match result {
```

node/core/bitfield-signing/src/lib.rs
```diff
@@ -103,20 +108,32 @@ async fn get_core_availability(
...
+
+		tracing::trace!(
+			target: LOG_TARGET,
+			para_id = %core.para_id,
+			availability = ?res,
+			?candidate_hash,
+			"Candidate availability",
+		);
+
+		return res
```
montekki, line 136:
```suggestion
		res
```

drahnr (approved):

node/core/bitfield-signing/src/lib.rs
bkchr: That was wrong..

rphmeier:
I will merge this. Further tests can come in a follow-up PR.

rphmeier:
bot merge

node/network/availability-distribution/src/lib.rs
```diff
@@ -955,8 +905,7 @@ where
 		AvailabilityStoreMessage::QueryDataAvailability(candidate_hash, tx),
 	)).await;
 
-	rx.await
-		.map_err(|e| Error::QueryAvailabilityResponseChannel(e))
+	rx.await.map_err(|e| Error::QueryAvailabilityResponseChannel(e))
 }
```
pepyakin, line 915: Can't this be handled by `slice::windows`?

node/network/availability-distribution/src/lib.rs
```diff
@@ -807,96 +756,97 @@ where
...
 {
-	let iter = relay_parents.into_iter();
-	let hint = iter.size_hint();
+	let mut live_candidates = HashMap::new();
+
+	// fetch and fill out cache for each of these
+	for relay_parent in relay_blocks {
+		let receipts_for = match receipts.entry(relay_parent) {
+			Entry::Occupied(e) => {
+				live_candidates.extend(
```
pepyakin, line 792: this clone is actually redundant: we own the `ccr` and we destruct it by taking out `descriptor`.

node/network/availability-distribution/src/lib.rs
```diff
@@ -807,96 +756,97 @@ where
...
-		for para in paras {
+		for para in query_para_ids(ctx, relay_parent).await? {
 			if let Some(ccr) = query_pending_availability(ctx, relay_parent, para).await? {
-				live_candidates.insert(ccr);
+				let receipt_hash = ccr.hash();
+				let descriptor = ccr.descriptor().clone();
+
+				// unfortunately we have no good way of telling the candidate was
+				// cached until now. But we don't clobber a `Cached` entry if there
+				// is one already.
```
pepyakin, line 807: What does it mean to be "paired"?

rphmeier:
bot merge

node/network/availability-distribution/src/lib.rs
```diff
@@ -157,164 +154,116 @@ struct PerCandidate {
...
 
-		self.per_relay_parent
-			.entry(relay_parent)
-			.or_default()
-			.ancestors = ancestors;
-
-		Ok(())
+		let per_relay_parent = self.per_relay_parent.entry(relay_parent).or_default();
+		per_relay_parent.ancestors = ancestors;
+		per_relay_parent.live_candidates.extend(candidate_hashes);
```
pepyakin, line 238: tip: you can move this before the for loop and then do the `extend` without intermediate `collect`ing.

pepyakin:
Oops, apparently I am late to the party : )

node/network/availability-distribution/src/lib.rs
```diff
@@ -129,19 +137,8 @@ struct ProtocolState {
...
-	/// Note that the presence of this is not sufficient to determine if deletion is OK, i.e.
-	/// two histories could cover this.
-	reverse: HashMap<CandidateHash, Hash>,
-
-	/// Keeps track of which candidate receipts are required due to ancestors of which relay parents
-	/// of our view.
-	/// Maps ancestor -> relay parents in view
-	ancestry: HashMap<Hash, HashSet<Hash>>,
+	/// Maps relay parent / ancestor -> candidate receipts.
+	receipts: HashMap<Hash, HashSet<CandidateHash>>,
```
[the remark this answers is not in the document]
drahnr: It used to be the full receipt of a candidate, but it's not anymore, so yes, a rename is in order.

node/network/availability-distribution/src/lib.rs
rphmeier: probably addressed in #2134

node/network/availability-distribution/src/lib.rs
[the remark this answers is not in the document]
rphmeier: > I suppose there are reasons not put it in the per_relay_parent?

Reason was time, mostly. I felt that I'd done enough refactoring of this code and still feel that way :)

PR #2108, opened 2020-12-10
add availability dist test case for duplicate candidate hashes and ancestors

drahnr:
Adds a test case that relies on the interior state of the module.


drahnr:
Blockbox test already impl'd by @rphmeier so this is obsolete.

PR #2110, opened 2020-12-11
Fix an off-by-one error in creation of ValidationData

pepyakin:
In order to create or validate a candidate we read the state from a
relay-parent. That state is obtained after execution of a block at a
certain height. However, for some reason we are substracting one from
that height here.

### DIFF
--- a/runtime/parachains/src/util.rs
+++ b/runtime/parachains/src/util.rs
@@ -47,7 +47,7 @@ pub fn make_transient_validation_data<T: paras::Config + dmp::Config>(
 	para_id: ParaId,
 ) -> Option<TransientValidationData<T::BlockNumber>> {
 	let config = <configuration::Module<T>>::config();
-	let relay_parent_number = <frame_system::Module<T>>::block_number() - One::one();
+	let relay_parent_number = <frame_system::Module<T>>::block_number();
 
 	let freq = config.validation_upgrade_frequency;
 	let delay = config.validation_upgrade_delay;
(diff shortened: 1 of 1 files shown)


coriolinus:
Isn't that to get the parent's number, instead of the current block number?

pepyakin:
Exactly. However, I feel that it is wrong. My logic was something like this:

1. When we collate a block on the collator, we base it off some relay-parent with height `P`.
1. In order to form the inherent we request persistent validation data for `P`
1. The control flow gets to the code in question, and collects all the data at state of `P`. Or more specifically, the state obtained **after** `P`'s execution.
1. Then we set `relay_parent_number` to `P - 1`, whereas it should be `P`.

Does this sound right?

coriolinus:
It sounds plausible, but I don't feel like I know enough to answer definitively. Maybe write a test case? That should prove it one way or the other, I think.

pepyakin:
Apparently, the tests are saying that the logic is wrong somewhere. And I've realized that this path is also called from the inclusion, where we do need to offset everything to the previous block.

I am confused.

bkchr:
> Exactly. However, I feel that it is wrong. My logic was something like this:
> 
>     1. When we collate a block on the collator, we base it off some relay-parent with height `P`.
> 
>     2. In order to form the inherent we request persistent validation data for `P`
> 
>     3. The control flow gets to the code in question, and collects all the data at state of `P`. Or more specifically, the state obtained **after** `P`'s execution.
> 
>     4. Then we set `relay_parent_number` to `P - 1`, whereas it should be `P`.
> 
> 
> Does this sound right?

That should be right. The docs for `block_number` also say that it is the number of the block this data was created in. I would assume the tests are incorrect.

pepyakin:
Closing for now, expect another PR for this subject.

pepyakin:
Subsumed by https://github.com/paritytech/polkadot/pull/2123

PR #2115, opened 2020-12-14
refactor avail dist test

drahnr:
Contains the solution to always adding a message to the message vault plus a unit test verifying that. Splits the gigantomanic integration test into two large ones as well.


node/network/availability-distribution/src/lib.rs
```diff
@@ -365,13 +365,7 @@ where
 			.map(|(peer, _view)| peer.clone())
 			.collect();
 
-		// If we have no peers that are interested, skip the rest.
-		if peers.is_empty() {
```
coriolinus, line 369: It's not clear what value we get from not `continue`ing here.

node/network/availability-distribution/src/lib.rs
```diff
@@ -451,10 +445,11 @@ async fn send_tracked_gossip_messages_to_peers<Context>(
 where
 	Context: SubsystemContext<Message = AvailabilityDistributionMessage>,
 {
-	if peers.is_empty() {
```
coriolinus, line 454: Is the goal just to modify the internal state in a way that the test can detect? In that case, it might be better to write it like 

```rust
if peers.is_empty() && !cfg!(test) {
```

node/network/availability-distribution/src/lib.rs
```diff
@@ -704,13 +697,21 @@ impl AvailabilityDistributionSubsystem {
...
+		Context: SubsystemContext<Message = AvailabilityDistributionMessage>,
+	{
+		let mut state = ProtocolState::default();
+		self.run_inner(ctx, &mut state).await
+	}
+
+	/// Start processing work.
 	#[tracing::instrument(skip(self, ctx), fields(subsystem = LOG_TARGET))]
-	async fn run<Context>(self, mut ctx: Context) -> Result<()>
+	async fn run_inner<Context>(self, mut ctx: Context, state: &mut ProtocolState) -> Result<()>
```
coriolinus, line 710: :+1: This is a useful pattern. Might be worth adding a note that this function is mainly for testing purposes and general usage should use just `run`.

coriolinus (approved):
I looked more at syntax/style than at the semantics of the added tests. If you want me to take a deeper look there, just let me know. Obviously this needs its merge conflicts and test failures to be resolved, but I don't see any particular issue with merging this once green.

node/network/availability-distribution/src/lib.rs
drahnr: This was the lengthy discussion about adding the message to the `message_vault`, which is a goto cache for seen and messages to redistribute. Not having it in the store, because receiving it at a time where none of the peers has interest in the particular candidate can lead to duplicate gossips which will harm reputation.

node/network/availability-distribution/src/lib.rs
drahnr: Will add a comment.

bkchr:
There are still merge conflicts.

node/network/availability-distribution/src/tests.rs
```diff
@@ -22,18 +22,25 @@ use polkadot_node_subsystem_util::TimeoutExt;
...
+use maplit::{hashset, hashmap};
 
 macro_rules! view {
-	( $( $hash:expr ),* $(,)? ) => { View(vec![ $( $hash.clone() ),* ]) };
+	( $( $hash:expr ),* $(,)? ) => {
+		View(vec![ $( $hash.clone() ),* ])
+	};
+	[ $( $hash:expr ),* $(,)? ] => {
+		View(vec![ $( $hash.clone() ),* ])
+	};
```
bkchr, line 43: These both branches are doing the same?

drahnr: one for `[]` one for `()`, but other than that, yes.

drahnr:
> There are still merge conflicts.

You too fast..

node/network/availability-distribution/src/tests.rs
bkchr: This is not required. Please remove the second branch.

node/network/availability-distribution/src/tests.rs
```diff
@@ -22,18 +22,25 @@ use polkadot_node_subsystem_util::TimeoutExt;
...
+use sp_keyring::Sr25519Keyring;
 use std::{sync::Arc, time::Duration};
-use maplit::hashmap;
+use maplit::{hashset, hashmap};
 
 macro_rules! view {
-	( $( $hash:expr ),* $(,)? ) => { View(vec![ $( $hash.clone() ),* ]) };
+	( $( $hash:expr ),* $(,)? ) => {
+		View(vec![ $( $hash.clone() ),* ])
+	};
```
drahnr, line 40:
```suggestion
```

rphmeier:
@drahnr superseded by #2104 ?

drahnr:
> @drahnr superseded by #2104 ?

It would add two more unit tests, both asserting against the internal state. The base pr #2104 just moved faster (and with that the test fn signatures). If the other PR already covers that case, I think we can close this.

rphmeier:
Yeah, let's keep this open then. Sorry about the rebases - the other PR was blocking.

bkchr:
> > @drahnr superseded by #2104 ?
> 
> It would add two more unit tests, both asserting against the internal state. The base pr #2104 just moved faster (and with that the test fn signatures). If the other PR already covers that case (I'd have to check again, I gave up on keeping up during the second rebase today), I think we can close this.

Can we not rewrite them to be blackboxing tests?

drahnr:
Technically yes, but at the time of writing that was not clear to me that this was a not entirely clear to me that that was a requirement.

PR #2121, opened 2020-12-16
Add logging to collation-generation

pepyakin:
Right now if the collation is not happening one will have to sprinkle
log statements and then recompile the code. It's doubly annoying if that
happens when working with Cumulus: that means one has to resort to
.cargo/config's `paths` or `diener`, which both are not ideal.

This just adds some verbose logging to save the investigators some time
when looking why the collations are not happening


bkchr (approved):

bkchr:
bot merge

pepyakin:
bot merge

PR #2122, opened 2020-12-16
Availability recovery subsystem

### LINKED ISSUE #1974: Implement Availability Recovery Subsystem
Depends on #1597

montekki:
TODO:
- [x] chunk requests in interactions should timeout properly
- [ ] probably add more testcases
__

Closes #1974


node/network/availability-recovery/src/lib.rs
```diff
@@ -0,0 +1,724 @@
...
+	f + 1
+}
+
+/// The Availability Recovery Subsystem.
+pub struct AvailabilityRecoverySubsystem;
+
+type ChunkResponse = Result<(PeerId, ErasureChunk), RecoveryError>;
+
+/// Data we keep around for every chunk that we are awaiting.
+struct AwaitedChunk {
```
coriolinus, line 82: The guide also has an `issued_at` field. Either the guide or this struct should be updated, so they're in sync.

node/network/availability-recovery/src/lib.rs
```diff
@@ -0,0 +1,724 @@
...
+
+	/// Report a peer.
+	ReportPeer(
+		PeerId,
+		Rep,
+	),
+}
+
+/// A state of a single interaction reconstructing an available data.
+struct Interaction {
```
coriolinus, line 120: In the guide, this struct also has a `validator_authority_keys` field. Either the guide or this struct should be updated, so they're in sync.

node/network/availability-recovery/src/lib.rs
```diff
@@ -0,0 +1,724 @@
...
+						"Failed to send Concluded result from interaction",
+					);
+				}
+				return;
+			}
+		}
+	}
+}
+
+struct State {
```
coriolinus, line 238: This struct differs from the guide: the guide has `connecting_validators`, while this has `connection_requests`. These have different types, but I suspect that they serve the same purpose. Either the guide or this struct should be updated so they're in sync.

coriolinus (commented):
Basically all I know about this subsystem is what's written in the guide, so that's what most of the comments are about. Still, there are a few places where it's out of sync; in those cases, the guide should be updated.

Overall, the code is very clean and maps neatly onto the guide, so :+1: for that.

I have not yet attempted to review the tests.

coriolinus:
It would also likely be valuable to add some Jaeger tracing to this subsystem.

montekki:
Yep, I am planning to add metrics and jaeger in a followup, it is already quite bloated

node/network/availability-recovery/src/lib.rs
```diff
@@ -0,0 +1,756 @@
...
+
+// Size of the LRU cache where we keep recovered data.
+const LRU_SIZE: usize = 16;
+
+// A timeout for a chunk request.
+const CHUNK_REQUEST_TIMEOUT: Duration = Duration::from_secs(3);
+
+/// Compute the threshold from the number of validators.
+///
+/// It should be f + 1, where n = 3f + k, where k in {1, 2, 3}, and n is the number of validators.
```
rphmeier, line 74: This should be computed the same way as `data_shards` in `polkadot-erasure-coding`.

node/subsystem/src/errors.rs
```diff
@@ -60,6 +60,23 @@ impl core::fmt::Display for ChainApiError {
...
+#[derive(PartialEq, Debug, Clone)]
+pub enum RecoveryError {
+	/// A chunk is recovered but is invalid.
+	Invalid,
+
+	/// A requested chunk is unavailable.
+	Unavailable,
+}
+
+impl core::fmt::Display for RecoveryError {
```
rphmeier, line 73: why `core` as opposed to `std`?

node/network/availability-recovery/src/error.rs
```diff
@@ -0,0 +1,33 @@
...
+
+// You should have received a copy of the GNU General Public License
+// along with Polkadot.  If not, see <http://www.gnu.org/licenses/>.
+
+//! The `Error` and `Result` types used by the subsystem.
+
+use thiserror::Error;
+
+#[derive(Debug, Error)]
+pub enum Error {
```
rphmeier, line 22: missing docs

node/network/availability-recovery/src/lib.rs
rphmeier: The types are actually the same `ConnectionRequests` is our implementation of the stream mentioned there. But yeah, the names appear different.

node/network/availability-recovery/src/lib.rs
rphmeier: I think I wanted `issued_at` to handle request timeouts. Looking over the guide text now, I don't see where that is referenced. I suppose it was not included here because it wasn't used.

I suggest that we alter the guide and the implementation here to keep `issued_at`, but to prune out any entries from both `discovering_validators` and `live_chunk_requests` which have been around longer than some constant `REQUEST_TIMEOUT`. We'll have to poll periodically in order to do this garbage collection.

Dropping them from the state should lead to the interaction receiving a notification from `requesting_chunks` of `oneshot::Canceled`. Within the interaction, we should roughly just ignore such failures as the loop will request from other validators as a replacement.

node/network/availability-recovery/src/lib.rs
```diff
@@ -0,0 +1,756 @@
...
+	/// The chunk requests that are waiting to complete.
+	requesting_chunks: FuturesUnordered<Timeout<oneshot::Receiver<ChunkResponse>>>,
+}
+
+impl Interaction {
+	async fn run(mut self) {
+		loop {
+			// If it's empty and requesting_chunks is empty,
+			// break and issue a FromInteraction::Concluded(RecoveryError::Unavailable)
+			if self.requesting_chunks.is_empty() && self.shuffling.is_empty() {
```
rphmeier, line 157: We could even break earlier: as soon as `received_chunks + requesting_chunks + shuffling.len() < threshold`

node/network/availability-recovery/src/lib.rs
rphmeier: Now I see that there's a `CHUNK_REQUEST_TIMEOUT`. Still need to see how this is used to clean up the `state`.

node/network/availability-recovery/src/lib.rs
```diff
@@ -0,0 +1,756 @@
...
+			// If received_chunks has more than threshold entries, attempt to recover the data.
+			// If that fails, or a re-encoding of it doesn't match the expected erasure root,
+			// break and issue a FromInteraction::Concluded(RecoveryError::Invalid).
+			// Otherwise, issue a FromInteraction::Concluded(Ok(())).
+			if self.received_chunks.len() >= self.threshold {
+				let concluded = match polkadot_erasure_coding::reconstruct_v1(
+					self.validators.len(),
+					self.received_chunks.values().map(|c| (&c.chunk[..], c.index as usize)),
+				) {
+					Ok(data) => FromInteraction::Concluded(self.candidate_hash.clone(), Ok(data)),
```
rphmeier, line 253: I don't see where it checks the re-encoding matches the expected erasure root. It would be great to have a unit test for that case as well.

rphmeier: Since this loop consists of a few different pieces of functionality, maybe refactor them into separate functions that may be more unit-testable.

node/network/availability-recovery/src/lib.rs
```diff
@@ -0,0 +1,756 @@
...
+		OverseerSignal::Conclude => Ok(true),
+		OverseerSignal::ActiveLeaves(ActiveLeavesUpdate { activated, .. }) => {
+			// if activated is non-empty, set state.live_block_hash to the first block in Activated.
+			if let Some(hash) = activated.get(0) {
+				state.live_block_hash = *hash;
+			}
+
+			Ok(false)
+		}
+		_ => Ok(false),
```
rphmeier, line 349: I prefer exhaustive match to future-proof

node/network/availability-recovery/src/lib.rs
```diff
@@ -0,0 +1,756 @@
...
+							tracing::debug!(
+								target: LOG_TARGET,
+								err = ?e,
+								"A chunk request ended with an error",
+							);
+						}
+						None => {
+							tracing::debug!(
+								target: LOG_TARGET,
+								"A chunk request has timed out",
```
rphmeier, line 237: We need some way of triggering cleanup of the state which holds the other side of the channel, as mentioned above. A periodic poll of the state could work, because `oneshot::Sender` has an `is_cancelled` function.

node/network/availability-recovery/src/lib.rs
```diff
@@ -0,0 +1,756 @@
...
+				session_info,
+				receipt,
+				response_sender,
+			).await
+		}
+		None => {
+			tracing::warn!(
+				target: LOG_TARGET,
+				"SessionInfo is `None` at {}", state.live_block_hash,
+			);
```
rphmeier, line 461: supposed to return `Unavailable` on the sender

node/network/availability-recovery/src/lib.rs
```diff
@@ -0,0 +1,756 @@
...
+) -> error::Result<()> {
+	use rand::seq::SliceRandom;
+	use rand::thread_rng;
+
+	let threshold = recovery_threshold(session_info.validators.len());
+	let to_state = state.from_interaction_tx.clone();
+	let candidate_hash = receipt.hash();
+	let erasure_root = receipt.descriptor.erasure_root;
+	let validators = session_info.validators.clone();
+	let mut shuffling: Vec<_> = (0..validators.len() as u32).collect();
```
rphmeier, line 380: Does `as ValidatorIndex` work?

node/network/availability-recovery/src/lib.rs
```diff
@@ -0,0 +1,756 @@
...
+							// Send the chunk response on the awaited_chunk for the interaction to handle.
+							if let Some(chunk) = chunk {
+								awaited_chunk.response
+									.send(Ok((peer_id, chunk)))
+									.map_err(|_| oneshot::Canceled)?;
+							}
+						}
+						Some(a) => {
+							// If the peer in the entry doesn't match the sending peer,
+							// reinstate the entry, report the peer, and return
```
rphmeier, line 603: peer not reported here so far

rphmeier:
I'd like to see more tests:
  * Test that timeouts work and state is cleaned up
  * Test that back-up validators are requested from even if unavailable
  * Test that erasure-root is checked after recovery

rphmeier:
Generally looks good and seems to match guide! thanks

node/network/availability-recovery/src/lib.rs
burdges: As a rule, we could just move on to another chunk whenever one looks sticky, which amounts to some balancing act between waiting for timeouts and preemptively adding more requests.

node/network/availability-recovery/src/lib.rs
burdges: Yes, rechecking the erasure root calculation is extremely important.  If it fails then the candidate is invalid so we escalate.

node/network/availability-recovery/src/lib.rs
rphmeier: @burdges in practice that would mean having two timeouts a 'sticky' timeout and a 'hard-fail' timeout. seems worth exploring but not for the initial PR i'd say

node/network/availability-recovery/src/lib.rs
rphmeier: Yep - escalation will happen elsewhere. This subsystem just needs to return an "invalid" result to be handled

node/network/availability-recovery/src/lib.rs
```diff
@@ -0,0 +1,756 @@
...
+			// Take the validators we are already connecting to and merge them with the
+			// newly requested one to create a new connection request.
+			let new_discovering_validators_set = state.discovering_validators.keys()
+				.cloned()
+				.chain(std::iter::once(id.clone()))
+				.collect();
+
+			// Issue a NetworkBridgeMessage::ConnectToValidators.
+			// Add the stream of connected validator events to state.connecting_validators.
+			match validator_discovery::connect_to_validators(
```
rphmeier, line 520: Should use `connect_to_past_session_validators`, I believe

rphmeier: it's a little bit of a misnomer because it also works in the current session, but we need to get data for past sessions as well.

node/network/availability-recovery/src/lib.rs
```diff
@@ -0,0 +1,756 @@
...
+		}
+		return Ok(());
+	}
+
+	if let Some(interaction) = state.interactions.get_mut(&candidate_hash) {
+		interaction.awaiting.push(response_sender);
+		return Ok(());
+	}
+
+	let session_index = request_session_index_for_child_ctx(state.live_block_hash, ctx).await?.await??;
```
rphmeier, line 443: we're supposed to get the session index of the candidate, not the session index of the most recent block. might not actually be that easily available here, but the caller of the subsystem should have it. can alter API to accept a `SessionIndex` as a parameter.

node/network/availability-recovery/src/lib.rs
montekki: I believe that it just computes the necessary session index every time, so it would compute the past session for `relay_parent` if needed.

https://github.com/paritytech/polkadot/blob/2c6920c40fb5ffb607cd2d3607dd182a76cee5d3/node/subsystem-util/src/validator_discovery.rs#L40-L47

node/network/availability-recovery/src/lib.rs
```diff
@@ -150,96 +147,111 @@ struct Interaction {
 }
 
 impl Interaction {
-	async fn run(mut self) {
-		loop {
-			// If it's empty and requesting_chunks is empty,
-			// break and issue a FromInteraction::Concluded(RecoveryError::Unavailable)
-			if self.requesting_chunks.is_empty() && self.shuffling.is_empty() {
-				if self.to_state.send(FromInteraction::Concluded(
+	async fn maybe_done(&mut self) -> error::Result<bool> {
```
rphmeier, line 150: style: this function mixes the computation of being done with the effect that's meant to occur upon being done. It would be better to write this as a pure function and dispatch the message at the point it's called.

It also doesn't quite do what the name suggests. maybe rename to `is_impossible` or `is_unavailable`?

node/network/availability-recovery/src/lib.rs
```diff
@@ -150,96 +147,111 @@ struct Interaction {
...
-								}
-							} else {
-								if self.to_state.send(FromInteraction::ReportPeer(
-										peer_id,
+		Ok(())
+	}
+
+	async fn wait_for_chunks(&mut self) -> error::Result<()> {
+		// Check if the requesting chunks is not empty not to poll to completion.
+		if !self.requesting_chunks.is_empty() {
```
rphmeier, line 188: we could avoid a level of indentation by doing an early return instead

node/network/availability-recovery/src/lib.rs
```diff
@@ -250,21 +262,48 @@ impl Interaction {
...
-					);
-				}
-				return;
+				self.to_state.send(concluded).await.map_err(error::Error::ClosedToState)?;
+				return Ok(());
 			}
 		}
 	}
+
+	fn reconstructed_data_matches_root(
```
rphmeier, line 287: style: this could be a free function that accepts `n_validators` and `erasure_root` as parameters, to avoid reliance on `self`.

node/network/availability-recovery/src/lib.rs
rphmeier: Yes, but we're using `relay_parent=state.live_block_hash` here, which is some arbitrary recent block we believe the state to still be present for. It's not the relay-parent of the candidate we're requesting, which we shouldn't use anyway because the state may not be available.

So my original comment is still correct

node/network/availability-recovery/src/lib.rs
```diff
@@ -0,0 +1,843 @@
...
+		return Ok(());
+	}
+
+	let session_index = request_session_index_for_child_ctx(
+		receipt.descriptor.relay_parent,
+		ctx,
+	).await?.await.map_err(error::Error::CanceledSessionIndex)??;
+
+	let session_info = request_session_info_ctx(
+		receipt.descriptor.relay_parent,
```
rphmeier, line 497: We might easily race with state pruning here. It'd be better to use the `live_block_hash` as the relay parent.

node/network/availability-recovery/src/lib.rs
```diff
@@ -0,0 +1,843 @@
...
+		return Ok(());
+	}
+
+	if let Some(interaction) = state.interactions.get_mut(&candidate_hash) {
+		interaction.awaiting.push(response_sender);
+		return Ok(());
+	}
+
+	let session_index = request_session_index_for_child_ctx(
+		receipt.descriptor.relay_parent,
```
rphmeier, line 492: Here we can't use the `live_block_hash`, for the reason discussed elsewhere. I suggest altering the API of availability recovery to accept the session index as an argument. The primary consumer of this API is approval voting, which has the session index for each candidate at the point where it makes the request.

rphmeier: Actually I wrote that and then went to check the guide, and saw that the guide already defined `session_index` to be one of the parameters.

node/subsystem/src/messages.rs
```diff
@@ -237,6 +237,18 @@ pub enum AvailabilityDistributionMessage {
 	NetworkBridgeUpdateV1(NetworkBridgeEvent<protocol_v1::AvailabilityDistributionMessage>),
 }
 
+/// Availability Recovery Message.
+#[derive(Debug)]
+pub enum AvailabilityRecoveryMessage {
+	/// Recover available data from validators on the network.
+	RecoverAvailableData(
+		CandidateReceipt,
+		oneshot::Sender<Result<AvailableData, crate::errors::RecoveryError>>,
```
rphmeier, line 246: the guide has a `SessionIndex` here

rphmeier:
Looks good except for the grumble about session indices. It's annoying how the API of `validator_discovery` and the underlying `NetworkBridgeMessage::ConnectToValidators` have diverged. The interaction in the guide is defined as storing `Vec<AuthorityDiscoveryId>` which is all you need for the raw `NetworkBridgeMessage`. But the `validator_discovery` doesn't support that.

I suggest that we should make `connect_to_authorities` because we're doing unnecessary mapping from validator<>authorityid that the interaction handles. But we can make that future work.

node/network/availability-recovery/src/lib.rs
```diff
@@ -0,0 +1,843 @@
...
+
+			// Issue a NetworkBridgeMessage::ConnectToValidators.
+			// Add the stream of connected validator events to state.connecting_validators.
+			match validator_discovery::connect_to_validators(
+				ctx,
+				relay_parent,
+				new_discovering_validators_set,
+			).await {
+				Ok(new_connection_request) => {
+					state.connection_requests.put(relay_parent, new_connection_request);
```
rphmeier, line 588: The doc on `connection_requests` says that this will drop any other request from the `relay_parent`. I think this API isn't really suitable to our needs. Maybe we should just use `ConnectToValidators` directly and save ourselves the headache.

rphmeier: The main question I have is about ownership of the receiver, which unfortunately I overlooked when writing the guide page. I'd guess that we want the receiver to drop as soon as we either fulfill the request or time out. At the same time, we want the main subsystem state to be the one polling the receiver. I'm sure there's a workaround, but maybe kind of ugly and involves more channels.

node/network/availability-recovery/src/lib.rs
```diff
@@ -564,55 +565,22 @@ async fn handle_from_interaction(
...
-					);
-				}
-			}
+			state.discovering_validators.entry(id).or_default().push(AwaitedChunk {
+				validator_index,
+				candidate_hash,
+				response,
+			});
+
+			state.connecting_validators.push(rx);
```
rphmeier, line 583: With this `StreamUnordered` I do wonder about when the `rx` is actually dropped as that is how the network bridge is actually informed.

node/network/availability-recovery/src/lib.rs
```diff
@@ -774,18 +742,20 @@ impl AvailabilityRecoverySubsystem {
...
-						v.validator_id,
-						v.peer_id,
-					).await {
-						tracing::warn!(
-							target: LOG_TARGET,
-							err = ?e,
-							"Failed to handle a newly connected validator",
-						);
+				v = state.connecting_validators.next() => {
+					if let Some((StreamYield::Item(v), _)) = v {
```
rphmeier, line 746: It seems that we should match on `StreamYield::Finished` and invoke `remove` on that, otherwise the stream will be kept alive internally. 

https://docs.rs/streamunordered/0.5.1/src/streamunordered/lib.rs.html#650

Default behavior is `keep`: https://docs.rs/streamunordered/0.5.1/src/streamunordered/lib.rs.html#672 .

node/network/availability-recovery/src/lib.rs
rphmeier: still stands

rphmeier (approved):

rphmeier:
bot merge

PR #2135, opened 2020-12-17
adjust span names

rphmeier:
https://github.com/paritytech/polkadot/pull/2132#discussion_r545333423


ordian (approved):
Thanks!

What about https://github.com/paritytech/polkadot/blob/f714f5f485c71b25f32aad21429fec78656116a7/node/network/bitfield-distribution/src/lib.rs#L184 and https://github.com/paritytech/polkadot/blob/f714f5f485c71b25f32aad21429fec78656116a7/node/core/bitfield-signing/src/lib.rs#L254?

rphmeier:
those seem to fit the scheme, IDK

PR #2140, opened 2020-12-18
Fix Jaeger service name

tomaka:
Services would no longer be named `rococo-3-validator-5-rococo-3-validator-5` but `polkadot-rococo-3-validator-5`.
`polkadot` prefix is there in order to not create confusion if someone wants to connect additional non-Polkadot services to the same Jaeger agent.


bkchr:
bot merge

PR #2143, opened 2020-12-18
Alter behavior of `max_validators_per_core`

### LINKED ISSUE #2138: Max-per-core shouldn't leave any validators inactive for an entire session
Rather, we should create additional cores which are "Inactive" to accommodate all validators.

Then, even if only 5 or 10 validators are active in backing at any given time, over the course of a session almost all validators will have still rotated onto one of the active cores and been rewarded for backing.

The main open question is about how the `availability_cores` runtime API should function. Higher level code will have to be aware of the possibility of inactive cores. Also, whether we need an inactive variant at all or whether the `Free` variant is already sufficient.

rphmeier:
Closes #2138 

Previously, we would truncate the number of validators to `n_cores * max_validators_per_core`. This implied that only those validators (e.g. 5 / 800 on Kusama, for one parachain) would receive reward points for backing in the current session, which are substantial and would reap the bulk of the rewards.

Now, we simply allocate extra cores but ensure that only `n_validators / max_validators_per_core` are assigned to each core.

So even if we have 1 parachain + 1 parathread, with 100 validators and a maximum of 5, we'd have 20 cores, where only the first 2 see any activity. However, the 20 validator groups would rotate over those 2 cores as normal, and thus would all see rewards over the course of the session.

There are many places in the node-side where we iterate over the cores and/or validator groups, but we typically take action only on `Occupied` cores and in some instances, `Scheduled` cores. These extra cores will always be `Free`, and iterating over some extra contiguous memory should be very cheap.

### DIFF
--- a/roadmap/implementers-guide/src/runtime/scheduler.md
+++ b/roadmap/implementers-guide/src/runtime/scheduler.md
@@ -176,8 +176,8 @@ Actions:
 1. Clear all `Some` members of `AvailabilityCores`. Return all parathread claims to queue with retries un-incremented.
 1. Set `configuration = Configuration::configuration()` (see [`HostConfiguration`](../types/runtime.md#host-configuration))
 1. Determine the number of cores & validator groups as `n_cores`. This is the maximum of
-  1. `Paras::parachains().len() + configuration.parathread_cores`
-  1. `n_validators / max_validators_per_core` > N if `configuration.max_validators_per_core` is `Some` and non-zero.
+   1. `Paras::parachains().len() + configuration.parathread_cores`
+   1. `n_validators / max_validators_per_core` if `configuration.max_validators_per_core` is `Some` and non-zero.
 1. Resize `AvailabilityCores` to have length `n_cores` with all `None` entries.
 1. Compute new validator groups by shuffling using a secure randomness beacon
    - We obtain "shuffled validators" `SV` by shuffling the validators using the `SessionChangeNotification`'s random seed.
--- a/roadmap/implementers-guide/src/runtime/scheduler.md
+++ b/roadmap/implementers-guide/src/runtime/scheduler.md
@@ -175,13 +175,14 @@ Actions:
 1. Set `SessionStartBlock` to current block number.
 1. Clear all `Some` members of `AvailabilityCores`. Return all parathread claims to queue with retries un-incremented.
 1. Set `configuration = Configuration::configuration()` (see [`HostConfiguration`](../types/runtime.md#host-configuration))
-1. Resize `AvailabilityCores` to have length `Paras::parachains().len() + configuration.parathread_cores with all `None` entries.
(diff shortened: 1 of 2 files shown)


pepyakin:
One of the places that I've found to be suspicious is in bitfield-signing in `construct_availability_bitfield` we iterate over all availability cores in parallel and request a chunk for it. 

Even though it tries to run these requests in parallel, due to the impl of the subsystem all the requests will be handled serially. So I think it would be good to fix that before landing this.

roadmap/implementers-guide/src/runtime/scheduler.md
```diff
@@ -175,13 +175,14 @@ Actions:
 1. Set `SessionStartBlock` to current block number.
 1. Clear all `Some` members of `AvailabilityCores`. Return all parathread claims to queue with retries un-incremented.
 1. Set `configuration = Configuration::configuration()` (see [`HostConfiguration`](../types/runtime.md#host-configuration))
-1. Resize `AvailabilityCores` to have length `Paras::parachains().len() + configuration.parathread_cores with all `None` entries.
+1. Determine the number of cores & validator groups as `n_cores`. This is the maximum of
+  1. `Paras::parachains().len() + configuration.parathread_cores`
+  1. `n_validators / max_validators_per_core` > N if `configuration.max_validators_per_core` is `Some` and non-zero.
```
pepyakin, line 180: What is `N` here? And I am not clear why do we have `>` here.

roadmap/implementers-guide/src/runtime/scheduler.md
```diff
@@ -175,13 +175,14 @@ Actions:
 1. Set `SessionStartBlock` to current block number.
 1. Clear all `Some` members of `AvailabilityCores`. Return all parathread claims to queue with retries un-incremented.
 1. Set `configuration = Configuration::configuration()` (see [`HostConfiguration`](../types/runtime.md#host-configuration))
-1. Resize `AvailabilityCores` to have length `Paras::parachains().len() + configuration.parathread_cores with all `None` entries.
+1. Determine the number of cores & validator groups as `n_cores`. This is the maximum of
+  1. `Paras::parachains().len() + configuration.parathread_cores`
```
pepyakin, line 179: Is it me, or there is not enough indentation. GH preview shows these two elements inline with the rest.

roadmap/implementers-guide/src/runtime/scheduler.md
rphmeier: Vestiges from a previous edit I made. Thanks for catching it

roadmap/implementers-guide/src/runtime/scheduler.md
rphmeier: added more indentation

rphmeier:
@pepyakin all good now?

pepyakin (approved):
Yup!

PR #2145, opened 2020-12-18
avoid creating duplicate unbacked spans when we see extra statements

rphmeier:
Receiving a `CandidateBackingMessage::Second` or seeing a `Statement::Seconded` after we'd already seen requisite votes for backing led us to create extra unbacked spans that were never cleaned up. 

image

### DIFF
--- a/node/core/backing/src/lib.rs
+++ b/node/core/backing/src/lib.rs
@@ -373,7 +373,7 @@ struct BackgroundValidationParams<F> {
 	pov: Option<Arc<PoV>>,
 	validator_index: Option<ValidatorIndex>,
 	n_validators: usize,
-	span: JaegerSpan,
+	span: Option<JaegerSpan>,
 	make_command: F,
 }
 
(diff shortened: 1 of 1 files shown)


rphmeier:
honestly, this is related to a larger bug in candidate backing (hooray) where we validate the candidate a second time as well.

rphmeier:
will merge this so we can redeploy; touches no actual logic, just logging

burdges:
I suppose a signature that seconds already references who proposed?

rphmeier:
@burdges the receipt contains the collator address and signature

coriolinus:
@rphmeier Just following up on

> this is related to a larger bug in candidate backing

Does that larger bug have its own issue? If so, I can't seem to find it: the most recent issue created which contains the phrase "candidate backing" seems to have been from 10 days before this PR.

rphmeier:
@coriolinus I didn't file one, no, but please go ahead if you like. It's not a terrible bug since the sanity checks before statement issuance will protect us from issuing conflicting statements. Just wasted a little bit of CPU

PR #2146, opened 2020-12-19
Fix statement distribution: forward statements to other peers.

rphmeier:
Previously, we weren't forwarding new statements to our peers unless they changed their view after we logged the statement. I added a test to check this and then fixed the issue. Also amended the guide accordingly and cleaned up the jaeger spans for statement distribution.


node/network/statement-distribution/src/lib.rs
```diff
@@ -493,27 +500,6 @@ fn check_statement_signature(
 		.and_then(|v| statement.check_signature(&signing_context, v))
 }
 
-type StatementListeners = Vec<mpsc::Sender<SignedFullStatement>>;
```
bkchr, line 496: Why did you remove this?!

Please bring it back. That is required by cumulus.

bkchr (changes requested):

node/network/statement-distribution/src/lib.rs
rphmeier: Will reinstate - misunderstood the origin of this code and thought it was unused.

node/subsystem/src/jaeger.rs
```diff
@@ -159,9 +159,13 @@ pub fn pov_span(pov: &PoV, span_name: impl Into<String>) -> JaegerSpan {
 
 /// Creates a `Span` referring to the given hash. All spans created with [`hash_span`] with the
 /// same hash (even from multiple different nodes) will be visible in the same view on Jaeger.
+///
+/// This span automatically has the `relay-parent` tag set.
 #[inline(always)]
 pub fn hash_span(hash: &Hash, span_name: impl Into<String>) -> JaegerSpan {
-	INSTANCE.read_recursive().span(|| { *hash }, span_name).into()
+	let mut span: JaegerSpan = INSTANCE.read_recursive().span(|| { *hash }, span_name).into();
+	span.add_string_tag("relay-parent", &format!("{:?}", hash));
```
rphmeier, line 167: I added a `relay-parent` tag to all `hash_span` invocations. Probably helps make jaeger searchable.

node/network/statement-distribution/src/lib.rs
```diff
@@ -1586,4 +1627,162 @@ mod tests {
...
+						protocol_v1::ValidationProtocol::StatementDistribution(
+							protocol_v1::StatementDistributionMessage::Statement(r, s)
+						),
+					)
+				) => {
+					assert_eq!(recipients, vec![peer_b.clone()]);
+					assert_eq!(r, hash_a);
+					assert_eq!(s, statement);
+				}
+			);
```
bkchr, line 1780: I think it would be nice to check also b sending this message afterwards.

node/network/statement-distribution/src/lib.rs
```diff
@@ -532,6 +539,15 @@ async fn circulate_statement_and_dependents(
 		None => return,
 	};
 
+	let _span = {
+		let mut span = active_head.span.child("circulate-statement");
+		span.add_string_tag(
```
bkchr, line 544: I'm no fan of all of these calls. We will hash the payload and generate a hash, even if there is no Jaeger instance running, meaning we waste a lot of cpu power.

bkchr (approved):

node/network/statement-distribution/src/lib.rs
rphmeier: https://github.com/paritytech/polkadot/issues/2150

node/network/statement-distribution/src/lib.rs
rphmeier: We're not actually emulating either `peer_a` or `peer_b` in this test. Are you suggesting that we should emulate `peer_b` and also see how it reacts?

PR #2161, opened 2020-12-23
Add relay storage root to persisted validation data

pepyakin:
This PR adds the relay storage root of the relay parent, to persisted validation data. A PDK such as Cumulus can create storage proofs of the relay chain state and the PVF can verify those proofs against the relay parent's storage root.

In runtime APIs it's pretty straightforward, since we can just obtain the storage root directly.

During inclusion it's a bit more tricky, since we need to create persisted validation data in the middle of the block where the storage root is already tainted. 

To overcome this, we obtain the storage root by the submitting the parent block's header with the inclusion inherent. Then, in the inherent logic we make sure that that is really the previous block header by checking its has against the saved block hashes in the frame_system module. 

Alternatively, we could just supply the block hash through PVD, but that would spread that work to each parachain and we are currently planning on implementing contextual execution using merkle proofs.

The inclusion inherent was chosen to host the parent header (in contrast to a special inherent) is to avoid problems with ordering issues related to inherents and also for the future contextual execution.

### DIFF
--- a/node/core/proposer/src/lib.rs
+++ b/node/core/proposer/src/lib.rs
@@ -213,9 +213,9 @@ where
 			drop(_span);
 
 			let inclusion_inherent_data = (
-				self.parent_header,
 				provisioner_data.0,
 				provisioner_data.1,
+				self.parent_header,
 			);
 			inherent_data.put_data(
 				polkadot_primitives::v1::INCLUSION_INHERENT_IDENTIFIER,
--- a/node/core/proposer/src/lib.rs
+++ b/node/core/proposer/src/lib.rs
@@ -98,11 +98,13 @@ where
 		// data to be moved into the future
 		let overseer = self.overseer.clone();
 		let parent_header_hash = parent_header.hash();
+		let parent_header = parent_header.clone();
 
 		async move {
 			Ok(Proposer {
 				inner: proposer?,
 				overseer,
+				parent_header,
 				parent_header_hash,
 			})
 		}.boxed()
@@ -116,6 +118,7 @@ where
 pub struct Proposer<TxPool: TransactionPool<Block = Block>, Backend, Client> {
 	inner: sc_basic_authorship::Proposer<Backend, Block, Client, TxPool>,
 	overseer: OverseerHandler,
+	parent_header: Header,
 	parent_header_hash: Hash,
 }
 
@@ -209,9 +212,14 @@ where
 
 			drop(_span);
 
+			let inclusion_inherent_data = (
+				self.parent_header,
+				provisioner_data.0,
+				provisioner_data.1,
+			);
 			inherent_data.put_data(
 				polkadot_primitives::v1::INCLUSION_INHERENT_IDENTIFIER,
-				&provisioner_data,
+				&inclusion_inherent_data,
 			)?;
 
 			let _span = span.child("authorship-propose");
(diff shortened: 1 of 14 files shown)


runtime/parachains/src/runtime_api_impl/v1.rs
```diff
@@ -190,31 +190,42 @@ fn with_assumption<Config, T, F>(
...
 	let relay_parent_number = <frame_system::Module<T>>::block_number();
-	with_assumption::<T, _, _>(
-		para_id,
-		assumption,
-		|| Some(ValidationData {
-			persisted: crate::util::make_persisted_validation_data::<T>(para_id, relay_parent_number)?,
-			transient: crate::util::make_transient_validation_data::<T>(para_id, relay_parent_number)?,
-		}),
-	)
+	let relay_storage_root = Hash::decode(&mut &sp_io::storage::root()[..])
```
rphmeier, line 196: How heavy is this?

pepyakin: I believe this depends on the size of the delta. Since that's a fresh runtime call - should be no-op.

node/core/proposer/src/lib.rs
```diff
@@ -209,9 +212,14 @@ where
 
 			drop(_span);
 
+			let inclusion_inherent_data = (
+				self.parent_header,
```
rphmeier, line 216: If we put this at the end, it'll be easier to restart Rococo.

As is, I'm not sure how we should plan to upgrade Rococo. Do you think we should just restart it? Do you have more breaking changes planned?

Whether we upgrade the node-side or the runtime first, the `inclusion_inherent::create_inherent` will panic when it fails to decode the inclusion inherent.

pepyakin: That's the last one (I hope :) ). I have some stuff in pipeline but it's not critical and we can get away with runtime only upgrade.

roadmap/implementers-guide/src/runtime/inclusioninherent.md
```diff
@@ -16,11 +16,13 @@ Included: Option<()>,
...
+    1. Hash the parent header and make sure that it corresponds to the block hash of the parent (tracked by the `frame_system` FRAME module),
     1. The `Bitfields` are first forwarded to the `Inclusion::process_bitfields` routine, returning a set of freed cores. Provide a `Scheduler::core_para` as a core-lookup to the `process_bitfields` routine. Annotate each of these freed cores with `FreedReason::Concluded`.
     1. If `Scheduler::availability_timeout_predicate` is `Some`, invoke `Inclusion::collect_pending` using it, and add timed-out cores to the free cores, annotated with `FreedReason::TimedOut`.
     1. Invoke `Scheduler::schedule(freed)`
-	1. Invoke the `Inclusion::process_candidates` routine with the parameters `(backed_candidates, Scheduler::scheduled(), Scheduler::group_validators)`.
+    1. Extract `parent_storage_root` from the parent header,
+	1. Invoke the `Inclusion::process_candidates` routine with the parameters `(parent_storage_root, backed_candidates, Scheduler::scheduled(), Scheduler::group_validators)`.
```
coriolinus, line 25: Nit: it looks like this file is using spaces for indentation, but this line uses a tab.

coriolinus (approved):
Agree that this PR does what its description says. The workaround to get a fresh inherent during inclusion makes sense to me. I agree that this approach is better than the proposed alternative. Agree that it makes sense to add the parent header to the inclusion inherent instead of creating a new inherent for the purpose.

rphmeier (approved):

PR #2162, opened 2020-12-23
Approval Voting Database

rphmeier:
This change-set became large on its own. This is implementing the described database and operations from the Approval Voting subsystem, against the `AuxStore` trait. 

Main points to review:
  - The structs match the "Database Schema" section
  - The start-up logic is fulfilled by `fn clear`
  - The finalization logic is fulfilled by `fn canonicalize`
  - The `ActiveLeavesUpdate` logic is fulfilled by `fn add_block_entry`

Related to #1975


rphmeier:
The bot added `pleasereview` before I could :)

node/core/approval-voting/src/aux_schema/mod.rs
```diff
@@ -0,0 +1,531 @@
...
+#[derive(Debug, Clone, Encode, Decode, PartialEq)]
+pub(crate) struct ApprovalEntry {
+	tranches: Vec<TrancheEntry>,
+	backing_group: GroupIndex,
+	// When the next wakeup for this entry should occur. This is either to
+	// check a no-show or to check if we need to broadcast an assignment.
+	next_wakeup: Tick,
+	our_assignment: Option<OurAssignment>,
+	// `n_validators` bits.
+	assignments: BitVec<BitOrderLsb0, u8>,
```
burdges, line 76: We could discuss this feature further, and even if we keep it caching the assignments remains fine, but in my design no-shows giving approval votes can unassign some no show replacements.

node/core/approval-voting/src/aux_schema/mod.rs
```diff
@@ -0,0 +1,531 @@
...
+}
+
+/// Metadata regarding approval of a particular candidate.
+#[derive(Debug, Clone, Encode, Decode, PartialEq)]
+pub(crate) struct CandidateEntry {
+	candidate: CandidateReceipt,
+	session: SessionIndex,
+	// Assignments are based on blocks, so we need to track assignments separately
+	// based on the block we are looking at.
+	block_assignments: BTreeMap<Hash, ApprovalEntry>,
```
burdges, line 87: I suppose the `Hash` here is the hash of the relay chain block where inclusion happens?  It's ambiguious form this comment.

node/core/approval-voting/src/aux_schema/mod.rs
```diff
@@ -0,0 +1,531 @@
...
+	session: SessionIndex,
+	slot: SlotNumber,
+	relay_vrf_story: RelayVRF,
+	// The candidates included as-of this block and the index of the core they are
+	// leaving. Sorted ascending by core index.
+	candidates: Vec<(CoreIndex, CandidateHash)>,
+	// A bitfield where the i'th bit corresponds to the i'th candidate in `candidates`.
+	// The i'th bit is `true` iff the candidate has been approved in the context of this
+	// block. The block can be considered approved if the bitfield has all bits set to `true`.
+	approved_bitfield: BitVec<BitOrderLsb0, u8>,
```
burdges, line 105: Actually either all bits for which `self.candidates.binary_search_by(|probe| probe.0.cmp(&i)).is_ok()` holds or else all bits but we set any missing cores initially.

node/core/approval-voting/src/aux_schema/mod.rs
rphmeier: That's still maintained in the logic PR that follows, because we always compute a mask over this bit-vec based on the tranches we take.

node/core/approval-voting/src/aux_schema/mod.rs
rphmeier: Yes, that's right.

node/core/approval-voting/src/aux_schema/mod.rs
rphmeier: This bitfield is by candidates, not by cores. So if there were candidates included from cores `[1, 3, 7]`, then the bit-field would have 3 bits corresponding to each of those, respectively. I think all bits is correct with that in mind, but I might misunderstand the purpose of the binary search snippet you posted if it's not for checking existence of the core in the included candidates set.

It does mean that every time we want to mark a candidate as approved, we have an extra search of the candidate set for each block entry a candidate appears in, but that's contiguous memory and would be dwarfed by the cost of writing that metadata to disk.

node/core/approval-voting/src/aux_schema/mod.rs
rphmeier: https://github.com/paritytech/polkadot/blob/b854306311e5872bd5840b26e3e5d2d514ea438f/node/core/approval-voting/src/aux_schema/mod.rs#L83

rphmeier: (from the next PR)

node/core/approval-voting/src/aux_schema/mod.rs
burdges: I only meant that `approved_bitfield` could be initialized to zero for bits representing existent candidates and one for bits representing non-existent candidates.  I just assumed candidates would always be indexed by core in this stage.  Appears you're indexing by some other candidate index, which yes makes `BitVec`'s length is equivalent to presetting non-existent candidates to one.  I see..

rphmeier: From the outside of this struct, we do lookup by core, but we have to do an internal look-up via the `candidates` vec first, which is cheap.

roadmap/implementers-guide/src/node/approval/approval-voting.md
```diff
@@ -143,7 +142,7 @@ Main loop:
...
+On receiving an `OverseerSignal::BlockFinalized(h)`, we fetch the block number `b` of that block from the ChainApi subsystem. We update our `StoredBlockRange` to begin at `b+1`. Additionally, we remove all block entries and candidates referenced by them up to and including `b`. Lastly, we prune out all descendents of `h` transitively: when we remove a `BlockEntry` with number `b` that is not equal to `h`, we recursively delete all the `BlockEntry`s referenced as children. We remove the `block_assignments` entry for the block hash and if `block_assignments` is now empty, remove the `CandidateEntry`. We also update each of the `BlockNumber -> Vec<Hash>` keys in the database to reflect the blocks at that height, clearing if empty.
```
burdges, line 145: We keep an checks running however I guess.  We should also keep the erasure coded pieces and the outgoing messages sent by blocks for another 24 hours or whatever our XCMP window is.

rphmeier: We keep around the data but don't do any checks after finality anymore.

node/core/approval-voting/src/aux_schema/mod.rs
coriolinus: Would it be worthwhile to define a `BlockHash` newtype adjacent to `CandidateHash`? It won't change anything re: type safety, but it might make it even harder to confuse what each hash represents.

node/core/approval-voting/src/aux_schema/mod.rs
```diff
@@ -0,0 +1,531 @@
...
+//!
+//! With tens or hundreds of parachains, hundreds of validators, and parablocks
+//! in every relay chain block, there can be a humongous amount of information to reference
+//! at any given time.
+//!
+//! As such, we provide a function from this module to clear the database on start-up.
+//! In the future, we may use a temporary DB which doesn't need to be wiped, but for the
+//! time being we share the same DB with the rest of Substrate.
+
+// TODO https://github.com/paritytech/polkadot/1975: remove this
```
coriolinus, line 33: bad link

```suggestion
// TODO https://github.com/paritytech/polkadot/issues/1975: remove this
```

node/core/approval-voting/src/aux_schema/mod.rs
coriolinus, line 87: Slight divergence from the guide here: `HashMap` vs `BTreeMap`. I'm curious why the change; does it prove useful to sort the assignments or have a consistent iteration order or something?

node/core/approval-voting/src/aux_schema/mod.rs
```diff
@@ -0,0 +1,531 @@
...
+	// Assignments are based on blocks, so we need to track assignments separately
+	// based on the block we are looking at.
+	block_assignments: BTreeMap<Hash, ApprovalEntry>,
+	approvals: BitVec<BitOrderLsb0, u8>,
+}
+
+/// Metadata regarding approval of a particular block, by way of approval of the
+/// candidates contained within it.
+#[derive(Debug, Clone, Encode, Decode, PartialEq)]
+pub(crate) struct BlockEntry {
```
coriolinus, line 94: It's fine that this omits `rotation_offset` from the guide, as I can't see any documentation of what that field would actually be used for, but we should probably update the guide to remove it there also.

node/core/approval-voting/src/aux_schema/mod.rs
```diff
@@ -0,0 +1,531 @@
...
+	// block. The block can be considered approved if the bitfield has all bits set to `true`.
+	approved_bitfield: BitVec<BitOrderLsb0, u8>,
+	children: Vec<Hash>,
+}
+
+/// A range from earliest..last block number stored within the DB.
+#[derive(Debug, Clone, Encode, Decode, PartialEq)]
+pub(crate) struct StoredBlockRange(BlockNumber, BlockNumber);
+
+// TODO https://github.com/paritytech/polkadot/1975: probably in lib.rs
```
coriolinus, line 113:
```suggestion
// TODO https://github.com/paritytech/polkadot/issues/1975: probably in lib.rs
```

node/core/approval-voting/src/aux_schema/mod.rs
```diff
@@ -0,0 +1,531 @@
...
+	let deleted_keys: Vec<_> = deleted_block_keys.iter().map(|k| &k[..])
+		.chain(deleted_height_keys.iter().map(|k| &k[..]))
+		.chain(deleted_candidates.iter().map(|k| &k[..]))
+		.collect();
+
+	// Update the values on-disk.
+	store.insert_aux(
+		inserted_keys.iter(),
+		deleted_keys.iter(),
+	)?;
```
coriolinus, line 288: `Vec<_>` implements `IntoIterator` already.

```suggestion
	store.insert_aux(inserted_keys, deleted_keys)?;
```

node/core/approval-voting/src/aux_schema/tests.rs
```diff
@@ -0,0 +1,517 @@
...
+
+
+#[test]
+fn canonicalize_works() {
+	let store = TestStore::default();
+
+	//   -> B1 -> C1 -> D1
+	// A -> B2 -> C2 -> D2
+	//
+	// We'll canonicalize C1. Everytning except D1 should disappear.
```
coriolinus, line 334:
```suggestion
	// We'll canonicalize C1. Everything except D1 should disappear.
```

typo

coriolinus (approved):
Some nits, but nothing essential. Looks like a reasonable implementation of a data store backed by `AuxStore` for the types described in the guide.

node/core/approval-voting/src/aux_schema/mod.rs
rphmeier: `HashMap` is not encodable, it seems

node/core/approval-voting/src/aux_schema/mod.rs
rphmeier: I already removed it from the guide in this PR

node/core/approval-voting/src/aux_schema/mod.rs
coriolinus: Ah, I see. Well, I certainly agree that `insert_aux`'s type signature could be more ergonomic!

node/core/approval-voting/src/aux_schema/mod.rs
rphmeier: Actually it was in the next pr 😅

node/core/approval-voting/src/aux_schema/mod.rs
rphmeier: "worthwhile" is always a question of opportunity cost :) . I wouldn't object to such a change, but I don't think it's in the scope of this PR.

coriolinus: Agree, that would be its own thing. I'll extract an issue.

Issue #1634, opened 2020-08-25
Crowdfund refactor


xlc:
Found some code that can be improved in crowdfund pallet

There are no more additional transfer fees

https://github.com/paritytech/polkadot/blob/102aa6fd72b7fcb092d5f8d64ce3d98b78b0a286/runtime/common/src/crowdfund.rs#L449-L453

https://github.com/paritytech/polkadot/blob/102aa6fd72b7fcb092d5f8d64ce3d98b78b0a286/runtime/common/src/crowdfund.rs#L480-L483

Many places can be updated using `mutate` or `try_mutate` to improve performance

https://github.com/paritytech/polkadot/blob/102aa6fd72b7fcb092d5f8d64ce3d98b78b0a286/runtime/common/src/crowdfund.rs#L272-L274

https://github.com/paritytech/polkadot/blob/102aa6fd72b7fcb092d5f8d64ce3d98b78b0a286/runtime/common/src/crowdfund.rs#L367-L373

bkchr:
CC @shawntabrizi

shawntabrizi:
Yeah I expect all the Polkadot pallets will need to be reviewed and updated.

Here is another issue I opened about crowdfund: https://github.com/paritytech/polkadot/issues/662

shawntabrizi:
One more thing, if I remember correctly, crowdfund does not guarantee that funds make it back to the end users. Since the funds become property of the parachain origin, the owner of that origin can move the funds anywhere a the end of the slot.

Probably something we want to patch.

xlc:
Another thing that will be helpful to improve is that currently the participants have to call withdraw before `RetirementPeriod` passed, or they will lost their funds. It is likely after two years, people will forgot about this so some may not be able to withdraw in a timely manner.

An easy improvement will be allow anyone to trigger withdraw on behalf others so at least the crowdfund owner can choose to refund everyone before calling `dissolve`.

PR #2166, opened 2020-12-24
Crowdloan Updates

shawntabrizi:
* Renames pallet to crowdloan
* Allows any user to withdraw contributions of a closed crowdloan on behalf of another user
* Uses append API for optimized vec mutation
* Introduces partial removal of contribution trie for weight reasons
* Moved `on_finalize` logic to `on_initialize`
* benchmarks
* changes withdraw/create pattern with simple transfer now that extra transfer fee does not exist

Fixes: #1634


runtime/common/src/crowdloan.rs
```diff
@@ -304,23 +304,23 @@ decl_module! {
 		/// slot. It will be withdrawable in two instances: the parachain becomes retired; or the
 		/// slot is unable to be purchased and the timeout expires.
 		#[weight = 0]
-		fn contribute(origin, #[compact] index: FundIndex, #[compact] value: BalanceOf<T>) {
+		fn contribute(origin, #[compact] index: FundIndex, #[compact] value: BalanceOf<T>, return_to: T::AccountId) {
```
xlc, line 307: the parachain AccountId type may not be same as Polkadot. e.g. it could be H160 for evm chain.

shawntabrizi: Look at how we approached the claims pallet in Polkadot. A user can sign a message with their polkadot account and then assign the recipient for the claim. I dont think it is up to our pallet here to understand all ways that downstream chains may use this data.

xlc: Make it `Vec<u8>` is enough to represent all different kinds of account types and we don't need another claim process

shawntabrizi: Im not yet convinced, but will keep it in mind. Should be easy enough to update if we need to, but like, feels wrong instinctually

runtime/common/src/crowdloan.rs
```diff
@@ -482,23 +491,30 @@ decl_module! {
...
+					Self::deposit_event(RawEvent::Dissolved(index));
+				},
+				child::KillOutcome::SomeRemaining => {
+					Self::deposit_event(RawEvent::PartiallyDissolved(index));
+				}
+			}
 		}
 
-		fn on_finalize(n: T::BlockNumber) {
+		fn on_initialize(n: T::BlockNumber) -> frame_support::weights::Weight {
```
thiolliere, line 517: it is hard to me to say if moving this on_initialize can be an issue, I don't know how intricated slots and crowdfunds are.

thiolliere (approved):
apart from the on_initialize change which I cannot easily know the consequences, everything else is good to me.

Also I guess the weight info generation is coming in follow-up PR ?

runtime/common/src/crowdloan.rs
shawntabrizi: all tests pass :)

it should only affect 1 block difference with any auction/crowdfund timelines so not a big deal.

runtime/common/src/crowdloan.rs
```diff
@@ -304,23 +313,23 @@ decl_module! {
 		/// slot. It will be withdrawable in two instances: the parachain becomes retired; or the
 		/// slot is unable to be purchased and the timeout expires.
 		#[weight = 0]
-		fn contribute(origin, #[compact] index: FundIndex, #[compact] value: BalanceOf<T>) {
+		fn contribute(origin, #[compact] index: FundIndex, #[compact] value: BalanceOf<T>, return_to: T::AccountId) {
```
gavofyork, line 316: i don't like this change. it facilitates a "shadow transfer", i.e. allowing transfer of funds between accounts while being ostensibly for some other use. this means you can't e.g. introduce the crowdfund module when a chain (for legal reasons) cannot enable transfers, and you can't have a proxy type which allows crowdloans to be made but disallowed transfers.

runtime/common/src/crowdloan.rs
gavofyork: it might be an issue... care should be taken regarding the state transitioning of the loans. in general i'd prefer to ensure that any time the current block number is used within this call graph, it is first decremented by 1, which should make it exactly equivalent to before.

runtime/common/src/crowdloan.rs
xlc: Just requires transfer permission in order for proxy to participate crowdloan. Otherwise because proxy setup will not exist on parachain, it will simply make the rewards inaccessible.

runtime/common/src/crowdloan.rs
shawntabrizi: I have checked the logic here again, and I do not see issue with the change here from `on_finalize` to `on_initialize`.

I don't think it would be right to manipulate the block number in the `on_initialize` as you hinted. Functions like `<slots::Module<T>>::is_ending(n)` would report wrong information when given the wrong block number. For example, if we tell it `is_ending(n-1)` it might report true, when `is_ending(n)` would be false because the ending period is already finished in the current block.

The only different I see between `on_initialize` vs `on_finalize` is that crowdloans do not consider donations that come in on that last block for bidding on that block. But it would increment the bid in the next block no problem.

joepetrowski:
Regarding @xlc 's [comment](https://github.com/paritytech/polkadot/issues/1634#issuecomment-680289691) to allow `fn withdraw` for a `target: AccountId`, is there any issue with that? If not, should I make a separate issue or can you add it to this PR?

shawntabrizi:
@joepetrowski done!

runtime/common/src/crowdloan.rs
```diff
@@ -513,17 +519,21 @@ decl_module! {
...
 						fund.first_slot,
 						fund.last_slot,
 						fund.raised,
 					);
+
+					Self::deposit_event(RawEvent::HandleBidResult(index, result));
 				}
 			}
+
+			0
```
athei, line 536: You added a benchmark for this. Why not using the results here?

runtime/common/src/crowdloan.rs
```diff
@@ -440,8 +444,8 @@ decl_module! {
 
 		/// Withdraw full balance of a contributor to an unsuccessful or off-boarded fund.
 		#[weight = 0]
```
athei, line 446: You added a benchmark for this. Why not using the results here?

runtime/common/src/crowdloan.rs
kianenigma: What was the incentive to move to on_inititalize in the first place?

runtime/common/src/crowdloan.rs
```diff
@@ -482,23 +484,27 @@ decl_module! {
...
+					T::OrphanedFunds::on_unbalanced(imbalance);
 
-			Self::crowdfund_kill(index);
-			<Funds<T>>::remove(index);
+					<Funds<T>>::remove(index);
 
-			Self::deposit_event(RawEvent::Dissolved(index));
+					Self::deposit_event(RawEvent::Dissolved(index));
+				},
+				child::KillOutcome::SomeRemaining => {
```
kianenigma, line 501: There must be an explanation for this that I haven't reached yet, but I can't help but ask: where/when is the rest of this being deleted, if we happen to encounter it? I wasn expecting something similar to contracts pallet where we gradually wipe on_initialize.

runtime/common/src/crowdloan.rs
```diff
@@ -440,8 +444,8 @@ decl_module! {
 
 		/// Withdraw full balance of a contributor to an unsuccessful or off-boarded fund.
 		#[weight = 0]
-		fn withdraw(origin, #[compact] index: FundIndex) {
-			let who = ensure_signed(origin)?;
+		fn withdraw(origin, who: T::AccountId, #[compact] index: FundIndex) {
+			let _ = ensure_signed(origin)?;
```
athei, line 448: The `?` does already suppress the warning you are trying to avoid AFAIK.

```suggestion
			ensure_signed(origin)?;
```

runtime/common/src/crowdloan.rs
```diff
@@ -1085,184 +1115,250 @@ mod tests {
...
+				assert_eq!(Crowdloan::contribution_get(0, &i), 100);
+			}
+		});
+
+		ext.commit_all().unwrap();
+		ext.execute_with(|| {
+			// Partially dissolve the crowdloan, again
+			assert_ok!(Crowdloan::dissolve(Origin::signed(1), 0));
+			for i in 0 .. 20 {
+				println!("{:?}", i);
```
kianenigma, line 1286:
```suggestion
```

runtime/common/src/crowdloan.rs
athei: I think the sender is expected to send this dispatchable until is is completely resolved in order to get the deposit back. We need to make sure that the deposit is high enough so that the sender is incentivized to do so.

kianenigma: sensible, thanks. 

> We need to make sure that the deposit is high enough so that the sender is incentivized to do so.

👍

runtime/common/src/crowdloan.rs
shawntabrizi: You need to move logic which may have variable/unknown weight into the beginning of the block where possible. You don't want to be left with more logic than you can fit at the end of the block. In general, `on_finalize` should only be small pieces of known size logic, not something like this crowdloan management.

runtime/common/src/crowdloan.rs
shawntabrizi: I didnt put any benchmark integration yet because I have not run them. I still have updates to slots which this pallet is tightly coupled to, so benchmarks wont matter until that is also done.

runtime/common/src/crowdloan.rs
shawntabrizi: Anyone can call this extrinsic, so as long as someone wants to maintain the network, it will be called. Even if it costs someone some tokens to do it, you can imagine Parity or Web3 calling this to clean up storage on the blockchain.

runtime/common/src/crowdloan.rs
shawntabrizi: same as above

athei (approved):

shawntabrizi:
bot merge

PR #2169, opened 2020-12-24
Improve Network Spans

rphmeier:
The main change here revolves around adding a span to the live candidates we track in the availability distribution subsystem which is removed after we receive our availability piece. In the case that we were one of the backers of the candidate and have all availability pieces, the span measures the time it takes to load all the pieces from the availability store and dispatch network messages to other peers awaiting them.

I was going to remove the `bitfield-gossip-received` trace, because it seems very spammy. But I noticed that almost all the traces are originating from the same peer - something that shouldn't happen if  a peer isn't sending duplicate messages. So instead I amended the trace to track the validator-index of the bitfields that are sent. We can use that to determine if the reputation system isn't correctly disconnecting the peer who's spamming all our validators.

### DIFF
--- a/erasure-coding/src/lib.rs
+++ b/erasure-coding/src/lib.rs
@@ -120,17 +120,28 @@ impl CodeParams {
 	}
 }
 
-fn code_params(n_validators: usize) -> Result<CodeParams, Error> {
+const fn n_faulty(n_validators: usize) -> Result<usize, Error> {
 	if n_validators > MAX_VALIDATORS { return Err(Error::TooManyValidators) }
 	if n_validators <= 1 { return Err(Error::NotEnoughValidators) }
 
-	let n_faulty = n_validators.saturating_sub(1) / 3;
+	Ok(n_validators.saturating_sub(1) / 3)
+}
+
+fn code_params(n_validators: usize) -> Result<CodeParams, Error> {
+	let n_faulty = n_faulty(n_validators)?;
 	let n_good = n_validators - n_faulty;
 
 	Ok(CodeParams {
 		data_shards: n_faulty + 1,
 		parity_shards: n_good - 1,
 	})
+}
+
+/// Obtain a threshold of chunks that should be enough to recover the data.
+pub fn recovery_threshold(n_validators: usize) -> Result<usize, Error> {
+	let n_faulty = n_faulty(n_validators)?;
+
+	Ok(n_faulty + 1)
 }
 
 /// Obtain erasure-coded chunks for v0 `AvailableData`, one for each validator.
--- a/erasure-coding/src/lib.rs
+++ b/erasure-coding/src/lib.rs
@@ -119,7 +119,9 @@ impl CodeParams {
 			.expect("this struct is not created with invalid shard number; qed")
 	}
 }
-
+/// Returns the maximum number of allowed, faulty chunks
+/// which does not prevent recovery given all other pieces
+/// are correct.
 const fn n_faulty(n_validators: usize) -> Result<usize, Error> {
 	if n_validators > MAX_VALIDATORS { return Err(Error::TooManyValidators) }
 	if n_validators <= 1 { return Err(Error::NotEnoughValidators) }
--- a/node/network/availability-distribution/src/lib.rs
+++ b/node/network/availability-distribution/src/lib.rs
@@ -167,7 +167,7 @@ struct PerCandidate {
 	/// The set of relay chain blocks this appears to be live in.
 	live_in: HashSet<Hash>,
 
-	/// A Jaeger span relating to this candidate. Dropped after we obtain our availability piece.
+	/// A Jaeger span relating to this candidate.
 	span: jaeger::JaegerSpan,
 }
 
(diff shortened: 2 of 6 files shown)


eskimor (approved):
LGTM

node/network/availability-distribution/src/lib.rs
```diff
@@ -218,16 +232,28 @@ impl ProtocolState {
...
+							received_messages: HashMap::new(),
+							sent_messages: HashMap::new(),
+							validators: validators.clone(),
+							validator_index,
+							descriptor,
+							live_in: HashSet::new(),
+							span: jaeger::candidate_hash_span(&receipt_hash, "pending-availability"),
+						})
+					} else {
+						tracing::warn!(target: LOG_TARGET, "No `per_candidate` but not fresh. logic error");
```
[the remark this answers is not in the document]
rphmeier: It seems correct as-is. There's no `per_candidate` since it's a vacant entry. So it would be very strange for `FetchedLiveCandidate` to be `Cached` as there appears to be no bookkeeping for `receipt_hash` yet.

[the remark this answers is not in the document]
eskimor: Oh yeah. Read it the wrong way round and that more than once :1st_place_medal:

erasure-coding/src/lib.rs
```diff
@@ -120,11 +120,15 @@ impl CodeParams {
 	}
 }
```
drahnr, line 122:
```suggestion
/// Returns the maximum number of allowed, faulty chunks
/// which does not prevent recovery given all other pieces
/// are correct.
```

node/subsystem/src/jaeger.rs
```diff
@@ -148,7 +154,11 @@ impl From<mick_jaeger::Span> for JaegerSpan {
 /// Shortcut for [`candidate_hash_span`] with the hash of the `Candidate` block.
 #[inline(always)]
```
drahnr, line 155:
```suggestion
```

drahnr (commented):
I just had  a sneak peek in here, yet to double down on the logic. I think the separation of sending and adding messages to the vault is a good idea.

node/network/availability-distribution/src/lib.rs
```diff
@@ -259,6 +290,18 @@ impl ProtocolState {
 	}
 }
 
+fn cached_live_candidates_unioned<'a>(
```
drahnr, line 293: It would be preferable to add a `_inner` suffix.

node/network/availability-distribution/src/lib.rs
```diff
@@ -166,13 +166,30 @@ struct PerCandidate {
...
+	/// Add a chunk to the message vault. Overwrites anything that was already present.
+	fn add_message(&mut self, chunk_index: u32, message: AvailabilityGossipMessage) {
+		let _ = self.message_vault.insert(chunk_index, message);
+	}
+
+	/// Clean up the span if we've got our own chunk.
+	fn drop_span_after_own_availability(&mut self) {
+		if let Some(validator_index) = self.validator_index {
+			if self.message_vault.contains_key(&validator_index) {
+				self.span = jaeger::JaegerSpan::Disabled;
```
drahnr, line 190: A bit of a quirk, I'll add a PR to handle this as part of the `JaegerSpan` API.

node/network/availability-distribution/src/lib.rs
```diff
@@ -557,105 +611,140 @@ where
...
+				modify_reputation(ctx, origin, COST_PEER_DUPLICATE_MESSAGE).await;
+				return Ok(());
+			}
+		}
+
+		// check that the message content matches what we have already before rewarding
+		// the peer.
+		{
+			let _span = span.child("check-accurate");
+			if existing == &message {
```
drahnr, line 635: :+1:

node/network/availability-distribution/src/lib.rs
```diff
@@ -557,105 +611,140 @@ where
...
+		let _span = span.child("check-chunk-hash");
+		let erasure_chunk_hash = BlakeTwo256::hash(&message.erasure_chunk.chunk);
+		if anticipated_hash != erasure_chunk_hash {
 			tracing::trace!(
 				target: LOG_TARGET,
 				candidate_hash = ?message.candidate_hash,
 				peer = %origin,
-				error = ?e,
-				"Failed to calculate chunk merkle proof",
+				"Peer send chunk with invalid merkle proof",
```
drahnr, line 682:
```suggestion
				"Peer sent chunk with invalid merkle proof",
```

node/network/availability-distribution/src/lib.rs
```diff
@@ -681,7 +770,14 @@ impl AvailabilityDistributionSubsystem {
 	where
 		Context: SubsystemContext<Message = AvailabilityDistributionMessage>,
 	{
-		let mut state = ProtocolState::default();
+		let mut state = ProtocolState {
```
drahnr, line 773: Not sure what we gain with the explicit initialization of all members?

drahnr (approved):
A few nits, other than that LGTM. The refactored logic looks excellent.

node/network/availability-distribution/src/lib.rs
drahnr: After a few attempts, it seems this is the most idiomatic for now.

coriolinus: You could maybe write an API like

```rust
impl JaegerSpan {
  fn drop_if(&mut self, condition: impl FnOnce() -> bool) { 
    if condition() {
      *self = JaegerSpan::Disabled;
    }
}
```

Which would mean that we could rewrite `drop_span_after_own_availability` as

```rust
self.span.drop_if(|| match self.validator_index {
  None => false,
  Some(validator_index) => self.message_vault.contains_key(&validator_index),
});
```

It's not a huge improvement over what currently exists, but it's something.

node/network/availability-distribution/src/lib.rs
rphmeier: I removed the `Default` because `JaegerSpan` is not `Default`

node/network/availability-distribution/src/lib.rs
rphmeier: @coriolinus Your snippet showing use of `drop_if` will not compile due to a borrow conflict on `self`. I think a simple `disable()` function would be sufficient to avoid passing things around in closures.

node/network/availability-distribution/src/lib.rs
rphmeier: It's not just an "inner" function - I introduced this to get around borrowing issues in another function elsewhere.

rphmeier:
bot merge

PR #2171, opened 2020-12-27
Migrate Claims Pallet to WeightInfo

shawntabrizi:
This PR updates the Claims pallet to use the generated WeightInfo from the Benchmarking Bot.


runtime/kusama/src/weights/claims.rs
```diff
@@ -0,0 +1,83 @@
...
+
+#![allow(unused_parens)]
+#![allow(unused_imports)]
+
+use frame_support::{traits::Get, weights::Weight};
+use sp_std::marker::PhantomData;
+
+/// Weight functions for claims.
+pub struct WeightInfo<T>(PhantomData<T>);
+impl<T: frame_system::Config> claims::WeightInfo for WeightInfo<T> {
```
shawntabrizi, line 45:
```suggestion
impl<T: frame_system::Config> runtime_common::claims::WeightInfo for WeightInfo<T> {
```

shawntabrizi:
Note that some extrinsic weights (the one with prevalidate logic) have increased by 2 reads weight. This is because of duplication in reading certain storage items. This should be fine.

runtime/common/src/claims.rs
```diff
@@ -47,6 +49,30 @@ pub trait Config: frame_system::Config {
...
+impl WeightInfo for TestWeightInfo {
+	fn claim() -> Weight { 0 }
+	fn validate_unsigned_claim() -> Weight { 0 }
+	fn mint_claim() -> Weight { 0 }
+	fn claim_attest() -> Weight { 0 }
+	fn validate_unsigned_claim_attest() -> Weight { 0 }
+	fn attest() -> Weight { 0 }
+	fn validate_prevalidate_attests() -> Weight { 0 }
+	fn move_claim() -> Weight { 0 }
 }
```
athei, line 76: Shouldn't that the generated by the bot inside the weights module?

shawntabrizi: All that doesn't really work when these pallets are not their own crates.

runtime/common/src/claims.rs
```diff
@@ -1268,7 +1287,13 @@ mod benchmarking {
 
 		// Benchmark the time it takes to execute `validate_unsigned` for `claim`
 		validate_unsigned_claim {
```
athei, line 1289: Aren't signed extension not executed during benchmarks?

runtime/common/src/claims.rs
```diff
@@ -241,7 +267,7 @@ decl_module! {
 		/// - Write: Vesting Vesting, Account, Balance Lock, Total, Claim, Claims Vesting, Signing
 		/// Validate Unsigned: +188.7 µs
 		/// </weight>
-		#[weight = T::DbWeight::get().reads_writes(7, 7) + 270_000_000 + 190_000_000]
+		#[weight = T::WeightInfo::claim().saturating_add(T::WeightInfo::validate_unsigned_claim())]
```
thiolliere, line 270: at some point wouldn't be better to have all the unsigned and dispatchable logic is the same benchmark ? to avoid counting 2 times some storages ?

Anyway it doesn't really matter here.

thiolliere (approved):
Looks good to me

runtime/common/src/claims.rs
thiolliere: if benchmark use `_($args)` or `$dispatch($args)` then it only execute:
```rust
				<
					Call<T $(, $instance)? > as $crate::frame_support::traits::UnfilteredDispatchable
				>::dispatch_bypass_filter(
					Call::<T $(, $instance)? >::$dispatch($($arg),*), $origin.into()
				)?;
```

shawntabrizi: def not. the only logic executed during the benchmark is what is define in the dispatchable function or whatever other logic the benchmark wraps

runtime/common/src/claims.rs
shawntabrizi: I could also use whitelisting here, but I think it doesn't matter, and is probably better to keep things simple.

athei (approved):

runtime/common/src/claims.rs
shawntabrizi: took your suggestion!

shawntabrizi:
bot merge

PR #2188, opened 2021-01-04
fix `availability_cores`

JNhua:
When `schedule_para_cleanup` become effective, an error will occur. The log as follows:
`Thread 'tokio-runtime-worker' panicked at 'index out of bounds: the len is 1 but the index is 1', /xxx/polkadot-4038f27d5e4ea2e8/0df0195/runtime/parachains/src/runtime_api_impl/v1.rs:154`.

### DIFF
--- a/runtime/parachains/src/scheduler.rs
+++ b/runtime/parachains/src/scheduler.rs
@@ -726,10 +726,11 @@ mod tests {
 	use frame_support::traits::{OnFinalize, OnInitialize};
 	use keyring::Sr25519Keyring;
 
-	use crate::mock::{new_test_ext, Configuration, Paras, System, Scheduler, GenesisConfig as MockGenesisConfig};
+	use crate::mock::{new_test_ext, Configuration, Paras, System, Scheduler, GenesisConfig as MockGenesisConfig, Test};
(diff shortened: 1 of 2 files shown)


bkchr (commented):
Could you add some test that validates that this will not happen again?

bkchr:
I think there is more going on, as `scheduled()` should probably not return anything for empty slots.

bkchr:
@JNhua this breaks some other test :(

runtime/parachains/src/scheduler.rs
```diff
@@ -2038,4 +2051,119 @@ mod tests {
...
+			});
+
+			assert_eq!(Paras::parachains(), vec![chain_b]);
+
+			{
+				assert_eq!(AvailabilityCores::get().len(), 1);
+
+				let scheduled = Scheduler::scheduled();
+
+				assert_eq!(scheduled.len(), 2);
```
rphmeier, line 2146: Why should `scheduled` have 2 items, including `chain_a`, if `chain_a` has been de-registered?

rphmeier:
This is a bug stemming from #2123 but this is not the correct fix. Thank you @JNhua for alerting. I will prepare a new PR with a different fix.

PR #2192, opened 2021-01-04
guide: disputes

drahnr:
`VotesDB` and `DisputeParticipation` design in form of a guide update plus the runtime component


roadmap/implementers-guide/src/node/disputes/participation.md
```diff
@@ -0,0 +1,94 @@
+# Dispute Participation
+
+Tracks all open disputes in the active leaves set.
+
+## Remarks
+
+`UnsignedTransaction`s are OK to be used, since the inner
+element `CommittedCanddidateReceipt` is verifiable.
```
coriolinus, line 8:
```suggestion
element `CommittedCandidateReceipt` is verifiable.
```

typo

roadmap/implementers-guide/src/node/disputes/participation.md
```diff
@@ -0,0 +1,94 @@
...
+
+## ToDo
+
+1. Specify what is being gossiped, what messages and what they must contain.
```
coriolinus, line 12: unresolved todo

roadmap/implementers-guide/src/node/disputes/participation.md
```diff
@@ -0,0 +1,94 @@
...
+
+
+## IO
+
+Inputs:1. Receive `DisputeParticipationMessage::Resolution`
```
coriolinus, line 17:
```suggestion
Inputs:

1. Receive `DisputeParticipationMessage::Resolution`
```

Formatting

roadmap/implementers-guide/src/node/disputes/participation.md
```diff
@@ -0,0 +1,94 @@
...
+1. Delete all relevant data associated with this dispute
+
+
+* `DisputeParticipationMessage::Detection`
+* `DisputeParticipationMessage::Resolution`
```
coriolinus, line 22: Not yet specified: brief steps to take on `DisputeParticipationMessage::Detection`

Redundantly specified: `DisputeParticipationMessage::Resolution`

roadmap/implementers-guide/src/node/disputes/participation.md
```diff
@@ -0,0 +1,94 @@
...
+
+Outputs:
+
+* `AvailabilityRecoveryMessage::RecoverAvailableData`
+* `CandidateValidationMessage::ValidateFromChainState`
+* `...::Blacklist`
```
coriolinus, line 28: I believe the ellipsis here counts as an unresolved TODO. I think this is blacklisting a particular parachain block, right? In that case, possibly Candidate Backing could maintain the blacklist.

roadmap/implementers-guide/src/node/disputes/votesdb.md
```diff
@@ -0,0 +1,137 @@
...
+
+Votes must be persisted. Disputes might happen long after
+there was a vote, as such just keeping things in memory
+is thus insufficient, as every blip (OOM, bug, maintanance)
+could cause the node to forget the state.
+
+## ToDo
+
+* Missing Definition of Governance Mode.
+* Is a `CommittedCandidateReceipt` enough or should the whole `PoV` be included? There was talk about detachable validator signatures.
```
rphmeier, line 19: `PoV` should not be included. We only store `PoV`s as erasure-coded to save on space

roadmap/implementers-guide/src/node/disputes/votesdb.md
```diff
@@ -0,0 +1,137 @@
...
+    pro: Vec<(ValidatorIndex, SessionIndex, ValidityAttestation)>,
+    /// all entries must be unique
+    cons: Vec<(ValidatorIndex, SessionIndex, ValidityAttestation)>,
+}
+```
+
+## Session Change
+
+A sweep clean is to be done to remove stored attestions
+of all validators that are not slashable / have no funds
```
rphmeier, line 94: In fact, our dispute window is somewhat smaller than the window for slashing in general. We will just remove everything older than a fixed number of sessions e.g. a day's worth.l

roadmap/implementers-guide/src/node/disputes/participation.md
```diff
@@ -0,0 +1,94 @@
...
+    /// The block is valid
+    Valid,
+    /// The block is valid
+    Invalid,
+}
+```
+
+## Session Change
+
+> TODO
```
coriolinus, line 60: Unresolved TODO.

Sketch: on a session change, mostly in other subsystems we just drop pending work. However, we can't do that in the case of disputes, because it would allow malefactors to get away with doing bad things. Therefore, we need a mechanism for outgoing validators to hand off work to incoming validators. This might be a signed transaction accumulating the data in possession of the outgoing validator.

coriolinus: A transaction would be redundant if many outgoing validators posted essentially the same VotesDB. Maybe something more analogous to PoV distribution would be more appropriate.

roadmap/implementers-guide/src/node/disputes/votesdb.md
```diff
@@ -0,0 +1,137 @@
...
+
+## Sequence
+
+### Incoming message
+
+1. Incoming request to store a dispute gossip message.
+1. Validate message.
+1. Check if the dispute already has a supermajority.
+    1. iff:
+        1. Craft and enqueue an unsigned transaction to reward/slash the validator of the proof, that is part of the incoming message, directly.
```
rphmeier, line 115: I do believe this slashing should happen automatically on-chain. e.g. The goal of the node-side should just be to provide information to the various forks of the chain, which can act on it as necessary.

roadmap/implementers-guide/src/node/disputes/participation.md
```diff
@@ -0,0 +1,94 @@
...
+
+### Blacklisting bad block and invalid decendants
+
+In case a block was disputed successfully, and is now deemed invalid.
+
+1. Query `ChainApiMessage::Descendants` for all descendants of the disputed `Hash`.
+1. Blacklist all descendants of the disputed block `ChainApiMessage::Blacklist`.
+1. Check if the dispute block was finalized
+1. iff:
+    1. put the chain into governance mode
```
coriolinus, line 93: I don't think "governance mode" has been formally defined anywhere. Pretty sure it just means "halt the chain until restarted by some entity with superuser permissions", but searching the current impl of the guide doesn't give me that definition.

roadmap/implementers-guide/src/node/disputes/participation.md
```diff
@@ -0,0 +1,94 @@
...
+1. Craft and enqueue an unsigned transaction to reward/slash the validator of the proof, that is part of the incoming message, directly.
```
coriolinus, line 94: Presumably: reward the validator of the proof, slash the party or parties identified by the tx as misbehaving?

roadmap/implementers-guide/src/node/disputes/participation.md
drahnr: There is already an API on the 
```
	pub fn unsafe_revert(
		&mut self,
		n: NumberFor<Block>,
		blacklist: bool,
	) -> sp_blockchain::Result<NumberFor<Block>> {
```

so that logic has to be made persistent, but is already there.

roadmap/implementers-guide/src/node/disputes/votesdb.md
drahnr: And that is due to the fact of the information spreads through the network (plus some generous margin) or is there another reason?

roadmap/implementers-guide/src/node/disputes/participation.md
drahnr: So far the agreement was, that disputes never cross a session change. If they do, they are restarted. This still leaves the question what can really be dropped here without losing messages / votes.

roadmap/implementers-guide/src/node/disputes/votesdb.md
drahnr: It feels odd to split that up and defer it to the runtime when all the information to do so is already present.

drahnr: But `active_heads` is not enough I guess - that covers all the local forks, for forks that are not present locally (`remote` ones) how would those be best handled? (or is that the reason to do in the runtime?)

roadmap/implementers-guide/src/node/disputes/participation.md
drahnr: I think I left a TODO for exactly that at the very end :)

roadmap/implementers-guide/src/node/disputes/participation.md
```diff
@@ -0,0 +1,106 @@
...
+
+## Remarks
+
+`UnsignedTransaction`s are OK to be used, since the inner
+element `CommittedCandidateReceipt` is verifiable.
+
+## Assumptions
+
+* If each validator gossips its own vote regarding the disputed block, upon receiving the first vote (the initiating dispute vote) there is only
+a need for a very small gossip only containing the `Vote` itself.
```
burdges, line 13: I've trouble parsing this: 

I presume you are not saying they should wait to gossip the dispute until they vote?  That'd take too long.  

Are you saying politeness is simple because they could assume their gossip partners got their previous gossiped message about the dispute?  Yes this seems reasonable.  We could've some churn in the gossip pattern though, so maybe one wants a way to update this later, although probably not anytime soon.  I'm unsure how robust politeness is on this sort of point elsewhere.

roadmap/implementers-guide/src/node/disputes/participation.md
burdges: I'd assume they'd keep working on the dispute until resolved, even if no longer technically a validator.

roadmap/implementers-guide/src/node/disputes/participation.md
```diff
@@ -0,0 +1,106 @@
...
+    1. Store our vote to the `VotesDB` via `VotesDBMessage::StoreVote`
+    1. Gossip our vote on the network.
+
+### Resolution
+
+1. Receive `DisputeParticipationMessage::Resolution`
+1. Craft an unsigned transaction with all votes cast
+  1. includes `CandidateReceipts`
+  1. includes `Vote`s
+1. Delete all relevant data associated with this dispute
```
burdges, line 93: We do not care overly if `CandidateReceipts` lives on the current chain, although if it does then we could impact the fork choice rule later.

roadmap/implementers-guide/src/node/disputes/participation.md
burdges: It likely means pausing all parachains and maybe halts transactions on the relay chain too.  In the short term, we mostly just want to minimize our own headache whenever we cause this accidentally like by some mistake in a wasm upgrade or whatever.  ;)

roadmap/implementers-guide/src/node/disputes/votesdb.md
burdges: It permits validators to clean up all the candidate blocks, sent messages, and erasure coded pieces they retain.

We need the validators to (a) support one another in checking candidates, but they're all done long before this, and (b) provide the XCMP messages sent by candidates they validated, meaning parachains must empty their sent messages within a day. 

In fact, we elsewhere argue that parachains should build a block per day or every 6 hour or something, but we never say they must empty they messages so fast, but actually this requires they do so, given we use sending checkers for message transport.

roadmap/implementers-guide/src/node/disputes/participation.md
coriolinus: Restarting disputes on session change makes sense to me; it's more compact than persisting the whole VotesDB. 

I think it's important that only validators can resolve disputes. Otherwise, we have to accept that an arbitrary node which may not have any slashable stake at all can just announce "I've resolved this dispute (you may not have heard of it, it's very indie), and we need to slash That Guy." 

More seriously, once a node falls out of the validator set, I don't know of any reason it couldn't just immediately unbond its stake, which removes what I believe to be an important part of the trust model for validators. A node without stake shouldn't be making significant network decisions IMO.

burdges: They're slashable for 28 more days.  Approval votes are part of a block from when they were a validator, so slashing for messages after the epoch end makes sense.  I agree the optics suck of course.

We constrain grandpa votes by approval votes, so yes approval votes remaining with the previous validator set creates a strange scenario where the previous session's validators control the grandpa voting behavior for the current session.   Yes, we should rediscuss this with Alistair and Rob of course :) but it's looks fine to me. 

Aside from keeping the current assignments, we've two other options:

We could simply abandon any non-approved blocks at session boundaries, meaning grandpa rolls back to the last approved block.  We dislike this because it couples too tightly with other systems, like by impacting the randomness more.
 
We could rerun assignments with the new validator set, but we need assignments to be unpredictable, so we've at least three options:
3. We include a authority set id aka session number along with the VRF inputs.  This is fine, but forces all validators to restart their validation work, making this somewhat similar to a chain rollback. 
4. We force all validators to rotate approval assignment VRF keys every session.  Although doable, it prevents the network from ever benefiting from validator operators using remote signers.  And causes work restarts like 3 so really just do 3 instead.
5. We actually preserve all assignment announcements from the previous session, except now any incoming validators have double the workload, and any assignment announcements from the previous session become invalid.  An adversary might gain an advantage in knowing which candidates loose approval checkers, but exploiting this sounds messy.

We're left with three out of these five options:
1. we keep assignments like I suggest but this interacts with grandpa and complicates your data model but gives the smoothest performance, 
3. we redo assignments entirely with fresh VRF inputs ala 3 above which requires parachains kinda pause, or
5. we keep any assignments for which the VRF yields the same result ala 5 above, which complicates our soundness analysis.  It appears both simplest and most risky.

We'll discuss this, thanks.  :) 

As an aside, all these approval assignment VRF keys should ideally be announced two sessions in advance, maybe three for sassafras, like all VRF keys, although it's much less important here than with babe/sassafras.

roadmap/implementers-guide/src/node/disputes/participation.md
drahnr: I am not saying they should not gossip _other_ votes of other validators, but a validator should cast its own vote upon reception of the first validity challenging vote.
I added the vote definition of a `Vote` here which was missing previously https://github.com/paritytech/polkadot/pull/2192/files#diff-87b3e890a1e89a1ec02e48556fa509020e0ca6c88cd447e3de9ca84f57af9647R59-R85

roadmap/implementers-guide/src/node/disputes/votesdb.md
drahnr: > I do believe this slashing should happen automatically on-chain. e.g. The goal of the node-side should just be to provide information to the various forks of the chain, which can act on it as necessary.

In other words, the node sideonly crafts unsigned transactions for each active head and lets the runtime logic (block import) do whatever the runtime deems necessary.

roadmap/implementers-guide/src/node/disputes/participation.md
drahnr: For me this is more about the mechanics, of how to halt the chain while technically staying operational such that sudo/governance decisions can still "resolve it", which implies halting the relay chain does not work, but halting all para chains should be fine™.

Given that this is very fuzzy, I'd be in favour of extracting that into a separate PR.

roadmap/implementers-guide/src/node/disputes/votesdb.md
```diff
@@ -0,0 +1,137 @@
...
+    1. Lookup all `CommittedCandidateReceipt`s
+    1. Reconstruct the `PoV` via `AvailabilityStoreMessage::QueryAvailableData`
+
+### Incoming dispute vote
+
+1. Check if the dispute already has a supermajority.
+1. iff:
+    1. store the resolution and keep the resolution
+    1. send `DisputeMessage::Resolution` to the overseer
+    1. TODO when to delete stuff from the DB?
```
drahnr, line 135: This a question on how long to retain the data in order to properly handle incoming messages _after_ the quorum is reached.

roadmap/implementers-guide/src/node/disputes/participation.md
burdges: Yes fine, they do begin their checking process whenever they discover a challenge.

We mostly keep our gossip protocols "polite" in that nodes track one anothers' knowledge and does not forward messages until we know the recipient possesses the precursor knowledge.  It's a higher level back pressure mechanism from a developers perspective that reliably bounds network bandwidth and avoids messy internal queuing or resequencing logic for handling out-of-context messages.  We design complex protocols more easily using politeness.  In fact, politeness requires attention to protocol gossip assumptions, like around messages delays, but so far it never proved too tricky there.

drahnr: > We mostly keep our gossip protocols "polite" in that nodes track one anothers' knowledge and does not forward messages until we know the recipient possesses the precursor knowledge. It's a higher level back pressure mechanism from a developers perspective that reliably bounds network bandwidth and avoids messy internal queuing or resequencing logic for handling out-of-context messages. We design complex protocols more easily using politeness. In fact, politeness requires attention to protocol gossip assumptions, like around messages delays, but so far it never proved too tricky there.

That makes sense and I did not associate that with politeness :) but rather "not-spam". Essentially this mimics the availability distribution pretty well.

roadmap/implementers-guide/src/node/disputes/participation.md
drahnr: Right, we always craft the unsigned transaction to all active heads.

roadmap/implementers-guide/src/node/disputes/votesdb.md
AlistairStewart: It doesn't quite require that they act on the messages that fast, it just means that someone on the freceiving chain has to have got them, as they can't rely on validators to get them now.

roadmap/implementers-guide/src/node/disputes/participation.md
burdges: I think politeness provides an extremely useful design pattern that simplifies the interaction between research and development, but that said...

We've considered non-polite options for announcing approval assignment VRFs as possible defenses against highly targeted DoS attacks on soundness, which ideally want all validators to learn the assignments whenever malicious ones do. 
1. Assignment announcements could be sent directly to everyone in single UDP packets, which perhaps precludes knowing that everyone saw the relay chain block, and definitely creates hiccups in the gossip protocol.  
2. An erasure coded reliable gossip protocol could ensure assignment announcements reached all validators if they reached malicious ones, which maybe could ensure knowledge, but gets noisy for such small messages, and requires slashing to prevent false message spam.
3. If we adopted the ring VRF here, then a mix network could transport assignment announcements with less noise, but adds latency, damages liveness, and is not as polite.

We might adopt 2 or 3 one day, which remain somewhat polite if messy, but that's a problem for another day.

roadmap/implementers-guide/src/node/disputes/participation.md
burdges: We've recalled our earlier decision that approval checks run from the previous session and do not restart.  Of course if a dispute happens in the new session then those new validators could slash the approval checker or backer form the previous session, which is fine.

rphmeier: I don't believe that we should restart disputes. Rather, the disputes should continue but we should always follow the same rule that approvals does of selecting the validator set based on the session of the candidate.

roadmap/implementers-guide/src/node/disputes/votesdb.md
rphmeier: Slashing is something that's computed by the chain in our code.

> In other words, the node side only crafts unsigned transactions for each active head and lets the runtime logic (on block import? or another entry point?) do whatever the runtime deems necessary such as slashing.

Yes. And these transactions should be generated based on the data in that active head - if there are live disputes, or that chain has included blocks that have been disputed elsewhere. What the chain does with those can be coded in the runtime.

roadmap/implementers-guide/src/node/disputes/participation.md
rphmeier: It's definitely a runtime concern. The way that we would implement this in practice is by disabling all parachain import from this point onwards by setting a flag in the runtime. i.e. it would happen automatically from the perspective of the node-side code.

roadmap/implementers-guide/src/node/disputes/participation.md
rphmeier: I figure we can probably just piggyback on the existing transaction gossip that exists in Substrate. If validators create regular old operational Unsigned transactions, then other nodes will gossip and include those. No need for additional messaging.

roadmap/implementers-guide/src/node/disputes/votesdb.md
rphmeier: 12 / 24 hours, i'd say.

roadmap/implementers-guide/src/node/disputes/votesdb.md
rphmeier: Just to give a little bit of intuition on slashing. Here is my epistemology:
* Slashing has to happen somewhere
* Account balances and staking ledger are stored on-chain, so slashing must happen on-chain
* Slashing on-chain must be justified - the chain must be able to interpret data to determine that some offence has occurred
* Therefore the off-chain code should be concerned with producing justification as opposed to commands to slash
* In this case, justification takes the form of statements from many validators.

roadmap/implementers-guide/src/runtime/disputes.md
```diff
@@ -19,34 +19,34 @@ We account for these requirements by having the validity module handle two kinds
 
 ## Approval
 
-We begin approval checks upon any candidate immediately once it becomes available.  
+We begin approval checks upon any candidate immediately once it becomes available.
```
rphmeier, line 22: Section on approval votes could be removed from this module

roadmap/implementers-guide/src/runtime/disputes.md
```diff
@@ -104,3 +104,83 @@ The goal of the dispute is to garner a `>2/3` (`2f + 1`) supermajority either in
...
+
+### Storage
+
+layout:
+
+The set of disputed canddidates
+
+```rust
+decl_storage! {
+  Disputed {
```
rphmeier, line 118: What is this `Disputed` type? Is it a `Map`, or a single value?

roadmap/implementers-guide/src/node/disputes/README.md
```diff
@@ -2,5 +2,48 @@
 
 Handles everything around disputes, ranging from persisting and retrieving
 votes to notifications of other peers and subsystems about the resolution
-as well as black listing invalid blocks and creating a transaction to
-persist the dispute resolution.
\ No newline at end of file
+as well as passing on information to the runtime.
+
+
+`VotesDB` is soley responsible for data persistance in a manner,
```
rphmeier, line 8:
```suggestion
`VotesDB` is solely responsible for data persistence in a manner
```

roadmap/implementers-guide/src/node/disputes/README.md
```diff
@@ -2,5 +2,48 @@
...
+that allows easy retrieval and evaluates the state of particular
+disputes and generates `StoreVoteResult`s.
+
+`Participation` uses runtime `Events` to facility information
```
rphmeier, line 12:
```suggestion
`Participation` uses runtime `Events` to facilitate information
```

roadmap/implementers-guide/src/node/disputes/README.md
```diff
@@ -2,5 +2,48 @@
...
+distribution via the transaction gossip.
+As such it queries all `DisputeEvents` on view change
+and extracts `Vote`s on which it operates on.
+
+If a new dispute is detected, it does the `PoV` retrieval and verification based on the runtime verification code and notifies
```
rphmeier, line 17: I guess this should also mention that this issues new votes based on active disputes, right?

roadmap/implementers-guide/src/node/disputes/README.md
```diff
@@ -2,5 +2,48 @@
...
+    alt double vote
+        Participation->>*: DoubleVote
+    else already present
+        Note right of Participation: nop
+    else detection
+        Note right of Participation: Cast our own vote<br/> by verifying the claim<br/> first
+    else resolution
+        Note right of Participation: Craft an event and put in an extrinsic
+    else stored
+        Note right of Participation: Craft an event and put in an extrinsic.
```
rphmeier, line 45: I don't follow what the extrinsic is here. What is it that we want each validator to submit to the chain here and above?

roadmap/implementers-guide/src/node/disputes/participation.md
```diff
@@ -104,7 +103,7 @@ Nothing to store, everything lives in `VotesDB`.
 		1. else:
 			1. iff: result is `Detection`
 				1. Request `AvailableData` via `RecoverAvailableData` to obtain the `PoV`.
-				1. Call `ValidateFromChainState` to validate the `PoV` with the runtime logic
+				1. Call `ValidateFromChainState` to validate the `PoV` with the runtime validation logic
```
rphmeier, line 106: I believe we would typically want to call `ValidateFromExhaustive` because we have requested the `AvailableData`.

roadmap/implementers-guide/src/node/disputes/votesdb.md
burdges: https://github.com/paritytech/polkadot/issues/2203

PR #2194, opened 2021-01-04
Parachains well known keys and abridged primitives

pepyakin:
This PR:

- introduces a set of `well_known_keys`, for the storage of parachains modules: Configuration, UMP, HRMP,
- Introduces two abridged versions of structs: `HostConfiguration` and `HrmpChannel` as well as reordering the fields. The trick with reordering the fields is needed so that we can safely add the fields to these structs without breaking the decoding.

We need this change so that cumulus (as well as any other PDK) would be able to reference these structures through merkle proofs. It's possible to define these structures on the cumulus side verbatim, but after discussion with @bkchr I decided to do define them here. Turned out that this way we can do some testing so that should the structure change in incompatible way the tests should hopefully catch that.

It's also worth noting that I chose to take the way of mimicking the keys generated by the decl_storage, instead of trying to store the values under fixed hardcoded keys.  The reasons for that:

- mimicking of the keys is very simple (see the code)
- one of arguments against this approach was that it's easy to inadvertently change the location of the values so that the merkle proofs are broken. However, that's not entirely true, because the changing the storage type or location is already a change that requires intervention. 
- moreover, we can (and do) test to ensure that the values reside in the expected places, because the locations of the values are determined locally (in contrast to be determined by inclusion in `construct_runtime`)
- and finally it requires less code change without resorting to these pesky unhashed storage APIs.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5382,6 +5382,7 @@ version = "0.8.27"
 dependencies = [
  "bitvec",
  "frame-system",
+ "hex-literal",
  "parity-scale-codec",
  "polkadot-core-primitives",
  "polkadot-parachain",
@@ -5393,6 +5394,7 @@ dependencies = [
  "sp-authority-discovery",
  "sp-core",
  "sp-inherents",
+ "sp-io",
  "sp-keystore",
  "sp-runtime",
  "sp-serializer",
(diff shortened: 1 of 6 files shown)


rphmeier (approved):

pepyakin:
bot merge

Issue #2186, opened 2021-01-02
Metrics: track time for messages to be placed into recipient channel


rphmeier:
I suspect we don't want to do this for every message that crosses the overseer, but probably some fraction of them. The `OverseerSubsystemContext` should carry a metrics instance and a cheap RNG (ChaCha) and tag randomly selected messages with a timer that is dropped after `route_message()` in the overseer. Over time our random sample will show how long messages take to be delivered.

rphmeier:
Tracking how long it takes from message sending to message receipt is more difficult so we shouldn't do that yet. If we have metrics on how many messages are received by each subsystem then we don't need that, IMO.

PR #2201, opened 2021-01-05
Add metrics timing message passing from OverseerSubsystemContext to Overseer::route_message

coriolinus:
Closes https://github.com/paritytech/polkadot/issues/2186.

### DIFF
--- a/node/overseer/src/lib.rs
+++ b/node/overseer/src/lib.rs
@@ -74,6 +74,7 @@ use futures::{
 	Future, FutureExt, SinkExt, StreamExt,
 };
 use futures_timer::Delay;
+use oorandom::Rand32;
 use streamunordered::{StreamYield, StreamUnordered};
 
 use polkadot_primitives::v1::{Block, BlockNumber, Hash};
@@ -100,8 +101,8 @@ const CHANNEL_CAPACITY: usize = 1024;
 const STOP_DELAY: u64 = 1;
 // Target for logs.
 const LOG_TARGET: &'static str = "overseer";
-// 1 message out of each of this many is timed.
-const MESSAGE_TIMER_METRIC_CAPTURE_RATE: u32 = 256;
+// Rate at which messages are timed.
+const MESSAGE_TIMER_METRIC_CAPTURE_RATE: f64 = 0.005;
 
 /// A type of messages that are sent from [`Subsystem`] to [`Overseer`].
 ///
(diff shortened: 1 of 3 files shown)


coriolinus:
Promoting the commit message from 0ab8594:

> This deserves a bit of explanation, as the motivating issue explicitly
> requested randomness. In short, it's hard to get randomness to compile
> for `wasm32-unknown-unknown` because that is explicitly intended to be
> as deterministic as practical. Additionally, even though it would never
> be used for consensus purposes, it still felt offputting to intentionally
> introduce randomness into a node's operations. Except, it wasn't really
> random, either: it was a deterministic PRNG varying only in its state,
> and getting the state to work right for that target would have required
> initializing from a constant.
>
> Given that it was a deterministic sequence anyway, it seemed much simpler
> and more explicit to simply select one of each N messages instead of
> attempting any kind of realistic randomness.

node/overseer/src/lib.rs
```diff
@@ -302,7 +324,83 @@ struct SubsystemInstance<M> {
...
+	tx: mpsc::Sender<MaybeTimed<ToOverseer>>,
+	metrics: Metrics,
+	capture_each: u32,
+	capture_state: u32,
+}
+
+impl<M> OverseerSubsystemContext<M> {
+	/// Create a new `OverseerSubsystemContext`.
+	///
+	/// `capture_each` determines which messages are timed, i.e. a value of 256 means that every 256th
```
drahnr, line 336: This is not a good statistical approach. In the good case (which is the case we care about most for timing I assume) a lot of messages have deterministic sequences / follow up messages. 

As such the pick of message should have a true statistical property, such as using a very cheap hash function on `MaybeTimer` creation, i.e. Marsaglia's Xorshift.

At the very least, we should pick a prime for which messages to pick.

node/overseer/src/lib.rs
```diff
@@ -302,7 +324,83 @@ struct SubsystemInstance<M> {
...
+	/// message is timed. A value of `0` has a special meaning: no messages are timed.
```
drahnr, line 337: I'd be in favour of either encoding the parameter as en enum with a `From<usize>` implementation for the common case, where the `Disabled` state is explicit. Special meaning numbers tend to become a burden when reading code in the future.

coriolinus: Resolved in ea59309: went back to the simpler `capture_rate` style, which has no special cases.

node/overseer/src/lib.rs
coriolinus: Agree about the statistics; restored a very simple RNG implementation in ea59309.

coriolinus:
Note to reviewers: the `oorandom::Rand32` implementation added in https://github.com/paritytech/polkadot/pull/2201/commits/ea593098feb8a1de368861be00aab837752551f4 is not suitable for cryptographic purposes. That should be fine because we do not use it for cryptographic purposes.

drahnr (approved):
LGTM!

coriolinus:
bot merge

PR #2206, opened 2021-01-05
Improve the `unbacked-span`

bkchr:
This ensures that we also record the first `import_statement` that will
lead to the creation of the `unbacked-span`.


rphmeier (approved):

bkchr:
bot merge

PR #2207, opened 2021-01-06
Minor refactor + doc fixes/improvements

eskimor:
By having everything peer set related depend directly on the enum the
code becomes more clear and it is also straight forward to add more
peersets/protocols as the compiler will complain if you forget to
implement parts of it.


eskimor:
Forgot something.

PR #2215, opened 2021-01-06
Make order consistent

tomaka:
Let's please make the life easier of people doing burn-ins.


bkchr (approved):
Just use https://github.com/bkchr/diener to make your life even more easier :P

PR #2218, opened 2021-01-07
Use Reserved Balances in Slots

shawntabrizi:
I am actually not 100% sure this is the best way to approach the problem.

The biggest issue I was facing here was that `repatriate_reserved` does not allow repatriation into a dead account. This means that you cannot use repatriate reserved from a bidder into a new parachain account, and from a parachain account into a new offboarding account.

As such, I needed to hack around the balances API to take the funds from one user and deposit it into another users reserved balance.

### DIFF
--- a/runtime/common/src/slots.rs
+++ b/runtime/common/src/slots.rs
@@ -25,7 +25,7 @@ use sp_runtime::traits::{
 use parity_scale_codec::{Encode, Decode, Codec};
 use frame_support::{
 	decl_module, decl_storage, decl_event, decl_error, ensure, dispatch::DispatchResult,
-	traits::{Currency, ReservableCurrency, WithdrawReasons, ExistenceRequirement, Get, Randomness},
+	traits::{Currency, ReservableCurrency, WithdrawReasons, ExistenceRequirement, Get, Randomness, Imbalance},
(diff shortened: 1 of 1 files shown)


runtime/common/src/slots.rs
```diff
@@ -659,16 +673,14 @@ impl<T: Config> Module<T> {
...
+						amount.checked_sub(&deposit_held)
 					{
-						if T::Currency::withdraw(
+						if T::Currency::reserve(
 							&para_id.into_account(),
-							additional,
-							WithdrawReasons::FEE,
-							ExistenceRequirement::AllowDeath
+							amount.max(deposit_held),
 						).is_err() {
```
shawntabrizi, line 684: The logic here actually changes, and while it makes more sense now, I am not sure how it was working before. I think there is or was a bug...

gavofyork: i don't think this is correct. it's meant to withdraw/reserve just the extra, if any, to brgin the total held for the slot up to the new bid. this new code would reserve not just the extra, but also the original amount too (which is already reserved from before).

shawntabrizi: ill have a second look. At the time, I justified things were right in my head, but might have been mistaken

PR #2222, opened 2021-01-07
misbehavior: report multiple offenses per validator as necessary

### LINKED ISSUE #1387: Improve misbehavior reports in CandidateBackingSubsystem
as noted in https://github.com/paritytech/polkadot/pull/1312#discussion_r451846093 we have to be able to slash for a _set_ of offences rather than for a _single_ offence. To do that we likely need to

* Unify the types used in subsystems and in `statement-table`.
* Improve `statement-table` so it would report a set of offences for an `AuthorityId`.

coriolinus:
Closes #1387.


burdges:
We should report multiple offenses of course.  In fact, slashes for wrong reports should be smallish and use the slashing system design, so correlation matters there.   We'll keep slashes invalid blocks at 100% so the second report cannot actually do much, but it's still good we record it obviously

node/core/backing/src/lib.rs
```diff
@@ -644,36 +641,17 @@ impl CandidateBackingJob {
...
-			self.send_to_provisioner(report).await?
+		// collect the misbehaviors to avoid double mutable self borrow issues
+		let misbehaviors: Vec<_> = self.table.drain_misbehaviors().collect();
+		for (validator_id, report) in misbehaviors {
+			self.send_to_provisioner(
+				ProvisionerMessage::ProvisionableData(
+					self.parent,
+					ProvisionableData::MisbehaviorReport(self.parent, validator_id, report)
+				)
+			).await?
```
montekki, line 654: `join_all`?

montekki (approved):

node/core/backing/src/lib.rs
coriolinus: Not sure how helpful that'll be. I'd expect the vast majority of cases to have 0 or 1 misbehaviors. Seems more work than it's worth to fan-out and fan-in again.

I was actually thinking of going the other way: bypassing `self.send_to_provisioner` and sending directly so that we didn't have to collect them into a vector. Same objection, though: for these quantities, it's probably just not worth it.

coriolinus:
bot merge

PR #2223, opened 2021-01-07
Optimizations of av-store

### LINKED ISSUE #2211: Optimize AV Store subsystem.
`av-store` initial implementation lacks optimizations and would ideally benefit from some kind of benchmarking.

Fixes requested on the  internal channel: We always only store the chunk that belongs to our validator when we store the availability data, while the availability distribution will fetch all chunks later Than we also cache nothing. We always go straight to the db. This results in recalculating the data later. And separately storing each chunk as one transaction in the db.

montekki:
Fixes #2211 

1. Compute and store all chunks at once.
2. Use LruCache for chunks
3. Also add two metrics for the sizes of pruning records.

### DIFF
--- a/node/core/av-store/src/lib.rs
+++ b/node/core/av-store/src/lib.rs
@@ -29,6 +29,7 @@ use std::time::{Duration, SystemTime, SystemTimeError, UNIX_EPOCH};
 use parity_scale_codec::{Encode, Decode};
 use futures::{select, channel::oneshot, future::{self, Either}, Future, FutureExt};
 use futures_timer::Delay;
+use lru::LruCache;
 use kvdb_rocksdb::{Database, DatabaseConfig};
 use kvdb::{KeyValueDB, DBTransaction};
 
@@ -46,6 +47,8 @@ use polkadot_subsystem::messages::{
 };
 
 const LOG_TARGET: &str = "availability";
+
+const CHUNK_CACHE_SIZE: usize = 32;
 
 mod columns {
 	pub const DATA: u32 = 0;
@@ -315,6 +318,7 @@ impl PartialOrd for ChunkPruningRecord {
 pub struct AvailabilityStoreSubsystem {
 	pruning_config: PruningConfig,
 	inner: Arc<dyn KeyValueDB>,
+	chunks_cache: LruCache<(CandidateHash, u32), ErasureChunk>,
 	metrics: Metrics,
 }
 
@@ -468,6 +472,7 @@ impl AvailabilityStoreSubsystem {
 		Ok(Self {
 			pruning_config: PruningConfig::default(),
 			inner: Arc::new(db),
+			chunks_cache: LruCache::new(CHUNK_CACHE_SIZE),
 			metrics,
 		})
 	}
@@ -477,6 +482,10 @@ impl AvailabilityStoreSubsystem {
 		Self {
 			pruning_config,
 			inner,
+			// In testing we would want to test actual deletions from DB,
+			// the values in the cache will prevent that so here cache size is
+			// set to `0`.
+			chunks_cache: LruCache::new(0),
 			metrics: Metrics(None),
 		}
 	}
@@ -975,6 +984,8 @@ fn store_chunks(
 	let mut chunk_pruning = chunk_pruning(&subsystem.inner).unwrap_or_default();
 
 	for chunk in chunks {
+		subsystem.chunks_cache.put((*candidate_hash, chunk.index), chunk.clone());
+
 		let prune_at = PruningDelay::into_the_future(subsystem.pruning_config.keep_stored_block_for)?;
 		if let Some(delay) = prune_at.as_duration() {
 			tx.put_vec(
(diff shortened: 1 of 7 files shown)


node/core/av-store/src/lib.rs
```diff
@@ -315,6 +318,7 @@ impl PartialOrd for ChunkPruningRecord {
 pub struct AvailabilityStoreSubsystem {
 	pruning_config: PruningConfig,
 	inner: Arc<dyn KeyValueDB>,
+	chunks_cache: LruCache<(CandidateHash, u32), ErasureChunk>,
```
bkchr, line 321: Why is that not per candidate?

bkchr: However I think a lru cache is a bad idea.

We should keep the chunks as long as a candidate is seen as active and not included or time out. 

The random size of 32 chunks will fail in the moment we increase the number of parachains/validators.

bkchr (changes requested):

node/core/av-store/src/lib.rs
montekki: done. Cache will be cleared on either candidate inclusion or timeout together with pruning.

node/core/av-store/src/lib.rs
```diff
@@ -477,6 +483,10 @@ impl AvailabilityStoreSubsystem {
 		Self {
 			pruning_config,
 			inner,
+			// In testing we would want to test actual deletions from DB,
+			// the values in the cache will prevent that so here cache size is
+			// set to `0`.
```
bkchr, line 488: Why can you not prune properly in tests to ensure that the cache is empty as well?

montekki: yep, you are right, this is leftover after LRU where the cache and store could diverge in their contents. Updated code to remove this

node/core/av-store/src/lib.rs
```diff
@@ -742,11 +753,11 @@ where
 
 			tx.send(result?).map_err(|_| oneshot::Canceled)?;
 		}
-		StoreChunk { candidate_hash, relay_parent, validator_index, chunk, tx } => {
```
bkchr, line 745: The `validator_index` can be removed from the message

node/core/av-store/src/lib.rs
```diff
@@ -966,54 +980,57 @@ fn store_available_data(
 }
 
 #[tracing::instrument(level = "trace", skip(subsystem), fields(subsystem = LOG_TARGET))]
-fn store_chunk(
+fn store_chunks(
 	subsystem: &mut AvailabilityStoreSubsystem,
 	candidate_hash: &CandidateHash,
-	_n_validators: u32,
-	chunk: ErasureChunk,
+	chunks: &[ErasureChunk],
```
bkchr, line 986:
```suggestion
	chunks: Vec<ErasureChunk>,
```

node/core/av-store/src/lib.rs
```diff
@@ -966,54 +980,57 @@ fn store_available_data(
...
-	let mut chunk_pruning = chunk_pruning(&subsystem.inner).unwrap_or_default();
-	let prune_at = PruningDelay::into_the_future(subsystem.pruning_config.keep_stored_block_for)?;
+		let prune_at = PruningDelay::into_the_future(subsystem.pruning_config.keep_stored_block_for)?;
+		if let Some(delay) = prune_at.as_duration() {
+			tx.put_vec(
+				columns::META,
+				&NEXT_CHUNK_PRUNING,
+				NextChunkPruning(delay).encode(),
+			);
+		}
```
bkchr, line 1004: Why is this done for each chunk? That makes no sense or?

node/core/av-store/src/lib.rs
```diff
@@ -966,54 +980,57 @@ fn store_available_data(
...
 
-	if let Some(delay) = prune_at.as_duration() {
-		tx.put_vec(
-			columns::META,
-			&NEXT_CHUNK_PRUNING,
-			NextChunkPruning(delay).encode(),
-		);
-	}
+		let pruning_record = ChunkPruningRecord {
+			candidate_hash: candidate_hash.clone(),
```
bkchr, line 1007:
```suggestion
			candidate_hash,
```

node/core/av-store/src/lib.rs
```diff
@@ -966,54 +980,57 @@ fn store_available_data(
...
+		);
+	}
 
 	tx.put_vec(
 		columns::META,
 		&CHUNK_PRUNING_KEY,
 		chunk_pruning.encode(),
 	);
 
+
```
bkchr, line 1033:
```suggestion
```

node/core/av-store/src/lib.rs
```diff
@@ -966,54 +980,57 @@ fn store_available_data(
...
 ) -> Result<(), Error> {
-	let _timer = subsystem.metrics.time_store_chunk();
+	let _timer = subsystem.metrics.time_store_chunks();
 
 	let mut tx = DBTransaction::new();
+	let mut chunk_pruning = chunk_pruning(&subsystem.inner).unwrap_or_default();
 
-	let dbkey = erasure_chunk_key(candidate_hash, chunk.index);
+	for chunk in chunks {
+		subsystem.chunks_cache.entry(*candidate_hash).or_default().insert(chunk.index, chunk.clone());
```
bkchr, line 995: This has to be moved at the end of loop and changed to
```suggestion
		subsystem.chunks_cache.entry(*candidate_hash).or_default().extend(chunks.into_iter().map(|c| (c.index, c));
```

bkchr (approved):
Some last nitpicks, otherwise looks good.

montekki:
bot merge

PR #2225, opened 2021-01-07
HRMP channel deposits

### LINKED ISSUE #1907: Hook up deposits to HRMP
Currently, the router module doesn't take any deposits since there is no machinery for that yet.

There should be an interface that allows to put down a deposit and to return that deposit (or equivalently the same amount).

pepyakin:
This PR implements a taking deposits for the HRMP channels.

Along the way this fixes two bugs that were found during the implementation as well as tests that cover this behavior.

I recommend reviewing this PR by checking the commits individually.

Closes #1907

### DIFF
--- a/parachain/src/primitives.rs
+++ b/parachain/src/primitives.rs
@@ -231,7 +231,7 @@ impl<T: Encode + Decode + Default> AccountIdConversion<T> for Id {
 /// that we use the first item tuple for the sender and the second for the recipient. Only one channel
 /// is allowed between two participants in one direction, i.e. there cannot be 2 different channels
 /// identified by `(A, B)`.
-#[derive(Clone, PartialEq, Eq, Encode, Decode, RuntimeDebug)]
+#[derive(Clone, PartialEq, Eq, PartialOrd, Ord, Encode, Decode, RuntimeDebug)]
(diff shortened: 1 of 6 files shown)


coriolinus (approved):
Looked through the commits. Everything LGTM. The only real nit I could come up with: you mention in the documentation for 3f5ab9d that it had previously been possible to violate the index invariant. AFAICT the tests added do not check that invariant. It would be nice to have a test which would have failed before 3f5ab9d, which passes now. However, that's not essential; it seems ancillary at best to this PR.

pepyakin:
Yeah, that's a good observation!

So the reason is that this case is exercised by one of the tests, IIRC, [here](https://github.com/paritytech/polkadot/pull/2225/commits/cf3e2386aca97ae0a3c54d4f6febbb0694a71ea1#diff-eb22a633f28e45a4f4b303b7d24ca298e96882492cb3267015e39b7368fc0c24R1837). I didn't add a dedicated test just to try minimize the number of tests. And in turn, the reason for that, is that I hold an opinion that it's impractical to have isolated tests that target only one path in a non trivial module, because that gives you more test code to fix should they break and they do not bring that much of speed up in diagnosing the failure.

pepyakin:
bot merge

PR #2231, opened 2021-01-08
repair Dockerfile /data path

5dao:
/data path should chmod to polkadot and link it to ./local/share/polkadot


TriplEight (approved):
Looks good, thanks!

PR #2236, opened 2021-01-08
subsystems have an unbounded channel to the overseer

rphmeier:
There is an upper bound on the number of messages that subsystems will send in actuality. It is a linear function of the number of peers, the number of validators, and the number of parachains. This will prevent any deadlocking issues we've observed so far.

### DIFF
--- a/node/overseer/src/lib.rs
+++ b/node/overseer/src/lib.rs
@@ -69,13 +69,12 @@ use std::collections::{hash_map, HashMap};
 use futures::channel::{mpsc, oneshot};
 use futures::{
 	poll, select,
-	future::BoxFuture,
-	stream::{self, FuturesUnordered},
+	future::{BoxFuture},
+	stream::{FuturesUnordered, Fuse},
 	Future, FutureExt, SinkExt, StreamExt,
 };
 use futures_timer::Delay;
 use oorandom::Rand32;
-use streamunordered::{StreamYield, StreamUnordered};
 
 use polkadot_primitives::v1::{Block, BlockNumber, Hash};
 use client::{BlockImportNotification, BlockchainEvents, FinalityNotification};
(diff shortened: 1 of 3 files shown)


node/overseer/src/lib.rs
```diff
@@ -69,13 +69,12 @@ use std::collections::{hash_map, HashMap};
 use futures::channel::{mpsc, oneshot};
 use futures::{
 	poll, select,
-	future::BoxFuture,
-	stream::{self, FuturesUnordered},
+	future::{BoxFuture},
```
drahnr, line 72:
```suggestion
	future::BoxFuture,
```

node/overseer/src/lib.rs
```diff
@@ -1767,14 +1729,19 @@ where
 fn spawn<S: SpawnNamed, M: Send + 'static>(
 	spawner: &mut S,
 	futures: &mut FuturesUnordered<BoxFuture<'static, SubsystemResult<()>>>,
-	streams: &mut StreamUnordered<mpsc::Receiver<MaybeTimed<ToOverseer>>>,
+	to_overseer: mpsc::UnboundedSender<MaybeTimed<ToOverseer>>,
```
drahnr, line 1732: Should we potentially keep the separate channels per subsystem to overseer in order to eventually being able to measure channel fill states? Just an idea.

drahnr (approved):
LGTM, one Q and a nit

node/overseer/src/lib.rs
rphmeier: @drahnr I think unless prometheus shows that this channel is really full, we don't need to do that. It'd be a simple change later on.

drahnr: Fair point.

tomaka (approved):
I'm not familiar with the code, but I think the approach of unbounded channels makes sense in this case

eskimor (approved):
Will think a bit more, but this should definitely improve things!

rphmeier:
bot merge

PR #2243, opened 2021-01-11
Adds propose parachain pallet

bkchr:


runtime/rococo/src/propose_parachain.rs
```diff
@@ -0,0 +1,383 @@
...
+		/// This requires:
+		/// - `para_id`: The id of the parachain.
+		/// - `name`: The name of the parachain.
+		/// - `validation_function`: The wasm runtime of the parachain.
+		/// - `initial_head_state`: The genesis state of the parachain.
+		/// - `validators`: Validators that will validate for the relay chain, needs to be at least one.
+		/// - `balance`: The initial balance of the parachain on the relay chain.
+		///
+		/// It will reserve a deposit from the sender account over the lifetime of the chain.
+		#[weight = 1000_000]
```
shawntabrizi, line 178:
```suggestion
		#[weight = 1_000_000]
```

runtime/rococo/src/propose_parachain.rs
```diff
@@ -0,0 +1,383 @@
...
+				validators.iter().all(|v| !active_validators.contains(v)),
+				Error::<T>::ValidatorAlreadyRegistered,
+			);
+			Proposals::<T>::iter().try_for_each(|(_, prop)|
+				if validators.iter().all(|v| !prop.validators.contains(v)) {
+					Ok(())
+				} else {
+					Err(Error::<T>::ValidatorAlreadyRegistered)
+				}
+			)?;
```
shawntabrizi, line 214: This is pretty aggressive, if I were writing a production pallet, i would try to catch this error on approval. If i approve a pallet, an the validator is already registered in the system, then it would fail. This should prevent any duplicate registrations from getting into the pallet.

shawntabrizi: Versus looking through all proposals proactively, which really has no bounds as far as I can tell.

bkchr: I see what you mean. However, even on approval we should still iterate over all approved proposals to search for validators.

I hope that this pallet never gets used for anything than for this test network, so I think it is okay to be "aggressive" :see_no_evil:

runtime/rococo/src/propose_parachain.rs
```diff
@@ -0,0 +1,383 @@
...
+		///
+		/// This also unreserves the deposit.
+		#[weight = 100_000]
+		fn cancel_proposal(origin, para_id: ParaId) {
+			let who = match EnsurePriviledgedOrSigned::<T>::ensure_origin(origin)? {
+				Either::Left(_) => None,
+				Either::Right(who) => Some(who),
+			};
+
+			Self::is_approved_or_scheduled(para_id)?;
```
shawntabrizi, line 259: This seems like it should be the last check among these checks since there is an iterator inside

shawntabrizi (approved):
I didn't look at this like a production pallet. For test pallet, at high level, seems fine.

PR #2230, opened 2021-01-08
Add an optional verifier to crowdloan

xlc:
This will allow crowd loan owner to enforce any custom policy offchain for contributors by requiring a valid signature before accept a contribution.

TODOs:
- [x] docs
- [ ] test
- [x] think about replay attack

Based on #2166

PR #2248, opened 2021-01-11
Add an optional verifier to crowdloan

xlc:
This will allow crowd loan owner to enforce any custom policy offchain for contributors by requiring a valid signature before accept a contribution.

TODOs:
- [x] docs
- [x] test
- [x] think about replay attack

Replaces #2230


runtime/common/src/crowdloan.rs
```diff
@@ -1554,7 +1658,7 @@ mod benchmarking {
 			let caller: T::AccountId = whitelisted_caller();
 
 			T::Currency::make_free_balance_be(&caller, BalanceOf::<T>::max_value());
-		}: _(RawOrigin::Signed(caller), cap, first_slot, last_slot, end)
+		}: _(RawOrigin::Signed(caller), cap, first_slot, last_slot, end, None)
```
shawntabrizi, line 1661: you need to benchmark with a signature here

runtime/common/src/crowdloan.rs
```diff
@@ -1565,7 +1669,7 @@ mod benchmarking {
 			let caller: T::AccountId = whitelisted_caller();
 			let contribution = T::MinContribution::get();
 			T::Currency::make_free_balance_be(&caller, BalanceOf::<T>::max_value());
-		}: _(RawOrigin::Signed(caller.clone()), fund_index, contribution)
+		}: _(RawOrigin::Signed(caller.clone()), fund_index, contribution, None)
```
shawntabrizi, line 1672: same here

xlc: What is the best way for me to generate a signature here for benchmarking purpose? Can't find similar examples.

shawntabrizi: Probably you can use the crypto libraries to sign the message you expect, and then verify the signature all at once.

If you want to make the benchmarking a bit more simple, you could make a benchmark that only does signature verification on its own, and then add that to all the extrinsics which have some signature verification part.

xlc: The problem is the benchmark will only work if the runtime using this pallet is configured to use the exact signing algorithm.

I guess not a major block here. So the next question is what signature type will kusama/polkadot use? `MultiSignature`?

shawntabrizi: I think SR25519 and ED25519 will be all you need here.

xlc:
I've update the code so it benchmark with the signature case, so it will overestimate for non signature case but I think that's fine.

But because crowdloan is not added to any runtime so I can't run the benchmark code to verify if my change actually works.

xlc:
Actually my benchmark code doesn't work because the signing API is not available in wasm unless `full_crypto` feature is enabled. Is there any workaround or we just need to enable `full_crypto` when benchmarking? @shawntabrizi

bkchr:
> Actually my benchmark code doesn't work because the signing API is not available in wasm unless `full_crypto` feature is enabled. Is there any workaround or we just need to enable `full_crypto` when benchmarking? @shawntabrizi

What signing api do you need?

xlc:
https://github.com/paritytech/substrate/blob/26f3d6826a181674257b8e9fc708aa04b49dc73c/primitives/core/src/ed25519.rs#L424-L425

So I can sign something and get it verified in the benchmarking code

bkchr:
Why don't you use https://github.com/paritytech/substrate/blob/26f3d6826a181674257b8e9fc708aa04b49dc73c/primitives/io/src/lib.rs#L468?

xlc:
How do I inject key into the crypto store?

bkchr:
https://github.com/paritytech/substrate/blob/26f3d6826a181674257b8e9fc708aa04b49dc73c/primitives/io/src/lib.rs#L456

xlc:
@bkchr CI failed with error:

```
---- crowdloan::benchmarking::tests::test_benchmarks stdout ----
783thread 'crowdloan::benchmarking::tests::test_benchmarks' panicked at 'No `keystore` associated for the current context!', /usr/local/cargo/git/checkouts/substrate-7e08433d4c370a21/09ba69f/primitives/io/src/lib.rs:459:14
784stack backtrace:
785   0: rust_begin_unwind
786             at /rustc/e1884a8e3c3e813aada8254edfa120e85bf5ffca/library/std/src/panicking.rs:495:5
787   1: core::panicking::panic_fmt
788             at /rustc/e1884a8e3c3e813aada8254edfa120e85bf5ffca/library/core/src/panicking.rs:92:14
789   2: core::option::expect_failed
790             at /rustc/e1884a8e3c3e813aada8254edfa120e85bf5ffca/library/core/src/option.rs:1260:5
791   3: <&mut dyn sp_externalities::Externalities as sp_io::crypto::Crypto>::ed25519_generate_version_1
792   4: std::thread::local::LocalKey<T>::with
793   5: tracing::span::Span::in_scope
794   6: sp_io::crypto::ed25519_generate_version_1
795   7: sp_io::crypto::ed25519_generate
796   8: polkadot_runtime_common::crowdloan::benchmarking::create_fund
797   9: <polkadot_runtime_common::crowdloan::benchmarking::SelectedBenchmark as frame_benchmarking::utils::BenchmarkingSetup<T>>::instance
798  10: polkadot_runtime_common::crowdloan::benchmarking::test_benchmark_create::{{closure}}
799  11: std::thread::local::LocalKey<T>::with
800  12: polkadot_runtime_common::crowdloan::benchmarking::tests::test_benchmarks
801  13: core::ops::function::FnOnce::call_once
802             at /rustc/e1884a8e3c3e813aada8254edfa120e85bf5ffca/library/core/src/ops/function.rs:227:5
```

xlc:
I have the keystone working in test but not sure if it will work in runtime benchmark.

Also another issue is in the real runtime, we will likely be using `MultiSignature` but `TestSignature` is used in mock.

But there is no appropriate interface to create signature so I am using `let sig = T::ContributionSignature::decode(&mut &sig.encode()[..]).unwrap();` but it only works for MultiSignature, not TestSignature.

Maybe we need to add some new benchmarking only method to `trait Verify`? @shawntabrizi

xlc:
Updated to only support ed25519 to make it simpler

xlc:
This is ready for review

gavofyork (approved):

gavofyork:
@shawntabrizi you ok with this?

shawntabrizi:
@xlc ive made it support MultiSigner and MultiSignature for ed25519, sr25519, and edcsa

shawntabrizi (approved):

bkchr (approved):

PR #2257, opened 2021-01-12
scheduler: handle re-scheduling around finalization correctly

rphmeier:
Alternative to #2188 .

#2123 Didn't update scheduler to be coherent with session changes happening after finalization instead of before initialization.

The main bit of weirdness I had to add was to ensure that parathreads don't get retries incremented twice on session change


runtime/parachains/src/scheduler.rs
```diff
@@ -1553,21 +1590,24 @@ mod tests {
...
+
 			// first rotation.
 			run_to_block(rotation_frequency + 1, |_| None);
 			assert_groups_rotated(1);
+			println!("before second");
 
 			// one block before second rotation.
 			run_to_block(rotation_frequency * 2, |_| None);
 			assert_groups_rotated(1);
+			println!("second");
```
pepyakin, line 1610: Do you intend keeping these or are these good to go?

runtime/parachains/src/scheduler.rs
```diff
@@ -337,6 +320,12 @@ impl<T: Config> Module<T> {
 			}
 		});
 		ParathreadQueue::set(thread_queue);
+
+		// We want to re-schedule as though it is the first block of the session.
+		let now = <frame_system::Module<T>>::block_number() + One::one();
+		<SessionStartBlock<T>>::set(now);
```
pepyakin, line 326: At this point I think we should alter the comment on `SessionStartBlock` to emphasize this subtlety, specifically that the session change is signalled at B and that B+1 is actually the first block to be executed in that session.

roadmap/implementers-guide/src/runtime/scheduler.md
```diff
@@ -206,12 +207,12 @@ No finalization routine runs for this module.
   - The core used for the parathread claim is the `next_core` field of the `ParathreadQueue` and adding `Paras::parachains().len()` to it.
   - `next_core` is then updated by adding 1 and taking it modulo `config.parathread_cores`.
   - The claim is then added to the claim index.
-- `schedule(Vec<(CoreIndex, FreedReason)>)`: schedule new core assignments, with a parameter indicating previously-occupied cores which are to be considered returned and why they are being returned.
+- `schedule(Vec<(CoreIndex, FreedReason)>, now: BlockNumber)`: schedule new core assignments, with a parameter indicating previously-occupied cores which are to be considered returned and why they are being returned.
```
pepyakin, line 210: the inclusion inherent call site is not updated

roadmap/implementers-guide/src/runtime-api/availability-cores.md
```diff
@@ -1,14 +1,14 @@
...
 
 ```rust
 fn availability_cores(at: Block) -> Vec<CoreState>;
 ```
 
-This is all the information that a validator needs about scheduling for the current block. It includes all information on [Scheduler](../runtime/scheduler.md) core-assignments and [Inclusion](../runtime/inclusion.md) state of blocks occupying availability cores. It includes data necessary to determine not only which paras are assigned now, but which cores are likely to become freed after processing bitfields, and exactly which bitfields would be necessary to make them so.
+This is all the information that a validator needs about scheduling for the current block. It includes all information on [Scheduler](../runtime/scheduler.md) core-assignments and [Inclusion](../runtime/inclusion.md) state of blocks occupying availability cores. It includes data necessary to determine not only which paras are assigned now, but which cores are likely to become freed after processing bitfields, and exactly which bitfields would be necessary to make them so.  The implementation of this runtime API should invoke `Scheduler::clear_and_reschedule(current_block_number + 1)` to ensure
```
rphmeier: I fixed this trailing sentence in a new commit

runtime/parachains/src/scheduler.rs
```diff
@@ -337,6 +327,10 @@ impl<T: Config> Module<T> {
 			}
 		});
 		ParathreadQueue::set(thread_queue);
+
+		// We want to re-schedule as though it is the first block of the session.
```
pepyakin, line 331: well now we don't want to reschedule anything here it looks like.

```suggestion
```

runtime/parachains/src/scheduler.rs
```diff
@@ -375,7 +367,10 @@ impl<T: Config> Module<T> {
 	/// Schedule all unassigned cores, where possible. Provide a list of cores that should be considered
 	/// newly-freed along with the reason for them being freed. The list is assumed to be sorted in
 	/// ascending order by core index.
-	pub(crate) fn schedule(just_freed_cores: impl IntoIterator<Item = (CoreIndex, FreedReason)>) {
+	pub(crate) fn schedule(
```
pepyakin, line 370: Need not to be a `pub(crate)`

In fact, it is not used anywhere except a single callsite in `clear_and_reschedule` and then in tests. I am wondering if we should inline this function into `clear_and_reschedule` (or the other way around) and use everywhere only one entrypoint.

Otherwise, I think it sucks a bit that we enter through `clear_and_reschedule` but test `schedule`.

pepyakin (approved):
Otherwise, the logic seems to be OK

rphmeier:
doing a local test then will merge

rphmeier:
bot merge

PR #2261, opened 2021-01-13
Add warning when dropping inclusion inherent

coriolinus:


runtime/parachains/src/inclusion_inherent.rs
```diff
@@ -202,6 +202,10 @@ impl<T: Config> ProvideInherent for Module<T> {
 					{
 						Call::inclusion(signed_bitfields, backed_candidates, parent_header)
 					} else {
+						sp_runtime::print(
+							"WARN: dropping signed_bitfields and backed_candidates because they produced \
```
coriolinus, line 206: If there's a better way to emit a log message from the runtime, I'd be glad to learn it.

runtime/parachains/src/inclusion_inherent.rs
```diff
@@ -202,6 +202,10 @@ impl<T: Config> ProvideInherent for Module<T> {
...
+							an invalid inclusion inherent. Probably this is related to a session change."
```
pepyakin, line 207: Hm, I am not yet sure if we can say if that's the case

runtime/parachains/src/inclusion_inherent.rs
pepyakin: it would've been great if we printed the error it outputs: that would allow us to tell what kind of error happened

I think we can use `frame_support::debug::print!` for that. I am not sure if dispatch error has a proper `Debug` implementation

burdges:
When does this happen?

runtime/parachains/src/inclusion_inherent.rs
rphmeier: Yeah, let's either avoid speculating in this message or give a different indication when the current block is the first block of the session (`Scheduler::SessionStartBlock`) might help.

rphmeier:
@burdges We do a sanity check when constructing the block that the inclusion inherent data actually passes the checks in the runtime to avoid accidentally creating bad blocks because of a bug in node-side code. This warning will trigger when that sanity check fails, which is a low, but unfortunately non-zero proportion of the time.

runtime/parachains/src/inclusion_inherent.rs
coriolinus: 6e5bb037

runtime/parachains/src/inclusion_inherent.rs
coriolinus: 6e5bb03 avoids speculation. eeb79ab has some (broken) work towards adding a note when the current block is the first block of the session, but it's not compiling for reasons which seem tedious to work through. I think it's not worth it, so I reverted that commit.

rphmeier (approved):

rphmeier:
@bkchr Please merge if looks OK

bkchr (approved):

Issue #2244, opened 2021-01-11
Consider cleaning up `PersistedValidationData`


pepyakin:
With shifting of the way how we obtain required information in cumulus to merkle proofs, some things became redundant in `PersistedValidationData`.

It looks right now like this:

```rust
pub struct PersistedValidationData<N = BlockNumber> {
	/// The parent head-data.
	pub parent_head: HeadData,
	/// The relay-chain block number this is in the context of.
	pub block_number: N,
	/// The relay-chain block storage root this is in the context of.
	pub relay_storage_root: Hash,
	/// The list of MQC heads for the inbound channels paired with the sender para ids. This
	/// vector is sorted ascending by the para id and doesn't contain multiple entries with the same
	/// sender.
	pub hrmp_mqc_heads: Vec<(Id, Hash)>,
	/// The MQC head for the DMQ.
	///
	/// The DMQ MQC head will be used by the validation function to authorize the downward messages
	/// passed by the collator.
	pub dmq_mqc_head: Hash,
	/// The maximum legal size of a POV block, in bytes.
	pub max_pov_size: u32,
}
```

Having only the `relay_storage_root` we can reach to every other field, including `block_number` since we now have that in the storage.

bkchr:
Just as a remark, changing this will make all parachains fail to validate and will require some upgrade/re-registration.

rphmeier:
Yes, I think if we will do this we should do it soon

pepyakin:
I am not sure if I follow: both block_number and relay_storage_root should refer to the same relay-chain block - the relay parent. Why do we need plus 1 and why don't we do it now?

rphmeier:
@pepyakin Another bug introduced by #2123, I think :). It used to be that `persisted_validation_data` would deal with the child block number.

pepyakin:
I guess that's a sign that we should rename `block_number` → `relay_parent_number`

rphmeier:
I agree

PR #2272, opened 2021-01-14
Remove TransientValidationData

### LINKED ISSUE #2085: Consider removing TransientValidationData
`TransientValidationData` is found to be not so useful as initially thought of.

It was initially conceived as a mechanism for checking the candidate outputs on the node side, but that was [replaced](https://github.com/paritytech/polkadot/pull/1842) by a runtime API. We thought that we should leave `TransientValidationData` for the time being because it may prove to be useful on the collator side, to help pass the acceptance criteria checks.

However, after taking a deeper look, it turns out that `TransientValidationData` can be used only for only a subset of these use-cases, specifically, to inform only the collator node. It cannot be used for informing the PVF/STF of a parachain, because the data is fully controlled by the collator.

For instance, the max_code_size used for upgrade cannot be passed through `TransientValidationData` since it's unauthorized and collator can send anything it wants. That allows the collator to deceive the PVF/STF to not perform upgrade by stating that `max_code_size` is 0 or 1.

On the other hand, it may be fine to pass the max pov size through `TransientValidationData`, because the collator node is fully responsible for creating a block anyway (the collator can create a block devoid of any transactions and that's totally fine) , and if the PoV of the block exceeds the maximum size the collator has an invalid block and it cannot do anything with it. It should be fine unless the PVF explicitly has some conditions that depends on such a value (which is not advisable under this construction).

There are some ways for rescuing it though. Here are a couple for completeness.

One way, we can also persist it as `PersistedValidationData`. That's basically mean merging the two and we would end up with a single `ValidationData`. That has downsides that it passes some redundant data, which the parachain is not interested. It contains all the data which doesn't change during the session. Besides that, should the collator include validation data in runtime, then we will make it available twice (once as part of PoV and then separately) (although, that can be mitigated a bit).

Second way is to pass the hash through the `TransientValidationData` through the `PersistedValidationData`. This is a bit better since it is at least persisted only once, but all other downsides stay.

For the time being, it would be better to completely avoid `TransientValidationData` and just mandate the collator to use merkle-proofs for the state the parachain is interested in is the way to go for now. As another observation, Contextual Execution may also depend on the merkle proofs.

pepyakin:
After https://github.com/paritytech/cumulus/pull/279 we don't need transient validation data for anything important (those use cases are replaced by the merkle proofs of the relay chain storage). This PR removes it.

I recommend reviewing this PR by checking the commits individually.

Related #2244 
Closes #2085

### DIFF
--- a/node/collation-generation/src/lib.rs
+++ b/node/collation-generation/src/lib.rs
@@ -32,7 +32,7 @@ use polkadot_node_subsystem::{
 	FromOverseer, SpawnedSubsystem, Subsystem, SubsystemContext, SubsystemResult,
 };
 use polkadot_node_subsystem_util::{
-	request_availability_cores_ctx, request_full_validation_data_ctx,
+	request_availability_cores_ctx, request_persisted_validation_data_ctx,
 	request_validators_ctx,
 	metrics::{self, prometheus},
 };
@@ -247,7 +247,7 @@ async fn handle_new_activations<Context: SubsystemContext>(
 			// we get validation data synchronously for each core instead of
 			// within the subtask loop, because we have only a single mutable handle to the
 			// context, so the work can't really be distributed
-			let validation_data = match request_full_validation_data_ctx(
+			let validation_data = match request_persisted_validation_data_ctx(
 				relay_parent,
 				scheduled_core.para_id,
 				assumption,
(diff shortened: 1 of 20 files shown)


coriolinus (approved):
LGTM. Seems straightforward.

rphmeier (approved):

PR #2276, opened 2021-01-15
use fill_level gauges instead of histograms

drahnr:
Minor change, in order to make the metrics more meaningful.

* queue length for `to` and `from` overseer are now `Gauges`
* collection time is reduced to `950ms` (`< 1s`, to avoid missing values on scheduling jitter)


coriolinus (approved):
Looks good, though based on the information we're already getting from the [histograms](http://grafana.parity-mgmt.parity.io/d/nfadbnBGz/overseer-queue-size?orgId=1), it looks like queue length isn't currently an issue.

PR #2277, opened 2021-01-15
perf measurements: piggyback coz on jaeger

drahnr:
Implements coz as an injection into existing jaeger spans and also using the tracing annotationats, to avoid addition code clutter.

This is an initial impl idea, to be refined.

Tool ref https://github.com/plasma-umass/coz

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -685,6 +685,16 @@ dependencies = [
 ]
 
 [[package]]
+name = "chashmap"
+version = "2.2.2"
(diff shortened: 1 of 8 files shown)


drahnr:
The time required to accumulate sufficient data is _very_ long, so we are talking weeks if not months which is maybe feasable at some point in the feature on a dedicated tracing node, but not today.

drahnr:
Jaeger also got refactored significantly in the meantime, closing.

PR #2286, opened 2021-01-18
Message and Data Type Flow Visualizations

coriolinus:
- [x] add graph of messages between subsystems
- [x] add graph of container relations `polkadot_primitives::v1` data types 
- [x] add graph of container relations between `polkadot::parachain::primitives` data types
- [ ] ~~add graph of transforms between data types~~ (i.e. Backing has a `CommittedCandidateReceipt` but sends only a `CandidateReceipt` to `Provisioner`).
- [x] Document node side data flow

Note that this PR uses mermaid for its sequence diagrams. This implies the need for all book builders to install the `mdbook-mermaid` package.

I couldn't figure out a good way to illustrate a graph of transforms between data types, so I've dropped that objective. In its place, I've added a section documenting the node side data flow from the collator to the inclusion inherent.


coriolinus:
Interested in opinions on what's the best layout available for the subsystem message flow graph.

- Concentrated horizontal: 
subsystems_and_jobs_0 generated

- Unconcentrated horizontal:
subsystems_and_jobs_0 generated(1)

- Circular:
subsystems_and_jobs_0 generated(2)

IMO the first of those reads the most cleanly, mainly because of the edge labels, but if there's a strong feeling in favor of any of the others, it's easy to change.

infinity0:
I like the first one Concentrated horizontal. Have you also tried vertical? Might be more space-efficient, since each label takes up only 1 row, and the graph would flow vertically.

infinity0:
Also - any ideas on how to keep the diagram updated when the code changes?

coriolinus:
The problem with vertical orientations is that Graphviz turns out to be catastrophically bad at routing a large number of flow arrows for our dataset in that orientation. I'd prefer it also, but it's practically unusable.

Unfortunately, there isn't a really reliable way to keep it automatically updated when code changes other than manual updates. I judged that it would take longer to write a script generating these graphs than to just do the data entry manually, and though my carpal tunnels aren't happy with that choice, I still think it was the right call from a time-efficiency standpoint.

roadmap/implementers-guide/src/node/subsystems-and-jobs.md
```diff
@@ -11,3 +11,427 @@ Since this goal of determining when to start and conclude work relative to a spe
...
+    provisioner -> cand_back    [arrowhead = "diamond", label = "GetBackedCandidates"]
+    provisioner -> chn_api      [arrowhead = "diamond", label = "BlockNumber"]
+
+    stmt_dist   -> net_brdg     [arrowhead = "onormal", label = "SendValidationMessage"]
+    stmt_dist   -> net_brdg     [arrowhead = "onormal", label = "ReportPeer"]
+    stmt_dist   -> cand_back    [arrowhead = "onormal", label = "Statement"]
+    stmt_dist   -> runt_api     [arrowhead = "onormal", label = "Request::Validators"]
+    stmt_dist   -> runt_api     [arrowhead = "onormal", label = "Request::SessionIndexForChild"]
+}
+```
```
drahnr, line 124: :+1: very nice and structured!

roadmap/implementers-guide/src/node/subsystems-and-jobs.md
```diff
@@ -11,3 +11,427 @@ Since this goal of determining when to start and conclude work relative to a spe
...
+
+        % fn circulate_statement
+        opt if we know of peers who haven't seen this message, gossip it
+            SD ->> NB: SendValidationMessage
+        end
+    end
+```
+
+But who are these `Listener`s who've asked to be notified about incoming `SignedStatement`s?
+Nobody, as yet.
```
drahnr, line 288: A possible candidate that consume this information could be the `Disputes`  module, but that's far from set in stone.

drahnr (approved):
Very nice :100:

lamafab:
The CI for the live version now runs with `mdbook-mermaid`: https://github.com/w3f/parachain-implementers-guide/commit/1307607372af4eea2351e4c95939e68560f65f97

Should be fine on this side.

PR #2291, opened 2021-01-19
Glossary updates

pepyakin:
I've noticed that some of the terms that we use widely don't have definitions in the glossary. This PR defines them.


bkchr (approved):

pepyakin:
bot merge

PR #2294, opened 2021-01-20
New slots/auctions architecture

gavofyork:
Refactor the slots/leases and auctions code

- split Slots into two pallets - Slots (which handles leases) and Auctions, separated by a trait.
- require `ParaId`s to be claimed before any bids can be made.
- allow anyone/thing to bid for any para to gain a parachain slot in the future.
- allow multiple funders to lease a parachain slot for the same parachain in different periods.
- use reserve system throughout, rather than burn/mint for the leasing.
- guarantee that the funds never leave ownership of the leaser.

TODO: 
- [x] Reenable Crowdloan tests
- [x] Finalise parachains lifecycle stuff (https://github.com/paritytech/polkadot/pull/2354)
- [x] Benchmarks:
  - [x] Slots
  - [x] Auctions
  - [x] Crowdloan
  - [x] Registrar
- [ ] Double check all Swapping Logic

Relies on:
- https://github.com/paritytech/substrate/pull/8180


runtime/common/src/slots.rs
```diff
@@ -44,257 +43,70 @@ pub trait Config: frame_system::Config {
...
-		pub AuctionCounter get(fn auction_counter): AuctionIndex;
-
-		/// Ordered list of all `ParaId` values that are managed by this module. This includes
-		/// chains that are not yet deployed (but have won an auction in the future).
-		pub ManagedIds get(fn managed_ids): Vec<ParaId>;
+		/// Amount held on deposit for each para and the original depositor.
+		///
+		/// The given account ID is responsible for registering the code and initial head data, but may only do
+		/// so if it isn't yet registered. (After that, it's up to governance to do so.)
+		pub Paras: map hasher(twox_64_concat) ParaId => Option<(T::AccountId, BalanceOf<T>)>;
```
thiolliere, line 65: this storage seems only used by crownloans, what is the reason to have it here ?
Also it seems a bit duplication of `paras_registrar::Paras` but I don't have the full picture yet.

runtime/common/src/paras_registrar.rs
```diff
@@ -52,131 +66,130 @@ pub trait Config: paras::Config + dmp::Config + ump::Config + hrmp::Config {
...
+				},
+			};
 
-			let debtor = <Debtors<T>>::take(id);
-			let _ = <T as Config>::Currency::unreserve(&debtor, T::ParathreadDeposit::get());
+			if let Some(info) = Paras::<T>::take(&id) {
+				<T as Config>::Currency::unreserve(&info.manager, info.deposit);
+			}
 
 			runtime_parachains::schedule_para_cleanup::<T>(id);
```
thiolliere, line 188: This should also clean in the PendingSwap

runtime/common/src/paras_registrar.rs
```diff
@@ -188,66 +201,91 @@ decl_module! {
 		#[weight = 0]
 		fn swap(origin, other: ParaId) {
```
thiolliere, line 202: sidenote: what if a parachain wants to swap, but the other never swap in return, should there be a mecanism to unrequest swap. Or this is consider never to happen because parachain should have trusted each over with other mecanism.

shawntabrizi: to "remove swap" you can claim a swap with yourself right?

thiolliere: ah true

runtime/parachains/src/dmp.rs
```diff
@@ -223,6 +223,12 @@ impl<T: Config> Module<T> {
 	}
 }
 
+impl<T: Config> crate::ParachainCleanup for Module<T> {
```
gavofyork, line 226: what's this for?

runtime/common/src/paras_registrar.rs
```diff
@@ -69,6 +68,9 @@ pub trait Config: paras::Config + dmp::Config + ump::Config + hrmp::Config {
 	/// Runtime hook for when a parachain and parathread swap.
 	type OnSwap: crate::traits::OnSwap;
 
+	/// Runtime hook for when a parachain should be cleaned up.
```
gavofyork, line 70: What does "cleaned up" mean here? Documentation could do to be expanded on this somewhere.

runtime/parachains/src/dmp.rs
shawntabrizi: This is to remove tight coupling between the registrar pallet and hrmp, dmp, and ump pallets.

When a parachain is deregistered, all parachain data needs to be removed at the end of the next session.

This hook calls into all the pallets which may have some data for a parachain, for example the queued messages in dmp, ump, or hrmp, and tells those pallets to clean up that data at the end of the next session.

gavofyork: Ok - might be better named `OnParachainDecommissioned`

runtime/common/src/crowdloan.rs
```diff
@@ -281,15 +277,15 @@ decl_module! {
 			ensure!(last_slot <= first_slot + 3u32.into(), Error::<T>::LastSlotTooFarInFuture);
 			ensure!(end > <frame_system::Module<T>>::block_number(), Error::<T>::CannotEndInPast);
 
-			let index = FundCount::get();
-			let next_index = index.checked_add(1).ok_or(Error::<T>::Overflow)?;
+			let manager = paras_registrar::Paras::<T>::get(index).ok_or(Error::<T>::InvalidParaId)?.manager;
```
gavofyork, line 280: prefer for this to be done via a trait.

runtime/common/src/crowdloan.rs
```diff
@@ -308,25 +304,25 @@ decl_module! {
...
 
-			T::Currency::transfer(&who, &Self::fund_account_id(index), value, AllowDeath)?;
+			CurrencyOf::<T>::transfer(&who, &Self::fund_account_id(index), value, AllowDeath)?;
 
 			let balance = Self::contribution_get(index, &who);
 			let balance = balance.saturating_add(value);
 			Self::contribution_put(index, &who, &balance);
 
-			if <slots::Module<T>>::is_ending(now).is_some() {
+			if auctions::Module::<T>::is_ending(now).is_some() {
```
gavofyork, line 325: this, too.

runtime/common/src/crowdloan.rs
```diff
@@ -505,31 +445,25 @@ decl_module! {
...
-						/// ever changes, then some sort of conversion will be needed here.
-						sub: index,
-					});
-
+				for (fund, para_id) in NewRaise::take().into_iter().filter_map(|i| Self::funds(i).map(|f| (f, i))) {
 					// Care needs to be taken by the crowdloan creator that this function will succeed given
 					// the crowdloaning configuration. We do some checks ahead of time in crowdloan `create`.
-					let result = <slots::Module<T>>::handle_bid(
-						bidder,
+					let result = auctions::Module::<T>::handle_bid(
```
gavofyork, line 457: ...and this

runtime/common/src/crowdloan.rs
```diff
@@ -270,27 +269,28 @@ decl_module! {
...
-			T::Currency::transfer(&owner, &Self::fund_account_id(index), deposit, AllowDeath)?;
-			FundCount::put(next_index);
+			CurrencyOf::<T>::transfer(&depositor, &Self::fund_account_id(index), deposit, AllowDeath)?;
 
-			<Funds<T>>::insert(index, FundInfo {
-				parachain: None,
-				owner,
+			Funds::<T>::insert(index, FundInfo {
+				parachain: index,
+				is_leased: false,
```
shawntabrizi, line 292: `is_leased` is totally unused and unmanaged

runtime/common/src/crowdloan.rs
```diff
@@ -354,101 +354,43 @@ decl_module! {
...
 			ensure_signed(origin)?;
 
-			let mut fund = Self::funds(index).ok_or(Error::<T>::InvalidFundIndex)?;
-			let parachain_id = fund.parachain.take().ok_or(Error::<T>::NotParachain)?;
+			let mut fund = Self::funds(index).ok_or(Error::<T>::InvalidParaId)?;
+			let parachain_id = fund.parachain;
+
 			// No deposit information implies the parachain was off-boarded
-			ensure!(<slots::Module<T>>::deposits(parachain_id).len() == 0, Error::<T>::ParaHasDeposit);
+			ensure!(!T::Registrar::is_registered(parachain_id), Error::<T>::ParaAlive);
```
shawntabrizi, line 371: This is not a valid check. `is_registered` can return true for a parathread, which should still allow a fund to be retired.

I think we need to also import the `Leaser` trait, and check `deposit_held` is zero

runtime/common/src/mock.rs
```diff
@@ -0,0 +1,173 @@
...
+use frame_support::dispatch::DispatchResult;
+use primitives::v1::{HeadData, ValidationCode, Id as ParaId};
+use crate::traits::Registrar;
+
+thread_local! {
+	static OPERATIONS: RefCell<Vec<(ParaId, u32, bool)>> = RefCell::new(Vec::new());
+	static PARACHAINS: RefCell<Vec<ParaId>> = RefCell::new(Vec::new());
+	static PARATHREADS: RefCell<Vec<ParaId>> = RefCell::new(Vec::new());
+	static MANAGERS: RefCell<HashMap<ParaId, Vec<u8>>> = RefCell::new(HashMap::new());
+}
```
kianenigma, line 31: parameter_types already supports this as well: 

```
parameter_types! {
    static Operations: Vec<(ParaId, u32, bool)> = Vec::new();
}

will generate both the `thead_local` stuff with SCREAM_CASE, and a struct named `Operations` that implements `Get`, and you can easily do `Operations::set(value)` to set the value of the `thread_local` value.

Comment (kianenigma) [reply runtime/common/src/mock.rs:31]: See example here: https://crates.parity.io/frame_support/macro.parameter_types.html

File: runtime/common/src/claims.rs [hunk 12]
```diff
@@ -1141,7 +1141,7 @@ mod benchmarking {
 
 	const SEED: u32 = 0;
 
-	const MAX_CLAIMS: u32 = 10_000;
+	const MAX_CLAIMS: u32 = 10;
```
shawntabrizi, line 1144: TODO: undo this

Made this change cause it made tests slow

runtime/westend/src/weights/crowdloan.rs
```diff
@@ -0,0 +1,75 @@
...
+	}
+	fn withdraw() -> Weight {
+		(120_984_000 as Weight)
+			.saturating_add(T::DbWeight::get().reads(4 as Weight))
+			.saturating_add(T::DbWeight::get().writes(4 as Weight))
+	}
+	fn dissolve() -> Weight {
+		(210_160_000 as Weight)
+			.saturating_add(T::DbWeight::get().reads(3 as Weight))
+			.saturating_add(T::DbWeight::get().writes(103 as Weight))
```
shawntabrizi, line 64: TODO: Better weight refund or have them tell us about weight

runtime/common/src/auctions.rs
```diff
@@ -0,0 +1,1229 @@
...
+/// The module's configuration trait.
+pub trait Config: frame_system::Config {
+	/// The overarching event type.
+	type Event: From<Event<Self>> + Into<<Self as frame_system::Config>::Event>;
+
+	/// The number of blocks over which a single period lasts.
+	type Leaser: Leaser<AccountId=Self::AccountId, LeasePeriod=Self::BlockNumber>;
+
+	/// The number of blocks over which an auction may be retroactively ended.
+	type EndingPeriod: Get<Self::BlockNumber>;
```
rphmeier, line 59: A comment here on configuration would be helpful. When using BABE randomness we need at least 1 full session between the time the bids are collected and the time we end the auction, otherwise the auction end is predictable. 

So if bids are collected in session S, we can use the current epoch randomness (`Babe::random`) in session S+2. This would be configured by having this value be at least `epoch_len_in_slots` blocks, which will take at least that many slots.

rphmeier: Might even need to be S+3 as the randomness for S+2 is almost or fully determined at the end of session S. Will let @burdges elaborate.

Alternatively we can change the `Babe::random` implementation to use `NextRandomness` but that has its own set of footguns.

rphmeier: https://github.com/paritytech/substrate/pull/8180 will also help with configuration.

burdges: https://github.com/paritytech/substrate/pull/8180

runtime/westend/src/weights/auctions.rs
```diff
@@ -44,13 +44,18 @@ use sp_std::marker::PhantomData;
...
 	fn bid() -> Weight {
-		(104_608_000 as Weight)
+		(101_899_000 as Weight)
 			.saturating_add(T::DbWeight::get().reads(7 as Weight))
 			.saturating_add(T::DbWeight::get().writes(4 as Weight))
 	}
+	fn on_initialize() -> Weight {
+		(69_339_808_000 as Weight)
+			.saturating_add(T::DbWeight::get().reads(14424 as Weight))
+			.saturating_add(T::DbWeight::get().writes(14421 as Weight))
```
shawntabrizi, line 59:
```suggestion
			.saturating_add(T::DbWeight::get().writes(14421 as Weight))
```

lmao wtf

gavofyork:
CC @shawntabrizi ; this delays the actual winner calculation for some indeterminate period after the auction must have ended. The actual time that the winner is calculated will depend on the randomness provider and if the `BabeRandomnessWithOneEpochDelay` is used then it'll happen at the end of the epoch following the point where the auction must have ended. Worst case it is one whole session afterwards. The benchmarking will need to take this into account.

runtime/common/src/claims.rs
shawntabrizi:
```suggestion
	const MAX_CLAIMS: u32 = 10_000;
```

shawntabrizi (approved):
Looking to merge these unaudited changes so that we and other users can start to test things locally.

PR #2305, opened 2021-01-22
WIP: Generic peer sets, for availability distribution

eskimor:
I discovered that it would actually make more sense to go
request/response with availability distribution, therefore this branch
is on hold. It is work in progress of making the network bridge generic
over peer sets, to make it easier to add more. If the need arises to have
more peer sets again, I will resurrect this branch, hopefully it won't
be too bitrotten by then ;-)

It also became clear that we will keep something like WireMessage around, but get rid of it when we build `Action`s already, so it won't complicate code any more.


rphmeier:
@eskimor is this still relevant?

eskimor:
Well, there is a chance that we need dummy peer sets for the request response protocols to ensure connections stay live. So at the moment I would absolutely not be surprised if I need to continue this work in a month or so. That being said, I can close this PR for cleanliness and just keep the branch locally, for sure.

PR #2308, opened 2021-01-22
[CI] Matrix release-notifications redux

s3krit:
This changes the matrix notification CI job to take a list of channels and iterate over them, rather than having 1 step per channel. This also means we can remove the channel 'internal IDs' from secrets (I verified on #matrix-dev:matrix.org that internal IDs are not considered secrets, and cannot be used to gain any more information than 'a channel with this ID once existed on this server, maybe'). Any additional channels we wish to send release notifications for can now easily be appended to the strategy.

@wpank if you think there are any other channels we should add to this list, let me know. It also means you can stop manually notifying the polkadot channels every release ;)

### DIFF
--- a/.github/workflows/release-bot.yml
+++ b/.github/workflows/release-bot.yml
@@ -1,4 +1,4 @@
-name: Push release notes to internal release notes channel
+name: Send new release notification to matrix channels
 on:
   release:
     types:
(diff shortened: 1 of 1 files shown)


TriplEight (approved):
LGTM, feel free to ping me to remove those channels names after you merge.

PR #2320, opened 2021-01-25
Add a `convert_payload` function to `Signed`

bkchr:


rphmeier:
Looks OK but I think it would be better to discourage the owned conversion of the payload. i.e. this encourages `statement.clone().convert_payload()`. A simple change to add an `as_ref` function on Signed and changing to `From<&Statement>` would encourage cheaper behavior.

bkchr:
> Looks OK but I think it would be better to discourage the owned conversion of the payload. i.e. this encourages `statement.clone().convert_payload()`. A simple change to add an `as_ref` function on Signed and changing to `From<&Statement>` would encourage cheaper behavior.

Done

rphmeier (approved):

rphmeier:
bot merge

PR #2126, opened 2020-12-16
Use Extensible Multiaddress in all Polkadot Runtimes

shawntabrizi:
This PR is a breaking change for the transaction version. cc @s3krit @joepetrowski 

This PR introduces the `Multiaddress` type into the Polkadot, Kusama, Westend, Cumulus, and "Test" runtimes.

You can learn more about the `Multiaddress` type here: https://github.com/paritytech/substrate/pull/7380

With this PR, all of the Polkadot runtimes are now able to more easily adopt future address formats like those provided by pallet-evm or even a future name-service pallet, or those we have not even thought of.

PR #2337, opened 2021-01-27
Bump tx version for westend and test-runtime, bump substrate

s3krit:
Bumping TX versions for westend and test-runtime required as per this PR:
https://github.com/paritytech/polkadot/pull/2126/files

@bkchr do we also need to bump Rococo's TX version due to this change? Or are we not currently bumping rococo's tx version?

In addition, bump Substrate to pull in the following change:
https://github.com/paritytech/substrate/pull/7993

Once merged, this PR will be backported to the `v0.8.28` release branch.

### DIFF
--- a/runtime/test-runtime/src/lib.rs
+++ b/runtime/test-runtime/src/lib.rs
@@ -100,7 +100,7 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
 	spec_version: 1055,
 	impl_version: 0,
 	apis: RUNTIME_API_VERSIONS,
-	transaction_version: 1,
+	transaction_version: 2,
 };
 
 /// Native version.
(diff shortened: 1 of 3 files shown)


runtime/test-runtime/src/lib.rs
```diff
@@ -100,7 +100,7 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
 	spec_version: 1055,
 	impl_version: 0,
 	apis: RUNTIME_API_VERSIONS,
-	transaction_version: 1,
+	transaction_version: 2,
```
bkchr, line 103:
```suggestion
	transaction_version: 1,
```

This is only the test runtime. No bumping required.

bkchr (approved):

PR #2342, opened 2021-01-28
use executor cache in polkadot-parachain executor

rphmeier:
The `sc_executor` cache isn't really built for this purpose, but we can reuse it.

I'm not super familiar with the lower-level executor code but my understanding is it'll allow reuse of the Wasmtime-compiled code across runs of the program. As things were implemented before, I think we'd be re-compiling the code every time we wanted to execute a parachain block.

The changes I made here only really help the child processes. It doesn't reuse any cache for in-main-process candidate validation.

This might be high in memory as we'll have a separate cache for every one of our child processes for executing the parachain code. It would be nice to find a way to share the cache among these child processes for two reasons:
  1. Less memory
  2. Pre-populate the cache based on our assignments in approval voting or our backing group rotation

### DIFF
--- a/parachain/src/wasm_executor/mod.rs
+++ b/parachain/src/wasm_executor/mod.rs
@@ -181,7 +181,6 @@ impl Default for ExecutorCache {
 }
 
 /// Validate a candidate under the given validation code.
-/// The provided cache is only used if the isolation strategy is local.
 ///
 /// This will fail if the validation code is not a proper parachain validation module.
 pub fn validate_candidate(
--- a/parachain/src/wasm_executor/mod.rs
+++ b/parachain/src/wasm_executor/mod.rs
@@ -160,8 +160,28 @@ pub enum InternalError {
 	WasmWorker(String),
 }
 
+/// A cache of executors for different parachain Wasm instances.
+///
+/// This should be reused across candidate validation instances.
+pub struct ExecutorCache(sc_executor::WasmExecutor);
+
+impl Default for ExecutorCache {
+	fn default() -> Self {
+		ExecutorCache(sc_executor::WasmExecutor::new(
+			#[cfg(all(feature = "wasmtime", not(any(target_os = "android", target_os = "unknown"))))]
+			sc_executor::WasmExecutionMethod::Compiled,
+			#[cfg(any(not(feature = "wasmtime"), target_os = "android", target_os = "unknown"))]
+			sc_executor::WasmExecutionMethod::Interpreted,
+			// TODO: Make sure we don't use more than 1GB: https://github.com/paritytech/polkadot/issues/699
+			Some(1024),
+			HostFunctions::host_functions(),
+			8
+		))
+	}
+}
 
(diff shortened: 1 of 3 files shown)


burdges:
We could use [memmap2-rs](https://github.com/RazrFalcon/memmap2-rs) to memory map files, including executable instructions.  It's the maintained version of [memmap-rs](https://github.com/danburkert/memmap-rs) and wraps linux's [mmap](https://linux.die.net/man/2/mmap) in a windows compatible way.  

It seemingly always uses `MAP_SHARED` from mmap calls except when using `map_copy`, meaning linux shared the memory if multiple processes mmap the same file, and likely windows behaves similarly.  We only run into trouble if wasmtime further modifies compiled code before or during execution, which then requires `map_copy`.

> Pre-populate the cache based on our assignments in approval voting or our backing group rotation

We should avoid fetching code over the network based upon unannounced approval assignments, but pre-fetching for all parachains works and pre-fetching for backing groups works.  In fact, these amount to the same thing if parachains only code changes only take effect in an upcoming epoch.  I'm unsure if we need or want that constraint, since presumably SPREE modules encapsulate any parameters that require a session delay.

parachain/src/wasm_executor/mod.rs
```diff
@@ -172,7 +192,12 @@ pub fn validate_candidate(
 ) -> Result<ValidationResult, ValidationError> {
 	match isolation_strategy {
 		IsolationStrategy::InProcess => {
-			validate_candidate_internal(validation_code, &params.encode(), spawner)
+			validate_candidate_internal(
+				&ExecutorCache::default(),
```
coriolinus, line 196: If we create a new default `ExecutorCache` here, how is that different from calling `WasmExecutor::new` within `validate_candidate_internal`? AFAICT the only difference is that we're creating the struct one frame deeper in the stack.

bkchr: Did you read the description of the pr? It is about a cache for the external process and not for when being in process.

coriolinus: Yes, but missed that. Makes more sense now.

parachain/src/wasm_executor/mod.rs
```diff
@@ -160,8 +160,28 @@ pub enum InternalError {
...
+			// TODO: Make sure we don't use more than 1GB: https://github.com/paritytech/polkadot/issues/699
+			Some(1024),
+			HostFunctions::host_functions(),
+			8
+		))
+	}
+}
 
 /// Validate a candidate under the given validation code.
+/// The provided cache is only used if the isolation strategy is local.
```
pepyakin, line 184: But the caller cannot provide the cache here, isn't it?

bkchr, line 184:
```suggestion
```

bkchr (approved):
As said in element, for now it should be okay, but we clearly need to improve this.

rphmeier:
bot merge

PR #2353, opened 2021-01-29
Clean up PersistedValidationData

### LINKED ISSUE #2244: Consider cleaning up `PersistedValidationData`

pepyakin:
Closes #2244

I didn't touch `max_pov_size` since it turns out that it is used for basic candidate checks and the alternative is not clear.


primitives/src/v1.rs
```diff
@@ -119,6 +119,22 @@ pub mod well_known_keys {
 		})
 	}
 
+	/// The list of inbound channels for the given para.
+	///
+	/// The storage entry stores a `Vec<ParaId>`
+	pub fn hrmp_ingress_channel_index(para_id: Id) -> Vec<u8> {
+		let prefix = hex!["6a0da05ca59913bc38a8630590f2627c1d3719f5b0b12c7105c073c507445948"];
```
rphmeier, line 126: what is this?

rphmeier (approved):

primitives/src/v1.rs
pepyakin: That's a prefix. It's generated by the `decl_storage` for [this](https://github.com/paritytech/polkadot/blob/7aa88e366239edbca7dcb4ff73527a3efdde0fab/runtime/parachains/src/hrmp.rs%23L285) storage item. This is essentially the value you get when calling `StorageMap::prefix_hash`. Since we are so upstream here this is copied verbatim. 

Perhaps, a more friendlier version would be to generate this from the components, i.e. the module name and the field name.

rphmeier: Gotcha - I think a comment explaining that would be helpful!

pepyakin: There actually is [one](https://github.com/paritytech/polkadot/blob/deea097a3e3f4580f1195a3b2ac66c1ca422bcbb/primitives/src/v1.rs#L69-L79), but above. Do you think we should comment on each `prefix`?

Maybe it's possible to extract a function and slap a comment there.

rphmeier: Oh, that's fine. It was just omitted from the diff

rphmeier:
Not sure if we should merge this yet as it means restarting Rococo again

pepyakin:
Ok, I was under impression that we are OK to merge into master because we have a separate branch for rococo-v1. Any ideas when the merge window opens?

cc @bkchr

bkchr:
If we ensure that we don't backport it for now, I don't see any reason why not to merge this. Would there be any immediate work on top of this in the next time?

primitives/src/v1.rs
```diff
@@ -360,18 +393,9 @@ pub struct PersistedValidationData<N = BlockNumber> {
 	/// The parent head-data.
 	pub parent_head: HeadData,
 	/// The relay-chain block number this is in the context of.
-	pub block_number: N,
+	pub relay_parent_number: N,
```
bkchr, line 396: Nitpick, why do we call it `relay_storage_root` and this here `relay_parent_number`? :D

bkchr (approved):

primitives/src/v1.rs
pepyakin: It seemed to me that `relay_parant_storage_number` exceeds the length that I am comfort with and on the other hand, `relay_storage_root` still conveys (as it seems to me) what it is about.

I don't mind renaming it though.

bkchr: Why not just call it `relay_number`?

pepyakin: I felt that this is confusing.

rphmeier: `relay_parent_storage_root` would be more clear. I think referencing the fact that every parachain block has a "relay chain parent" more explicitly is a good thing, so `relay_parent_number` is good.

PR #2354, opened 2021-01-30
Explicit Para Lifecycle w/ Upgrades and Downgrades

shawntabrizi:
This PR introduces an explicit lifecycle tracking for paras within the `runtime_parachains::paras` module.

Changes to parachains are delayed by a session, and as such, tracking exactly what is happening to a Para ID was not very easy.

This introduces a new storage item `ParaLifecycles` which tracks the current lifecycle of any known Para Id. This replaces the `Parathreads` storage item since that data is now redundant.

The possible lifecycle states are as follows:

```
/// The possible states of a para, to take into account delayed lifecycle changes.
pub enum ParaLifecycle {
	/// Para is new and is onboarding as a Parathread.
	OnboardingAsParathread,
	/// Para is new and is onboarding as a Parachain.
	OnboardingAsParachain,
	/// Para is a Parathread.
	Parathread,
	/// Para is a Parachain.
	Parachain,
	/// Para is a Parathread which is upgrading to a Parachain.
	UpgradingToParachain,
	/// Para is a Parachain which is downgrading to a Parathread.
	DowngradingToParathread,
	/// Parachain is being offboarded.
	OutgoingParathread,
	/// Parachain is being offboarded.
	OutgoingParachain,
}
```

Additionally, this PR introduces the workflow for upgrading a parathread into a parachain, and downgrading a parachain into a parathread.

This will be used for both the Slots/Auction module (which will perform upgrades/downgrades based on parachain auctions and offboarding slots), and also for swaps between parachains and parathreads.

### DIFF
--- a/roadmap/implementers-guide/src/runtime/paras.md
+++ b/roadmap/implementers-guide/src/runtime/paras.md
@@ -200,7 +200,8 @@ UpcomingDowngrades: Vec<ParaId>;
 * `lifecycle(ParaId) -> Option<ParaLifecycle>`: Return the `ParaLifecycle` of a para.
 * `is_parachain(ParaId) -> bool`: Returns true if the para ID references any live parachain, including
   those which may be transitioning to a parathread in the future.
-* `is_parathread(ParaId) -> bool`: Returns true if the para ID references any live parathread, including those which may be transitioning to a parachain in the future.
+* `is_parathread(ParaId) -> bool`: Returns true if the para ID references any live parathread,
+  including those which may be transitioning to a parachain in the future.
 * `is_valid_para(ParaId) -> bool`: Returns true if the para ID references either a live parathread
   or live parachain.
 * `last_code_upgrade(id: ParaId, include_future: bool) -> Option<BlockNumber>`: The block number of
--- a/roadmap/implementers-guide/src/runtime/paras.md
+++ b/roadmap/implementers-guide/src/runtime/paras.md
@@ -182,9 +182,9 @@ UpcomingDowngrades: Vec<ParaId>;
   session. Noop if para is already registered in the system with some `ParaLifecycle`.
 * `schedule_para_cleanup(ParaId)`: Schedule a para to be cleaned up at the next session.
 * `schedule_parathread_upgrade(ParaId)`: Schedule a parathread to be upgraded to a parachain. Noop
-  if `ParaLifecycle` is not `Parachain`.
+  if `ParaLifecycle` is not `Parathread`.
 * `schedule_parachain_downgrade(ParaId)`: Schedule a parachain to be downgraded to a parathread.
-  Noop if `ParaLifecycle` is not `Parachain`. `ParaLifecycle` is not `Parathread`.
+  Noop if `ParaLifecycle` is not `Parachain`.
 * `schedule_code_upgrade(ParaId, ValidationCode, expected_at: BlockNumber)`: Schedule a future code
   upgrade of the given parachain, to be applied after inclusion of a block of the same parachain
   executed in the context of a relay-chain block with number >= `expected_at`.
--- a/roadmap/implementers-guide/src/runtime/paras.md
+++ b/roadmap/implementers-guide/src/runtime/paras.md
@@ -56,10 +56,8 @@ struct ParaGenesisArgs {
 
 /// The possible states of a para, to take into account delayed lifecycle changes.
 pub enum ParaLifecycle {
-  /// Para is new and is onboarding as a Parathread.
-  OnboardingAsParathread,
-  /// Para is new and is onboarding as a Parachain.
-  OnboardingAsParachain,
+  /// A Para is new and is onboarding.
+  Onboarding,
   /// Para is a Parathread.
   Parathread,
   /// Para is a Parachain.
@@ -87,10 +85,10 @@ None                 Parathread                  Parachain
  |    (Session Delay)     |                          |
  |                        |                          |
  +----------------------->+                          |
- | OnboardingAsParathread |                          |
+ |       Onboarding       |                          |
  |                        |                          |
  +-------------------------------------------------->+
- | OnboardingAsParachain  |                          |
+ |       Onboarding       |                          |
  |                        |                          |
  |                        +------------------------->+
  |                        |  UpgradingToParachain    |
--- a/roadmap/implementers-guide/src/runtime/paras.md
+++ b/roadmap/implementers-guide/src/runtime/paras.md
@@ -1,8 +1,12 @@
 # Paras Module
 
-The Paras module is responsible for storing information on parachains and parathreads. Registered parachains and parathreads cannot change except at session boundaries. This is primarily to ensure that the number of bits required for the availability bitfields does not change except at session boundaries.
+The Paras module is responsible for storing information on parachains and parathreads. Registered
+parachains and parathreads cannot change except at session boundaries. This is primarily to ensure
+that the number of bits required for the availability bitfields does not change except at session
+boundaries.
 
-It's also responsible for managing parachain validation code upgrades as well as maintaining availability of old parachain code and its pruning.
+It's also responsible for managing parachain validation code upgrades as well as maintaining
+availability of old parachain code and its pruning.
 
 ## Storage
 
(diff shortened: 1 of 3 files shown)


runtime/parachains/src/paras.rs
```diff
@@ -412,6 +447,12 @@ impl<T: Config> Module<T> {
 			return weight;
 		}
 
+		if genesis.parachain {
```
rphmeier, line 450: We might want to check if `ParaLifecycles::contains(&id)` first here.

rphmeier (commented):
Looks fine but needs a guide change.

runtime/parachains/src/paras.rs
rphmeier: We don't do it now, and I think that's because we have an assumption that the callers of these functions don't double-register. I think that'd be a reasonable assumption to maintain, but we should document that.

shawntabrizi:
This should be ready to review

runtime/parachains/src/paras.rs
```diff
@@ -408,40 +555,128 @@ impl<T: Config> Module<T> {
...
 		UpcomingParasGenesis::insert(&id, &genesis);
+		weight = weight.saturating_add(T::DbWeight::get().writes(2));
 
-		T::DbWeight::get().reads_writes(1, 2)
+		weight
 	}
 
 	/// Schedule a para to be cleaned up at the start of the next session.
+	///
+	/// Noop if para is transitioning or already outgoing.
```
rphmeier, line 576: Can a chain avoid being cleaned up (e.g. by the auctions system) by sneakily triggering a downgrade right before the cleanup is scheduled? As their `ParaLifecycle` would be `Downgrading`.

I think that a) they can, but b) it's not a big deal, since we expect that the auctions system would trigger upgrades and downgrades on behalf of paras. Although swaps could be used to get around that:
  * Para A's slot is almost up
  * Para A and Para B decide to swap
  * A = Downgrading, B = Upgrading
  * Auctions triggers downgrade of A, which is already downgrading so no-op
  * B stays as a parachain

As long as we intermediate swaps through the auctions module, this also can be prevented

shawntabrizi: I think this is a good point, i see no reason not to allow clean up of a parachain that may be transitioning.

It would be the more "safe" thing to do

shawntabrizi: Yeah, these kinds of issues would be exactly why explicit lifecycles are needed. I think as long as the downstream pallets call the lifecycle check, we should be free of issues, but I am open to making this function more on the "safe side" and have it also switch a transitioning chain to outgoing.

roadmap/implementers-guide/src/runtime/paras.md
```diff
@@ -49,15 +53,69 @@ struct ParaGenesisArgs {
   /// True if parachain, false if parathread.
   parachain: bool,
 }
+
+/// The possible states of a para, to take into account delayed lifecycle changes.
+pub enum ParaLifecycle {
+  /// Para is new and is onboarding as a Parathread.
+  OnboardingAsParathread,
```
gavofyork, line 60: is it really worth complicating things by enabling onboarding directly to parachain? i expect that 99% of the time, teams will onboard to parathread, upgrade to parachain, downgrade to parathread and deregister the parathread. i very much doubt there will be any need (or much desire) for `OnboardingAsParachain` or `OutgoingParathread` states.

roadmap/implementers-guide/src/runtime/paras.md
```diff
@@ -1,12 +1,16 @@
 # Paras Module
 
-The Paras module is responsible for storing information on parachains and parathreads. Registered parachains and parathreads cannot change except at session boundaries. This is primarily to ensure that the number of bits required for the availability bitfields does not change except at session boundaries.
+The Paras module is responsible for storing information on parachains and parathreads. Registered
+parachains and parathreads cannot change except at session boundaries. This is primarily to ensure
+that the number of bits required for the availability bitfields does not change except at session
```
gavofyork, line 5: why would the number of bitys alter when upgrading/downgrading between parthread and parachain?

gavofyork (approved):
Seems ok. Would be nice to revisit at a later stage when we're thinking about more exotic scheduling systems (e.g. static slot scheduling where two or more parachains occupy slots in a fixed and repetitive pattern).

gavofyork:
Would it be convenient to build in parachain/thread swap logic in here now?

rphmeier:
> Would it be convenient to build in parachain/thread swap logic in here now?

@gavofyork In a way, that could be done already with the higher-level code (slots) triggering a downgrade of one and an upgrade of the other.

shawntabrizi:
@rphmeier @gavofyork yeah, this is what I did in Registrar. A swap is just a scheduled upgrade and downgrade. I think that is granular enough.

roadmap/implementers-guide/src/runtime/paras.md
shawntabrizi: This pallet has no such restrictions with the onboarding story. Those limitations will be imposed by the Registrar pallet.

rphmeier: @gavofyork It'd be helpful for testing / Rococo where we onboard teams as parachains. Otherwise, I'd request that this PR come with a change to the `ProposeParachain` pallet to onboard proposals as parathreads first.

roadmap/implementers-guide/src/runtime/paras.md
rphmeier: Every availability core gets its own bit in the bitfield. Parathreads are multiplexed over a fixed number of cores (configured per-session), whereas Parachains each get their own dedicated core. Upgrading from parathread to parachain would alter the amount of availability cores, which would alter the division of validators into groups, which could lead to consensus issues across forks.

roadmap/implementers-guide/src/runtime/paras.md
gavofyork: > This pallet has no such restrictions with the onboarding story. Those limitations will be imposed by the Registrar pallet.

Yeah, but the extra two enum items and the logic that is needed to handle them complicate things, and I don't think they're going to be needed in reality (except, apparently, for testing & Rococo).

shawntabrizi: I could also imagine direct parachain registration outside of the registrar module for system level parachains. Having the API handle that kind of scneario through root calls for example sounds relevant to me...

Please tell me if you feel strongly here, I will happily oblige, but I have made changes in this pallet independent of any parachain onboarding story we have. More as a general API backend.

roadmap/implementers-guide/src/runtime/paras.md
gavofyork, line 5:
```suggestion
that the number and meaning of bits required for the availability bitfields does not change except at session
```

roadmap/implementers-guide/src/runtime/paras.md
gavofyork: I have no burning desire to changes things *right now*, however when we eventually add a new kind of para (or two), this enum will quickly become unwieldy if the model of switching directly between any kind of para (or no para) persists, so I view a parathread-centric transition model to be the only way forward long-term.

roadmap/implementers-guide/src/runtime/paras.md
```diff
@@ -83,38 +141,73 @@ FutureCodeUpgrades: map ParaId => Option<BlockNumber>;
...
-* `is_parathread(ParaId) -> bool`: Returns true if the para ID references any live parathread.
-* `is_valid_para(ParaId) -> bool`: Returns true if the para ID references either a live parathread or live parachain.
-
-* `last_code_upgrade(id: ParaId, include_future: bool) -> Option<BlockNumber>`: The block number of the last scheduled upgrade of the requested para. Includes future upgrades if the flag is set. This is the `expected_at` number, not the `activated_at` number.
-* `persisted_validation_data(id: ParaId) -> Option<PersistedValidationData>`: Get the PersistedValidationData of the given para, assuming the context is the parent block. Returns `None` if the para is not known.
+* `schedule_para_initialize(ParaId, ParaGenesisArgs)`: Schedule a para to be initialized at the next
+  session. Noop if para is already registered in the system with some `ParaLifecycle`.
+* `schedule_para_cleanup(ParaId)`: Schedule a para to be cleaned up at the next session.
+* `schedule_parathread_upgrade(ParaId)`: Schedule a parathread to be upgraded to a parachain. Noop
+  if `ParaLifecycle` is not `Parachain`.
```
gavofyork, line 185:
```suggestion
  if `ParaLifecycle` is not `Parathread`.
```

roadmap/implementers-guide/src/runtime/paras.md
```diff
@@ -83,38 +141,73 @@ FutureCodeUpgrades: map ParaId => Option<BlockNumber>;
...
+* `schedule_parachain_downgrade(ParaId)`: Schedule a parachain to be downgraded to a parathread.
+  Noop if `ParaLifecycle` is not `Parachain`. `ParaLifecycle` is not `Parathread`.
```
gavofyork, line 187:
```suggestion
  Noop if `ParaLifecycle` is not `Parachain`. `ParaLifecycle` is not `Parachain`.
```

roadmap/implementers-guide/src/runtime/paras.md
```diff
@@ -83,38 +141,73 @@ FutureCodeUpgrades: map ParaId => Option<BlockNumber>;
...
+  where the new head was executed in the context of a relay-chain block with given number. This will
+  apply pending code upgrades based on the block number provided.
+* `validation_code_at(ParaId, at: BlockNumber, assume_intermediate: Option<BlockNumber>)`: Fetches
+  the validation code to be used when validating a block in the context of the given relay-chain
+  height. A second block number parameter may be used to tell the lookup to proceed as if an
+  intermediate parablock has been included at the given relay-chain height. This may return past,
+  current, or (with certain choices of `assume_intermediate`) future code. `assume_intermediate`, if
+  provided, must be before `at`. If the validation code has been pruned, this will return `None`.
+* `lifecycle(ParaId) -> Option<ParaLifecycle>`: Return the `ParaLifecycle` of a para.
+* `is_parachain(ParaId) -> bool`: Returns true if the para ID references any live parachain, including those which may be transitioning to a parathread in the future.
```
gavofyork, line 201:
```suggestion
* `is_parachain(ParaId) -> bool`: Returns true if the para ID references any live parachain, including
  those which may be transitioning to a parathread in the future.
```

roadmap/implementers-guide/src/runtime/paras.md
```diff
@@ -83,38 +141,73 @@ FutureCodeUpgrades: map ParaId => Option<BlockNumber>;
...
+* `is_parathread(ParaId) -> bool`: Returns true if the para ID references any live parathread, including those which may be transitioning to a parachain in the future.
```
gavofyork, line 203:
```suggestion
* `is_parathread(ParaId) -> bool`: Returns true if the para ID references any live parathread,
  including those which may be transitioning to a parachain in the future.
```

runtime/parachains/src/paras.rs
```diff
@@ -92,6 +92,73 @@ enum UseCodeAt<N> {
...
+	OnboardingAsParachain,
+	/// Para is a Parathread.
+	Parathread,
+	/// Para is a Parachain.
+	Parachain,
+	/// Para is a Parathread which is upgrading to a Parachain.
+	UpgradingToParachain,
+	/// Para is a Parachain which is downgrading to a Parathread.
+	DowngradingToParathread,
+	/// Parachain is being offboarded.
```
gavofyork, line 110:
```suggestion
	/// Parathread is being offboarded.
```

runtime/parachains/src/paras.rs
```diff
@@ -92,6 +92,73 @@ enum UseCodeAt<N> {
...
+	OutgoingParathread,
+	/// Parathread is being offboarded.
```
gavofyork, line 112:
```suggestion
	/// Parachain is being offboarded.
```

roadmap/implementers-guide/src/runtime/paras.md
```diff
@@ -49,15 +53,69 @@ struct ParaGenesisArgs {
...
+  OnboardingAsParachain,
+  /// Para is a Parathread.
+  Parathread,
+  /// Para is a Parachain.
+  Parachain,
+  /// Para is a Parathread which is upgrading to a Parachain.
+  UpgradingToParachain,
+  /// Para is a Parachain which is downgrading to a Parathread.
+  DowngradingToParathread,
+  /// Parachain is being offboarded.
```
gavofyork, line 71:
```suggestion
  /// Parathread is being offboarded.
```

roadmap/implementers-guide/src/runtime/paras.md
```diff
@@ -49,15 +53,69 @@ struct ParaGenesisArgs {
...
+  OutgoingParathread,
+  /// Parathread is being offboarded.
```
gavofyork, line 73:
```suggestion
  /// Parachain is being offboarded.
```

runtime/parachains/src/paras.rs
```diff
@@ -300,21 +366,68 @@ impl<T: Config> Module<T> {
 	fn apply_incoming(parachains: &mut Vec<ParaId>) {
 		let upcoming = <Self as Store>::UpcomingParas::take();
 		for upcoming_para in upcoming {
+			let state = match ParaLifecycles::get(&upcoming_para) {
+				Some(ParaLifecycle::OnboardingAsParachain) => ParaLifecycle::OnboardingAsParachain,
+				Some(ParaLifecycle::OnboardingAsParathread) => ParaLifecycle::OnboardingAsParathread,
+				_ => continue,
+			};
+
 			let genesis_data = match <Self as Store>::UpcomingParasGenesis::take(&upcoming_para) {
```
gavofyork, line 375: why is the information contained in here being duplicated in `ParaLifecycle`? Why not collapse both onboarding enum items into just `ParaLifecycle::Onboarding` and let `UpcomingParasGenesis` determine whether it is a parachain/parathread?

roadmap/implementers-guide/src/runtime/paras.md
rphmeier: @gavofyork That's reasonable, especially as you mention other ways that parachains might share cores in the future. However, anything we do there would need to come with significant changes in the `Scheduler` module.

rphmeier (approved):

runtime/parachains/src/paras.rs
```diff
@@ -52,6 +52,9 @@ pub trait Config: frame_system::Config + configuration::Config {
 		+ Into<result::Result<Origin, <Self as Config>::Origin>>;
 }
 
+// Wait until the session index is 2 larger then the current index to apply any changes.
+const SESSION_DELAY: SessionIndex = 2;
```
rphmeier: Can you add some elaboration here or link to somewhere that does elaborate more?

shawntabrizi: targeting this for another PR

runtime/parachains/src/paras.rs
shawntabrizi: I have collapsed the onboarding state 👍

shawntabrizi:
There is more that needs to be updated in future PRs, but this is a reasonable checkpoint i think for the direction we are headed.

Some more notes on what is next here: https://hackmd.io/XDkKzaYMSYumhNWzRMEvCQ

rphmeier (approved):

PR #2360, opened 2021-02-02
Paras Module Directly Cleans Up Outgoing Parachain Data

shawntabrizi:
I noticed while reviewing all the code around parachain offboarding that each of the messaging pallets had their own offboarding queue which would would trigger a parachain cleanup on the next session.

This was really strange to me since we need to basically maintain a copy of the same queue logic in each pallet... especially if there was ever a situation where a parachain was placed in offboarding, but then later would be removed from the offboarding queue, it would be a lot more work to make sure there isn't any state issue.

This PR introduces the `ParachainCleanup` trait which can be used by any pallet to clean up some data about a Parachain when it is being offboarded by the paras module.

As a result of this change, the Paras module is the definitive source of the outgoing parachains queue, and the trigger to remove data for the parachain will happen exactly when the paras module actually cleans up the data about the parachain.

Thus, there is only one source of the queue and there is no chance to be some kind of synchronization issue among all the pallets.

### DIFF
--- a/runtime/test-runtime/src/lib.rs
+++ b/runtime/test-runtime/src/lib.rs
@@ -463,6 +463,7 @@ impl parachains_session_info::Config for Runtime {}
 
 impl parachains_paras::Config for Runtime {
 	type Origin = Origin;
+	type ParachainCleanup = (Dmp, Hrmp, Ump);
 }
 
 impl parachains_dmp::Config for Runtime {}
@@ -522,6 +523,9 @@ construct_runtime! {
 		Scheduler: parachains_scheduler::{Module, Call, Storage},
 		ParasSudoWrapper: paras_sudo_wrapper::{Module, Call},
 		SessionInfo: parachains_session_info::{Module, Call, Storage},
+		Dmp: parachains_dmp::{Module, Call, Storage},
+		Ump: parachains_ump::{Module, Call, Storage},
+		Hrmp: parachains_hrmp::{Module, Call, Storage},
 
 		Sudo: pallet_sudo::{Module, Call, Storage, Config<T>, Event<T>},
 	}
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5712,6 +5712,7 @@ dependencies = [
  "frame-system",
  "futures 0.3.12",
  "hex-literal",
+ "impl-trait-for-tuples 0.2.0",
  "libsecp256k1",
  "log",
  "pallet-authority-discovery",
--- a/roadmap/implementers-guide/src/runtime/dmp.md
+++ b/roadmap/implementers-guide/src/runtime/dmp.md
@@ -4,14 +4,6 @@ A module responsible for Downward Message Processing (DMP). See [Messaging Overv
 
 ## Storage
 
-General storage entries
-
-```rust
-/// Paras that are to be cleaned up at the end of the session.
-/// The entries are sorted ascending by the para id.
-OutgoingParas: Vec<ParaId>;
-```
-
 Storage layout required for implementation of DMP.
 
 ```rust
--- a/roadmap/implementers-guide/src/runtime/hrmp.md
+++ b/roadmap/implementers-guide/src/runtime/hrmp.md
@@ -4,14 +4,6 @@ A module responsible for Horizontally Relay-routed Message Passing (HRMP). See [
 
 ## Storage
 
-General storage entries
-
-```rust
-/// Paras that are to be cleaned up at the end of the session.
-/// The entries are sorted ascending by the para id.
-OutgoingParas: Vec<ParaId>;
-```
-
 HRMP related structs:
 
 ```rust
--- a/roadmap/implementers-guide/src/runtime/ump.md
+++ b/roadmap/implementers-guide/src/runtime/ump.md
@@ -4,14 +4,6 @@ A module responsible for Upward Message Passing (UMP). See [Messaging Overview](
 
 ## Storage
 
-General storage entries
-
-```rust
-/// Paras that are to be cleaned up at the end of the session.
-/// The entries are sorted ascending by the para id.
-OutgoingParas: Vec<ParaId>;
-```
-
 Storage related to UMP
 
 ```rust
--- a/runtime/common/src/paras_registrar.rs
+++ b/runtime/common/src/paras_registrar.rs
@@ -470,6 +470,7 @@ mod tests {
 
 	impl paras::Config for Test {
 		type Origin = Origin;
+		type ParachainCleanup = ();
 	}
 
 	impl configuration::Config for Test { }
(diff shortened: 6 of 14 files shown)


runtime/test-runtime/src/lib.rs
```diff
@@ -522,6 +523,9 @@ construct_runtime! {
 		Scheduler: parachains_scheduler::{Module, Call, Storage},
 		ParasSudoWrapper: paras_sudo_wrapper::{Module, Call},
 		SessionInfo: parachains_session_info::{Module, Call, Storage},
+		Dmp: parachains_dmp::{Module, Call, Storage},
+		Ump: parachains_ump::{Module, Call, Storage},
+		Hrmp: parachains_hrmp::{Module, Call, Storage},
```
shawntabrizi, line 528: I assume someone just forgot to add these pallets to the test runtime

pepyakin:
I do welcome removing duplication of data and logic. However:

- Such a callback based approach requires the runtime to specify the order of clean-up. It is unnecessary and will involve duplication.
- There is now a bit of uncertainty from the perspective of the cleaned up module w.r.t. control-flow. Look for example at the prior version of the session change logic in the HRMP module. Then there was a straight sequence of state changes. Now, while the sequence of events is the same, it's another place.

I personally do not see this as a big improvement to the status quo and I would prefer something more explicit, e.g. paras returns from it's new session handler the list of paras that were offboarded and then that is passed to the session change handlers of the modules that are interested in the change (or something along these lines). 

Now, I do understand that the callback approach may be useful in future for something (e.g. clean data from some 3rd party module when a para offboards), but I am not a fan of such kind of reasoning because YAGNI and I don't buy this argument until the there are concrete use-cases (and still that doesn't change much for my reasoning above).

I wonder if @rphmeier has to say about this.

pepyakin (commented):
I do welcome removing duplication of data and logic. However:

- Such a callback based approach requires the runtime to specify the order of clean-up. It is unnecessary and will involve duplication.
- There is now a bit of uncertainty from the perspective of the cleaned up module w.r.t. control-flow. Look for example at the prior version of the session change logic in the HRMP module. Then there was a straight sequence of state changes. Now, while the sequence of events is the same, it's another place.

I personally do not see this as a big improvement to the status quo and I would prefer something more explicit, e.g. paras returns from it's new session handler the list of paras that were offboarded and then that is passed to the session change handlers of the modules that are interested in the change (or something along these lines). 

Now, I do understand that the callback approach may be useful in future for something (e.g. clean data from some 3rd party module when a para offboards), but I am not a fan of such kind of reasoning because YAGNI and I don't buy this argument until the there are concrete use-cases (and still that doesn't change much for my reasoning above).

I wonder if @rphmeier has to say about this.

runtime/test-runtime/src/lib.rs
pepyakin: Well, yeah, why not.

rphmeier:
We've avoided this trait-based callback architecture because it leads to spaghetti. I consider it an anti-pattern. Except in cases where we don't know what upstream code will do. i.e. `Session`, `Staking` or other pluggable modules.

However, that's not the case for Polkadot's parachains - everything in the `runtime-parachains` really makes up one module, that we've broken down into disparate pieces for better readability and testability. As we have and can define the clear execution order between modules, we shouldn't be using the callback trait pattern and should instead do something that is more explicitly aware of the interactions between the code.

shawntabrizi:
Sure, we do not need to use the callback pattern established here, but the main issue is that you have literally 4 different tracking queues across 4 storage items which all need to be in sync and describe the outgoing parachains.

If you want to treat them all as one module, there should be exactly 1 storage item in Paras, which has the list of outgoing parachains, and all the other modules use that.

rphmeier:
> If you want to treat them all as one module, there should be exactly 1 storage item in Paras, which has the list of outgoing parachains, and all the other modules use that.

The problem is initialization order. If paras is triggered before module X, then the outgoing may be cleared before other modules even get a chance to look at it.

shawntabrizi:
I really might be missing some key pieces of architecture design here, but this is exactly why I think a PR like I have done here would be helpful.

End of the day, the Paras module is the one which has ultimately say in when a parachain is actually offboarded. It may be that many modules in the future, outside of your `runtime_parachains` super module will have some parachain specific data which should be cleaned up upon the offboarding of a parachain. The message modules are a clear example, but you can imagine some kind of friendly parachain name registry service pallet or something else.

Allowing Paras to call into any number of other hooks telling those modules to clean up data is a very natural pattern in my eyes... How, given the current architecture, would you plan to support other pallets which may want to also clean up data when a parachain is removeed? Is that just not a concern for the moment?

But again, I may have the "substrate" mindset here, if you feel my comments still have not convinced you of my direction, feel free to close this PR. I dont see it as a major issue in the short term either way, just an observation made during a code review session.

rphmeier:
I think the direction of your PR is fine, and it seems like a reasonable refactoring, but we should do it without a trait. This code is going to need to support a lot of value and the hook traits are an anti-pattern for reasons I mentioned before. I wish we didn't use them nearly as often in FRAME either. I didn't even mention the possibility for cyclical dependency earlier, which brings in even more complexity. Take a look at almost any module in FRAME and see if you can actually pinpoint the invariants of the functions w.r.t. session boundaries or module initialization. This is a potential source of bugs that, although likely fine here, I would like to categorically avoid, and it doesn't seem that you've actually done the leg-work of thinking about initialization order, which is exactly why I'd like to avoid it.

We're building a product, not a library, so we don't have to worry about generality and reusability by other projects. If we need to trigger more things to happen when a para gets deregistered or downgraded or anything, we can just do it directly.

shawntabrizi:
@rphmeier I need a change like this for making https://github.com/paritytech/polkadot/pull/2406 work correctly...

but cannot for the life of me think of a clever way to do this without using a trait.

The problem is ultimately cyclic dependencies.

We want `paras` to be the controller of this cleanup action... but we cannot have `paras` depend on `hrmp` because `hrmp` is already dependent on `paras`.... so what are the choices?

pepyakin:
I wrote in my first answer a strawman solution how we can do this.

> I personally do not see this as a big improvement to the status quo and I would prefer something more explicit, e.g. paras returns from it's new session handler the list of paras that were offboarded and then that is passed to the session change handlers of the modules that are interested in the change (or something along these lines).

What I referred there, is something like this. The initializer invokes the session change handlers. Paras is one of the first modules there. From it's session handler it could return a set of paras that were offboarded and require cleanup. Each module that needs that information receives it as a parameter to its session change handler.

It's not that I am saying how this should be done. Rather, it's just an idea how it could be done. As you can see, there is no cyclic dependency.

rphmeier:
> I personally do not see this as a big improvement to the status quo and I would prefer something more explicit, e.g. paras returns from it's new session handler the list of paras that were offboarded and then that is passed to the session change handlers of the modules that are interested in the change (or something along these lines).

@shawntabrizi ^ This!

Use the initializer!

shawntabrizi:
closed in favor of #2413

PR #2366, opened 2021-02-02
Expose validation specific host function.

cheme:
This PR is required for https://github.com/paritytech/cumulus/pull/295

It exposes a runtime interface 'Validation'. It only contains a function 'validation_code' that return the wasm code use in a validation context.

This is use in https://github.com/paritytech/cumulus/pull/295 to insert the validation code in place of the runtime code in proof (note that this only work with cumulus because currently cumulus uses the same wasm for validation and runtime).


parachain/src/wasm_executor/validation_host.rs
```diff
@@ -169,22 +203,30 @@ pub fn run_worker(mem_id: &str) -> Result<(), String> {
...
-			let mut slice = memory.wlock_as_slice(0)
-				.map_err(|e| format!("Error locking shared memory: {:?}", e))?;
-
 			let result = {
-				let data: &mut[u8] = &mut **slice;
-				let (header_buf, rest) = data.split_at_mut(1024);
+				// Locking as read only to allow access to code memory from
+				// host validation extension.
+				// Note that this only work if two process are communicating only.
+				// Would be racy otherwhise.
```
cheme, line 210: I am unsure of this assertion: I did switch the lock to read to be able to acces the wasm slice directly, but this would be racy if the shared memory is consumed by multiple client.
CC\ @arkpar

bkchr: I don't understand why you are doing all this, when we read the code anyway.

cheme: The code is passed as a slice and the validation host function uses an extension which require 'static lifetime, so I cannot use a slice in it.

bkchr: I think it is better to clone than doing the magic here.

cheme: I can also try  to access it with 'let validation_code = runtime_api_request( ctx, descriptor.relay_parent, RuntimeApiRequest::ValidationCode( descriptor.para_id, assumption, code_tx, ), code_rx, ).await?;', not sure it if it is accessible.

arkpar: Multiple parallel validation are supported, but each process is run in a separate instance of `ValidationHost`. So aliasing code memory slice should be safe.

cheme:
Closing, I am reverting cumulus PR to only compact proof (I remove the value skip to keep thing simple, so no need for this pr).

cheme:
This branch and cumulus pre refactor https://github.com/paritytech/cumulus/tree/d300493f36150011437fe4e65b5eb25453eb6d85 and trie https://github.com/paritytech/trie/pull/126 .
Could probably be a start in https://github.com/paritytech/polkadot/issues/3310 .

Here the point was to avoid the runtime code in pov, which was done by removing from proof and allowing access from validation function (current).
For 3310, the validation code from inherent would replace the one here (current become next).

PR #2369, opened 2021-02-03
Fix bug in statement table

bkchr:
We only checked all validity votes, ignoring invalid votes. So, the
condidition could not hold. Besides fixing the panic, I removed some old
cruft that isn't required anymore.

Fixes: https://github.com/paritytech/polkadot/issues/2368

### DIFF
--- a/statement-table/src/generic.rs
+++ b/statement-table/src/generic.rs
@@ -267,40 +267,26 @@ impl<Ctx: Context> CandidateData<Ctx> {
 			Ctx::GroupId, Ctx::Candidate, Ctx::AuthorityId, Ctx::Signature,
 		>>
 	{
-		if self.can_be_included(validity_threshold) {
-			let validity_votes: Vec<_> = self.validity_votes.iter()
-				.filter_map(|(a, v)| match *v {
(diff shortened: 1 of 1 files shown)


statement-table/src/generic.rs
```diff
@@ -267,40 +267,26 @@ impl<Ctx: Context> CandidateData<Ctx> {
...
-			let validity_votes: Vec<_> = self.validity_votes.iter()
-				.filter_map(|(a, v)| match *v {
-					ValidityVote::Invalid(_) => None,
-
-					ValidityVote::Valid(ref s) =>
-						Some((a, ValidityAttestation::Explicit(s.clone()))),
-					ValidityVote::Issued(ref s) =>
-						Some((a, ValidityAttestation::Implicit(s.clone()))),
-				})
-				.take(validity_threshold)
```
bkchr, line 280: If someone comes up with a good explanation why we only want to have `validity_threshold` votes forwarded, please tell it. Otherwise, I don't see any reason to restrict this.

rphmeier: legacy

rphmeier (approved):

PR #2374, opened 2021-02-03
[Burn-in] Test wasmtime

haikoschol:


athei:
Do we need to add labels so that `gitlab-check-labels` succeeds in order to start the burnin?

bkchr:
Burnin is already running.

bkchr:
Done

PR #2377, opened 2021-02-03
Refactored the build steps for building with cargo

saki-osive:
Hi, the steps for `Install via Cargo` at **https://github.com/paritytech/polkadot#install-via-cargo** are incomplete in the regard that they miss important pre-requisites for getting the project up and running with cargo.

This PR addresses this.

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,12 +67,19 @@ dnf install polkadot
 
 ### Install via Cargo
 
+Before building, make sure you have the following installed on your system :
+
+- The `build-essential` package (install with ` apt install build-essential`)
+- Nightly Rust version (set it with `rustup override set nightly`)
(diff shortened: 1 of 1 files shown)


bkchr:
These steps are documented below your changes. So, this change isn't required.

saki-osive:
@bkchr So, shouldn't they be earlier so that the user knows that they're needed. I'm talking specifically about the Cargo install.

bkchr:
You could add a comment that the prerequisites from below need to be installed.

saki-osive:
Yes, for the `build-essential` package, it can be added, but some dependencies lie in the scripts of the code which need to be manually installed in the **Cargo build** method. So they have to be explicitly added, further, it's better the Cargo build method is self-sufficient in itself and can be followed from the get-go.

Issue #2300, opened 2021-01-21
Parachain Configuration and Parachain Registration must wait at least 1 full session


rphmeier:
Due to BABE, the validators and randomness of a session are already fixed in advance by one session.
We would like the validator groups to obey the same constraint, as it simplifies the process of tracking the state of parachain consensus of many forks across session changes, as done in the Approval Voting subsystem or in networking code that seeks to connect to the validators of the current and next session. Having consistency across sessions in small forks is desirable for simplicity, as we can expect a single longest chain to prevail within one session's time, as is assumed within BABE.

For this to work, we need to have fixed the number of Availability Cores in the scheduler, as this dictates the splitting of validators into groups. This is determined by the combination of the amount of registered parachains and configuration values - the number of parathread multiplexers, the maximum number of validators per core, and the maximum number of validators to be used in parachain consensus as a whole. We do not want any of these values, or indeed other values within `SessionInfo` to vary across forks.

To more clearly restate the problem, we would like to have the property that two blocks `B_a` and `B_b`, which each initiate a session `S + 1`, both descending from a block `B_s` which initiates session `S`, have the same parachain consensus state for `S + 1` - specifically, the validator groups, the validators' keys themselves, the number of cores, the parachains assigned to each core, among others.

This is clearly not the case under the status quo. Imagine that `B_a` and `B_b` differ in their chain only by their parent block `B_k` and `B_k'`, respectively. In `B_k'` if a new parachain is registered or a configuration change is made, it may affect the number of cores or the number of validators selected for each core, leading to an inconsistency that complicates the node-side code more than it is already.

```
          /--- B_k  -|- B_a
B_s .....
          \--- B_k' -|- B_b
                     |
                 session change
```

The solution is to have parachain registration and configuration updates take at least one full session. More specifically, it means that if the changes are submitted at any time during session `S`, then the changes will take effect at the start of session `S + 2`. The actual duration is somewhere between 1 and 2 full sessions, depending on at what point the change is submitted.

Some notes on parathreads:
  * Registering a parathread is OK at any time, as long as it's at least a full session before the parathread can submit any claims. In practice, it would be unwieldy to try and track this at the level of the `Scheduler` module where claims are submitted instead of the level of the `Paras` module where parathreads are submitted for registration.
  * Upgrading a parathread to a parachain would have to undergo the same "at least one session", as a parachain gets a reserved core.
  * De-registering a parachain would require the same.

[...]

PR #2406, opened 2021-02-09
Session Delayed Para Changes / Actions Queue

shawntabrizi:
Fixes: https://github.com/paritytech/polkadot/issues/2300

Design Notes: https://hackmd.io/XDkKzaYMSYumhNWzRMEvCQ

This PR introduces a full session delay to any changes to parachains and parathreads in the para module and any changes to the parachains runtime configuration.

Changes:

* Shared Module for tracking the current session index and safe session to schedule changes.
* Remove multiple different queues in favor of a single `ActionQueue` which tracks all paras that need to be updated at the start of a specific session.
* `ActionQueue` is a map from `SessionIndex -> Vec<ParaId>` so we can schedule the change of a para for a future session.
* Actions are queued for `current_session + 2` session index in the future, solving the problem in issue 2300.
* Cannot queue any other changes to a para when there is already a change in the queue. As such, any para scheduling now returns a result to handle the case where we are not allowed to schedule some task.
* Configurations are queued for future session.


shawntabrizi:
@rphmeier and @pepyakin 

What should be considered for `is_valid_para` given the delayed state transitions?

For example, if a para is being offboarded, should they immediately not be considered a valid para? How about in the context of HRMP channel registration?

pepyakin:
`is_valid_para` is used exclusively in HRMP. The primary reason why it is there has to do with pruning of stale items such as channels.

So channels are opened on the nearest session boundary. If during that moment it turns out that the channel is offboarded (and HRMP's session change handler gets control after Paras') the request is declined. If this check wasn't there then the channel will be left as junk, because there will be no further signals that the para is offboarded. 

The second place where we check that is before accepting a channel open request, which just makes sure that we don't open a channel to a non-existent para, but it is kind of optional because the recipient may request offboarding anyway later in which case the first check will handle this.

I think as long as we get `is_valid_para(P) = true` for the cases when we can expect a further notification of removal is fine.

But now I am more worried about another aspect. Right now, we foresee a possibility that parachains and parathread may differ in configuration. For instance, a parachain can have N channels and a parathread can have only M channels, where N > M. For HRMP it doesn't matter that much in practice since we limit them as much as possible and do not make a difference between parachains and parathreads, but for XCMP that may become a problem.

rphmeier:
> For instance, a parachain can have N channels and a parathread can have only M channels, where N > M. For HRMP it doesn't matter that much in practice since we limit them as much as possible and do not make a difference between parachains and parathreads, but for XCMP that may become a problem.

I think the `Initializer` should notify modules of paras deregistering, upgrading, downgrading, etc. and we can handle it in those modules. We could just cut off all channels when swapping for instance, or just cut off `N - M` when downgrading

rphmeier:
This should not close #2300 unless it does something similar for `Configuration` updates. (disclaimer: haven't looked)

shawntabrizi:
I am happy to update for `PendingConfig` too.

Probably best to do it in a follow up PR unless you want me to make this PR larger.

shawntabrizi:
@rphmeier should be good to go and completely close #2300

shawntabrizi:
One issue with this PR as it currently stands is that it is possible that a parachain cannot be queued to be offboarded. It is possible to imagine that some combinations of states could be used to grief the parachains system and prevent an offboarding.

That being said, we will always return the result of trying to schedule offboarding, so it can always be handled, but maybe we need to create a more low level API to make sure that a parachain will be offboarded in the future.

runtime/parachains/src/session_info.rs
```diff
@@ -218,24 +225,14 @@ mod tests {
 	}
 
 	fn session_changes(n: BlockNumber) -> Option<SessionChangeNotification<BlockNumber>> {
-		match n {
-			100 => Some(SessionChangeNotification {
-				session_index: 10,
+		println!("block {:?}", n);
```
rphmeier, line 228: stray println

rphmeier (approved):
Looks good to me! Thanks for the cleanup of tests and addition of `Shared` module

runtime/parachains/src/lib.rs
```diff
@@ -49,11 +50,11 @@ pub use paras::ParaLifecycle;
 pub fn schedule_para_initialize<T: paras::Config>(
 	id: primitives::v1::Id,
 	genesis: paras::ParaGenesisArgs,
-) {
-	<paras::Module<T>>::schedule_para_initialize(id, genesis);
+) -> Result<(), ()> {
+	<paras::Module<T>>::schedule_para_initialize(id, genesis).map_err(|_| ())
```
rphmeier, line 54: An API nit. When you want to do swapping I assume you'd call `onboard()` on one and `offboard()` on the other. The first might succeed and the second might fail. You could either have some kind of atomic updating API where you update multiple paras at once or you just expose the `can_*` functions here. Something for a follow-up?

shawntabrizi: ack, will follow up in another PR and with more consideration specifically with the swap integration in registrar

runtime/common/src/paras_registrar.rs
```diff
@@ -309,7 +308,7 @@ mod tests {
 	parameter_types! {
 		pub const BlockHashCount: u32 = 250;
 		pub BlockWeights: limits::BlockWeights =
-			limits::BlockWeights::with_sensible_defaults(4 * 1024 * 1024, NORMAL_RATIO);
+			frame_system::limits::BlockWeights::simple_max(1024);
```
ggwpez, line 311: Just wondering: what was the motivation here?

PR #2413, opened 2021-02-10
Initializer + Paras Clean Up Messages When Offboarding

shawntabrizi:
This is a replacement for https://github.com/paritytech/polkadot/pull/2360

This PR updates the Paras module to return the list of offboarding parachains to the Initializer module, which then calls into the appropriate messaging modules to clean up any existing state for those parachains.

The main purpose of this PR is to remove multiple queues for performing actions during parachain offboarding, and instead have such actions controlled by one module (paras), and all clean up actions are directly executed once the parachain is actually offboarded.


pepyakin (commented):
Well, personally, I would prefer if we passed the list of outgoing directly into relevant `initializer_on_new_session`s.

The reason is: each extra lifecycle callback makes it harder to reason about within a module. Namely, you need to think about the ordering the ordering. E.g. when you you do something in `initializer_on_new_session` you need keep in mind that `clean_*_after_outgoing` comes before and may already clobber some data. 

Moreover, it also makes the system less flexible since it's already decided that the cleanup always goes before the session change, whereas if we just passed the offboarded paras in `initializer_on_new_session`.

shawntabrizi:
@pepyakin hows this?

runtime/parachains/src/paras.rs
```diff
@@ -317,21 +317,24 @@ impl<T: Config> Module<T> {
...
 		Self::apply_incoming(&mut parachains);
 		Self::apply_upgrades(&mut parachains);
 		Self::apply_downgrades(&mut parachains);
 		<Self as Store>::Parachains::set(parachains);
+		notification.outgoing_paras.append(&mut outgoing);
 	}
 
 	/// Cleans up all outgoing paras. Returns the new set of parachains
-	fn clean_up_outgoing(now: T::BlockNumber) -> Vec<ParaId> {
+	fn clean_up_outgoing(now: T::BlockNumber) -> (Vec<ParaId>, Vec<ParaId>) {
```
pepyakin, line 333: Needs update of the doc comment.

runtime/parachains/src/paras.rs
```diff
@@ -317,21 +317,24 @@ impl<T: Config> Module<T> {
...
+	/// Returns the list of outgoing parachains for this session.
+	pub(crate) fn initializer_on_new_session(notification: &mut SessionChangeNotification<T::BlockNumber>) {
 		let now = <frame_system::Module<T>>::block_number();
-		let mut parachains = Self::clean_up_outgoing(now);
+		let (mut parachains, mut outgoing) = Self::clean_up_outgoing(now);
 		Self::apply_incoming(&mut parachains);
 		Self::apply_upgrades(&mut parachains);
 		Self::apply_downgrades(&mut parachains);
 		<Self as Store>::Parachains::set(parachains);
+		notification.outgoing_paras.append(&mut outgoing);
```
pepyakin, line 329: I'd just assign it or `mem::replace` with assert. That's because we know that this field is assigned only by paras.

runtime/parachains/src/initializer.rs
```diff
@@ -46,6 +46,8 @@ pub struct SessionChangeNotification<BlockNumber> {
 	pub random_seed: [u8; 32],
 	/// New session index.
 	pub session_index: sp_staking::SessionIndex,
+	/// Outgoing paras for the session.
+	pub outgoing_paras: Vec<ParaId>,
```
pepyakin, line 50: I would say that this is supposed to be assigned by the Paras module

pepyakin (commented):
That's better.

I'd prefer if `ougoing_paras` were a separate parameter, i.e. not within the SessionChangeNotification. This way we would avoid making SessionChangeNotification mutable and we would avoid this annoying comment saying that outgoing paras is only assigned at Paras. But that's not a big deal.

runtime/parachains/src/paras.rs
rphmeier: I'm fine with making `initializer_on_new_session` return a `Vec<_>` for this module only, if that makes things easier.

rphmeier (commented):
Guide updates? When design is settled

shawntabrizi:
@pepyakin and @rphmeier looking gucci now?

I will update guide once I get an ack you are happy with this design. Thanks for patience from both of you.

rphmeier:
Design looks good! Code changes look fine but I haven't gone really deep

runtime/parachains/src/dmp.rs
```diff
@@ -229,23 +217,26 @@ mod tests {
...
 	pub(crate) fn run_to_block(to: BlockNumber, new_session: Option<Vec<BlockNumber>>) {
 		while System::block_number() < to {
 			let b = System::block_number();
+			Paras::initializer_finalize();
 			Dmp::initializer_finalize();
 			if new_session.as_ref().map_or(false, |v| v.contains(&(b + 1))) {
-				Dmp::initializer_on_new_session(&Default::default());
+				let notification = Default::default();
+				let outgoing_paras = Paras::initializer_on_new_session(&notification);
+				Dmp::initializer_on_new_session(&notification, &outgoing_paras);
```
pepyakin, line 232: Nit: not strictly necessary to involve paras here

runtime/parachains/src/dmp.rs
```diff
@@ -270,39 +261,26 @@ mod tests {
...
-			assert!(!<Dmp as Store>::DownwardMessageQueues::get(&b).is_empty());
-			assert!(!<Dmp as Store>::DownwardMessageQueues::get(&c).is_empty());
-
-			Dmp::schedule_para_cleanup(b);
-
-			// run to block changing the session.
-			run_to_block(3, Some(vec![3]));
+			let notification = crate::initializer::SessionChangeNotification {
+				..Default::default()
+			};
```
pepyakin, line 277: Nit: `crate::initializer::SessionChangeNotification::default()`

runtime/parachains/src/scheduler.rs
```diff
@@ -754,8 +754,8 @@ mod tests {
 			Scheduler::initializer_finalize();
 			Paras::initializer_finalize();
 
-			if let Some(notification) = new_session(b + 1) {
-				Paras::initializer_on_new_session(&notification);
+			if let Some(mut notification) = new_session(b + 1) {
+				Paras::initializer_on_new_session(&mut notification);
```
pepyakin, line 758: Nit: we do not need to take mutable reference here (and other places), right?

runtime/parachains/src/paras.rs
```diff
@@ -845,7 +851,7 @@ mod tests {
 			let b = System::block_number();
 			Paras::initializer_finalize();
 			if new_session.as_ref().map_or(false, |v| v.contains(&(b + 1))) {
-				Paras::initializer_on_new_session(&Default::default());
+				Paras::initializer_on_new_session(&mut Default::default());
```
pepyakin, line 854: Nit: Not needed

runtime/parachains/src/initializer.rs
```diff
@@ -21,7 +21,7 @@
 
 use sp_std::prelude::*;
 use frame_support::weights::Weight;
-use primitives::v1::ValidatorId;
+use primitives::v1::{ValidatorId};
```
pepyakin, line 24: Nit: not needed

runtime/parachains/src/ump.rs
```diff
@@ -207,44 +203,36 @@ impl<T: Config> Module<T> {
...
-				v.insert(i, id);
-			}
-		});
-	}
-
-	fn clean_ump_after_outgoing(outgoing_para: ParaId) {
-		<Self as Store>::RelayDispatchQueueSize::remove(&outgoing_para);
-		<Self as Store>::RelayDispatchQueues::remove(&outgoing_para);
+	/// Remove all relevant storage items for an outgoing parachain.
+	fn clean_ump_after_outgoing(outgoing_para: &ParaId) {
```
pepyakin, line 220: Nit: not strictly necessary to change to a shared reference: `ParaId` is a copy type

pepyakin (approved):
Looks good. Some clean up wouldn't hurt

runtime/parachains/src/ump.rs
shawntabrizi: is there a reason to not be a shared reference? This would be the natural type coming from a slice.

pepyakin: it would be natural either way: again because it's a `Copy` type. You can think of it as similar to a `u32`, so it's always possible to "move it out" from the slice by basically making a copy.

FWIW, on a lower lever, I am pretty sure there is no difference because LLVM will be able to note that using a pointer doesn't make sense because, the pointer requires a dereference and twice as big as the value itself.

↑ But that's a nit, feel free to ignore anyway

shawntabrizi:
Anyone else need to review this to merge?

rphmeier (approved):

rphmeier:
bot merge

PR #2417, opened 2021-02-10
companion for substrate/pull/7929 (24 nominators)

kianenigma:
https://github.com/paritytech/substrate/pull/7929

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -21,9 +21,9 @@ dependencies = [
 
 [[package]]
 name = "adler"
-version = "0.2.3"
+version = "0.2.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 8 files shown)


Cargo.lock
```diff
@@ -1107,35 +1155,35 @@ dependencies = [
...
  "digest 0.8.1",
  "rand_core 0.5.1",
- "subtle 2.2.3",
+ "subtle 2.4.0",
  "zeroize",
 ]
 
 [[package]]
 name = "curve25519-dalek"
-version = "3.0.0"
```
thiolliere, line 1123: I'm not sure update of other crate should be done in this PR,
usually companion only update substrate

thiolliere (approved):

Cargo.lock
kianenigma: I am not sure how that has came to be 🤔  I only merged master into both branches.

kianenigma: reverted them, maybe they were the cause of red CI as well.

PR #2429, opened 2021-02-12
CancelProxy uses `reject_announcement` instead of `remove_announcement`

rphmeier:
Cancel proxies were designed for a situation where we have 3 accounts:
1. `Alice`, a cold storage account which is meant to be controlled via its proxies
2. `Bob`, a time-delay proxy with `Any` permissions. This account must _announce_ its intentions ahead of time.
3. `Charlie`, a `CancelProxy`, which can observe and cancel `Bob`'s announcements.

In practice, we may have multiple proxies in the place of `Bob`, but this minimal example suffices to explain. Also in practice, `Bob` and `Charlie` may be multi-sigs, but that's irrelevant to the example, again. And also in practice, `Alice` might be an anonymous proxy, and this also doesn't matter.

If the `Bob` account is compromised, we'd like `Charlie` to be able to issue a transaction to cancel `Bob`'s announcements.

Thus we have the following desired control flow to detect, mitigate, and deter future attacks:

1. `Bob` issues `proxy::Announce(Alice, bad_call_hash)`.
2. `Charlie` does not recognize the `bad_call_hash`, and does ????, canceling the call by `Bob`.
3. `Bob` is replaced as a proxy by `Alice`.

The purpose of this PR is to address what ???? stands for.

In the status quo, `Charlie` has access to issue `proxy::RemoveAnnouncement` calls on behalf of `Alice`.  However, `proxy::RemoveAnnouncement` is meant to be called by a _proxy_ in order to retract a previously announced call. As `Alice` is not a proxy, it isn't useful for `Charlie` to issue `proxy::Proxy(Alice, proxy::RemoveAnnouncement(Alice, bad_call_hash))`. This is a no-op. 

In fact, unless `Charlie` is equal to `Bob`, `Charlie` has no way of retracting calls by `Bob`. If `Charlie` is equal to `Bob`, `Charlie` can issue a `proxy::RemoveAnnouncement(Alice, bad_call_hash)` - note, this is not a proxy call on behalf of Alice - and successfully remove `Bob`'s announcement. The reason I bring up this `Bob == Charlie` case is because I suspect it was used during testing.

This PR gives `CancelProxy` access to `proxy::RejectAnnouncement` which is intended to be called _by_ a delegate account to reject the announcements by its proxies. With this change, `Charlie` can issue `proxy::Proxy(Alice, proxy::RejectAnnouncement(Bob, bad_call_hash))` to cause the `Alice` account to reject Bob's bad call hash. I believe this is the desired behavior.

It's also possible that I'm completely missing the point; I haven't found anyone able to explain this feature adequately to me yet and I wasn't part of the initial design discussions.

### DIFF
--- a/runtime/kusama/src/lib.rs
+++ b/runtime/kusama/src/lib.rs
@@ -918,7 +918,7 @@ impl InstanceFilter<Call> for ProxyType {
 				Call::Utility(..)
 			),
 			ProxyType::CancelProxy => matches!(c,
-				Call::Proxy(pallet_proxy::Call::remove_announcement(..))
+				Call::Proxy(pallet_proxy::Call::reject_announcement(..))
 			)
 		}
 	}
--- a/runtime/kusama/src/lib.rs
+++ b/runtime/kusama/src/lib.rs
@@ -918,8 +918,7 @@ impl InstanceFilter<Call> for ProxyType {
 				Call::Utility(..)
 			),
 			ProxyType::CancelProxy => matches!(c,
-				Call::Proxy(pallet_proxy::Call::reject_announcement(..)) |
-				Call::Proxy(pallet_proxy::Call::remove_announcement(..))
+				Call::Proxy(pallet_proxy::Call::reject_announcement(..))
 			)
 		}
 	}
--- a/runtime/kusama/src/lib.rs
+++ b/runtime/kusama/src/lib.rs
@@ -918,7 +918,8 @@ impl InstanceFilter<Call> for ProxyType {
 				Call::Utility(..)
 			),
 			ProxyType::CancelProxy => matches!(c,
-				Call::Proxy(pallet_proxy::Call::reject_announcement(..))
+				Call::Proxy(pallet_proxy::Call::reject_announcement(..)) |
+				Call::Proxy(pallet_proxy::Call::remove_announcement(..))
 			)
 		}
 	}
--- a/runtime/polkadot/src/lib.rs
+++ b/runtime/polkadot/src/lib.rs
@@ -922,7 +922,7 @@ impl InstanceFilter<Call> for ProxyType {
 				Call::Utility(..)
 			),
 			ProxyType::CancelProxy => matches!(c,
-				Call::Proxy(pallet_proxy::Call::remove_announcement(..))
+				Call::Proxy(pallet_proxy::Call::reject_announcement(..))
 			)
 		}
 	}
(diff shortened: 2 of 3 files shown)


rphmeier:
I added back `remove_announcement`. It can be somewhat useful in situations where you have two layers of proxies. But it's a footgun - none of those proxies can have `Any` or `NonTransfer` permissions as it'd be able to issue `proxy::removeProxy` calls.

```
             Cancel
                 |
                 v
Master <- Intermediate <- Alice


```

If this is an architecture where `removeAnnouncement` is desired, then we're trying to have `Cancel` issue `proxy::RemoveAnnouncement` on behalf of the `Intermediate`. If we only use `proxy::RemoveAnnouncement`, the only sensible time delay is between `Intermeidate` and `Master` as `Intermediate` (and therefore `Cancel`) can't do anything about calls in between `Alice` and `Intermediate` without using `proxy::RejectAnnouncement`. I still think `RejectAnnouncement` is better for this use-case, and I am willing to argue that with anybody ( @ddorgan and @wpank ?). But I guess this is somewhat (?) useful, although a footgun, because if `Alice` has the ability to execute `proxy::RemoveProxy(Cancel)` on behalf of `Intermediate`, it can just remove its own watchdog 🤦

gavofyork:
Cancel proxy type should just use `reject_announcement`. It was never meant to use `remove_announcement`.

gavofyork:
I think the original change is the one we should stick with.

shawntabrizi (approved):

wpank:
This looks to make a bit more sense. The workflow of using this would look something like the following:

Set up an account (Charlie)  bonded to itself and has both a Time Delay Staking Proxy (Dave) and a CancelProxy (Eve).  

proxies

First generate  the Call Hash: right now this must be constructed with some code, since there isn't any way to get the call hash from the apps ui yet. As an example, we want Charlie to nominate an account: `16kZJGPJ37uYxjs7adswyEHbPYeHS9jQHSaSUJhkfvWPcoeF`. The call hash would be: 

```javascript
const innerTx = api.tx.staking.nominate(['16kZJGPJ37uYxjs7adswyEHbPYeHS9jQHSaSUJhkfvWPcoeF']);
const callhash = innerTx.method.hash;
console.log(callhash);
```

or 

`0x82943ce618a1996e1837be4ee3faaa8e408eb6c9a65faf7c0aada29875fb9720`

Dave announces the call on behalf of Charlie

timedelayannounce

We can then see Dave has an announcement stored

announcement

Eve then rejects the announcement

rejectannouncement

And that previous announcement is removed

noannouncements

wpank (approved):

PR #2438, opened 2021-02-15
companion:  Babe remove inherent

thiolliere:
companion of https://github.com/paritytech/substrate/pull/8124
Remove inherent part of pallet babe in construct_runtime and check babe is in OnTimestampSet.


andresilva (approved):

node/test/client/src/block_builder.rs
```diff
@@ -67,6 +64,29 @@ impl InitPolkadotBlockBuilder for Client {
...
+			logs: vec![
+				DigestItem::PreRuntime(
+					BABE_ENGINE_ID,
+					PreDigest::SecondaryPlain(SecondaryPlainPreDigest {
+						slot,
+						authority_index: 42,
+					}).encode()
+				),
+			],
+		};
```
thiolliere, line 83: @andresilva I had to add this digest so that the test `grandpa_pause_voting_rule_works` succeed.
I'm not sure it is ok, or if it better to actually remove babe from OnTimestampSet directly.
But seems fine.

PR #2440, opened 2021-02-15
Mitigation of SIGBUS

pepyakin:
On Rococo we observe that some nodes constantly crash with SIGBUS signal. The coredump would show this stack trace

```
  * frame #0: 0x00005622f1602294 rococo-20210212`shared_memory::conf::SharedMemConf::create::h3b4bdaf4fd303997 + 1140
    frame #1: 0x00005622f15f3063 rococo-20210212`polkadot_parachain::wasm_executor::validation_host::ValidationHost::validate_candidate::h841dbe531d5b4b1f + 2371
    frame #2: 0x00005622f15f0837 rococo-20210212`polkadot_parachain::wasm_executor::validation_host::ValidationPool::validate_candidate_custom::hc64b3ac33eb43727 + 407
    frame #3: 0x00005622f15f0605 rococo-20210212`polkadot_parachain::wasm_executor::validation_host::ValidationPool::validate_candidate::h250300cc4b686d08 + 661
    frame #4: 0x00005622ef1f337b rococo-20210212`polkadot_parachain::wasm_executor::validate_candidate::h0dd7d3b4f578e7df + 379
    frame #5: 0x00005622f018aa86 rococo-20210212`polkadot_node_core_candidate_validation::validate_candidate_exhaustive::h601288254d015d57 + 1814
    frame #6: 0x00005622ef9d51df rococo-20210212`_$LT$core..future..from_generator..GenFuture$LT$T$GT$$u20$as$u20$core..future..future..Future$GT$::poll::h2c834a476e8bd245 + 335
    frame #7: 0x00005622f092f3e1 rococo-20210212`_$LT$sc_service..task_manager..prometheus_future..PrometheusFuture$LT$T$GT$$u20$as$u20$core..future..future..Future$GT$::poll::hb22887fdbf627ea5 + 65
    frame #8: 0x00005622f092ab4b rococo-20210212`_$LT$futures_util..future..select..Select$LT$A$C$B$GT$$u20$as$u20$core..future..future..Future$GT$::poll::hac8861c9054cdc28 + 75
    frame #9: 0x00005622f0917b66 rococo-20210212`_$LT$core..future..from_generator..GenFuture$LT$T$GT$$u20$as$u20$core..future..future..Future$GT$::poll::hb8a5b16daeea0b8a + 230
    frame #10: 0x00005622f091aba6 rococo-20210212`_$LT$tracing_futures..Instrumented$LT$T$GT$$u20$as$u20$core..future..future..Future$GT$::poll::h869f77cae2e9183b (.llvm.13681125076900316595) + 38
    frame #11: 0x00005622f092d4a7 rococo-20210212`_$LT$sc_service..task_manager..WithTelemetrySpan$LT$T$GT$$u20$as$u20$core..future..future..Future$GT$::poll::h7e367afd349789ec + 135
    frame #12: 0x00005622ef0b5b31 rococo-20210212`std::thread::local::LocalKey$LT$T$GT$::with::hac0311a348c6d2f7 + 97
    frame #13: 0x00005622eef8db9e rococo-20210212`futures_executor::local_pool::block_on::hc1ec056b085a3d59 + 62
    frame #14: 0x00005622ef092d5c rococo-20210212`_$LT$std..panic..AssertUnwindSafe$LT$F$GT$$u20$as$u20$core..ops..function..FnOnce$LT$$LP$$RP$$GT$$GT$::call_once::ha0d22998eaf8d416 + 252
    frame #15: 0x00005622ef01327d rococo-20210212`tokio::runtime::task::harness::Harness$LT$T$C$S$GT$::poll::h1dea41e2d7641c36 + 125
    frame #16: 0x00005622f1ebfee1 rococo-20210212`tokio::runtime::blocking::pool::Inner::run::hc38f319445b5d344 + 257
    frame #17: 0x00005622f1ed09a8 rococo-20210212`tokio::runtime::context::enter::h8923a2a871eec0e9 + 104
    frame #18: 0x00005622f1ecefef rococo-20210212`std::sys_common::backtrace::__rust_begin_short_backtrace::h7a3e99e8a2c2c4bd + 79
    frame #19: 0x00005622f1ecf8a0 rococo-20210212`core::ops::function::FnOnce::call_once$u7b$$u7b$vtable.shim$u7d$$u7d$::h895c3678b7c51e76 + 224
    frame #20: 0x00005622f227b59a rococo-20210212`std::sys::unix::thread::Thread::new::thread_start::h587efff279c68ba7 [inlined] _$LT$alloc..boxed..Box$LT$F$C$A$GT$$u20$as$u20$core..ops..function..FnOnce$LT$Args$GT$$GT$::call_once::h09ff301006f1aeca at boxed.rs:1307:9
    frame #21: 0x00005622f227b594 rococo-20210212`std::sys::unix::thread::Thread::new::thread_start::h587efff279c68ba7 [inlined] _$LT$alloc..boxed..Box$LT$F$C$A$GT$$u20$as$u20$core..ops..function..FnOnce$LT$Args$GT$$GT$::call_once::he79488c8f00b5f31 at boxed.rs:1307
    frame #22: 0x00005622f227b58b rococo-20210212`std::sys::unix::thread::Thread::new::thread_start::h587efff279c68ba7 at thread.rs:71
    frame #23: 0x00007f76b4431fa3 libpthread.so.0`setxid_mark_thread.isra.0 + 99
```

The error is coming from the fact that /dev/shm, the place where the shmem files are typically placed on Linux, is full. /dev/shm in that case was backed by a tmpfs, with limited amount of memory. It is actually ok to call `shm_open` and `ftruncate` on a full filesystem, but the first access to the region will lead to a SIGBUS.

In order to mitigate this issue we first decrease the amount of memory required, prior to this PR it was 1GiB and now it's around 32MiB. Secondly, as soon as the worker connected to the shared memory we unlink the file, which, if not prevents it from accounting towards the total occupied size of the /dev/shm then it makes sure that the shm file is not laying around if something goes wrong.

Ideal solution would be to avoid using shared memory. I see we should solve this problem with a better approach to caching, outlined [here](https://github.com/paritytech/polkadot/pull/2342)

### DIFF
--- a/parachain/src/wasm_executor/validation_host/workspace.rs
+++ b/parachain/src/wasm_executor/validation_host/workspace.rs
@@ -19,6 +19,7 @@
 //! worker.
 
 use crate::primitives::{ValidationParams, ValidationResult};
+use super::LOG_TARGET;
 use parity_scale_codec::{Decode, Encode};
 use raw_sync::{
 	events::{Event, EventImpl, EventInit, EventState},
@@ -428,11 +429,16 @@ pub fn open(id: &str) -> Result<WorkerHandle, String> {
 		// (Not sure, what alternative it has though).
 		unsafe {
 			// must be in a local var in order to be not deallocated.
-			let shmem_id =
+			let shmem_id_cstr =
 				std::ffi::CString::new(shmem_id).expect("the shmmem id cannot have NUL in it; qed");
 
-			if libc::shm_unlink(shmem_id.as_ptr()) == -1 {
+			if libc::shm_unlink(shmem_id_cstr.as_ptr()) == -1 {
 				// failed to remove the shmem file nothing we can do ¯\_(ツ)_/¯
+				log::warn!(
+					target: LOG_TARGET,
+					"failed to remove the shmem with id {}",
+					shmem_id,
+				);
 			}
 		}
 	}
--- a/parachain/src/wasm_executor/validation_host/workspace.rs
+++ b/parachain/src/wasm_executor/validation_host/workspace.rs
@@ -160,7 +160,7 @@ impl Inner {
 }
 
 fn align_up_to(v: usize, alignment: usize) -> usize {
-	(v + alignment - 1) & !(alignment - 1)
+	((v + alignment - 1) / alignment) * alignment
 }
 
 /// Initializes a new or attaches to an exising event.
--- a/parachain/src/wasm_executor/validation_host/workspace.rs
+++ b/parachain/src/wasm_executor/validation_host/workspace.rs
@@ -359,7 +359,8 @@ impl HostHandle {
 
 	/// Wait for the validation result from the worker with the given timeout.
 	///
-	/// Returns `Ok` if the response was received within the deadline or error otherwise. Returning
+	/// Returns `Ok` if the response was received within the deadline or error otherwise. An error
+	/// could also occur because of failing decoding the result from the worker. Returning
 	/// `Ok` doesn't mean that the candidate was successfully validated though, for that the client
 	/// needs to inspect the returned validation result header.
 	pub fn wait_for_result(
--- a/parachain/src/wasm_executor/workspace.rs
+++ b/parachain/src/wasm_executor/workspace.rs
@@ -1,0 +1,492 @@
+// Copyright 2021 Parity Technologies (UK) Ltd.
+// This file is part of Polkadot.
+
+// Polkadot is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+
+// Polkadot is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with Polkadot.  If not, see <http://www.gnu.org/licenses/>.
+
+use crate::primitives::{ValidationParams, ValidationResult};
+use parity_scale_codec::{Decode, Encode};
+use raw_sync::{
+	events::{Event, EventImpl, EventInit, EventState},
+	Timeout,
+};
+use shared_memory::{Shmem, ShmemConf};
+use std::{
+	error::Error,
+	fmt,
+	io::{Cursor, Write},
+	slice,
+	sync::atomic::AtomicBool,
+	time::Duration,
+};
+
+// maximum memory in bytes
+const MAX_RUNTIME_MEM: usize = 1024 * 1024 * 1024; // 1 GiB
+const MAX_CODE_MEM: usize = 16 * 1024 * 1024; // 16 MiB
+const MAX_VALIDATION_RESULT_HEADER_MEM: usize = MAX_CODE_MEM + 1024; // 16.001 MiB
+
+/// Params header in shared memory. All offsets should be aligned to WASM page size.
+#[derive(Encode, Decode, Debug)]
+struct ValidationHeader {
+	code_size: u64,
+	params_size: u64,
+}
+
+#[derive(Encode, Decode, Debug)]
+pub enum WorkerValidationError {
+	InternalError(String),
+	ValidationError(String),
+}
+
+#[derive(Encode, Decode, Debug)]
+pub enum ValidationResultHeader {
+	Ok(ValidationResult),
+	Error(WorkerValidationError),
+}
+
+#[derive(Debug, PartialEq, Eq, Copy, Clone)]
+enum Mode {
+	Initialize,
+	Attach,
+}
+
+fn stringify_err(err: Box<dyn Error>) -> String {
+	format!("{:?}", err)
+}
+
+struct Inner {
+	shmem: Shmem,
+	candidate_ready_ev: Box<dyn EventImpl>,
+	result_ready_ev: Box<dyn EventImpl>,
+	worker_ready_ev: Box<dyn EventImpl>,
+
+	/// Flag that indicates that the worker side is attached to this workspace.
+	///
+	/// While there are apparent problems attaching multiple workers to the same workspace, we don't
+	/// need that anyway. So to make our reasoning a little bit simpler just add a flag and check
+	/// it before attaching.
+	attached: *mut AtomicBool,
+
+	/// The number of bytes reserved by the auxilary stuff like events from the beginning of the
+	/// shared memory area.
+	///
+	/// We expect this to be way smaller than the whole shmem size.
+	consumed: usize,
+}
+
+impl Inner {
+	fn layout(shmem: Shmem, mode: Mode) -> Self {
+		unsafe {
+			let base_ptr = shmem.as_ptr();
+			let mut consumed = 0;
+
+			let candidate_ready_ev = add_event(base_ptr, &mut consumed, mode);
+			let result_ready_ev = add_event(base_ptr, &mut consumed, mode);
+			let worker_ready_ev = add_event(base_ptr, &mut consumed, mode);
+
+			// The size of AtomicBool is guaranteed to be the same as the bool, however, docs
+			// on the bool primitve doesn't actually state that the in-memory size is equal to 1 byte.
+			//
+			// AtomicBool requires hardware support of 1 byte width of atomic operations though, so
+			// that should be fine.
+			//
+			// We still assert here to be safe than sorry.
+			static_assertions::assert_eq_size!(AtomicBool, u8);
+
+			// SAFETY: `AtomicBool` is represented by an u8 thus will be happy to take any alignment.
+			let attached = base_ptr.add(consumed) as *mut AtomicBool;
+
+			let consumed = align_up_to(consumed, 64);
+
+			Self {
+				shmem,
+				attached,
+				consumed,
+				candidate_ready_ev,
+				result_ready_ev,
+				worker_ready_ev,
+			}
+		}
+	}
+
+	fn as_slice(&self) -> &[u8] {
+		unsafe {
+			let base_ptr = self.shmem.as_ptr().add(self.consumed);
+			let remaining = self.shmem.len() - self.consumed;
+			slice::from_raw_parts(base_ptr, remaining)
+		}
+	}
+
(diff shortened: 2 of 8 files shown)


parachain/src/wasm_executor/workspace.rs
```diff
@@ -0,0 +1,523 @@
...
+// (at your option) any later version.
+
+// Polkadot is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with Polkadot.  If not, see <http://www.gnu.org/licenses/>.
+
```
rphmeier, line 16: missing doc-string

parachain/src/wasm_executor/workspace.rs
```diff
@@ -0,0 +1,523 @@
...
+/// Params header in shared memory. All offsets should be aligned to WASM page size.
+#[derive(Encode, Decode, Debug)]
+struct ValidationHeader {
+	code_size: u64,
+	params_size: u64,
+}
+
+#[derive(Encode, Decode, Debug)]
+pub enum WorkerValidationError {
```
rphmeier, line 46: It would be nice, although I won't block on it, to have docs on public items

parachain/src/wasm_executor/workspace.rs
```diff
@@ -0,0 +1,523 @@
...
+			slice::from_raw_parts_mut(base_ptr, remaining)
+		}
+	}
+
+	/// Mark that this workspace has attached. Returns `true` if this instance was attached the
+	/// first.
+	fn declare_exclusive_attached(&self) -> bool {
+		unsafe {
+			// If this succeeded then the value was `false`, thus, we managed to attach exclusively.
+			(&*self.attached)
```
bkchr, line 143: I like this xD

parachain/src/wasm_executor/workspace.rs
```diff
@@ -0,0 +1,523 @@
...
+			.set(EventState::Signaled)
+			.map_err(stringify_err)
+			.map_err(RequestValidationErr::Signal)?;
+
+		Ok(())
+	}
+
+	/// Wait for the validation result from the worker with the given timeout.
+	///
+	/// Returns `Ok` if the response was received within the deadline or error otherwise. Returning
```
bkchr, line 335: Seems to indicate that an error means a timeout, which is not true here.

parachain/src/wasm_executor/workspace.rs
```diff
@@ -0,0 +1,523 @@
...
+/// Furthermore, this function should be called once for initialization.
+///
+/// Specifically, `consumed` should not be modified by the caller, it should be passed as is to this function.
+unsafe fn add_event(base_ptr: *mut u8, consumed: &mut usize, mode: Mode) -> Box<dyn EventImpl> {
+	// SAFETY: there is no safety proof since the documentation doesn't specify the particular constraints
+	//         besides requiring the pointer to be valid. AFAICT, the pointer is valid.
+	let ptr = base_ptr.add(*consumed);
+	let (ev, used_bytes) = match mode {
+		Mode::Initialize => Event::new(ptr, true).unwrap(),
+		Mode::Attach => Event::from_existing(ptr).unwrap(),
```
rphmeier, line 176: Remove `unwrap` here in favor of `expect`?

bkchr (commented):
Much better interface! :)

parachain/src/wasm_executor/workspace.rs
```diff
@@ -0,0 +1,523 @@
...
+			// on the bool primitve doesn't actually state that the in-memory size is equal to 1 byte.
+			//
+			// AtomicBool requires hardware support of 1 byte width of atomic operations though, so
+			// that should be fine.
+			//
+			// We still assert here to be safe than sorry.
+			static_assertions::assert_eq_size!(AtomicBool, u8);
+
+			// SAFETY: `AtomicBool` is represented by an u8 thus will be happy to take any alignment.
+			let attached = base_ptr.add(consumed) as *mut AtomicBool;
```
rphmeier, line 107: Should `consumed` have 1 added to it in the case that the subsequent `align_up_to` is a no-op?

parachain/src/wasm_executor/workspace.rs
```diff
@@ -0,0 +1,523 @@
...
+		unsafe {
+			let base_ptr = self.shmem.as_ptr().add(self.consumed);
+			let remaining = self.shmem.len() - self.consumed;
+			slice::from_raw_parts_mut(base_ptr, remaining)
+		}
+	}
+
+	/// Mark that this workspace has attached. Returns `true` if this instance was attached the
+	/// first.
+	fn declare_exclusive_attached(&self) -> bool {
```
rphmeier, line 140: It would help to understand a bit better what "attached" means

parachain/src/wasm_executor/workspace.rs
pepyakin: Ah, good catch. We should always add 1.

parachain/src/wasm_executor/workspace.rs
```diff
@@ -0,0 +1,523 @@
...
+		// shared-memory doesn't return file path to the shmem if get_flink_path is called, so we
+		// resort to `shm_unlink`.
+		//
+		// Additionally, even thouygh `fs::remove_file` is said to use `unlink` we still avoid relying on it,
+		// because the stdlib doesn't actually provide any gurantees on what syscalls will be called.
+		// (Not sure, what alternative it has though).
+		unsafe {
+			// must be in a local var in order to be not deallocated.
+			let shmem_id =
+				std::ffi::CString::new(shmem_id).expect("the shmmem id cannot have NUL in it; qed");
```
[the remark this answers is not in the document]
pepyakin: For some reason, \0 is referred as NUL, to a degree that the error returned by http://doc.rust-lang.org/1.48.0/std/ffi/struct.CString.html#method.new is called `NulError`

rphmeier (approved):
LGTM

parachain/src/wasm_executor/workspace.rs
```diff
@@ -0,0 +1,523 @@
...
+///
+/// Specifically, `consumed` should not be modified by the caller, it should be passed as is to this function.
+unsafe fn add_event(base_ptr: *mut u8, consumed: &mut usize, mode: Mode) -> Box<dyn EventImpl> {
+	// SAFETY: there is no safety proof since the documentation doesn't specify the particular constraints
+	//         besides requiring the pointer to be valid. AFAICT, the pointer is valid.
+	let ptr = base_ptr.add(*consumed);
+	let (ev, used_bytes) = match mode {
+		Mode::Initialize => Event::new(ptr, true).unwrap(),
+		Mode::Attach => Event::from_existing(ptr).unwrap(),
+	};
```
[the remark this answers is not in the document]
pepyakin: Well, yeah, that's is something I missed 🤐  We expect this to always succeed so I will change to a `expect`

parachain/src/wasm_executor/workspace.rs
pepyakin: I missed it 🤦

parachain/src/wasm_executor/workspace.rs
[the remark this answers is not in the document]
rphmeier: Yes I think this is an old convention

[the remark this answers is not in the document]
bkchr: Okay :D Never before came across this name, even not in my good old C++ times :D

parachain/src/wasm_executor/workspace.rs
```diff
@@ -0,0 +1,523 @@
...
+					std::sync::atomic::Ordering::SeqCst,
+					std::sync::atomic::Ordering::SeqCst,
+				)
+				.is_ok()
+		}
+	}
+}
+
+fn align_up_to(v: usize, alignment: usize) -> usize {
+	(v + alignment - 1) & !(alignment - 1)
```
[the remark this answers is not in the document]
pepyakin: Wait, shouldn't it be `((v + alignment - 1) / alignment) * alignment`?

parachain/src/wasm_executor/workspace.rs
```diff
@@ -0,0 +1,523 @@
...
+	consumed: usize,
+}
+
+impl Inner {
+	fn layout(shmem: Shmem, mode: Mode) -> Self {
+		unsafe {
+			let base_ptr = shmem.as_ptr();
+			let mut consumed = 0;
+
+			let candidate_ready_ev = add_event(base_ptr, &mut consumed, mode);
```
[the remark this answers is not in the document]
pepyakin: I figured that is not important since we control all the sizes and that those events (they are backed by pthread cond vars) are negligible small compared to the whole size.

parachain/src/wasm_executor/workspace.rs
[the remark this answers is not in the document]
bkchr: Yes :facepalm:

pepyakin:
- [x] localnet produces blocks with a validator built from this branch produces blocks OK

parachain/src/wasm_executor/workspace.rs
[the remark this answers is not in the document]
bkchr: Fine :)

parachain/src/wasm_executor/validation_host/workspace.rs
```diff
@@ -0,0 +1,556 @@
...
+	error::Error,
+	fmt,
+	io::{Cursor, Write},
+	slice,
+	sync::atomic::AtomicBool,
+	time::Duration,
+};
+
+// maximum memory in bytes
+const MAX_PARAMS_MEM: usize = 1024 * 1024; // 1 MiB
```
arkpar, line 39: Is 1 Mib hard limit on proof size really enough?

pepyakin: Yeah, good point. We over allocate the buffer so it will still fit, but this could be express more clearer.

bkchr: By how much do we over allocate? Or do you mean that the POV could be any size and we always can allocate enough memory for it?

pepyakin: Scratch that, even if we over allocate, the call data is rejected by the params size check. See the fix here https://github.com/paritytech/polkadot/pull/2445

PR #2447, opened 2021-02-16
Migration testing CLI

kianenigma:
Not the companion per se, but the counterpart of https://github.com/paritytech/substrate/pull/8038. Needs to revert some dependency changes after substrate is updated.


node/service/Cargo.toml
```diff
@@ -116,6 +116,11 @@ runtime-benchmarks = [
 	"westend-runtime/runtime-benchmarks",
 	"rococo-runtime/runtime-benchmarks"
 ]
+try-runtime = [
+	"polkadot-runtime/try-runtime",
+	"kusama-runtime/try-runtime",
+	# "westend-runtime/try-runtime",
```
kianenigma, line 122:
```suggestion
	"westend-runtime/try-runtime",
```

kianenigma:
```suggestion
	"westend-runtime/try-runtime",
```

runtime/westend/src/lib.rs
```diff
@@ -1030,6 +1029,75 @@ sp_api::impl_runtime_apis! {
...
+			let weight = match target {
+				frame_try_runtime::Target::All => {
+					frame_support::debug::info!("Dry-running all on-runtime-upgrades.");
+					Executive::try_runtime_upgrade()
+				},
+				frame_try_runtime::Target::Pallet(name) => {
+					let name = sp_std::str::from_utf8(&name).unwrap();
+					frame_support::debug::info!("Dry-running on-runtime-upgrade of {}.", name);
+
+					frame_try_runtime::match_pallet_on_runtime_upgrade!(name,
```
shawntabrizi, line 1046: Probably could use `AllModules` for this?

shawntabrizi:
Would love other opinions, but I am not a fan of the name `try-runtime`. Its not very descriptive imo.

cli/src/command.rs
```diff
@@ -276,6 +276,31 @@ pub fn run() -> Result<()> {
...
+
+				Ok((
+					cmd.run::<
+						service::kusama_runtime::Block,
+						service::KusamaExecutor,
+					>(config).map_err(|e| Error::SubstrateCli(e)),
+					task_manager
+				))
+				// NOTE: we fetch only the block number from the block type, the chance of disparity
+				// between kusama's and polkadot's block number is small enough to overlook this.
```
shawntabrizi, line 301: This is saying that you are getting `Block = u32` from Kusama Runtime configuration, but potentially applying it to any runtime?

There should be a better way I think.

runtime/westend/src/lib.rs
```diff
@@ -1030,6 +1029,75 @@ sp_api::impl_runtime_apis! {
...
+						System,
```
shawntabrizi, line 1047: Can we use `AllModules` rather than writing these all manually?

cli/src/command.rs
kianenigma: Yes, I'll look into it. I'll look into it, there must be a way. 

(fyi we do the same for benchmarking CLI as well)

runtime/westend/src/lib.rs
kianenigma: Easily, no, because `AllModules` is a tuple, and tuples are by definition heterogeneous and not enumerable. What would have worked was to also create a type alias in `construct_runtime!` called `AllModulesArray`, which has the type `[Box<dyn PalletCommonTrait>; <number-of-pallets>]` (or just `Vec` for simplicity, ain't matter), where `PalletCommonTrait` is a wrapper trait for all the common stuff that all `Pallet` structs implement: `trait PalletCommonTrait: OnRuntimeUpgrade + OnInitialize + OnFinalize + OffchainWorker + PalletInfo + <anything else>`. 

I think this is the best way, but I'd see it as a follow-up task. 

@thiolliere I disucssed this with you as well in DM, chime in if I am wrong please.

kianenigma:
> Would love other opinions, but I am not a fan of the name `try-runtime`. Its not very descriptive imo.

I'm happy to rename it while we can, just keep it in mind that I think the feature name and runtime api trait name should not be limited to just migration, nor just runtime upgrade operations. 

I foresee adding other functions to the same api trait that does other test/diagnostics operations on the runtime (i.e. import this block in a dry-run manner, build 10 empty block on top of that state, etc.). So `try-runtime` was an attempt at making it a generalized name.

cli/src/command.rs
shawntabrizi: Its probably fine then

runtime/westend/src/lib.rs
thiolliere: I think I have a solution for it, I'm trying

thiolliere: one solution could be to have another method:
```diff
diff --git a/frame/support/src/traits.rs b/frame/support/src/traits.rs
index 602bc1aa1a..d62f6340e3 100644
--- a/frame/support/src/traits.rs
+++ b/frame/support/src/traits.rs
@@ -1566,6 +1566,9 @@ pub trait OnRuntimeUpgrade {
        fn post_upgrade() -> Result<(), &'static str> {
                Ok(())
        }
+
+       #[cfg(feature = "try-runtime")]
+       fn try_individual_on_runtime_upgrade_filtered(filter: &str) -> Result<(), &'static str>;
 }
 
 #[impl_for_tuples(30)]
@@ -1589,6 +1592,13 @@ impl OnRuntimeUpgrade for Tuple {
                for_tuples!( #( result = result.and(Tuple::post_upgrade()); )* );
                result
        }
+
+       #[cfg(feature = "try-runtime")]
+       fn try_individual_on_runtime_upgrade_filtered(filter: &str) -> Result<(), &'static str> {
+               let mut result = Ok(());
+               for_tuples!( #( result = result.and(Tuple::try_individual_on_runtime_upgrade_filtered()); )* );
+               result
+       }
 }
 
 /// Off-chain computation trait.
```
And make pallet macro implementing it using PalletInfo name.

But actually I wonder the interest, why not running all runtime upgrade everytime ?
Also the order of runtime ugprades is important and some can break other.
(like if pallet A depend on pallet B, but pallet B hasn't upgraded yet.

kianenigma: This has been removed altogether.

runtime/rococo/src/lib.rs
```diff
@@ -145,7 +145,7 @@ pub type UncheckedExtrinsic = generic::UncheckedExtrinsic<Address, Call, Signatu
 /// Extrinsic type that has already been checked.
 pub type CheckedExtrinsic = generic::CheckedExtrinsic<AccountId, Nonce, Call>;
 /// Executive: handles dispatch to the various modules.
-pub type Executive = frame_executive::Executive<Runtime, Block, frame_system::ChainContext<Runtime>, Runtime, AllModules>;
+pub type Executive = frame_executive::Executive<Runtime, Block, frame_system::ChainContext<Runtime>, Runtime, AllModules, ()>;
```
bkchr, line 148:
```suggestion
pub type Executive = frame_executive::Executive<Runtime, Block, frame_system::ChainContext<Runtime>, Runtime, AllModules>;
```

runtime/westend/Cargo.toml
```diff
@@ -147,6 +148,7 @@ std = [
 	"sp-session/std",
 	"pallet-randomness-collective-flip/std",
 	"runtime-common/std",
+	"frame-try-runtime/std",
```
bkchr, line 151: Still not sure how this should work, but okay :P

cli/src/command.rs
```diff
@@ -276,6 +276,31 @@ pub fn run() -> Result<()> {
...
+			let chain_spec = &runner.config().chain_spec;
+			set_default_ss58_version(chain_spec);
+
+			runner.async_run(|config| {
+				use sc_service::TaskManager;
+				let registry = config.prometheus_config.as_ref().map(|cfg| &cfg.registry);
+				let task_manager = TaskManager::new(
+					config.task_executor.clone(),
+					registry,
+				).unwrap();
```
bkchr, line 291: Sir, an unwrap :P

cli/src/command.rs
```diff
@@ -276,6 +276,31 @@ pub fn run() -> Result<()> {
...
+
+				Ok((
+					cmd.run::<
+						service::kusama_runtime::Block,
+						service::KusamaExecutor,
```
bkchr, line 296: The correct way for this would be an if that checks which version we need to call.

bkchr (approved):

runtime/kusama/src/lib.rs
```diff
@@ -1071,7 +1072,7 @@ pub type Executive = frame_executive::Executive<
 	frame_system::ChainContext<Runtime>,
 	Runtime,
 	AllModules,
-	(UpgradeSessionKeys, PhragmenElectionDepositRuntimeUpgrade)
+	(UpgradeSessionKeys, PhragmenElectionDepositRuntimeUpgrade,),
```
shawntabrizi, line 1075:
```suggestion
	(UpgradeSessionKeys, PhragmenElectionDepositRuntimeUpgrade),
```

shawntabrizi (approved):

runtime/westend/Cargo.toml
kianenigma: yeah these can be removed, they do nothing afaik.

cli/src/command.rs
kianenigma: Hmm strangely then the two features will have incompatible return types (tries coercing both `Ok` and `Err` with `map_ok(|_| ())` and `map_err(|_| ())`). Will play a bit more with it later today, might make an issue for it and leave for some other time.

runtime/westend/Cargo.toml
thiolliere: they are needed if you want to compile the crate with std and try-runtime features.
the cli is able to compile it because it activate the frame-try-runtime/std feature through try-runtime-cli

runtime/kusama/src/lib.rs
```diff
@@ -1067,7 +1067,11 @@ pub type Executive = frame_executive::Executive<
 	frame_system::ChainContext<Runtime>,
 	Runtime,
 	AllModules,
+<<<<<<< HEAD
+	(UpgradeSessionKeys, PhragmenElectionDepositRuntimeUpgrade),
+=======
 	()
+>>>>>>> c66049ddda26c9609d1a2cf23c0975a1f9677712
```
thiolliere, line 1074:
```suggestion
```

PR #2449, opened 2021-02-16
[CI] Fix check-tx-version Gitlab job

s3krit:
Makes the artifacts available to the check-transactions-version job which was causing an issue reported by @andresilva . In addition, we now also fetch the latest released version of Polkadot and compare the metadata with *that*.

Prior to this PR, we compare the metadata with whatever is live *on-chain*, which means if we have released a new version of polkadot but the runtime on Westend, Kusama or Polkadot has not been upgraded yet (as is currently the case on Polkadot), it could potentially complain about changed extrinsics. 

I'll make a followup PR at some point that checks more than just the *ordering* of the extrinsics (like new/removed modules, what have you) but that's dependent on me getting a more machine-readable summary of metadata..

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -157,7 +157,7 @@ check-transaction-versions:
   <<:                              *docker-env
   needs:
     - job:                         test-linux-stable
-      artifacts:                   false
+      artifacts:                   true
   before_script:
     - npm install --ignore-scripts -g @polkadot/metadata-cmp
     - git fetch origin release
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -159,6 +159,7 @@ check-transaction-versions:
     - job:                         test-linux-stable
       artifacts:                   true
   before_script:
+    - apt-get -y update; apt-get -y install jq
     - npm install --ignore-scripts -g @polkadot/metadata-cmp
     - git fetch origin release
   script:
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -159,7 +159,7 @@ check-transaction-versions:
     - job:                         test-linux-stable
       artifacts:                   true
   before_script:
-    - apt-get -y update; apt-get -y install jq
+    - apt-get -y update; apt-get -y install jq lsof
     - npm install --ignore-scripts -g @polkadot/metadata-cmp
     - git fetch origin release
   script:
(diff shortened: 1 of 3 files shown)


TriplEight:
I somehow missed that it needs a binary

TriplEight (approved):
LGTM

scripts/gitlab/check_extrinsics_ordering.sh
```diff
@@ -1,17 +1,31 @@
 #!/bin/bash
```
bkchr, line 1:
```suggestion
#!/usr/bin/env bash
```

scripts/gitlab/check_extrinsics_ordering.sh
```diff
@@ -32,19 +46,17 @@ for RUNTIME in "${runtimes[@]}"; do
...
-
-  # Start running the local polkadot node in the background
-  $BIN --chain="$RUNTIME-local"  &
+  # Start running the nodes in the background
+  $HEAD_BIN --chain="$RUNTIME-local" --tmp &
+  $RELEASE_BIN --chain="$RUNTIME-local" --ws-port 9945 --tmp &
   jobs
 
+  # Sleep a little to allow the nodes to spin up and start listening
+  sleep 5
```
bkchr, line 55: What happens when it takes more than 5 seconds?

TriplEight: maybe it makes sense to wait for a certain reaction from a node.

s3krit: Good point. Added a check for listening ports that times out after X seconds (currently 5). In reality, if the nodes are working correctly they'll usually start listening after < 1s but this'll also pick up if something's wrong with the binary (though one would hope that would have been detected in test-linux-stable)

bkchr (approved):

.gitlab-ci.yml
```diff
@@ -159,6 +159,7 @@ check-transaction-versions:
     - job:                         test-linux-stable
       artifacts:                   true
   before_script:
+    - apt-get -y update; apt-get -y install jq
```
TriplEight: it's a bad, bad security practice to update the pkg manager within a container,
it's also bad to install something every time the job runs,
why not just add jq to the [CI image](https://github.com/paritytech/scripts/blob/master/dockerfiles/ci-linux/Dockerfile#L26)?

PR #2450, opened 2021-02-16
Make sure the adder collator fails when something is wrong

bkchr:
As futures most of the time are catching panics and we don't check this,
it could happen that we have some statement that isn't correct but the
test succeeds successfully.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -9215,7 +9215,6 @@ dependencies = [
 name = "test-parachain-adder-collator"
 version = "0.7.26"
 dependencies = [
- "assert_matches",
  "futures 0.3.12",
  "futures-timer 3.0.2",
  "log",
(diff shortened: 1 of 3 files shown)


ordian (approved):
> As futures most of the time are catching panics and we don't check this,
it could happen that we have some statement that isn't correct but the
test succeeds successfully.

This is a bit worrying, do we have other such places in tests?

rphmeier:
I didn't expect that futures will catch panics without an explicit `catch_panic()` future added. Is it something about the Substrate task executor?

bkchr:
> I didn't expect that futures will catch panics without an explicit `catch_panic()` future added. Is it something about the Substrate task executor?

Sort of this https://github.com/tokio-rs/tokio/issues/495 + what Substrate is doing.

bkchr:
> This is a bit worrying, do we have other such places in tests?

If you are using the single threaded executors it is fine, they return the panic as error.

This panic here is also returned as error, but we swallow it in Substrate.

bkchr:
Most often you also don't do what we do here in the test. We constantly spawn new futures to record some value. So, it could happen the value is once incorrect, but all the other times correct and the test would succeed. However, we want to catch this "once" the value was incorrect.

In most tests you probably spawn stuff and wait somewhere on a result. If now this spawned stuff panics somewhere, your test will not finish.

bkchr:
bot merge

PR #2451, opened 2021-02-16
Reduce Kusama Fees / Deposits by 50x

shawntabrizi:
This PR updated the value of `DOLLARS` by 50x, which should ultimately reduce all the fees in Kusama by 50x too.

Did I miss anything else that needs to be done for this?

Need to be careful that there are no runtime assumptions that would break if the dollar value changes (like what happened with elections).

cc @rrtti


shawntabrizi:
Note that this PR will NOT retroactively reduce the fees already taken by the runtime, only reduce new fees.

So for many people to take advantage of this PR, they would need to undo and then rdo some operation, for example unclaim an index and claim it again to take advantage of the new fees.

There may be a few pallets where we want to proactively return some fees. If people can identify those needs, we can write migration code to handle that.

kianenigma:
Phew (hoping that we have recorded deposits everywhere now).

Also, I proposed picking up the fee multiplier again as well for the next release https://github.com/paritytech/polkadot/issues/2464.

kianenigma:
Also wonder where the 50 is coming from and isn't it a bit too much for one upgrade?

xlc:
Should we have some governance discussion about this? Or did I missed it?

AlistairStewart:
It would really help if we had a list of where DOLLAR was used and the concrete values of the the fees/reserve amounts are, so we could judge whether this was a good idea.

I know @Lederstrumpf was looking into this.

Lederstrumpf:
> It would really help if we had a list of where DOLLAR was used and the concrete values of the the fees/reserve amounts are, so we could judge whether this was a good idea.
> 
> I know @Lederstrumpf was looking into this.

I'll tally up an external list later, but you can get an overview from these searches:

https://github.com/search?l=Rust&q=DOLLARS+repo%3Aparitytech%2Fsubstrate&type=Code
https://github.com/search?q=DOLLARS+repo%3Aparitytech%2Fpolkadot+path%3Aruntime%2Fkusama+language%3ARust+language%3ARust&type=Code&ref=advsearch&l=Rust&l=Rust

joshua-mir:
> There may be a few pallets where we want to proactively return some fees. If people can identify those needs, we can write migration code to handle that.

I can only think of frozen indices as something that users won't be able to proactively rebalance themselves

Lederstrumpf:
> It would really help if we had a list of where DOLLAR was used and the concrete values of the the fees/reserve amounts are, so we could judge whether this was a good idea.
> 
> I know @Lederstrumpf was looking into this.

List of its concrete uses: https://hackmd.io/kK2BfY9MR1ay54WzJPNSPg

Swader:
Can the community do something to speed this merge up?

kianenigma:
If there are no pallets that take any bond/fee without recoding it on-chain, I think we can merge this right now. 

I can't think of a problem with all of the pallets that I work closely with.

Lederstrumpf:
> If there are no pallets that take any bond/fee without recoding it on-chain, I think we can merge this right now.
> 
> I can't think of a problem with all of the pallets that I work closely with.

Err, didn't see this and created a post on polkassembly in parallel to poll support for merging this PR:  https://kusama.polkassembly.io/post/552. If merging is uncontroversial (it's a significant adjustment, but likely well supported) and should happen fast, I can edit it to signal deprecation (can't delete).

christinammiyar:
Almost more controversial to not pass it fast as this is the biggest complaint people have right now with cryptocurrencies. Moving them around shouldn't be cost prohibative.

kianenigma (approved):
As far as I know all pallets record their deposit, so we should have no issue with unreserving. I can't see any other obvious issues here, and we should try this.

TomaszWaszczyk:
Let's merge it!

PR #2476, opened 2021-02-18
Enable two phase election provider in runtimes in 'DryRun' mode

kianenigma:
Almost the companion of https://github.com/paritytech/substrate/pull/7909, but not quite. 

- [ ] Run benchmark one last time, for now I copied the weigths from substrate 🚨
- [x] Revert dep updates.


runtime/kusama/src/lib.rs
```diff
@@ -1037,6 +1074,9 @@ construct_runtime! {
 
 		// Tips module.
 		Tips: pallet_tips::{Module, Call, Storage, Event<T>} = 36,
+
+		// Election pallet. Only works with staking, but placed here to maintain indices.
```
thiolliere, line 1078: not sure the comment make sense, this is just a normal pallet. nothing special here

runtime/polkadot/src/lib.rs
```diff
@@ -337,6 +337,42 @@ impl pallet_session::historical::Config for Runtime {
 	type FullIdentificationOf = pallet_staking::ExposureOf<Runtime>;
 }
 
+parameter_types! {
+	// no signed phase for now, just unsigned.
+	pub const SignedPhase: u32 = 0;
+	// NOTE: length of unsigned phase is, for now, different than `ElectionLookahead` to make sure
+	// that we won't run OCW threads at the same time with staking. Once staking's election is out,
+	// replace with `/ 4`.
+	pub const UnsignedPhase: u32 = EPOCH_DURATION_IN_BLOCKS / 8;
```
thiolliere, line 346: polkadot staking 
```suggestion
	pub const UnsignedPhase: u32 = EPOCH_DURATION_IN_BLOCKS / 32;
```
or if it works:
```
	pub const UnsignedPhase: u32 = ElectionLookAhead::get() / 2;
```

runtime/westend/src/lib.rs
```diff
@@ -319,6 +319,39 @@ impl pallet_session::historical::Config for Runtime {
 	type FullIdentificationOf = pallet_staking::ExposureOf<Runtime>;
 }
 
+parameter_types! {
+	// no signed phase for now, just unsigned.
+	pub const SignedPhase: u32 = 0;
+	pub const UnsignedPhase: u32 = EPOCH_DURATION_IN_BLOCKS / 4;
```
thiolliere, line 325: FYI this is same as ElectionLookAhead

runtime/kusama/src/lib.rs
kianenigma: I'm trying to reason about where it should be located, because that will influence the order of `on_initialize` and such

runtime/polkadot/src/lib.rs
kianenigma: yeah setting it to `ElectionLookAhead  / 2 makes more sense`

runtime/westend/src/lib.rs
kianenigma: I treat westend as a testnet :P but will make it same as polkadot now.

runtime/kusama/src/lib.rs
thiolliere: actually the indices is given by `= 36`, so we could put if after staking if wanted.
But maybe it is better to place if after Session actually, because it will use session information indirectly.

Also on_initialize order is reversed from declaration: https://github.com/paritytech/substrate/issues/6280

thiolliere (approved):

runtime/kusama/src/lib.rs
kianenigma: hmm actually 9 is empty, can I put it there and there won't be a need to update the transaction version still?

kianenigma: For now, I'll keep it as-is. As you said, not a big deal if we're 1 block off.

thiolliere: > hmm actually 9 is empty, can I put it there and there won't be a need to update the transaction version still?

yes you can, but there is no real need to reuse index as well `= 36` is good.

Issue #2464, opened 2021-02-17
decrease on_initialize weight + increase min weight-multiplier


kianenigma:
A follow-up on https://github.com/paritytech/polkadot/pull/1838:

- Check and if the `AVERAGE_ON_INITIALIZE` is still too much, decrease it a bit further to allow more nominators. 
- If the impact on fees is still within reason and not too significant, bump the weight feel multiplier one more time.

PR #2481, opened 2021-02-19
Change a few fee/weight parameters

kianenigma:
Closes https://github.com/paritytech/polkadot/issues/2464

### Fee 

You can run the `test_fees` to get a sense of why this change is not a big deal as long as the chain is not congested. Nonetheless, this will make the chain react faster to an increase of weight, if it happens. To that end, I am also considering proposing increasing the `variability_factor` of `TargetedFeeAdjustment`. But I thik for that I should first propose it to the council for review. 

These logs are for the weight-fee of one `transfer_keep_alive` call:

```
// Polkadot 
weight = "189,060,000" // multiplier = FixedU128(0.001000000000000000) // full transfer fee = "46,015,124"
weight = "189,060,000" // multiplier = FixedU128(1.000000000000000000) // full transfer fee = "61,124,800"
weight = "189,060,000" // multiplier = FixedU128(0.001000000000000000) // full transfer fee = "46,015,124"
weight = "189,060,000" // multiplier = FixedU128(0.000001000000000000) // full transfer fee = "46,000,015"
weight = "189,060,000" // multiplier = FixedU128(0.000000001000000000) // full transfer fee = "46,000,000"

// Kusama
weight = "188,971,000" // multiplier = FixedU128(0.001000000000000000) // full transfer fee = "766,918,387"
weight = "188,971,000" // multiplier = FixedU128(1.000000000000000000) // full transfer fee = "1,018,627,759"
weight = "188,971,000" // multiplier = FixedU128(0.001000000000000000) // full transfer fee = "766,918,387"
weight = "188,971,000" // multiplier = FixedU128(0.000001000000000000) // full transfer fee = "766,666,677"
weight = "188,971,000" // multiplier = FixedU128(0.000000001000000000) // full transfer fee = "766,666,426"
```

### Weight

A new analysis of weights show that the average weight weight is still 6B and 8B for polkadot and kusama respectively: 

Screenshot 2021-02-19 at 11 05 33

Which means the average `on_initialize` is 0.4% and 0.3% respectively, so we set it now to 1%. Note that this is again mostly to make the nominator community of polkadot temporarily bigger, while PRs to fix this fundamentally arrive. Afterwards, we can go back to numbers like 5% again.

### DIFF
--- a/runtime/common/src/lib.rs
+++ b/runtime/common/src/lib.rs
@@ -54,9 +54,9 @@ pub type NegativeImbalance<T> = <pallet_balances::Module<T> as Currency<<T as fr
 /// valid wasm module.
 pub const WASM_MAGIC: &[u8] = &[0x00, 0x61, 0x73, 0x6d, 0x01, 0x00, 0x00, 0x00];
 
-/// We assume that an on-initialize consumes 2.5% of the weight on average, hence a single extrinsic
-/// will not be allowed to consume more than `AvailableBlockRatio - 2.5%`.
-pub const AVERAGE_ON_INITIALIZE_RATIO: Perbill = Perbill::from_perthousand(25);
+/// We assume that an on-initialize consumes 1% of the weight on average, hence a single extrinsic
+/// will not be allowed to consume more than `AvailableBlockRatio - 1%`.
+pub const AVERAGE_ON_INITIALIZE_RATIO: Perbill = Perbill::from_percent(1);
 /// We allow `Normal` extrinsics to fill up the block up to 75%, the rest can be used
 /// by  Operational  extrinsics.
 const NORMAL_DISPATCH_RATIO: Perbill = Perbill::from_percent(75);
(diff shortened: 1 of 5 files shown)


runtime/common/src/lib.rs
```diff
@@ -77,7 +77,7 @@ parameter_types! {
 	/// Minimum amount of the multiplier. This value cannot be too low. A test case should ensure
 	/// that combined with `AdjustmentVariable`, we can recover from the minimum.
 	/// See `multiplier_can_grow_from_zero`.
-	pub MinimumMultiplier: Multiplier = Multiplier::saturating_from_rational(1, 1_000_000_000u128);
+	pub MinimumMultiplier: Multiplier = Multiplier::saturating_from_rational(1, 1_000u128);
```
shawntabrizi, line 80: Even if this is the correct value, I might prefer we reduce it by 1,000 over 2 upgrades than by 1,000,000 at once :)

kianenigma: cowboy prevention system activated 🤠 

fair enough.

kianenigma, line 80:
```suggestion
	pub MinimumMultiplier: Multiplier = Multiplier::saturating_from_rational(1, 1_000_000u128);
```

shawntabrizi (approved):
cool

Maybe make another issue to reduce by 1,000 again if needed

thiolliere (approved):
I don't know if it is really needed to increase the minimum multiplier.
If chain is congested the fast mechanism is tipping.
Otherwise looks good to me

PR #2489, opened 2021-02-22
Add Xcm sender.

### LINKED ISSUE #2405: Rococo should be able to send XCM
I was trying to do a cross-chain DOT transfer on rococo local network, but it failed and the dest parachain cannot receive `Xcm::ReserveAssetDeposit` message. It looks like to me the reason is Rococo V1 has [no XCM sender](https://github.com/paritytech/polkadot/blob/57eb9d10d874a27f2661523ce4b533aa690589c3/runtime/rococo/src/lib.rs#L550).

Rococo should be able to send XCM to help parachains on cross-chain DOT transfer.

shaunxw:
Fixes #2405


shawntabrizi:
This doesn't look like the right place for this.

You should implement such a thing here: https://github.com/paritytech/polkadot/tree/master/xcm

With a more specific struct name saying how it should be used, and generic usage of the xcm executor.

shaunxw:
> This doesn't look like the right place for this.
> 
> You should implement such a thing here: https://github.com/paritytech/polkadot/tree/master/xcm
> 

I thought this place is only for fundamental XCM definitions that could be used across relay chain and parachain impl.

The XCM sender defined in this PR is only for relay chain's use, as it only sends DMP. That's why I put it in common runtime lib. The sender in cumulus, which sends UMP and HRMP, is for parachain collators. IMHO the relay chain XCM sender might be in a place where less chance developers would use a wrong sender.

> With a more specific struct name saying how it should be used, and generic usage of the xcm executor.

Will update the naming and documentation.

shaunxw:
Another thing I forgot to mention is that this sender depends on `dmp` and `configuration` module. It's not possible make it in https://github.com/paritytech/polkadot/tree/master/xcm unless defining traits to abstract `dmp` and `configuration` behaviors.

shawntabrizi (approved):
I think this is fine for now then, but long term, I suspect this will be refactored somewhere else.

I don't have a great immediate suggestion.

runtime/common/src/xcm_sender.rs
```diff
@@ -0,0 +1,42 @@
+// Copyright 2017-2021 Parity Technologies (UK) Ltd.
```
shawntabrizi, line 1:
```suggestion
// Copyright 2021 Parity Technologies (UK) Ltd.
```

shaunxw:
Looks like this PR needs more approvals as @bkchr added *A0-pleasereview* label.

Please help to add reviewer. Thanks!

shaunxw:
@shawntabrizi @bkchr Could you please help to add reviewers?

shawntabrizi:
bot merge

PR #2491, opened 2021-02-22
Fix XCM encoding in HRMP pallet.

### LINKED ISSUE #2490: HRMP channel setup results 'xcmHandler.BadFormat' event in collators
Looks like collators cannot decode HRMP channel notifications message.

shaunxw:
Fixes #2490

### DIFF
--- a/runtime/parachains/src/hrmp.rs
+++ b/runtime/parachains/src/hrmp.rs
@@ -934,14 +934,14 @@ impl<T: Config> Module<T> {
 		<Self as Store>::HrmpOpenChannelRequestsList::append(channel_id);
 
 		let notification_bytes = {
-			use xcm::v0::Xcm;
+			use xcm::{v0::Xcm, VersionedXcm};
 			use parity_scale_codec::Encode as _;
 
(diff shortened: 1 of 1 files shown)


shawntabrizi (approved):

bkchr (approved):
Could you add some test for this?

pepyakin (approved):
Oh right, this is better

I am not sure what kind of test we can have here: a test that checks what bytes these dispatchables enqueue doesn't seem super useful: it won't allow us to catch any issues.

shaunxw:
We might need integration tests for this, and further more to cover HRMP/UMP/DMP sending and handling, between rococo and parachain.

If versioned XCM encode/decode is shared knowledge for all message passing modules (which I'm not sure if it's a protocol level thing), HRMP module could assume that these bytes must be at versioned encoded. Then adding versioned decoding works assertion in unit tests could help.

shawntabrizi:
Yes integration tests are on the TODO list which does the end to end for a bunch of XCM+P scenarios

PR #2492, opened 2021-02-22
Substrate companion for #8163

bkchr:
https://github.com/paritytech/substrate/pull/8163


node/core/bitfield-signing/src/lib.rs
```diff
@@ -268,10 +268,20 @@ impl JobTrait for BitfieldSigningJob {
...
-				.map_err(|e| Error::Keystore(e))?;
+				.map_err(|e| Error::Keystore(e))?
+			{
+				Some(b) => b,
+				None => {
+					tracing::error!(
+						target: LOG_TARGET,
+						"Key was found at construction, but while signing it could not be found.",
+					);
+					return Ok(());
```
andresilva, line 281: Is this fine? I assume that the task will stop running but the node will not go down.

andresilva (approved):

PR #2496, opened 2021-02-22
Remove v0.8.28 runtime migrations

s3krit:
Prior to beginning the v0.8.29 release process, we need to remove the old migrations. It's my first time doing this properly, so advice and feedback VERY welcome :)

### DIFF
--- a/runtime/westend/src/lib.rs
+++ b/runtime/westend/src/lib.rs
@@ -735,7 +735,7 @@ pub type Executive = frame_executive::Executive<
 	frame_system::ChainContext<Runtime>,
 	Runtime,
 	AllModules,
-	CustomOnRuntimeUpgrade,
+	()
 >;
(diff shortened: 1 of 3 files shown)


rphmeier (approved):
It seems right to me.

runtime/westend/src/lib.rs
```diff
@@ -764,20 +735,11 @@ pub type Executive = frame_executive::Executive<
 	frame_system::ChainContext<Runtime>,
 	Runtime,
 	AllModules,
-	UpgradeSessionKeys,
+	CustomOnRuntimeUpgrade,
```
kianenigma, line 738:
```suggestion
	()
```

You can just remove `CustomOnRuntimeUpgrade` and pass either `()` here or just nothing will also work.

s3krit:
bot merge

PR #2500, opened 2021-02-23
Added RAM requirements for running tests

saki-osive:
The **running tests** given at :

```bash
cargo test --all
```

requires ~30 GiB of combined RAM/swap to run. This is nowhere mentioned for the user. This PR addresses this.

### DIFF
--- a/README.md
+++ b/README.md
@@ -183,6 +183,8 @@ cargo build # Builds all native code
 
 You can run the tests if you like:
 
+> :warning: **Running tests is a very memory-intensive process**. Before running the tests, make sure you have at least 30 GiB of RAM(plus swap) available.
+
 ```bash
 cargo test --all
 ```


shawntabrizi:
I don't think this is a requirement. It is just that rust will use as much as it can to make things run faster. So if you have a 32 GB machine, you will probably see near 100% utilization... but the same stuff will run on 8 GB machine too.

saki-osive:
Hi @shawntabrizi , thanks for your reply. Have you given this a try?
I thought the same way until I actually ran the tests, I tried it on 20 GiB RAM, plus 4GiB swap, my system halted and I had to reboot it and start all over again. Only when I increased my swap size to 20 GiB could I successfully run the tests. It is due to this very fact that I have created this PR. 

Everything else works with lower resources, just the _tests_ part I'm referring to specifically.

shawntabrizi:
@saki-osive no i haven't tried, but thanks for the clarification. I was making an assumption, but indeed if what you are saying is true then it is strange to me.

bkchr:
I have never seen anywhere in any project how much RAM tests will use. I also don't see any required on this. This depends completely on your system and how all tests are build. If you are unlucky as in your case and the late linking stages are done in parallel you will use more memory (inking is using a lot of RAM). However this can change between Rust versions or we add new tests.

TLDR: I don't see any reason to merge this and will close this.

saki-osive:
Hi @bkchr , thanks for your reply. If this is changed in future versions, we can update the documentation accordingly. For now, isn't it better to warn the user beforehand rather than depending on if more RAM will be used and potentially halting the system?

TL;DR: The warning is not to lure users away from running tests but to give them an indication that things can go bad and ask them to be prepared.

PR #2503, opened 2021-02-23
Burnin for Substrate#8171

tomaka:
Not intended to be merged.

This burnin is a bit special because I mostly want the `debug-assertions = true` in order to debug https://github.com/paritytech/substrate/issues/8171.


tomaka:
https://github.com/paritytech/substrate/pull/8181 has been merged and debug_assertions are now enabled by default on our canary nodes. As such, this burnin is no longer needed as it does the same thing as the canary nodes.

PR #2504, opened 2021-02-23
runtime: use babe as randomness source

andresilva:
Companion for https://github.com/paritytech/substrate/pull/8180. 

`RandomnessCollectiveFlip` is only used on test runtimes now. I did not remove the `RandomnessCollectiveFlip` modules from the runtimes but I think we should be able to now (is there any reason why we'd want to use `RandomnessCollectiveFlip`?).

Randomness source is set to `babe::RandomnessFromOneEpochAgo`, if we have stricter requirements we can change it to `RandomnessFromTwoEpochsAgo`.

### DIFF
--- a/runtime/kusama/src/lib.rs
+++ b/runtime/kusama/src/lib.rs
@@ -804,7 +804,7 @@ parameter_types! {
 impl pallet_society::Config for Runtime {
 	type Event = Event;
 	type Currency = Balances;
-	type Randomness = RandomnessCollectiveFlip;
+	type Randomness = pallet_babe::RandomnessFromOneEpochAgo<Runtime>;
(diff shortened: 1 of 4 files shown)


burdges:
> is there any reason why we'd want to use RandomnessCollectiveFlip

Nope

rphmeier (approved):

shawntabrizi (approved):

andresilva:
@tomusdrw This PR contains changes that should be upstreamed to the bridges repo. They are isolated on commit https://github.com/paritytech/polkadot/pull/2504/commits/bab31d2d914a26ad1c286766db7638fbbd6db57a.

rphmeier (approved):

andresilva:
bot merge

PR #2515, opened 2021-02-24
Adding Bridges code as git subtree.

tomusdrw:
This PR adds entire https://github.com/paritytech/parity-bridges-common repository as git subtree.

The goal of having Bridges code in Polkadot repo is to make it possible to integrate with polkadot-based runtimes (Rococo & Westend first, later Polkadot & Kusama), while achieving following properties:
1. Keeping Bridges & Polkadot development process independent and focused (especially reducing notifications fatigue).
2. Reducing maintenance costs for Polkadot developers (coordinating substrate version bumps or substrate version changes).

I expect this integration will have a bunch of bumps on it's way, but it's still the best what we get to satisfy above conditions. I know this is something we have never really tried before and the PR is quite significant, despite that I feel it's worth to try.

[How to work with `git subtree`?](https://github.com/paritytech/polkadot/pull/2515/files#diff-ddbac2576db135434cb3a397d15ac6a8099de53b109a0a68273b6438aec406a0)

#### Why git subtree?
<details>
We did a short overview of possible solutions internally, and a summary can be found below (ordered by how closely the code is integrated):

1. Depending on `crates.io` versioning - not feasibly for now, given our fast development pace.
2. Using git-based dependencies - high maintenance cost for substrate devs (any breaking change requires PRs to bridges repo first, before the companion PR can be merged).
3. Using `git submodule` - this may lead to including unreviewed (neither in Polkadot nor in Bridges) code (just referencing commit hash of bridges repo) and may be hard to track versioning-wise (i.e. "Does this hash contain that important change or is it done on some other fork of the code").
4. Using `git subtree` - the code is duplicated in both repos, there might be some unused/unreferenced code lingering in Polkadot repo.
5. Merging the repos - merging development practices, mixing notification/issues/prs which might make it hard to navigate and move fast.

</details>

#### Why not remove unused bridges files?
<details>
I expect this will make it much harder to contribute changes upsteam, cause you may run into many conflicts due to files being removed. I only removed `Cargo.toml`, because nested workspaces confuse `cargo`, but since it's just one file I don't think it will cause much trouble.
</details>

#### What about BEEFY / other projects
<details>
Bridges are already quite big and complex and break quite often when Substrate changes. I expect BEEFY to be referenced using git-based cargo dependencies short term and integrated with Substrate repo long term.
</details>

#### Why not integrate BRIDGES with Substrate repo?
<details>
Similar arguments apply if we were to integrate bridges to Substrate (see "Why git subtree?"). Moreover bridges have quite a lot of chain-specific components, which I feel makes them even less of a fit to generic framework repository.
</details>


BRIDGES.md
```diff
@@ -0,0 +1,48 @@
+# Using BRIDGES dependency (`git subtree`).
```
HCastano, line 1: I wouldn't stylize this in all caps - it's not an acronym. I'd write this as: "Using the Parity Bridges Common Dependency", and use either "Bridges" or `parity-bridges-common` in the rest of the doc.

BRIDGES.md
```diff
@@ -0,0 +1,48 @@
...
+
+# How to fix broken BRIDGES code?
+
+To fix BRIDGES code simply create a commit in current (`polkadot`) repo. Best if
+the commit is isolated to changes in `./bridges` sub-directory, because it makes
+it easier to import that change back to upstream repo.
+
+# How to pull latest BRIDGES code or contribute back?
+
+Note that it's totally fine to ping the Bridges Team to do that for you, the point
```
HCastano, line 14:
```suggestion
Note that it's totally fine to ping the Bridges Team to do that for you. The point
```

BRIDGES.md
```diff
@@ -0,0 +1,48 @@
...
+all into one.
+
+
+3. Contributing back to BRIDGES (creating upstream PR)
+```
+$ git subtree push --prefix=bridges my-bridges master
+```
+This command will push changes to your personal fork of Bridges repo, from where
+you can simply create a PR to the main repo.
+
```
HCastano, line 48:
```suggestion
```

BRIDGES.md
```diff
@@ -0,0 +1,48 @@
...
+it easier to import that change back to upstream repo.
+
+# How to pull latest BRIDGES code or contribute back?
+
+Note that it's totally fine to ping the Bridges Team to do that for you, the point
+of adding the code as `git subtree` is to **reduce maintenance cost** for Polkadot
+developers.
+
+If you still would like to either update the code to match latest code from the repo
+or create an upstream PR read below.
```
HCastano, line 19:
```suggestion
or create an upstream PR read below.

Note that the following commands should be run in the current (`polkadot`) repo.
```

Cargo.toml
```diff
@@ -26,6 +26,9 @@ tempfile = "3.2.0"
 
 [workspace]
 members = [
+	"bridges/primitives/kusama",
+	"bridges/primitives/polkadot",
+	"bridges/primitives/runtime",
```
HCastano, line 31: This reminds me - we have a branch called `bridges-integration` which already adds some Polkadot and Kusama primitives. While it's a bit outdated (last commit was Dec 22nd), we should probably use it as a reference for integration moving forward.

BRIDGES.md
```diff
@@ -0,0 +1,48 @@
...
+
+2. To update BRIDGES:
+```
+$ git fetch bridges master
+$ git subtree pull --prefix=bridges bridges master --squash
+````
+
+We use `--squash` to avoid adding individual commits and rather squashing them
+all into one.
+
```
HCastano, line 40:
```suggestion
```

Cargo.toml
tomusdrw: Yes, I'll add integration in a separate PR cause it should be properly reviewed. This one is 65k lines of code that was already reviewed in bridges repo, hence the separation.

andresilva (approved):
I don't really have any strong opinions on this, I haven't done any research on it anyway so I just trust your judgment. Either way, I think this is the way to go in the sense that we should just test this approach and if it doesn't work we can revert it later, otherwise there will be opinions floating around with nothing getting decided in practice.

bkchr (approved):
What @andresilva says

PR #2520, opened 2021-02-24
Make sure we generate a valid slot in tests

bkchr:
Currently the first last timestamp is `0` and that leads to the first
slot being `0` which is reserved for genesis. There is some debug assert
in BABE that complains about this in Cumulus :D


andresilva (approved):

andresilva:
bot merge

PR #2525, opened 2021-02-25
some initial spans for approval voting

rphmeier:


node/core/approval-voting/src/lib.rs
```diff
@@ -1131,7 +1148,18 @@ fn process_wakeup(
 	state: &State<impl DBReader>,
 	relay_block: Hash,
 	candidate_hash: CandidateHash,
+	expected_tick: Tick,
 ) -> SubsystemResult<Vec<Action>> {
+	let mut span = polkadot_node_jaeger::descriptor_span(
```
drahnr, line 1153: Wouldn't a `hash_span` work as well? Or is the intention to create a different descriptor than what `hash_span` create intentionally?

node/core/approval-voting/src/lib.rs
```diff
@@ -592,6 +595,10 @@ async fn handle_approved_ancestor(
 
 	if target_number <= lower_bound { return Ok(None) }
 
+	span.add_stage(JaegerStage::ApprovalChecking);
+	span.add_string_tag("target-number", &format!("{}", target_number));
```
drahnr, line 599: I'll do another sweep and see if we can use smth like `add_tag(&str, impl ToString)` and another for `fmt::Debug` impl. That would allow easier verification of consistent formatting and get rid of a lot of `&format!(..)` macros.

node/core/approval-voting/src/lib.rs
```diff
@@ -576,6 +578,7 @@ async fn handle_approved_ancestor(
 
 	use bitvec::{order::Lsb0, vec::BitVec};
 
+	let mut span = polkadot_node_jaeger::hash_span(&target, "approved-ancestor");
```
drahnr, line 581: If the span exists, it would make sense to add the `Stage` right away, mostly for consistency.

node/jaeger/src/lib.rs
```diff
@@ -381,6 +383,15 @@ pub fn hash_span(hash: &Hash, span_name: &'static str) -> Span {
 	span
 }
 
+/// Creates a `Span` referring to the given descriptor, which should be unique.
+#[inline(always)]
+pub fn descriptor_span(descriptor: impl Encode, span_name: &'static str) -> Span {
```
drahnr, line 388: :+1:

drahnr (approved):
Two tiny nits, a few ideas how to make the `&format!(..)` disappear, but that's for another PR

node/core/approval-voting/src/lib.rs
rphmeier: It wouldn't work , I assume, because it wouldn't be unique.

rphmeier:
bot merge

PR #2542, opened 2021-02-28
Update FRAME System Weights

shawntabrizi:
Follow up to: https://github.com/paritytech/polkadot/pull/2541


shawntabrizi:
I reran the benchmarks and looked at the raw results, the effect of the linear increase on the hash function overtakes any constant weight component. From the eyes of the min squares analysis, a best fit of 0 constant value is right. This should be fine since we add to any weight the "base weight" of a call, so a remark of no bytes will still have a weight, but yeah results are not exactly perfect

PR #2550, opened 2021-03-03
Add Participation Lotteries to Kusama

shawntabrizi:
This PR adds the Lottery Pallet to Kusama.

It has the following configuration:

* Randomness from BABE
* Managed by ROOT or Half of Council
* Up to 10 Lottery Tickets per Lotto (10 Calls Max)
* Lottery Storage used for Valid Calls
* Up to 10 random numbers generated to avoid modulo bias

After this pallet is included in the Kusama runtime upgrade, a Council vote needs to start which will batch two calls:

1. set_calls
2. start_lottery

I recommend the following configurations to start:

1. set_calls
    1. Staking -> Nominate (Encourage updating/setting nominations in staking)
    2. Democracy -> Vote (Encourage voting on Democracy votes)
    3. Elections Phragmen -> Vote (Encourage voting for council members)
2. start_lottery
    1. price: .1 KSM
    2. length: 1 or 2 weeks 
    3. delay: 1 hour
    4. repeat: true

More details about the Lottery Pallet can be found here:
* https://github.com/paritytech/substrate/blob/master/frame/lottery/src/lib.rs
* https://kusama.polkassembly.io/post/242


rphmeier:
How long does the lottery wait after all tickets are in before drawing?

shawntabrizi:
> How long does the lottery wait after all tickets are in before drawing?

This is configurable when starting the lottery. I suggested 1 HR which is the Epoch time on Kusama. That is what we need right?

rphmeier:
@shawntabrizi it depends which randomness you're using - are you using `RandomnessFromOneEpochAgo` or `RandomnessFromTwoEpochsAgo`?

rphmeier:
Well if you're at all worried about validators reorganizing the chain to influence the lottery, you should use `RandomnessFromTwoEpochsAgo`. Otherwise `RandomnessFromOneEpochAgo` is fine.

burdges:
We use `RandomnessFromTwoEpochsAgo` when we need finality on whatever commitments the random computation acts upon, so normally only consensus protocols like BABE.  

`RandomnessFromOneEpochAgo` suffices if we do not require finality.  It's true some actions this incentivizes require finality.  I doubt this lottery incentive requires finality however, unless I missed some off chain thing among the goals.

shawntabrizi:
stale

Issue #2203, opened 2021-01-05
XCMP vs PoW 


burdges:
We've chosen this XCMP message transport of requesting your incoming messages from the approval checkers for the sending block, which provides relative simplicity, performance, and reasonable security without nasty never-used-buggy fallback pathways.

A priori, it creates a DoS vector if all parachain nodes could demand message from validators all day, which we'll eventually plug by requests containing a proof that they've an upcoming block.  As you can imagine this makes sense for Aura, Babe, and Sassafras by adding a second entry point into the PVF that checks a block creation credential early.  It makes little sense for PoW however.

We'd eventually support XCMP for PoW chains, either via 
- some layer-two-ish payment tricks that satisfy this second PVF entry point with money, or else
- via attaching the messages outside fully formed blocks so that the next block producer processes them, so the second PVF entry point is satisfied by the block existing and then gives you stuff to append.  

It's kinda overly complex to add either right now, so..  

@rphmeier  Can we just ban PoW parachains from XCMP until we've more time to work out how we want this done, i.e. think about layer-two-ish options?

AlistairStewart:
Even for Sassafras/BABE, it's not ideal to reveal your credential early, because then you can be DOSed.

burdges:
It's trouble for praos/babe yes but you request anonymously in sassafras.

rphmeier:
> Can we just ban PoW parachains from XCMP until we've more time to work out how we want this done, i.e. think about layer-two-ish options?

I don't know how to do that :(

burdges:
An approach goes:  

Initially, we permit requests from validators freely, but support rate limiting by validators.  At this stage, any parachain could be captured, censored, or locked by adversaries hoarding XCMP requests.

We then later provide some recognition for the babe, aura, and sassafras block production capabilities as cumulus provides them.  At this state, we warn PoW parachains that we corrected the XCMP problem only for babe, aura, and sassafras, and suggest they change block production mechanisms.

burdges:
We should ensure validators only check one collation per collation authorization, meaning they should mostly ignore collators who equivocate in parachain block production.

burdges:
I think opaque works, but so does `(relay_parent_height,collator_id)`, no?

rphmeier:
outdated

PR #2555, opened 2021-03-04
Ideas about candidate selection

JoshOrndorff:
I've been thinking about candidate selection recently and had a few ideas I thought worth sharing. Feedback welcome!

cc @girazoki


JoshOrndorff:
Well, I guess the runtime API idea wouldn't be so straightforward. Validators don't know the parachain state, so they wouldn't be able to execute the runtime api (except in the case that it's a pure function). I guess that could be solved with some additional state proof, but that's adding more machinery than I initially thought.

burdges:
Yes, bridge candidates necessarily prove target chain finality.

bkchr:
This is nothing that the relay chain should care about. You can do whatever block author verification that you want.

girazoki:
The runtime API in cases like BABE might not be able to provide the set of approved authors right? I mean collator eligibility depends on each collator's private key.

girazoki:
I like the GRANDPA-style game but I think it might only work for a small and bounded set of collators (in the end validators most likely will need to verify a candidate was backed by 2/3 of the total collators).

burdges:
There are two basically unrelated questions here: 

There exist parachains with pre-collation finality like bridges, consortiums, or even faster payment chains, since polkadot trades finality speed for throughput.  You might place some finality proofs into block headers, but actually validating them proceeds like any other parachain logic, not our problem.

Avoid spam by too many parachain collators all opening connections with parachain backing checker groups and the approval checkers who hold their incoming XCMP messages.  We've discussed various solutions and now the conversation lives in https://github.com/paritytech/polkadot/issues/2203  

We're obviously limited in what block production strategies we'll support, initially aura and later sassafras, but we're only hostile to strategies without pre-authorization proofs, like proof-of-work and [fantomette](https://arxiv.org/pdf/1805.06786.pdf), and likely something works even there.

JoshOrndorff:
Thanks for the discussion and the links!

I wasn't aware that there may be some special treatment for engines that work similarly to Babe or Aura. That approach makes sense, and I'll follow the issue.

Thanks also for the observation that my grandpa-before-sending-to-validators idea can be conceived of as a bridge. I hadn't realized that.

burdges:
Yes, if a parchain has enough validators for grandpa then they can achieve 6ish second finality before even submitting their block to polkadot, which maybe they desire.  I've not thought about how this could interact with backing, but maybe doing something there avoids adding 6 seconds before polkadot submission.  If so, that's slightly different than a bridge.

PR #1719, opened 2020-09-14
Companion for sub/7040

kianenigma:

PR #2562, opened 2021-03-04
Fix wrong deposit amount in council voters.

kianenigma:
As pointed out correctly by @AurevoirXavier

In https://github.com/paritytech/polkadot/pull/1719/files#diff-e5e76e02c0d16e79c70b024cbe3c6ea56f3249382a0f987ba203c34fcb40ed66 we made a mistake and passed in an incorrect value (5 CENT) as the old voting bond to the migration, instead of the correct value, 5 DOLLARs. 

Thus, currently all council voters have 50 milli-dots in deposit recorded for them, whereas it should be 5 DOTs.

Screenshot 2021-03-04 at 18 50 17

- [ ] Do a cost analysis for the number of voters that we can support within 2 blocks, and increase bonds if needed.


kianenigma:
Once released, I'll do another damage analysis and find the voter's who got affected by this bug (anyone who changed or removed their vote sine runtime 27) and find ways to refund them.

shawntabrizi (approved):

runtime/kusama/src/lib.rs
```diff
@@ -955,13 +955,6 @@ impl pallet_proxy::Config for Runtime {
 	type AnnouncementDepositFactor = AnnouncementDepositFactor;
 }
 
-pub struct CustomOnRuntimeUpgrade;
-impl frame_support::traits::OnRuntimeUpgrade for CustomOnRuntimeUpgrade {
-	fn on_runtime_upgrade() -> frame_support::weights::Weight {
-		0
-	}
-}
```
thiolliere, line 963: this code wasn't use in executive 😮
We should really not keep dead code like this, either add it to executive, or like you did remove completly

runtime/polkadot/src/lib.rs
```diff
@@ -1088,9 +1081,17 @@ impl frame_support::traits::OnRuntimeUpgrade for FixPolkadotCouncilVotersDeposit
 			},
 		);
 
-		frame_support::debug::info!(target: "runtime::polkadot", "updated {} + {} voter's deposit.", updated, skipped);
+		log::info!(target: "runtime::polkadot", "updated {} + {} voter's deposit.", updated, skipped);
 		BlockWeights::get().max_block
 	}
+
+	fn pre_upgrade() -> Result<(), &'static str> {
+		todo!()
```
thiolliere, line 1089: it should have a small comment about why it is not implemented and why it is not important for now.

runtime/polkadot/src/lib.rs
```diff
@@ -1088,9 +1081,17 @@ impl frame_support::traits::OnRuntimeUpgrade for FixPolkadotCouncilVotersDeposit
...
+	}
+
+	fn post_upgrade() -> Result<(), &'static str> {
+		todo!()
```
thiolliere, line 1093: it should have a small comment about why it is not implemented and why it is not important for now.

runtime/polkadot/src/lib.rs
```diff
@@ -1063,11 +1056,44 @@ pub type Executive = frame_executive::Executive<
...
+	fn on_runtime_upgrade() -> Weight {
+		use pallet_elections_phragmen::Voter;
+		use frame_support::IterableStorageMap;
+		let mut updated = 0;
+		let mut skipped = 0;
+		pallet_elections_phragmen::Voting::<Runtime>::translate::<Voter<AccountId, Balance>, _>(
+			|_who, mut vote| {
+				if vote.deposit == 5 * CENTS {
+					// If their deposit is what we set by mistake
+					vote.deposit = 5 * DOLLARS;
```
thiolliere, line 1075: just to be sure CENTS hasn't changed since the wrong migration, and DOLLARS hasn't change since they reserved, isn't it ?

thiolliere (approved):
Just wonder about https://github.com/paritytech/polkadot/pull/2562/files#r594340154
Otherwise looks good

runtime/polkadot/src/lib.rs
kianenigma: I should double check that indeed.

runtime/kusama/src/lib.rs
kianenigma: yeah crazy.

runtime/polkadot/src/lib.rs
kianenigma: Line has not been changed for 10 months 


Screenshot 2021-03-17 at 12 24 32

runtime/polkadot/src/lib.rs
kianenigma: Removed actually.

kianenigma:
bot merge

kianenigma:
bot merge abort

kianenigma:
bot abort

kianenigma:
bot stop

kianenigma:
bot  please stop

kianenigma:
Well, I think I can no longer stop the bot. Not a big deal. I wanted to add a further test as well, but it is fine.

kianenigma:
bot help

kianenigma:
bot merge

xlc:
maybe you need to ask nicely. i.e. `sudo bot stop` 😆

kianenigma:
More description of what went wrong: 

In runtime [27](https://polkadot.subscan.io/block/3613564), we wrote a wrong deposit amount in `electionsPhragmen.voting` field of all voters. The real deposit (which was already reserved) was 5 DOTs, but we recorded 5 milli-DOTs. This means that the chain had reserved 5 DOTs, but the `electionsPhragmen` pallet _thought_ that it had reserved 5 milli-DOTs from all voters. 

This is problematic for two reasons: 

1. when you submit `vote`, [your new deposit is calculated based on the number of members that you vote for](https://github.com/paritytech/substrate/blob/e11a670b0310130c329ab32a6e16608f5712113d/frame/elections-phragmen/src/lib.rs#L713). Then the [difference of this value compared to your current recorded deposit](https://github.com/paritytech/substrate/blob/e11a670b0310130c329ab32a6e16608f5712113d/frame/elections-phragmen/src/lib.rs#L322) (in `electionsPhragmen.voting`) is either reserved or unreserved. 
2. when you submit `remove_voter`, the amount that you have in `electionsPhragmen.voting` is unreserved.

In runtime [30](https://polkadot.subscan.io/block/4876134) we partially fixed this. We iterated over all voters, and if any of them had 5 milli-DOTs recorded as deposit, we changed it to 5 DOTs. 

This means that: 

- if you never changed your voting between runtime 27 and 30, then you would be fine, and you should have correct deposit values now. 

- If you submitted `vote` in this period, then the chain computed a value for your deposit based on `fn deposit`, which is certainly more than 5 DOTs, and since that value is certainly more than your recorded 5 milli-DOTs, a large amount reserved again. Afterwards, a correct deposit value is written for you. Any further interaction is then correct. The damage here is: 4.95 DOTs are potentially left in the reserved that need to be moved to free balance now. Example: Alice has 5 DOTs reserved and the pallet thinks she has 0.05 DOTs reserved. She votes again, and the new deposit is 5.5 DOTs. The pallet should have reserved 0.5 DOTs, but in reality it did 5.45 DOTs, 4.95 of which is incorrect. 

- If you submitted `remove_voter`, you are basically facing the last part of the previous scenario. Alice has reserved 5 DOTs, but 0.05 is recorded in `electionsPhragmen.voting`. When she leaves, 0.05 DOTs are unreserved, and 0.95 are left.

emielvanderhoek:
We found 189 extrinsics of type ElectionsPhragmen.vote for polkadot runtimes 28, 29 & 30.

Data set here: https://gist.github.com/emielvanderhoek/0a6cf51393e8d22de364b777f98cd453

Please note we have not looked at nested calls, like: batches, proxies, multisig & schedules. These could theoretically have triggered council election votes too.

Lmk if this helps.

emielvanderhoek:
Updated gist: https://gist.github.com/emielvanderhoek/0a6cf51393e8d22de364b777f98cd453

We found 49 successfully executed extrinsics of type ElectionsPhragmen.vote for polkadot runtimes 28, 29 & 30.
We found 35 successfully executed extrinsics of type ElectionsPhragmen.remove_voter for polkadot runtimes 28, 29 & 30.

ical10:
Hi @kianenigma, I'd be happy to contribute, but as far I can digest, we will need to check those two types of actions related to voting. Am I correct? 
I can start looking at the extrinsics that @emielvanderhoek provided, but then I don't know what to look for and how to do it (perhaps using calls on the js api?).
Please advise on how to proceed.

kianenigma:
My description above is pretty clear about what we are looking for, can you tell me exactly which part of it you don't understand, or need help with? 

Note that I have done an analysis myself and don't want to push you too much in the direction of replicating what I did. Instead, Ideally, I want someone else to come up with a _different_ way of doing the analysis independently.

paradox-tt:
I counted 16 with balances of 0.05

jakehemmerle:
@kianenigma

Me and @a-jwc are on w3f teched and are looking into this

PR #2580, opened 2021-03-08
Replace XCM `Error::Undefined` usage

### LINKED ISSUE #2577: Replace XCM `Error::Undefined` usage with more explict errors
`Error::Undefined` is used in different places. It's hard to tell the exact reason of execution failure with this error returned. It should be replaced with explicit ones.

In currency adapter, I think it could be replaced with a `Error::FailedToTransactAsset(&'static str)` error, to represent all asset transact failures. As parachains are likely to have their own `TransactAsset` implementation, the `&'static str` could be helpful to provide details of parachain specific errors.

I've started [some work](https://github.com/shaunxw/polkadot/blob/sw%2Freplace-xcm-undefined/xcm/xcm-builder/src/currency_adapter.rs) to replace the `Undefined` error in currency adapter. If this direction makes sense, I'm happy to polish it and open a PR.

shaunxw:
Closes #2577

### DIFF
--- a/xcm/xcm-builder/src/currency_adapter.rs
+++ b/xcm/xcm-builder/src/currency_adapter.rs
@@ -15,10 +15,26 @@
 // along with Polkadot.  If not, see <http://www.gnu.org/licenses/>.
 
 use sp_std::{result, convert::TryInto, marker::PhantomData};
-use xcm::v0::{Error, Result, MultiAsset, MultiLocation};
+use xcm::v0::{Error as XcmError, Result, MultiAsset, MultiLocation};
 use sp_arithmetic::traits::SaturatedConversion;
 use frame_support::traits::{ExistenceRequirement::AllowDeath, WithdrawReasons};
 use xcm_executor::traits::{MatchesFungible, LocationConversion, TransactAsset};
+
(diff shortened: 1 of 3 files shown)


shawntabrizi (approved):

xcm/xcm-builder/src/currency_adapter.rs
```diff
@@ -15,11 +15,33 @@
 // along with Polkadot.  If not, see <http://www.gnu.org/licenses/>.
 
 use sp_std::{result, convert::TryInto, marker::PhantomData};
-use xcm::v0::{Error, Result, MultiAsset, MultiLocation};
+use xcm::v0::{Error as XcmError, Result, MultiAsset, MultiLocation};
 use sp_arithmetic::traits::SaturatedConversion;
 use frame_support::traits::{ExistenceRequirement::AllowDeath, WithdrawReasons};
 use xcm_executor::traits::{MatchesFungible, LocationConversion, TransactAsset};
 
+/// The asset transact errors.
```
shawntabrizi, line 23:
```suggestion
/// Asset transaction errors.
```

Issue #2579, opened 2021-03-08
Statement distribution should gossip to any full node


eskimor:
Right now, it only gossips with validators, but we also want collators to see statements.

rphmeier:
@bkchr You worked around this or planned to, relatively recently, right?

bkchr:
Is this again broken?

bkchr:
I fixed this after the last rococo restart.

bkchr:
https://github.com/paritytech/polkadot/pull/2430

rphmeier:
Not broken, just something that we are filing. Longer term, we do want collators to learn about _other_ blocks backed by validators which are not necessarily their own, so they can already import those blocks and build upon them. 

It's also related to something you've expressed in a few different places - that collators have only a small window to actually author a block. Observing backing statements and getting an idea of which parachain block will be backed will give the collator some more time to actually build upon the block, although the `collation-generation` subsystem will only gather it later.

rphmeier:
(re-categorizing as annoyance because the system will function without this being implemented, but we can make it better by addressing this)

bkchr:
> Not broken, just something that we are filing. Longer term, we do want collators to learn about _other_ blocks backed by validators which are not necessarily their own, so they can already import those blocks and build upon them. 


That is already solved. Collators start to import any block that is announced with the seconded statement attached. So that is nothing they need to extra ask the relay chain for. In the end we don't have any way to request a block, for that we don't know which node has this block. 

So from the parachain side, I don't see any need to see all seconded statements.

rphmeier:
Makes sense!

Will close this issue then.

eskimor:
Will fix the guide then, to reflect those changed requirements.

AlistairStewart:
> That is already solved. Collators start to import any block that is announced with the seconded statement attached. So that is nothing they need to extra ask the relay chain for. In the end we don't have any mechanism to request a block, for that we don't know any peer who has this block.
> 
> So from the parachain side, I don't see any need to see all seconded statements.

Well there may still be an issue. The next question is who announces the blocks with secondary statements attached? The backing validators? 

In the long run, we maybe don't want to just trust the backing validators or the original collator.

bkchr:
> Well there may still be an issue. The next question is who announces the blocks with secondary statements attached? The backing validators?

[...]

PR #2583, opened 2021-03-09
Mostly, let guide reflect #2579

eskimor:
Collators are now notified about their own collations being seconded, therefore statement distribution no longer needs to gossip to non validator nodes. See #2579

Also moved Arc creation into `if`, not that it should matter in practice.


rphmeier (approved):

rphmeier:
bot merge

PR #2585, opened 2021-03-09
Add some magic to signed statements and approval votes

rphmeier:
Different signed statements by validators weren't previously separable by magic. This change adds magic to the payloads so an upcoming dispute vote kind will not be used in place of backing statements.

I also changed the naming of `Candidate` statements in the table to match the `Seconded` we use elsewhere.

### DIFF
--- a/node/core/approval-voting/src/lib.rs
+++ b/node/core/approval-voting/src/lib.rs
@@ -697,7 +697,9 @@ fn approval_signing_payload(
 	approval_vote: ApprovalVote,
 	session_index: SessionIndex,
 ) -> Vec<u8> {
-	(approval_vote, session_index).encode()
+	const MAGIC: [u8; 4] = *b"APPR";
+
+	(MAGIC, approval_vote, session_index).encode()
 }
 
 // `Option::cmp` treats `None` as less than `Some`.
--- a/primitives/src/v0.rs
+++ b/primitives/src/v0.rs
@@ -715,7 +715,8 @@ impl From<CompactStatement> for CompactStatementInner {
 
 impl parity_scale_codec::Encode for CompactStatement {
 	fn size_hint(&self) -> usize {
-		CompactStatementInner::from(self.clone()).size_hint() + 4
+		// magic + discriminant + payload
+		4 + 1 + 32
 	}
 
 	fn encode_to<T: parity_scale_codec::Output + ?Sized>(&self, dest: &mut T) {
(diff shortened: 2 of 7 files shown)


primitives/src/v0.rs
```diff
@@ -677,27 +677,73 @@ pub struct ErasureChunk {
...
+			CompactStatement::Seconded(h) => CompactStatementInner::Seconded(h),
+			CompactStatement::Valid(h) => CompactStatementInner::Valid(h),
+			CompactStatement::Invalid(h) => CompactStatementInner::Invalid(h),
+		}
+	}
+}
+
+impl parity_scale_codec::Encode for CompactStatement {
+	fn size_hint(&self) -> usize {
+		CompactStatementInner::from(self.clone()).size_hint() + 4
```
drahnr, line 718:
```suggestion
		CompactStatementInner::from(self.clone()).size_hint() + MAGIC.len()
```

and we should assert in a unit test that the length is `4`, so that things never pass CI in the first place.

drahnr (approved):
Tiny nit, looks good!

node/core/approval-voting/src/lib.rs
```diff
@@ -697,7 +697,9 @@ fn approval_signing_payload(
 	approval_vote: ApprovalVote,
 	session_index: SessionIndex,
 ) -> Vec<u8> {
-	(approval_vote, session_index).encode()
+	const MAGIC: [u8; 4] = *b"APPR";
```
drahnr, line 700: Not sure if this is worth a type yet.

statement-table/src/generic.rs
```diff
@@ -66,7 +66,7 @@ pub enum Statement<Candidate, Digest> {
 	///
 	/// Broadcasting two different candidate messages per round is not allowed.
 	#[codec(index = 1)]
-	Candidate(Candidate),
+	Seconded(Candidate),
```
bkchr, line 69: This always confused me :D

primitives/src/v0.rs
bkchr: I would not call `CompactStatementInner::from(self.clone()).size_hint()` here.

You can just return `1 + 4 + 4` or completely ignore this function.

node/core/approval-voting/src/lib.rs
rphmeier: In what way is it different?

drahnr: Strike that, it would only different for enums.

rphmeier:
bot merge

PR #2586, opened 2021-03-09
Configurable maximum validators

rphmeier:
This adds a `configuration.max_validators` item which can be used to configure the parachains logic to use only some subset of all of the existing validators. Only this subset will appear within the session info. Validator indices will always refer to the subset.

### DIFF
--- a/roadmap/implementers-guide/src/runtime/shared.md
+++ b/roadmap/implementers-guide/src/runtime/shared.md
@@ -20,8 +20,14 @@ pub(crate) const SESSION_DELAY: SessionIndex = 2;
 ## Storage
 
 ```rust
-// The current session index within the Parachains Runtime system.
+/// The current session index within the Parachains Runtime system.
 CurrentSessionIndex: SessionIndex;
+/// All the validators actively participating in parachain consensus.
+/// Indices are into the broader validator set.
+ActiveValidatorIndices: Vec<ValidatorIndex>,
+/// The parachain attestation keys of the validators actively participating in parachain consensus.
+/// This should be the same length as `ActiveValidatorIndices`.
+ActiveValidatorKeys: Vec<ValidatorId>
 ```
 
 ## Initialization
@@ -35,8 +41,9 @@ them.
 
 ## Session Change
 
-During a session change, the Shared Module receives and stores the current Session Index for that
-block through the Session Change Notification.
+During a session change, the Shared Module receives and stores the current Session Index directly from the initializer module, along with the broader validator set, and it returns the new list of validators.
+
+The list of validators should be first shuffled according to the chain's random seed and then truncated. The indices of these validators should be set to `ActiveValidatorIndices` and then returned back to the initializer. `ActiveValidatorKeys` should be set accordingly.
 
 This information is used in the:
 
--- a/runtime/parachains/src/util.rs
+++ b/runtime/parachains/src/util.rs
@@ -50,7 +50,7 @@ pub fn take_active_subset<T: Clone>(active_validators: &[ValidatorIndex], set: &
 
 	if subset.len() != active_validators.len() {
 		log::warn!(
-			target: "parachains_runtime",
+			target: "runtime::parachains",
 			"Took active validators from set with wrong size",
 		);
 	}
--- a/runtime/parachains/src/util.rs
+++ b/runtime/parachains/src/util.rs
@@ -18,6 +18,7 @@
 //! on all modules.
 
 use primitives::v1::{Id as ParaId, PersistedValidationData, Hash, ValidatorIndex};
+use sp_std::vec::Vec;
 
 use crate::{configuration, paras, hrmp};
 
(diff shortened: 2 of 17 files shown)


roadmap/implementers-guide/src/runtime/shared.md
```diff
@@ -20,8 +20,14 @@ pub(crate) const SESSION_DELAY: SessionIndex = 2;
...
 
 ```rust
-// The current session index within the Parachains Runtime system.
+/// The current session index within the Parachains Runtime system.
 CurrentSessionIndex: SessionIndex;
+/// All the validators actively participating in parachain consensus.
+/// Indices are into the broader validator set.
+ActiveValidatorIndices: Vec<ValidatorIndex>,
+/// The parachain attestation keys of the validators actively participating in parachain consensus.
+/// This should be the same length as `ActiveValidatorIndices`.
```
ordian, line 29: we could store it as a `Vec` of pairs to enforce this, also it's a bit unexpected that tests set the validator keys, but not indices, which may result in some unexpected errors when running tests

ordian (approved):

runtime/parachains/src/util.rs
```diff
@@ -39,3 +40,8 @@ pub fn make_persisted_validation_data<T: paras::Config + hrmp::Config>(
 		max_pov_size: config.max_pov_size,
 	})
 }
+
+/// Take the active subset of a set containing all validators.
+pub fn take_active_subset<T: Clone>(active_validators: &[ValidatorIndex], set: &[T]) -> Vec<T> {
+	active_validators.iter().filter_map(|i| set.get(i.0 as usize)).cloned().collect()
```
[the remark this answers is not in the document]
rphmeier: Panic in the runtime means taking down the entire network with no avenue to recovery, even via governance, as it would be impossible to author any further blocks. I really don't want that.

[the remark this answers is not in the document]
eskimor: Ok, I can understand that. But just silently ignoring a logic error, makes it very easy to 1 make that logic error reach a production system, 2 make it potentially incredibly hard to debug the problem in the production system.

Uh! I have the solution - what about a debug_assert? Won't trigger in production, but will still help catch errors early.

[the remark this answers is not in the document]
rphmeier: We never run runtimes in debug mode anyway. I think we would notice quite quickly if the number of validators, when supposedly constrained to `max` is less than `max`.

roadmap/implementers-guide/src/runtime/shared.md
rphmeier: A `Vec` of pairs is just more unwieldy to use and may require additional allocation within the block execution

runtime/parachains/src/util.rs
```diff
@@ -39,3 +40,20 @@ pub fn make_persisted_validation_data<T: paras::Config + hrmp::Config>(
...
 }
+
+/// Take the active subset of a set containing all validators.
+pub fn take_active_subset<T: Clone>(active_validators: &[ValidatorIndex], set: &[T]) -> Vec<T> {
+	let subset: Vec<_> = active_validators.iter()
+		.filter_map(|i| set.get(i.0 as usize))
+		.cloned()
+		.collect();
+
+	if subset.len() != active_validators.len() {
```
ordian, line 51: the size doesn't have to match, right? what we want to check is that `subset.iter().max().map_or(false, |m| m >= active_validators.len())`

runtime/parachains/src/util.rs
```diff
@@ -39,3 +40,20 @@ pub fn make_persisted_validation_data<T: paras::Config + hrmp::Config>(
...
+		log::warn!(
+			target: "parachains_runtime",
```
ordian, line 53: we do this to allow matching by prefix, e.g. https://github.com/paritytech/polkadot/blob/f02d5e45ada77621954cb7ad3893cb81de657d15/runtime/parachains/src/runtime_api_impl/v1.rs#L88

```suggestion
			target: "runtime::parachains",
```

ordian: only on some builds https://github.com/paritytech/polkadot/pull/2522

runtime/parachains/src/util.rs
ordian: Ah, no, sorry, I confused a few names. Yes, it should be alright.

runtime/parachains/src/util.rs
eskimor: Uh, but default is on - that's good! So for Rococo, local tests and maybe even Westend we will have it switched on by default. I like.

runtime/parachains/src/util.rs
rphmeier: Isn't it the same thing? If any index is out of bounds, the subset will be missing entries due to the `filter_map` above.

eskimor (approved):

rphmeier:
bot merge

PR #2587, opened 2021-03-09
Initial integration of Gilts pallet (Kusama)

gavofyork:
Also bumps Substrate to use the latest Gilt pallet.


burdges:
We do need to slash gilts whenever validators get slashed, which never got implemented.

gavofyork:
Not sure what you mean here @burdges . The point of gilts is to ensure that the proportion of the capital remains the same over time. That's how this is implemented. With the inclusion of https://github.com/paritytech/substrate/pull/8299, then we can ensure that the treasury funds don't make it into the issuance calculation, so if stakers (or anyone else) on the system get slashed, then that slash won't form part of the issuance whose proportion is being honoured.

That said, if the treasury elects to spend the slashed funds and they go back into circulation, then the gilt holders will rightly "benefit", in so much as the issuance of which a proportion they hold will increase, no longer being in the treasury.

One thing that is added now is the `CappedTreasury` imbalance handler, which sends funds to the treasury only up to the amount that goes to the stakers. This means that even if the treasury spends its funds immediately, then the average staker will never do worse than the gilt-holder (and they could only do as bad as a gilt-holder in the rather degenerate case where the staking was so over-staked as to make the reward curve diip below 50% of maximum).

burdges:
I have not worked out any details, so maybe that suffices, but..

Inflation exists to pay stakers for validators, and staking incurs slashing risk, so either
1. we account for slashing risk in the delta between validator rewards and gilts inflation, or else
2. we slash gilts proportionally whenever any validator gets slashed.

We choose 2 if our goal is a smaller delta.  

We could also restate the question as:  Are gilts meant to be a "bet" that some slashing occurs?  If not, then slashing gilts ala 2 makes gilts slashing neutral, in the sense that if gilts get penalized like the average staker gets penalized. 

Actually there exist other options like changing the culture to refund more slashes, and thereby lower the slashing risks, but this dichotomy still roughly exists.

gavofyork:
> we account for slashing risk in the delta between validator rewards and gilts inflation, or else

this is one difference (and why staking rewards will generally be substantially greater merely retaining equal proportion of issuance). the other is that staking only requires a lockup of 7/28 days (kusama/polkadot). with gilts, the lockup could be many months or years, depending on the market. 

gilts are not a direct alternative to staking. whereas staking is expected to account for 50-75% of issuance and allow for as much capital to be placed in as is desired (with the only drawback being a very graceful reduction in rewards), gilts are strictly capped at some proportion of issuance (e.g. 10%) and their market bids instead on lock-time.

the deal cannot be characterised as "giving up rewards that outperform base inflation for the advantage of no slashing", since there is an auction for lock-duration. if market participants are overwhelmingly in favour of that particular tradeoff, then it just means that the lock-duration with be accordingly large. if the market prefers a small (and inherently avoidable) risk of slashing for rewards that outperform inflation then the lock-duration will be correspondingly smaller.

burdges:
> and their market bids instead on lock-time.

I missed this part before, interesting.  If Alice gilt locks for 1 year then can Bob kick Alice out by bidding a later expiry date?

runtime/kusama/src/lib.rs
```diff
@@ -350,13 +346,67 @@ pallet_staking_reward_curve::build! {
...
+			let fraction_over = amount_over / max_amount_over;
+			Perquintill::one() / 2 + fraction_over / 2
+		};
+		let reward_fraction = REWARD_CURVE.calculate_for_fraction_times_denominator(x, Perquintill::ACCURACY);
+		let validator_payout = portion * reward_fraction;
+
+//		let max_payout = portion * (REWARD_CURVE.maximum * total_issuance);
+//		let rest = max_payout.saturating_sub(validator_payout);
+
+		let (gilt_locked, _gilt_payout) = Gilt::issuance();
```
gavofyork, line 401: @AlistairStewart please take a look.

runtime/kusama/src/lib.rs
```diff
@@ -1016,6 +1131,9 @@ construct_runtime! {
 
 		// Election pallet. Only works with staking, but placed here to maintain indices.
 		ElectionProviderMultiPhase: pallet_election_provider_multi_phase::{Pallet, Call, Storage, Event<T>, ValidateUnsigned} = 37,
+
+		// Gilts module.
```
kianenigma, line 1135:
```suggestion
		// Gilts pallet.
```

runtime/kusama/src/lib.rs
```diff
@@ -361,36 +361,122 @@ sp_npos_elections::generate_solution_type!(
...
+
+	// 30% reserved for up to 60 slots.
+	let auction_proportion = Perquintill::from_rational(auctioned_slots.min(60), 200u64);
+	// The gilt proportion is just the gilt target adjusted down by 25%.
+	let gilt_proportion = Perquintill::from_percent(75) * gilt_target;
+	let ideal_stake = Perquintill::from_percent(75)
+		.saturating_sub(auction_proportion)
+		.saturating_sub(gilt_proportion);
+
+	let stake = Perquintill::from_rational(total_staked, non_gilt_issuance);
```
thiolliere, line 402: I don't completely understand why stake rate is `total_staked/non_gilt_issuance` and ideal_stake is 75% minus 75% of gilt target.
Gilt target says to be `"The target proportion of gilts within total issuance."`. Why isn't the stake rate `total_staked/total issuance (including gilt issuance)` then. (or why isn't ideal_stake just 75%)

But I think I misunderstand gilt target meaning.

thiolliere (approved):
apart from not understanding how we use gilt info in the ideal_stake https://github.com/paritytech/polkadot/pull/2587/files#r629483645 implementation is good to me

gavofyork:
@shawntabrizi final review appreciated.

runtime/kusama/src/lib.rs
```diff
@@ -361,36 +364,80 @@ sp_npos_elections::generate_solution_type!(
...
+}
+
+pub struct EraPayout;
+impl pallet_staking::EraPayout<Balance> for EraPayout {
+	fn era_payout(
+		total_staked: Balance,
+		_total_issuance: Balance,
+		era_duration_millis: u64,
+	) -> (Balance, Balance) {
+		const AUCTIONED_SLOTS: u64 = 0;
```
shawntabrizi, line 429:
```suggestion
		// TODO: Update with Auctions logic when auctions pallet added.
		const AUCTIONED_SLOTS: u64 = 0;
```

shawntabrizi (approved):
I did not dive deep into `ideal_stake`, but I assume Al will jump in.

Other parts look okay. Important TODO to make this stuff work w/ slots counter.

xlc:
How does teleported asset work with this? i.e. KSM teleported to Statemine. Am I right that teleported assert will be removed from total issuance?

PR #2603, opened 2021-03-10
Better error handling in approval-voting

rphmeier:
The subsystem would previously fail on spurious errors when the chain API or runtime API subsystems were unavailable. This changeset makes the subsystem more resilient to such errors by ignoring them. If we encounter such an error when importing a new section of the chain, we discard the entire section of the chain from the approval voting subsystem, with the expectation that upon the next block we'll surpass the spurious error.


ordian (approved):
Seems reasonable.

PR #2605, opened 2021-03-10
avoid reshuffling active validators

rphmeier:
There was a bug introduced in #2586 which caused a re-shuffle of validators in the `SessionInfo`, breaking everything that depends on `ValidatorIndex`.


ordian (approved):
Thanks for adding tests. I also missed that we pass shuffled validator in the notification in #2586.

burdges:
Cool.  It's handy if validators keep the same index from when they get elected to when they loose an election.  :)

rphmeier:
That'd be good, but this PR actually does not do that. The validator indices will change pretty much every session based on the random seed.

burdges:
We'll deal with that later then.  It'll be a while I think before we need it for anything.

PR #2613, opened 2021-03-11
Add a memo field to crowdloan::contribute

JoshOrndorff:
This PR adds a parameter `memo: Vec<u8>,` to the `contribute` function. From the perspective of the relay chain, this field is just graffiti like a `system::remark`.

This will be useful for parachains who intend to reward their crowdloan contributors on the Parachain (eg by giving them native tokens). The memo field is a place where contributors can specify an parachain-native address to be associated with the contribution. By including it in the parameters of the call, the original contributor signs over it cryptographically with the contributing relay chain key, and the Polkadot validators attest to it by including it in the chain. This allows contributors to **trustlessly bootstrap a parachain identity**.

Without this field parachain developers will have to resort to one of these alternatives:

* Require users to use the same key on the parachain as the relay chain. This is contradicts the best practice of not reusing keys on multiple chains. And it is not an option at all for parachains that use a different native identity type.
* Add a parachain dispatchable where the contributor signs a parachain address with their relay chain address. Similar to the polkadot claims pallet. This works, but is less elegant than the memo field approach imo.
* Rely on an offchain process like a web2.0-style form for associating addresses, and include the associations at parachain genesis.


runtime/common/src/crowdloan.rs
```diff
@@ -322,7 +322,8 @@ decl_module! {
 			origin,
 			#[compact] index: FundIndex,
 			#[compact] value: BalanceOf<T>,
-			signature: Option<MultiSignature>
+			signature: Option<MultiSignature>,
+			_memo: [u8; 32],
```
xlc, line 326: why fixed size memo? it will be a wasted space for people not using it

shawntabrizi:
Independent of whether this feature should exist, any changes to Crowdloan should target the `gav-new-slots` branch. Thanks!

runtime/common/src/crowdloan.rs
JoshOrndorff: Good point. Maybe it should be `Vec<u8>` and we can enforce a maximum size.

JoshOrndorff: Took care of that. Thanks

JoshOrndorff:
I've switched to variable length as @xlc suggested, and merged master which now includes the work @shawntabrizi mentioned. Further reviews or feedback welcome.

girazoki:
I agree this is the more elegant option for contributors to prove contributions in the parachain side. I would say that the most trustless solution without the memo is the native address signature with the relay chain key, which already adds quite a bit of complexity to the process. 

Comments on the PR: Shouldnt we store the `memo` as well in the `contribution_put` function?

JoshOrndorff:
Does anyone oppose adding this memo field to the extrinsic? I don't believe I can add labels myself, so I might need help with that.

Regarding storing the memo field in the state. I think that is more up for debate because we have to trade the utility it brings against the cost of the state size. I think we should consider that separately after https://github.com/paritytech/cumulus/issues/320 is addressed.

Even adding the memo to the extrinsic adds value because it doesn't require the contributor to make a separate signature over their parachain identity and submit it to the parachain to prove their contribution.

shawntabrizi:
closed by https://github.com/paritytech/polkadot/pull/2728

PR #2635, opened 2021-03-17
collator-protocol: add message authentication

andresilva:
The `Declare` and `AdvertiseCollation` messages of the collator protocol were not signed, which made it possible for any peer to impersonate a given collator. This PR makes it so that these messages include a signed payload to authenticate the peer.


andresilva:
This breaks the collator network protocol so we'd need to restart rococo.

rphmeier:
@andresilva isn't restarting all the nodes sufficient?

andresilva:
Restarting nodes works, I just assumed that it would be hard to coordinate since there are already community collators on the network. But I guess the worst thing that can happen is that they cannot collate until they upgrade their nodes.

node/network/collator-protocol/src/collator_side.rs
```diff
@@ -467,7 +495,16 @@ async fn advertise_collation(
...
-	let wire_message = protocol_v1::CollatorProtocolMessage::AdvertiseCollation(relay_parent, collating_on);
+	let advertise_collation_signature_payload =
+		protocol_v1::advertise_collation_signature_payload(&relay_parent, &collating_on);
+
+	let wire_message = protocol_v1::CollatorProtocolMessage::AdvertiseCollation(
+		relay_parent,
+		collating_on,
+		state
+			.collator_pair
+			.sign(&advertise_collation_signature_payload),
```
eskimor, line 506: Why do we sign `AdvertiseCollation` messages as well? We already reject messages if we have not received a `Declare` before and if we did, the collator has already proven its identity.

eskimor (approved):

node/network/collator-protocol/src/collator_side.rs
andresilva: Good point, will update the PR.

tomaka:
I don't really understand why this is needed?
When the networking layer says "peer X has sent us a message", we know for sure that it's indeed peer X, there's no extra signature required.

eskimor:
Yeah, but we don't know that peer x is collator y.

tomaka:
What is this "y" in "collator y"? Aren't collator identified only by their `PeerId`?

burdges:
We'll face issues with "high availability" collators too, like peer X and Y connect and both legitimately present certificates from collator C.  All solutions look okay off the top of my head.  

We're broadly speaking reinventing the certificate chain from TLS here, but over Noise IK and friends.  We'll likely one day discover that Noise IK only benefits us if some certificates should ideally be presented during the first handshake flow, as opposed to one the channel gets established like we'll choose here, but maybe best if we do not become overly bogged down by those details just yet.

node/network/protocol/src/lib.rs
```diff
@@ -404,4 +411,12 @@ pub mod v1 {
 	}
 
 	impl_try_from!(CollationProtocol, CollatorProtocol, CollatorProtocolMessage);
+
+	/// Get the payload that should be signed and included in a `Declare` message.
+	///
+	/// The payload is the local peer id of the node, which serves to prove that it
+	/// controls the collator key it is declaring an intention to collate under.
+	pub fn declare_signature_payload(peer_id: &sc_network::PeerId) -> Vec<u8> {
+		peer_id.to_bytes()
```
rphmeier, line 420: Should we include some magic in this payload?

rphmeier (approved):

node/network/protocol/src/lib.rs
andresilva: What kind of magic? :)

rphmeier: `b"COLL" or something?

Issue #2418, opened 2021-02-11
minimum nomination limit on Polkadot: Status of the off-chain election fix. 


pmensik:
I am creating this issue because I asked multiple times in various chats, unfortunately without any reply.

I would like to ask what's the status with fixing the off-chain nomination distribution so there's no minimal stake for nominators (it's a bit above 100 DOT at the moment). This is becoming a very serious issue because it prevents a lot of people to stake DOTs because the minimal amount has become about 2,500 USD - I am afraid this makes Polkadot almost a "permissioned" network. I created this issue as a "last resort" since I couldn't get the answer anywhere else - if this is being worked on and when we can expect it to be done. Thanks a lot for the insight!

kianenigma:
My short answer is keep an eye on these PRs and how they are deployed: 

https://github.com/paritytech/substrate/pull/7908
https://github.com/paritytech/substrate/pull/7909
https://github.com/paritytech/substrate/pull/7910

These works will optimize the whole process, allowing potentially _many times more_ nominators to be _active_. Moreover, I was planning on tweaking some parameters to see if we can allow more nominators (similar to [this one](https://github.com/paritytech/polkadot/pull/1838)) for the next release as a temporary mitigation in favour of Polkadot's stakers.

--- 

If you are interested in the matter a bit more: 

You might ask why we have this issue in the first place? The answer is that nomination is different from (so called) delegation in the sense that you are not merely replicating/supporting/backing the behaviour of someone else, but rather you are imposing _your direct own opinion_ on the system as well. This is why typically when you delegate, you name 1 entity (e.g. a validator) whom you wish to support. Whilst in nominating, you provide a list of preferences, and an election process needs to process all of these lists to come up with an optimal set of validators that optimizes a set of rules (that of course serve toward decentralization). At this point it should be clear the supporting an arbitrary large number of nominators is not trivial in a system with constrained resource like a blockchain. This was explained in the original blog post [here](https://polkadot.network/polkadot-staking-an-update/#:~:text=Since%20it%20went%20live%20on,it%20matures%20through%20its%20launch.&text=Polkadot%20is%20a%20Nominated%20Proof,nominators%20to%20maximize%20chain%20security.).

[...]

PR #2637, opened 2021-03-17
Increase payable nominators per validator

kianenigma:
Related to https://github.com/paritytech/polkadot/issues/2418, increases the payable nominators per validator to 256. This will ensure that if and when we allow more nominators in polkadot, we won't have that many oversubscribed nominators. 

I also changed Kusama for consistency, but we could keep it as-is. Kusama, has fewer nominators so less of an issue. 

needs re-benchmarking.


runtime/kusama/src/weights/pallet_staking.rs
```diff
@@ -44,155 +44,155 @@ use sp_std::marker::PhantomData;
...
-			.saturating_add((49_957_000 as Weight).saturating_mul(n as Weight))
+		(129_521_000 as Weight)
+			// Standard Error: 18_000
+			.saturating_add((49_353_000 as Weight).saturating_mul(n as Weight))
 			.saturating_add(T::DbWeight::get().reads(11 as Weight))
 			.saturating_add(T::DbWeight::get().reads((3 as Weight).saturating_mul(n as Weight)))
 			.saturating_add(T::DbWeight::get().writes(2 as Weight))
 			.saturating_add(T::DbWeight::get().writes((1 as Weight).saturating_mul(n as Weight)))
 	}
 	fn payout_stakers_alive_staked(n: u32, ) -> Weight {
```
shawntabrizi, line 158: This is the function which where this constant is used.

Did you check how much weight this function is when N is 256?

shawntabrizi: And that it fits into any of our block limits or anything like that?

kianenigma: Added a test to ensure this will never exceed 50% of the block weight. With 256, it'll take 0.06 of the block weight.

shawntabrizi (approved):

runtime/kusama/src/lib.rs
```diff
@@ -1415,6 +1415,23 @@ mod test_fees {
 	use pallet_transaction_payment::Multiplier;
 	use separator::Separatable;
 
+	#[test]
+	fn payout_weight_portion() {
```
shawntabrizi, line 1419: nice test

kianenigma:
bot merge

kianenigma:
> The PR has approval from a member of substrateteamleads.


@shawntabrizi is a team lead but actually is not a member there, manual merging for now.

PR #2642, opened 2021-03-18
restructure polkadot-node-jaeger

drahnr:
Make the jaeger API only usable in one way in order to avoid future confusion.

* [x] remove the builder pattern
* [x] use `jaeger::Span::new()` for arbitrary types that implement new `trait LazyIdentEval`. 
* [x] apply jaeger API changes to the codebase


drahnr:
New usage pattern is defined here: https://github.com/paritytech/polkadot/pull/2642/files#diff-0f31587cb20443f483bfad5818d9f7c782f8b93da7947f6670ca3f226add85a1R22-R46

eskimor (approved):

Issue #2433, opened 2021-02-13
Collator protocol should decrease peer reputation


bkchr:
The collator protocol collator side currently doesn't decrease peer reputation when someone sends us junk data that isn't expected.

rphmeier:
Maybe can be done with the reputation refactoring https://github.com/paritytech/polkadot/issues/2184

eskimor:
Apart from decreasing peer reputation on garbage, I think we should set [`in_peers`](https://github.com/paritytech/polkadot/blob/4377176446c17e46cb306d398cc21e435a8d0d0e/node/network/protocol/src/peer_set.rs#L59) to 0, as collators don't expect incoming connections right now, but just connect to the validators of its assigned group. This would mean that only staked validators we our selves connected to can send us garbage.

xlc:
Not sure if I understanding correctly, but why collator should not allow incoming connections? We want it to receive transactions broadcasted by other nodes right?

eskimor:
I am only talking about the collation peer set. The collator would still be able to accept connections on any other peer set. The collation peer set is only used for communicating with validators and there connections are opened by collators, as validators don't need to know what collators actually exist.

bkchr:
> Uh, we use the same config for both validators and collators - so that does not work. Although it is kinda stupid to allow incoming connections for collators if we actually don't want them.

We should be able to pass some parameter that says if the node is a collator

eskimor:
> We should be able to pass some parameter that says if the node is a collator

Yeah, currently looking into how to do this in a non hacky way.

eskimor:
Given that we are now only connected to validators of the current backing group and that groups rotate anyway, I think this can be closed.

PR #2644, opened 2021-03-18
Don't accept incoming connections for collators

eskimor:
on the `Collation` peer set. Part of #2433 - I will take care of the rest once, @andresilva finished his work on the collator protocol, to avoid more pointless merge conflicts.

### DIFF
--- a/node/network/protocol/src/peer_set.rs
+++ b/node/network/protocol/src/peer_set.rs
@@ -67,7 +67,7 @@ impl PeerSet {
 				notifications_protocol: protocol,
 				max_notification_size,
 				set_config: SetConfig {
-					// Collators don't need to accept incoming connections on this peer set:
+					// Non-authority nodes don't need to accept incoming connections on this peer set:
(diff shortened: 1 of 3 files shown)


node/network/protocol/src/peer_set.rs
```diff
@@ -56,10 +67,15 @@ impl PeerSet {
 				notifications_protocol: protocol,
 				max_notification_size,
 				set_config: SetConfig {
-					in_peers: 25,
+					// Collators don't need to accept incoming connections on this peer set:
```
bkchr, line 70:
```suggestion
					// Non-authority nodes don't need to accept incoming connections on this peer set:
```

Full nodes are also non authority nodes :D

bkchr: And at some point we may also support collators who are also relay chain validators.

bkchr (approved):

eskimor:
bot merge

PR #2646, opened 2021-03-18
Fix candidate timeout

bkchr:
Candidates would not timeout when we are too far into a rotation. This
pr fixes it by not checking how far we are into a rotation and instead
only checking if `now - pending_since >= availability_period`.

### DIFF
--- a/runtime/parachains/src/scheduler.rs
+++ b/runtime/parachains/src/scheduler.rs
@@ -576,54 +576,27 @@ impl<T: Config> Module<T> {
 		Some(GroupIndex(group_idx as u32))
 	}
 
-	/// Returns an optional predicate that should be used for timing out occupied cores.
+	/// Returns a predicate that should be used for timing out occupied cores.
 	///
-	/// If `None`, no timing-out should be done. The predicate accepts the index of the core, and the
+	/// The predicate accepts the index of the core, and the
 	/// block number since which it has been occupied, and the respective parachain and parathread
 	/// timeouts, i.e. only within `max(config.chain_availability_period, config.thread_availability_period)`
 	/// of the last rotation would this return `Some`, unless there are no rotations.
-	///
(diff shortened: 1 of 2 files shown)


runtime/parachains/src/scheduler.rs
```diff
@@ -576,54 +576,27 @@ impl<T: Config> Module<T> {
...
-						}
-					}
-				}
-			}))
+		move |core_index: CoreIndex, pending_since| {
+			match availability_cores.get(core_index.0 as usize) {
+				None => true, // out-of-bounds, doesn't really matter what is returned.
+				Some(None) => true, // core not occupied, still doesn't really matter.
+				Some(Some(CoreOccupied::Parachain)) =>
+					now.saturating_sub(pending_since) >= config.chain_availability_period,
```
drahnr, line 596: I think this alters behaviour quite a bit:

Before: Only within `rot..(rot + Δ)` the generated `Fn` could return `true` (where `rot` is the time of the _last_ rotation).

After: The only within `rot..(rot + Δ)` is removed.

As you stated, this clears the pending ones way more often and hence could have the desired effect.

drahnr (approved):

rphmeier:
This does alter the behavior. The intention, from Jeff & Al, was to have timeouts only occur in the first blocks of a new validator rotation.

bkchr:
> This does alter the behavior. The intention, from Jeff & Al, was to have timeouts only occur in the first blocks of a new validator rotation.

What is the reason behind this?

In the end you are right and this just hides bugs in our availabilityand system. So, I should better close this.

burdges:
We never thought overly carefully about these backer timeouts, and one could o many things, but our rotating cores constrains what happens and necessitates some backer timeouts.  

We'll just make backers start working for their new core or their core's net assignment early, well the rotations must happen sometime.

Issue #2641, opened 2021-03-18
error in auction/slot module


yrong:
```
polkadot --chain=westend-dev --alice
====================

Version: 0.8.29-6b474a44-x86_64-linux-gnu

   0: sp_panic_handler::panic_hook
             at /home/ronyang/.cargo/git/checkouts/substrate-7e08433d4c370a21/da7ca4d/primitives/panic-handler/src/lib.rs:150:18
   1: sp_panic_handler::set::{{closure}}
             at /home/ronyang/.cargo/git/checkouts/substrate-7e08433d4c370a21/da7ca4d/primitives/panic-handler/src/lib.rs:60:4
   2: std::panicking::rust_panic_with_hook
             at /rustc/9b471a3f5fe57e5c6e08acf665f2094422415a3d/library/std/src/panicking.rs:595:17
   3: std::panicking::begin_panic_handler::{{closure}}
             at /rustc/9b471a3f5fe57e5c6e08acf665f2094422415a3d/library/std/src/panicking.rs:495:13
   4: std::sys_common::backtrace::__rust_end_short_backtrace
             at /rustc/9b471a3f5fe57e5c6e08acf665f2094422415a3d/library/std/src/sys_common/backtrace.rs:141:18
   5: rust_begin_unwind
             at /rustc/9b471a3f5fe57e5c6e08acf665f2094422415a3d/library/std/src/panicking.rs:493:5
   6: core::panicking::panic_fmt
             at /rustc/9b471a3f5fe57e5c6e08acf665f2094422415a3d/library/core/src/panicking.rs:92:14
   7: core::panicking::panic
             at /rustc/9b471a3f5fe57e5c6e08acf665f2094422415a3d/library/core/src/panicking.rs:50:5
   8: polkadot_runtime_common::slots::Module<T>::manage_lease_period_start
             at runtime/common/src/slots.rs:235:5
   9: <polkadot_runtime_common::slots::Module<T> as frame_support::traits::OnInitialize<<T as frame_system::pallet::Config>::BlockNumber>>::on_initialize
             at runtime/common/src/slots.rs:149:5
  10: <(TupleElement0,TupleElement1) as frame_support::traits::OnInitialize<BlockNumber>>::on_initialize
             at /home/ronyang/.cargo/git/checkouts/substrate-7e08433d4c370a21/da7ca4d/frame/support/src/traits.rs:1581:1
[...]

PR #2656, opened 2021-03-22
Check that Para is Registered before Accepting a Bid

shawntabrizi:
This is technically not something the runtime needs to check for, since we correctly handle situations where the para is not registered and wins a bid (nothing happens), but this may be more user friendly to support.

Otherwise, we can look to have this be an entirely front-end check, but then multiple different UIs would all need to make sure they program this correctly.

cc @jacogr


shawntabrizi:
Would fix issues like:
https://github.com/paritytech/polkadot/issues/2641

jacogr:
UIs should do checks for these as well prior to submission, but think this is a good check to do on-chain anyway, especially if it avoids issues like reported in the GH link. 

(The crowdloan does do a similar check for parachain registration on creation)

joepetrowski:
From the current implementation of `Registrar`, it looks like it's not possible to update the head data/validation logic once a para has been registered (i.e. there is no `change_data` function and the registration [does not allow overwriting](https://github.com/paritytech/polkadot/blob/a5c5512/runtime/common/src/paras_registrar.rs#L335)).

This makes bidding for future slots pretty much impossible. What if you want to place a bid for slots 3-4 in the auction (i.e. a lease that starts 12 months in the future) with the plan that you would have your parachain ready by that time? Even if you had the para ready in the auction, a runtime upgrade could make your registered code invalid between the auction end and lease start.

I understand that for crowdloans, requiring registered code protects users from contributing to "vapor campaigns", but if someone is placing a bid from their own account, they should be free to participate in the auction and reserve their own DOT/KSM without the code registered yet.

Even with the crowdloan, I think the registration thing may be better handled by the front end, e.g. require a check box of "I acknowledge that this campaign has not registered any parachain code yet". Further, we could just make the creator make a sizable deposit and slash it if they don't upload valid code prior to the lease start.

jacogr:
Good points. Without a registration though, we also have the issue with future parachains, in that a bid is placed with a specific paraId. If you have not reserved that, it may also not be available to you when the time rolls around. So these are all linked as it stands.

joepetrowski:
Ah yeah, you need this `ParaId => ParaInfo` to make a claim on the ID. Could we split `register` into two functions, one that claims a `ParaId` for a fixed deposit, and then a `fn register_validation_info` to insert the validation code/genesis head? The latter could have a mechanism like `note_preimage` that takes a large fee for uploading invalid code but is low cost or free if registering a known winning `ParaId`.

If squatting on IDs indefinitely is a concern then we could also make them valid for some time period, after which a reporter can trigger a deposit slash.

shawntabrizi:
> it's not possible to update the head data/validation logic once a para has been registered

I am looking to add this functionality, but need to be careful what it may mean for someone who wins a crowdloan, and people expect a certain wasm/head data.

gavofyork (approved):

thiolliere (approved):
Code is good, I think the check is pretty cheap so better to have it on-chain IMO

runtime/common/src/auctions.rs
```diff
@@ -379,6 +384,8 @@ impl<T: Config> Module<T> {
 		last_slot: LeasePeriodOf<T>,
 		amount: BalanceOf<T>,
 	) -> DispatchResult {
+		// Ensure para is registered before placing a bid on it
```
kianenigma, line 387:
```suggestion
		// Ensure para is registered before placing a bid on it.
```

kianenigma (approved):

joepetrowski:
I would really like to see some way of changing your registered Wasm (or having a registration that doesn't require committing to some Wasm) due to the issues I mentioned before. Maybe that will be in a separate PR, but I want to raise the concern while this is still open.

> Need to be careful what it may mean for someone who wins a crowdloan, and people expect a certain wasm/head data.

I think we should be careful about the Relay Chain's responsibility here. For crowdloans, the top priority is that contributors are able to withdraw their contribution in the retirement period. I don't think the RC can enforce everything that the crowdloan creator promised off-chain. As long as we meet the first priority, the worst a para team can do is deny their users whatever reward they promised, but the Polkadot user will still get their contribution back at the end.

And for non-crowdloan bids, I don't think we should deny people the ability to bid on lease periods starting on 2 or higher, which this does in practice.

xlc:
I will suggest to opt-in the fixed wasm feature. It is good thing for crowdloan contributors to be sure what they getting but just really hard for many not initially fully decentralized projects.

The workaround plan we have is submit a simple wasm runtime with sudo pallet and once the parachain is live, use sudo to upgrade it to whatever we need.

shawntabrizi:
bot merge

PR #2660, opened 2021-03-22
Add Parachain Onboarding to Rococo

shawntabrizi:
This PR introduces the Parachain Onboarding Pallets (Slots, Auctions, Crowdloans) to the Rococo Runtime.

Reasonable configurations parameters have been selected to allow for people to test various parts of the lifecycle of the Parachain Onboarding Process.


runtime/rococo/src/lib.rs
```diff
@@ -418,7 +417,7 @@ impl pallet_session::Config for Runtime {
 	type ValidatorIdOf = ValidatorIdOf;
 	type ShouldEndSession = Babe;
 	type NextSessionRotation = Babe;
-	type SessionManager = pallet_session::historical::NoteHistoricalRoot<Self, ProposeParachain>;
+	type SessionManager = ();
```
shawntabrizi, line 420: I think this is right, but can I get a second confirmation?

I don't see that `SessionManager` is implemented anywhere else.

runtime/rococo/src/lib.rs
```diff
@@ -208,14 +207,14 @@ construct_runtime! {
 		Hrmp: parachains_hrmp::{Pallet, Call, Storage, Event},
 		SessionInfo: parachains_session_info::{Pallet, Call, Storage},
 
+		// Parachain Onboarding Pallets
 		Registrar: paras_registrar::{Pallet, Call, Storage, Event<T>},
-		ParasSudoWrapper: paras_sudo_wrapper::{Pallet, Call},
```
shawntabrizi, line 212: Do we need this still?

Offers:

* sudo_schedule_para_initialize
* sudo_schedule_para_cleanup
* sudo_queue_downward_xcm
* sudo_establish_hrmp_channel

xlc: still useful for testing purpose, at least for rococo-local we want those

shawntabrizi: ok, np

runtime/rococo/src/lib.rs
bkchr: How will we add new validators?

shawntabrizi: ah i must have misunderstood you.

I thought you said all validators are parity now, so i thought they might just be in genesis?

What function do we need here? Just a set validators function?

rphmeier: `ProposeParachain` also had some sudo functions for adding/removing validators manually. We have and will continue to use that to scale up Rococo over time.

shawntabrizi: I've updated `PrposeParachain` into `ValidatorManager`, which has only this functionality. I wanted to strip away any of the special management of Parachains so they are not accidentally used.

Let me know if this is all okay now, or if you think we should keep more features of propose Parachains.

Az111z (approved):

Az111z (approved):

runtime/rococo/src/lib.rs
```diff
@@ -636,16 +687,9 @@ impl EnsureOrigin<Origin> for PriviledgedOrigin {
...
-	pub const MaxNameLength: u32 = 20;
-}
-
-impl propose_parachain::Config for Runtime {
+impl validator_manager::Config for Runtime {
 	type Event = Event;
-	type MaxNameLength = MaxNameLength;
-	type ProposeDeposit = ProposeDeposit;
-	type PriviledgedOrigin = EnsureOneOf<AccountId, EnsureRoot<AccountId>, PriviledgedOrigin>;
+	type PrivilegedOrigin = EnsureOneOf<AccountId, EnsureRoot<AccountId>, PrivilegedOrigin>;
```
bkchr, line 692:
```suggestion
	type PrivilegedOrigin =  EnsureRoot<AccountId>;
```

This is some functionality we only require root for. You could also remove this type completly and just call `ensure_root`. As this is mainly test code, I don't care what you actually do.

runtime/rococo/src/lib.rs
```diff
@@ -600,23 +604,70 @@ impl paras_registrar::Config for Runtime {
...
+	type WeightInfo = slots::TestWeightInfo;
+}
+
+parameter_types! {
+	pub const CrowdloanId: ModuleId = ModuleId(*b"py/cfund");
+	pub const SubmissionDeposit: Balance = 100 * DOLLARS;
+	pub const MinContribution: Balance = 1 * DOLLARS;
+	pub const RetirementPeriod: BlockNumber = 6 * HOURS;
+	pub const RemoveKeysLimit: u32 = 500;
+
```
bkchr, line 646:
```suggestion
```

bkchr (approved):

PR #2661, opened 2021-03-22
Add HrmpChannelManagementHooks

4meta5:
As of now, to open, accept, and close channels on the relay chain, parachains must send the message `Xcm::Transact` with the formatted call data for `hrmp` dispatchables. Forming this call is hard from the parachain runtime; [this Acala demo](https://wiki.acala.network/build/development-guide/composable-chains/open-hrmp-channel) uses polkadot-js to form the call from the relay chain context and remove the prefix.

There is a [`xcm-format` PR](https://github.com/paritytech/xcm-format/pull/22) which includes three new `Xcm` variants to open, accept, and close channels. This PR adds those variants and the execution logic. This will make it unnecessary to use `Xcm::Transact` for those actions.


runtime/parachains/src/hrmp.rs
```diff
@@ -412,6 +412,25 @@ decl_module! {
 	}
 }
 
+use xcm_executor::traits::ExecuteHrmp;
+use xcm::v0::{Result as XcmResult};
+impl<T: Config> ExecuteHrmp for Module<T> {
+	fn hrmp_init_open_channel(sender: u32, recipient: u32, max_message_size: u32, max_capacity: u32) -> XcmResult {
+		Self::init_open_channel(sender.into(), recipient.into(), max_capacity, max_message_size).map_err(|_| ().into())
```
4meta5, line 419:
```suggestion
		Self::init_open_channel(sender.into(), recipient.into(), max_capacity, max_message_size).map_err(|_| ().into())
```

Right now, this is returning the XcmError::Undefined if a runtime error is returned. What error should it use?

runtime/rococo/src/lib.rs
```diff
@@ -559,6 +559,7 @@ impl xcm_executor::Config for XcmConfig {
 	type Call = Call;
 	type XcmSender = xcm_sender::RelayChainXcmSender<Runtime>;
 	type AssetTransactor = LocalAssetTransactor;
+	type HrmpExecutor = Hrmp;
```
4meta5, line 562: this only makes sense for the relay chain to have because it is managing the channels, maybe need to have an () impl for parachains

4meta5:
This PR is ready for review. The CI is red because there are no labels (and I cannot set this). Let me know of suggestions @pepyakin  @shawntabrizi

4meta5:
I received feedback that the `Xcm` message variants included in the referenced spec PR are not finalized so this PR will not be reviewed until that spec PR is merged.

I assumed the messages included in this PR were uncontroversial because they correspond directly to the actions for interacting with the `hrmp` pallet. This other spec PR https://github.com/paritytech/xcm-format/pull/18 was approved and it includes the change with the same intention of providing a proper way to dispatch HRMP actions from the parachain. The only way to do this now is to compute the encoded calls in polkadot-js from the relay chain context, remove the prefix, and pass that to whatever code is forming the Xcm::Transact message.

4meta5:
I received feedback that the `Xcm` message variants included in the referenced spec PR are not finalized so this PR will not be reviewed until that spec PR is merged.

I assumed the messages included in this PR were uncontroversial because they correspond directly to the actions for interacting with the `hrmp` pallet. This other spec PR https://github.com/paritytech/xcm-format/pull/18 was approved and it includes the change with the same intention of providing a proper way to dispatch HRMP actions from the parachain. The only way to do this now is to compute the encoded calls in polkadot-js from the relay chain context, remove the prefix, and pass that to whatever code is forming the Xcm::Transact message.

xcm/src/v0/traits.rs
```diff
@@ -67,3 +67,21 @@ impl SendXcm for () {
...
+	fn hrmp_init_open_channel(_sender: u32, _recipient: u32, _max_message_size: u32, _max_capacity: u32) -> Result {
+		Err(().into())
+	}
+	fn hrmp_accept_open_channel(_recipient: u32, _sender: u32) -> Result {
+		Err(().into())
+	}
+	fn hrmp_close_channel(_initiator: u32, _sender: u32, _recipient: u32) -> Result {
+		Err(().into())
+	}
+}
```
pepyakin, line 87: Missing trailing line

```suggestion
}

```

xcm/src/v0/traits.rs
```diff
@@ -67,3 +67,21 @@ impl SendXcm for () {
 		Err(Error::Unimplemented)
 	}
 }
+
+pub trait ExecuteHrmp {
```
pepyakin, line 71: I am not entirely sure about naming of this. This trait seems to be focused on the channel management of HRMP. Therefore, maybe we should name it such that it reflects that? A strawman: `HrmpChannelManagementHooks`?

xcm/src/v0/mod.rs
```diff
@@ -203,6 +203,53 @@ pub enum Xcm {
...
+	/// Safety: The message should originate directly from the sender or
+	/// recipient para (either member of the channel).
+	///
+	/// Kind: *Instruction*.
+	///
+	/// Errors:
+	HrmpCloseChannel {
+		#[codec(compact)] sender: u32,
+		#[codec(compact)] recipient: u32,
+	},
```
pepyakin, line 252: Recalling an open channel request is upcoming in https://github.com/paritytech/polkadot/pull/3543

I think it's worth to reconsider these messages in the light of that PR.

xcm/xcm-executor/src/config.rs
```diff
@@ -29,6 +29,9 @@ pub trait Config {
 	/// How to withdraw and deposit an asset.
 	type AssetTransactor: TransactAsset;
 
+	/// How to execute HRMP-related actions
+	type HrmpExecutor: ExecuteHrmp;
```
pepyakin, line 33: I think this could be improved by specifying what exactly actions are (i.e. the channel management). It is also better to take into account how the developers will approach it. I think this is mostly relevant for relay-chains which is a way less common case. Thus it would be good to guide a typical implementer of this to use the `()` dummy

xcm/src/v0/traits.rs
```diff
@@ -67,3 +67,21 @@ impl SendXcm for () {
...
+	fn hrmp_init_open_channel(sender: u32, recipient: u32, max_message_size: u32, max_capacity: u32) -> Result;
+	fn hrmp_accept_open_channel(recipient: u32, sender: u32) -> Result;
+	fn hrmp_close_channel(initiator: u32, sender: u32, recipient: u32) -> Result;
```
kianenigma, line 74: DQ: why are all parameters u32 all across the place?

pepyakin: I presumed it's the same reason why we use u32 within XCM definitions. The alternative you are thinking of is `ParaId`. That comes from the primitives crate. Unlike one may think, it can be pretty heavy, bringing all kinds of cruft. At the same time, we want this crate to be used pervasively thus we want to minimize the number of dependencies.

xcm/xcm-executor/src/lib.rs
```diff
@@ -234,6 +234,31 @@ impl<Config: config::Config> XcmExecutor<Config> {
...
+				let initiator = match origin {
+					MultiLocation::X1(Junction::Parachain { id }) => id,
+					_ => Err(XcmError::BadOrigin)?,
+				};
+				return Config::HrmpExecutor::hrmp_close_channel(initiator, sender, recipient)
+			}
+			(origin, Xcm::RelayTo { dest: MultiLocation::X1(Junction::Parachain { id }), inner }) => {
+				let msg = Xcm::RelayedFrom { superorigin: origin, inner }.into();
+				return Config::XcmSender::send_xcm(Junction::Parachain { id }.into(), msg)
+			},
```
4meta5, line 261:
```suggestion
```

xcm/src/v0/mod.rs
4meta5: I don't see the open channel request XCM in that PR (and no new XCM message variants).

Is there a plan to add XCM variants for (1) request open channel (2) request close channel (3) accept open channel ?

4meta5:
refactoring now, not ready for review as is

xcm/src/v0/traits.rs
4meta5: accepted the strawman for now

4meta5:
I would add tests, but I'd like feedback on the design first so I know this is a viable direction.

4meta5:
Current error, PRing substrate with impl

```
error[E0277]: the trait bound `(): GetDispatchInfo` is not satisfied
   --> xcm/src/v0/traits.rs:282:2
    |
271 |     type HrmpCall: Dispatchable + GetDispatchInfo;
    |                                   --------------- required by this bound in `HrmpChannelManagementHooks::HrmpCall`
...
282 |     type HrmpCall = ();
    |     ^^^^^^^^^^^^^^^^^^^ the trait `GetDispatchInfo` is not implemented for `()`

For more information about this error, try `rustc --explain E0277`.
error: could not compile `xcm`

To learn more, run the command again with --verbose.
```

4meta5:
branch is stale, I don't have time now to update/maintain it

PR #2662, opened 2021-03-22
Fungibles adapter and MultiLocation::X8

gavofyork:
Basic implementation allows the Assets pallet to be used in XCM. Some additional adapters/converters should allow more non-native assets to be sent & received.

Requires paritytech/substrate#8425.

This also includes some minor alterations to CI mainly for the renaming of a few of the labels.


xcm/src/v0/multi_location.rs
```diff
@@ -56,6 +56,14 @@ pub enum MultiLocation {
 	X3(Junction, Junction, Junction),
 	/// A relative path comprising four junctions.
 	X4(Junction, Junction, Junction, Junction),
+	/// A relative path comprising four junctions.
```
xlc, line 59: looks like we need some macro to generate those?

xcm/xcm-builder/src/fungibles_adapter.rs
```diff
@@ -0,0 +1,257 @@
+// Copyright 2020 Parity Technologies (UK) Ltd.
```
shaunxw, line 1:
```suggestion
// Copyright 2021 Parity Technologies (UK) Ltd.
```

xcm/xcm-builder/src/fungibles_adapter.rs
```diff
@@ -0,0 +1,257 @@
...
+		Ok((what, amount))
+	}
+}
+
+pub struct FungiblesAdapter<Assets, Matcher, AccountIdConverter, AccountId>(
+	PhantomData<(Assets, Matcher, AccountIdConverter, AccountId)>
+);
+
+impl<
+	Assets: Fungibles<AccountId>,
```
shaunxw, line 230: Only assets `deposit` and `withdraw` is used in in this impl, I think we could use a minimum trait bound here instead including only these two fn. In this way `FungiblesAdapter` would be easier to be reused with other assets impl like `orml-tokens`, without implementing all methods in `Fungibles`.

xcm/src/v0/multi_location.rs
gavofyork: yeah, that'd be nice

xcm/xcm-builder/src/fungibles_adapter.rs
gavofyork: with the possible exception of `transfer`, all will be needed for XCM eventually.

gavofyork:
cc @s3krit

s3krit (approved):
CI/label changes look good

xcm/xcm-builder/src/fungibles_adapter.rs
```diff
@@ -0,0 +1,267 @@
...
+	}
+}
+
+/// Generic third-party conversion trait. Use this when you don't want to force the user to use default
+/// impls of `From` and `Into` for the types you wish to convert between.
+///
+/// One of `convert`/`convert_ref` and `reverse`/`reverse_ref` MUST be implemented. If possible, implement
+/// `convert_ref`, since this will never result in a clone. Use `convert` when you definitely need to consume
+/// the source value.
+pub trait Convert<A: Clone, B: Clone> {
```
apopiak, line 54: DQ: What's the difference/motivation relative to `TryFrom` (see docs [here](https://doc.rust-lang.org/std/convert/trait.TryFrom.html))?

gavofyork: `TryFrom` is intrinsic to the (source) type in question and as such it is intended for the situation where there is and could only be one sensible way of converting from one to the other type.

In cases where there may be several reasonable ways of converting between two types, then a third-party "converter" type is needed which is able to select/define the means of conversion.

xcm/src/v0/multi_location.rs
shawntabrizi: https://github.com/paritytech/polkadot/issues/2750

xcm/xcm-builder/src/fungibles_adapter.rs
```diff
@@ -0,0 +1,267 @@
...
+/// Implementation of `Convert<Vec<u8>, _>` using the parity scale codec.
+pub struct Decoded;
+impl<T: Clone + Encode + Decode> Convert<Vec<u8>, T> for Decoded {
+	fn convert_ref(bytes: impl Borrow<Vec<u8>>) -> result::Result<T, ()> {
+		T::decode(&mut &bytes.borrow()[..]).map_err(|_| ())
+	}
+	fn reverse_ref(value: impl Borrow<T>) -> result::Result<Vec<u8>, ()> { Ok(value.borrow().encode()) }
+}
+
+/// Converter struct implementing `AssetIdConversion` converting a numeric asset ID (must by TryFrom/TryInto<u128>)
```
shawntabrizi, line 146:
```suggestion
/// Converter struct implementing `AssetIdConversion` converting a numeric asset ID (must be TryFrom/TryInto<u128>)
```

shawntabrizi (approved):

PR #2663, opened 2021-03-23
more approval voting instrumentation

rphmeier:
* fix typo
* fix `unapproved_candidates` which was actually returning approved candidates.
* add info about next wakeup for unapproved candidates. this should help me see if validators are failing to schedule
* add info about assignments/approvals to tracing

it would be nice to have info about `RequiredTranches` as well but that would get messy as the computation requires loading the `SessionInfo`.

### DIFF
--- a/node/core/approval-voting/src/lib.rs
+++ b/node/core/approval-voting/src/lib.rs
@@ -322,6 +322,11 @@ impl Wakeups {
 		self.wakeups.entry(tick).or_default().push((block_hash, candidate_hash));
 	}
 
+	// Get the wakeup for a particular block/candidate combo, if any.
+	fn wakeup_for(&self, block_hash: Hash, candidate_hash: CandidateHash) -> Option<Tick> {
+		self.reverse_wakeups.get(&(block_hash, candidate_hash)).map(|t| *t)
+	}
+
 	// Returns the next wakeup. this future never returns if there are no wakeups.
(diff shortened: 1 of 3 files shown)


arkpar (approved):

node/core/approval-voting/src/lib.rs
```diff
@@ -827,24 +835,42 @@ async fn handle_approved_ancestor(
...
+											candidate_hash,
+										);
+
 										tracing::debug!(
 											target: LOG_TARGET,
 											?candidate_hash,
 											?block_hash,
 											tranche,
+											?next_wakeup,
+											status,
```
bkchr, line 871:
```suggestion
								let status = || format!("{}/{}/{}",
									n_assignments,
									n_approvals,
									a_entry.n_validators(),
								);

								match a_entry.our_assignment() {
									None => tracing::debug!(
										target: LOG_TARGET,
										?candidate_hash,
										?block_hash,
										status = %status(),
										"no assignment."
									),
									Some(a) => {
										let tranche = a.tranche();
										let triggered = a.triggered();

										let next_wakeup = wakeups.wakeup_for(
											block_hash,
											candidate_hash,
										);

										tracing::debug!(
											target: LOG_TARGET,
											?candidate_hash,
											?block_hash,
											tranche,
											?next_wakeup,
											status = %status(),
```

As this is debug output, we should not always generate this string

bkchr (approved):

Issue #2664, opened 2021-03-23
Empty Crowdloan Can be Canceled by Anyone


shawntabrizi:
Right now dissolving a crowdloan is permissionless.

When dissolving a crowdloan, we check if the ending period is up or if the fund has received no contributions.

However, we should make sure that the latter only happens by the crowdloan owner, else they can get griefed by people closing other crowdloans.

PR #2665, opened 2021-03-23
Fix Crowdloan Dissolve and Add Auction Cancel

shawntabrizi:
Fixes https://github.com/paritytech/polkadot/issues/2664

If a crowdloan is trying to be dissolved before it is "ended" (but because funds are zero), we make sure that the caller is the fund depositor, which will avoid grief attacks.

This PR also adds an API to cancel an in-progress auction by Root.


runtime/common/src/auctions.rs
```diff
@@ -264,6 +264,21 @@ decl_module! {
...
+
+		/// Cancel an in-progress auction.
+		///
+		/// Can only be called by Root origin.
+		#[weight = 0]
+		pub fn cancel_auction(origin) {
+			ensure_root(origin);
+			// Unreserve all bids.
+			for ((bidder, para), amount) in ReservedAmounts::<T>::iter() {
+				ReservedAmounts::<T>::take((bidder.clone(), para));
```
bkchr, line 276:
```suggestion
			for ((bidder, _), amount) in ReservedAmounts::<T>::drain() {
```

bkchr (approved):

runtime/westend/src/weights/auctions.rs
```diff
@@ -44,18 +44,23 @@ use sp_std::marker::PhantomData;
...
 			.saturating_add(T::DbWeight::get().reads(7 as Weight))
 			.saturating_add(T::DbWeight::get().writes(4 as Weight))
 	}
 	fn on_initialize() -> Weight {
-		(602_940_000 as Weight)
-			.saturating_add(T::DbWeight::get().reads(75 as Weight))
-			.saturating_add(T::DbWeight::get().writes(71 as Weight))
+		(3_095_173_000 as Weight)
+			.saturating_add(T::DbWeight::get().reads(625 as Weight))
+			.saturating_add(T::DbWeight::get().writes(621 as Weight))
```
shawntabrizi, line 59: This weight is dependent on the number of samples we take in the ending period.

PR #2670, opened 2021-03-23
Sample Auction Winners Every `SampleLength` Blocks

shawntabrizi:
This PR adjusts the Auctions pallet so that winners are sampled every `SampleLength` blocks, rather than every block.

When `SampleLength == 1`, this code acts exactly the same as before.

If `SampleLength > 1`, we will only record the winners data every `SampleLength` blocks.

For example:

With `EndingPeriod == 100`, and `SampleLength == 10`, only 10 storage items will be placed to track all of the winning data over the ending period.


gavofyork (approved):

gavofyork:
Bit dirty, rewriting the same storage item `SampleSize` times. But it should work.

gavofyork (approved):

PR #2674, opened 2021-03-23
Rename constant to EPOCH_DURATION_IN_SLOTS

JoshOrndorff:
This PR makes a minor name change to the constant `EPOCH_DURATION_IN_BLOCKS` -> `EPOCH_DURATION_IN_SLOTS`.

This makes it more clear when reading the code that the duration of an epoch is actually defined in terms on the underlying consensus slots, and if there are slots without a block created, the actual number of blocks in the epoch will be different.

This arose recently when a private Polkadot chain stalled with the `Unexpected Epoch Change` error as reported in https://github.com/paritytech/substrate/issues/4464 and many other places. The natural question when you get that error is, "if the epoch length is defined as a number of blocks, how can you have no blocks in an epoch?". This naming change makes it clear that the epoch duration is not actually in number of blocks.

cc @crystalin

### DIFF
--- a/runtime/kusama/src/lib.rs
+++ b/runtime/kusama/src/lib.rs
@@ -180,7 +180,7 @@ impl pallet_scheduler::Config for Runtime {
 }
 
 parameter_types! {
-	pub const EpochDuration: u64 = EPOCH_DURATION_IN_BLOCKS as u64;
+	pub const EpochDuration: u64 = EPOCH_DURATION_IN_SLOTS as u64;
 	pub const ExpectedBlockTime: Moment = MILLISECS_PER_BLOCK;
 	pub const ReportLongevity: u64 =
 		BondingDuration::get() as u64 * SessionsPerEra::get() as u64 * EpochDuration::get();
@@ -311,7 +311,7 @@ impl pallet_session::historical::Config for Runtime {
 parameter_types! {
 	// no signed phase for now, just unsigned.
 	pub const SignedPhase: u32 = 0;
-	pub const UnsignedPhase: u32 = EPOCH_DURATION_IN_BLOCKS / 4;
+	pub const UnsignedPhase: u32 = EPOCH_DURATION_IN_SLOTS / 4;
 
 	// fallback: run election on-chain.
 	pub const Fallback: pallet_election_provider_multi_phase::FallbackStrategy =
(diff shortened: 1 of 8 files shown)


runtime/kusama/src/lib.rs
```diff
@@ -613,7 +613,7 @@ impl pallet_offences::Config for Runtime {
 impl pallet_authority_discovery::Config for Runtime {}
 
 parameter_types! {
-	pub const SessionDuration: BlockNumber = EPOCH_DURATION_IN_BLOCKS as _;
+	pub const SessionDuration: BlockNumber = EPOCH_DURATION_IN_SLOTS as _;
```
thiolliere, line 616:
```suggestion
	pub const SessionDuration: BlockNumber = EPOCH_DURATION_IN_SLOTS;
```

runtime/polkadot/src/lib.rs
```diff
@@ -659,7 +659,7 @@ impl pallet_offences::Config for Runtime {
 impl pallet_authority_discovery::Config for Runtime {}
 
 parameter_types! {
-	pub const SessionDuration: BlockNumber = EPOCH_DURATION_IN_BLOCKS as _;
+	pub const SessionDuration: BlockNumber = EPOCH_DURATION_IN_SLOTS as _;
```
thiolliere, line 662:
```suggestion
	pub const SessionDuration: BlockNumber = EPOCH_DURATION_IN_SLOTS;
```

runtime/westend/src/lib.rs
```diff
@@ -410,7 +410,7 @@ impl pallet_offences::Config for Runtime {
 impl pallet_authority_discovery::Config for Runtime {}
 
 parameter_types! {
-	pub const SessionDuration: BlockNumber = EPOCH_DURATION_IN_BLOCKS as _;
+	pub const SessionDuration: BlockNumber = EPOCH_DURATION_IN_SLOTS as _;
```
thiolliere, line 413:
```suggestion
	pub const SessionDuration: BlockNumber = EPOCH_DURATION_IN_SLOTS;
```

thiolliere (approved):
yes, new name is better

andresilva:
Feel free to create a PR to do the same rename in Substrate.

lotoussa (approved):
LGTM

JoshOrndorff:
Okay, I've rebased this onto the latest master. Can we please merge it? It feels like a waste of time to resolve conflicts and awaits CI again for a little one like this.

thiolliere:
bot merge

PR #2676, opened 2021-03-23
Remove unused constants

thiolliere:
those constants are never used.


shawntabrizi (approved):

shawntabrizi:
The label god has judged thee not worthy

Issue #2678, opened 2021-03-23
Validators should connect inactive collators


eskimor:
Right now, if a collator connects and just does nothing, we will keep the connection open indefinitely, taking up a slot, preventing good collators from connecting. The very least we should do, is disconnect them after some period of inactivity.

PR #2680, opened 2021-03-23
Evict inactive peers from the collator protocol peer-set

### LINKED ISSUE #2679: Malicious cost should be fatal
Cost for malicious is only 1_000_000, there is no reason to not immediately disconnect a malicious node.

rphmeier:
Closes #2679 , #2678 

...and ensure that cheap tricks to extend activity are punished with negative reputation.

This also exposes a wrapper around `NetworkBridge::disconnect_peer` for use by subsystems.

### DIFF
--- a/node/network/collator-protocol/src/validator_side.rs
+++ b/node/network/collator-protocol/src/validator_side.rs
@@ -673,7 +673,7 @@ pub(crate) async fn run<Context>(
 	let next_inactivity_stream = futures::stream::unfold(
 		Instant::now() + ACTIVITY_POLL,
 		|next_check| async move { Some(((), wait_until_next_check(next_check).await)) }
-	);
+	).fuse();
 
 	futures::pin_mut!(next_inactivity_stream);
 
--- a/node/network/collator-protocol/src/validator_side.rs
+++ b/node/network/collator-protocol/src/validator_side.rs
@@ -54,16 +54,9 @@ const COST_REQUEST_TIMED_OUT: Rep = Rep::CostMinor("A collation request has time
 const COST_REPORT_BAD: Rep = Rep::Malicious("A collator was reported by another subsystem");
 const BENEFIT_NOTIFY_GOOD: Rep = Rep::BenefitMinor("A collator was noted good by another subsystem");
 
-// After this long without activity, peers are disconnected.
-#[cfg(not(test))]
-const ACTIVITY_TIMEOUT: Duration = Duration::from_secs(5);
-
-#[cfg(test)]
-const ACTIVITY_TIMEOUT: Duration = Duration::from_millis(50);
-
 // How often to check all peers with activity.
 #[cfg(not(test))]
-const ACTIVITY_POLL: Duration = Duration::from_millis(2500);
+const ACTIVITY_POLL: Duration = Duration::from_secs(1);
 
 #[cfg(test)]
 const ACTIVITY_POLL: Duration = Duration::from_millis(10);
@@ -664,6 +657,7 @@ async fn wait_until_next_check(last_poll: Instant) -> Instant {
 #[tracing::instrument(skip(ctx, metrics), fields(subsystem = LOG_TARGET))]
 pub(crate) async fn run<Context>(
 	mut ctx: Context,
+	eviction_policy: crate::CollatorEvictionPolicy,
 	metrics: Metrics,
 ) -> Result<()>
 	where Context: SubsystemContext<Message = CollatorProtocolMessage>
@@ -711,7 +705,7 @@ pub(crate) async fn run<Context>(
 				continue
 			}
 			Some(Either::Right(())) => {
-				disconnect_inactive_peers(&mut ctx, &state.peer_data).await;
+				disconnect_inactive_peers(&mut ctx, eviction_policy, &state.peer_data).await;
 				continue
 			}
 			None => {}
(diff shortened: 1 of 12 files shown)


node/network/collator-protocol/src/validator_side.rs
```diff
@@ -590,23 +676,50 @@ where
...
-			let msg = msg?;
-			tracing::trace!(target: LOG_TARGET, msg = ?msg, "received a message");
-
-			match msg {
-				Communication { msg } => process_msg(&mut ctx, msg, &mut state).await,
-				Signal(BlockFinalized(..)) => {}
-				Signal(ActiveLeaves(_)) => {}
-				Signal(Conclude) => { break }
+		let res = {
+			let s = futures::future::select(ctx.recv().fuse(), next_inactivity_stream.next().fuse());
```
bkchr, line 688: Why don't you use the `select` macro? This code looks more complicated than it needs to.

And you should also fuse the stream outside the loop, otherwise the fuse doesn't make that much sense.

node/network/collator-protocol/src/validator_side.rs
```diff
@@ -590,23 +676,50 @@ where
 		..Default::default()
 	};
 
+	let next_inactivity_stream = futures::stream::unfold(
```
bkchr, line 679: Why is that a stream and not just a delay that you always [reset](https://docs.rs/futures-timer/3.0.2/futures_timer/struct.Delay.html#method.reset)?

node/network/collator-protocol/src/validator_side.rs
```diff
@@ -47,13 +51,26 @@ const COST_CORRUPTED_MESSAGE: Rep = Rep::CostMinor("Message was corrupt");
 /// Network errors that originated at the remote host should have same cost as timeout.
 const COST_NETWORK_ERROR: Rep = Rep::CostMinor("Some network error");
 const COST_REQUEST_TIMED_OUT: Rep = Rep::CostMinor("A collation request has timed out");
-const COST_REPORT_BAD: Rep = Rep::CostMajor("A collator was reported by another subsystem");
+const COST_REPORT_BAD: Rep = Rep::Malicious("A collator was reported by another subsystem");
 const BENEFIT_NOTIFY_GOOD: Rep = Rep::BenefitMinor("A collator was noted good by another subsystem");
 
+// After this long without activity, peers are disconnected.
+#[cfg(not(test))]
+const ACTIVITY_TIMEOUT: Duration = Duration::from_secs(5);
```
bkchr, line 59: Why should a collator send an advertise collation every 5 seconds?

And maybe these values should just be configurable, instead of having them hard coded here?

node/network/collator-protocol/src/validator_side.rs
rphmeier: This is different than the `select!` macro because it does not poll the future to completion but instead polls it just a single time. This is because of the `collation_requests` map polling logic below that. That logic should be refactored to use `FuturesUnordered` but I spent about an hour trying to do that refactoring and it is not that simple.

node/network/collator-protocol/src/validator_side.rs
rphmeier: @bkchr the point is that the collator shouldn't. this will keep the collation peer-set mostly empty for the purpose of accepting incoming connections from collators who are ready to collate immediately.

node/network/collator-protocol/src/validator_side.rs
rphmeier: Because of `futures::select` it must be `Unpin` which means we use `pin_mut!` which takes away our mutable access.

node/network/collator-protocol/src/validator_side.rs
rphmeier: https://github.com/paritytech/polkadot/issues/2682

i made it configurable but CLI refactor is its own issue

node/network/collator-protocol/src/validator_side.rs
bkchr: Yeah sorry I had not seen this logic below. Yeah that complicates the stuff :(

node/network/collator-protocol/src/validator_side.rs
bkchr: > @bkchr the point is that the collator shouldn't. this will keep the collation peer-set mostly empty for the purpose of accepting incoming connections from collators who are ready to collate immediately.

Fine, but why don't we disconnect a collator directly after we told him its collation result? And after connecting he gets like 1 second to send a collation or otherwise we ban him?

node/network/collator-protocol/src/validator_side.rs
bkchr: https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=5cc083bfc104416883d6e0f00b16185c should work or?

rphmeier: TIL.

But actually I think that this `stream::unfold` is more readable than intermingling "scheduler" and "scheduled" logic.

node/network/collator-protocol/src/validator_side.rs
rphmeier: Yeah, that sounds better. But I don't believe it should block this PR which will solve the basic version of the issue.

eskimor (approved):

PR #2686, opened 2021-03-24
Companion for substrate/pull/8445

kianenigma:
Companion for https://github.com/paritytech/substrate/pull/8445

@rrtti please apply all other changes that we need to the society.


runtime/kusama/src/lib.rs
```diff
@@ -791,6 +791,7 @@ parameter_types! {
 	pub const PeriodSpend: Balance = 500 * DOLLARS;
 	pub const MaxLockDuration: BlockNumber = 36 * 30 * DAYS;
 	pub const ChallengePeriod: BlockNumber = 7 * DAYS;
+	pub const MaxCandidateIntake: u32 = 1;
```
shawntabrizi, line 794: With intake of only 1, likely the only people to get into society will be those that bid 0.

just fyi

kianenigma: Society is full currently, and @rrtti wanted to make a few changes, including the above. I presume he has good reasons for it.

shawntabrizi: yeah, understandable, but in terms of what this change will do in the long term, that people with bid amounts probably don't get in anymore.

shawntabrizi (approved):

kianenigma:
whyy did the bot fail here again?

kianenigma:
thanks @shawntabrizi

kianenigma:
bot merge

PR #2688, opened 2021-03-24
Increase rotation period

rrtti:
Increase the rotation period for candidates to show PoI FROM 3,3 days to 7 days.

### DIFF
--- a/runtime/kusama/src/lib.rs
+++ b/runtime/kusama/src/lib.rs
@@ -787,7 +787,7 @@ parameter_types! {
 	pub const CandidateDeposit: Balance = 10 * DOLLARS;
 	pub const WrongSideDeduction: Balance = 2 * DOLLARS;
 	pub const MaxStrikes: u32 = 10;
-	pub const RotationPeriod: BlockNumber = 168 * HOURS;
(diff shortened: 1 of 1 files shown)


shawntabrizi (approved):

shawntabrizi:
I think this is "runtime worthy" since it will change some mechanics of Kusama

kianenigma:
indeed, @rrtti perhaps share some links ahead of time with the society members to be aware and object if they wish to do so, thourgh governance.

kianenigma (approved):

rrtti:
Changes have been announced on the Kappa Sigma Mu channel, will monitor response.

runtime/kusama/src/lib.rs
```diff
@@ -570,7 +570,7 @@ impl pallet_treasury::Config for Runtime {
 	type ProposalBondMinimum = ProposalBondMinimum;
 	type SpendPeriod = SpendPeriod;
 	type Burn = Burn;
-	type BurnDestination = Society;
+	type BurnDestination = ();
```
gavofyork, line 573:
```suggestion
	type BurnDestination = Society;
```

PR #2691, opened 2021-03-24
Unify maximum supported PoV size a bit.

eskimor:


primitives/src/v1.rs
```diff
@@ -458,6 +458,17 @@ impl PoV {
 #[derive(Clone, Encode, Decode, PartialEq, Eq)]
 pub struct CompressedPoV(Vec<u8>);
 
+/// Maximum PoV size we support right now.
+pub const MAX_POV_SIZE: usize = 52_428_800;
```
bkchr, line 462: We should use this value as well in the chain spec `HostConfiguration`

bkchr (approved):

eskimor:
bot merge

PR #2694, opened 2021-03-24
Port availability recovery to use req/res

rphmeier:
This reuses the chunk fetching request from availability distribution. This adds a new full data request type answered by the subsystem.


burdges:
We should be careful of full data requests not swamping validators, but it's probably fine since our number of approval checkers is small.  We could show our approval assignment in the request to ensure we get priority for being sent the data, both for erasure chunks and for the full data.

node/network/availability-recovery/src/lib.rs
```diff
@@ -520,13 +411,13 @@ impl RequestChunksPhase {
 	}
 }
 
-const fn is_unavailable(
+fn is_unavailable(
```
drahnr, line 414:
```suggestion
const fn is_unavailable(
```

node/network/availability-recovery/src/tests.rs
```diff
@@ -139,31 +141,48 @@ async fn overseer_recv(
 use sp_keyring::Sr25519Keyring;
 
 #[derive(Debug, Clone)]
-enum HasAvailableData {
+enum Has {
```
drahnr, line 144: Should clarify the intended use. `Has` is a tad too ambiguous.

drahnr:
Looks like a massive simplification, and LGTM

node/network/availability-recovery/src/tests.rs
rphmeier: it's localized to the test code where the use should be clear

drahnr (approved):

node/network/availability-recovery/src/lib.rs
```diff
@@ -314,24 +223,40 @@ impl RequestChunksPhase {
 		params: &InteractionParams,
 		to_state: &mut mpsc::Sender<FromInteraction>,
 	) -> Result<(), mpsc::SendError> {
-		while self.requesting_chunks.len() < N_PARALLEL {
+		let max_requests = std::cmp::min(N_PARALLEL, params.threshold);
+		while self.requesting_chunks.len() < max_requests {
 			if let Some(validator_index) = self.shuffling.pop() {
-				let (tx, rx) = oneshot::channel();
+				println!("requesting from {:?}", validator_index);
```
arkpar, line 229:
```suggestion
```

rphmeier:
bot merge

Issue #2513, opened 2021-02-24
Availability Distribution: Consider occupied cores from last k blocks within same session


rphmeier:
A Before and after of availability-recovery metrics on Kusama would be really interesting (number of `NoSuchChunk`):
https://grafana.parity-mgmt.parity.io/goto/sjkh8Upnz?orgId=1 .. currently I see 3-5 `NoSuchChunk` per second.

Discussion here: https://github.com/paritytech/polkadot/pull/2423#discussion_r580616520

This was the previous behavior before using request/response and should be reinstated. A permalink to the previous implementation can be found here: https://github.com/paritytech/polkadot/blob/c66049ddda26c9609d1a2cf23c0975a1f9677712/node/network/availability-distribution/src/lib.rs#L952 .

It is not crucial to the functioning of the network but is useful for maximizing availability and increasing throughput by increasing likelihood of peers having chunks later on.

It is important to only consider blocks within the same session, and this should recover gracefully from runtime API errors as historical state may be (but is unlikely to be) pruned.

eskimor:
[...]

Issue #2443, opened 2021-02-16
Integration tests for subsystems passing messages


rphmeier:
@drahnr Can you add some examples of tests here?

drahnr:
The idea is to have chunks of the whole system covered, this implies `net -> Mod1 -> Mod2 -> Mod3 -> net` while the transported messages might differ and characteristic ones must be selected manually @coriolinus overview slides might come in handy to pick viable candidates to cover a lot of subsets.

Implementation steps:

* [ ] reduce the block time to as little as a couple of 100ms
* [ ] replace the bridge subsystem with a configurable mock
* [ ] create a macro to spawn those "strain" integration tests more easily

At this stage, the majority of the work is already done.

* [ ] run `coz` for individual tests to find lock congestions and bottlenecks

drahnr:
This one has fallen a bit behind, it's essentially the integration of zombienet as cargo test runner with a pretty macro. It's on @pepoviola 's agenda (eventually)

drahnr:
Closing, since the value of adding such a thing at this point is questionable compared to the effort required, given that we can anneal the use cases from both ends already.

PR #2696, opened 2021-03-24
non-script adder collator test with mockable overseer

drahnr:
Immediate goals:

* [ ] unit test of adder collator with multiple validator groups

Extends to:

* [ ] add an overseer mock
   * [ ] allow arbitrary message modifications as a filter stage to the real overseer
   * [ ] ...

Ref #2513
Ref #2443


node/service/src/chain_spec.rs
```diff
@@ -1521,6 +1521,107 @@ pub fn rococo_testnet_genesis(
 	}
 }
 
+
+/// Helper function to create rococo GenesisConfig for testing
+pub fn mousetrap_testnet_genesis(
```
bkchr, line 1526: Why do we need a new runtime for this?

We already got the test-runtime that should be used for this kind of stuff.

drahnr: it would modify the behaviour significantly for all other tests as well and this is still early on exploration phase. Again, this is a DRAFT and tagged WIP.

bkchr: Yeah, I just wanted to warn you.

> behaviour significantly for all other tests as well

If you mean by this that you changed constants in the runtime. Exactly for this reason the constants in the test-runtime are read from the storage, to make them modifiable.

drahnr: That's a good point! Thanks

rphmeier:
Quite stale - close or stay open?

drahnr:
Closing, waaay to stale.

PR #2699, opened 2021-03-24
f+1 validators always approve

rphmeier:
Any set of f+1 validators will always count as an approval on a candidate, as it contains at least one honest approval.

In small testnets, like Rococo, this can prevent failures to approve candidates when all non-backers are in tranche 0 and at least 1 no-show occurs.

### DIFF
--- a/node/core/approval-voting/src/approval_checking.rs
+++ b/node/core/approval-voting/src/approval_checking.rs
@@ -66,6 +66,8 @@ pub enum Check {
 	Unapproved,
 	/// The candidate is approved, with the given amount of no-shows.
 	Approved(usize),
+	/// The candidate is approved by one third of all validators.
+	ApprovedOneThird,
 }
 
 impl Check {
@@ -74,6 +76,15 @@ impl Check {
 		match *self {
 			Check::Unapproved => false,
 			Check::Approved(_) => true,
+			Check::ApprovedOneThird => true,
+		}
+	}
+
+	/// The number of known no-shows in this computation.
+	pub fn known_no_shows(&self) -> usize {
+		match *self {
+			Check::Approved(n) => n,
+			_ => 0,
 		}
 	}
 }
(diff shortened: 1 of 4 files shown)


node/core/approval-voting/src/approval_checking.rs
```diff
@@ -119,6 +131,7 @@ pub fn check_approval(
 			if n_approved + tolerated_missing >= n_assigned {
 				Check::Approved(tolerated_missing)
 			} else {
+				println!("{} + {} < {}", n_approved, tolerated_missing, n_assigned);
```
pepyakin, line 134: Stray println?

node/core/approval-voting/src/lib.rs
```diff
@@ -1316,7 +1316,8 @@ fn check_and_apply_full_approval(
 			required_tranches.clone(),
 		);
 
-		if let approval_checking::Check::Approved(no_shows) = check {
+		let no_shows = check.known_no_shows();
+		if check.is_approved() {
```
pepyakin, line 1320: nit: why not move it under the if?

roadmap/implementers-guide/src/node/approval/approval-voting.md
```diff
@@ -360,8 +360,9 @@ Likewise, when considering how many tranches to take, the no-show depth should b
 #### Check Approval
   * Check whether a candidate is approved under a particular block.
   * Requires `(block_entry, candidate_entry, approval_entry, n_tranches)`
+  * If we have `3 * n_approvals > 2 * n_validators`, return true. This is because any set with f+1 validators must have at least one honest validator, who has approved the candidate.
```
pepyakin, line 363: why *2?

burdges:
At some point, we want a cut off in which we conclude block would never be approved if we reach say 128 assignments.  In essence, this says we abort contentious candidate, rewind the relay chain, and move forward without the contentious candidate.  We'd still complete existing approval checks on the contentious candidate and slash them if found invalid.

pepyakin (approved):
LGTM now

node/core/approval-voting/src/approval_checking.rs
```diff
@@ -74,6 +76,15 @@ impl Check {
...
 			Check::Approved(_) => true,
+			Check::ApprovedOneThird => true,
+		}
+	}
+
+	/// The number of known no-shows in this computation.
+	pub fn known_no_shows(&self) -> usize {
+		match *self {
+			Check::Approved(n) => n,
+			_ => 0,
```
drahnr, line 87:
```suggestion
	pub fn known_no_shows(&self) -> Option<std::num::NonZeroUsize> {
		match *self {
			Check::Approved(n) => std::num::NonZeroUsize::new(n),
			_ => None,
```

there seems to be exactly one use cases which treats `0` as special case, which really should be a `None`.

rphmeier: Why NonZero?

rphmeier: The option makes sense, but tbh we will always use it as unwrap_or(0), so this change would add more friction

rphmeier: The 'right' change would be to expose 'covered + covering' in the other RequiredTranches variants

PR #2700, opened 2021-03-24
use validators.len(), not shuffling

rphmeier:
`self.shuffling` is reduced in size over time


pepyakin (approved):

rphmeier:
actually this is wrong, although the logging change is right, but i fixed it in #2694

PR #2710, opened 2021-03-25
New PVF validation host

pepyakin:
This PR proposes a fresh implementation of PVF validation host. This new version tries to avoid shortcomings of the previous approach and, despite its size, only lays a foundation for the further evolution.

This basically implements the vision outlined here https://hackmd.io/c9GhdTjpT-y0W-w8v8CknA

This PR builds on top of https://github.com/paritytech/substrate/pull/8394 (atm still unmerged, hence inprogress). But note that this is not a companion though, since the mentioned change can be merged independently from this PR.

This PR incorporates the following coupled changes:

1. The new PVF validation host implemented by node/core/pvf. The root of that crate is a good entrypoint for reviewing the code, it has overview of how it works.
2. Integration of it and replacing the legacy validation host, in candidate-validation and CLI.
3. Removal of the legacy validation host.

I recommend approaching this PR in the following order:

1. For a start, you can look at https://github.com/paritytech/substrate/pull/8394 to see how exactly Substrate Executor was changed to provide compiled artifacts and accept them instead of the raw code.
1. Then, you can check out how is the node/core/pvf crate implemented. A high level description of the inner workings can be found in [node/core/pvf/src/lib.rs](https://github.com/paritytech/polkadot/pull/2710/files#diff-4d9f220ad316623151885b620fa1272914c871f5ba346a509d5b57ca565c9606R49-R76).
1. Knowing what is the public interface of the node/core/pvf crate you can see how is the integration into the node is implemented. Specifically, how it is integrated into `candidate-validation` instead of the legacy `polkadot-parachain`'s `wasm_executor`.

Suggestions for what to look for:

1. Race conditions. E.g. races between pruning and either preparation or execution.
1. Problems with and non idiomatic async code (I don't feel confident writing async code)
1. Portability and safety concerns (across POSIX platforms)

### This implementation is POSIX exclusive, why is that?

PVF validation only performed by the validators. Realistically, validators target only POSIX based systems. The same happens with development. Initially I thought that I may implement a fallback for other platforms, it turns out that we don't even have builders for them and also that on windows WSL2 works just fine. Therefore I didn't implement it and the project will no longer compile on Windows.

Should we ever want to port it on windows, it can be done relatively easily.

The project should still build for wasm and android as it did before.

### I see many allocations and a could be better optimized code, can you fix those?

I think I can, but I don't see a lot of sense in it. The code is very cold so I took liberty and optimized for code readability rather than performance.

### Why is caching always goes through files and doesn't attempt to cache in memory?

First of all, we cannot rely exclusively on in memory caches. The artifacts can swell in size and should it be 100 of them in memory that can be quite a big memory requirement. So there should be a fallback that persists the files on disk. Another thing is that we also want to be able to restart the node without recompiling all the artifacts. 

So why don't we use a combination of memory and a file cache? The answer is simplicity and no need. 
We can get away with only a file cache because the recently accessed files will reside in so called page cache, and thus, the disk hit will be avoided should be there enough memory.

One downside is that before execution the file contents should be copied from the page cache to the userland address space of the worker process. This could be addressed later by mmaping these files which would allow deserializing the artifacts directly from the page cache.

### Why worker now returns stringified errors?

Well, actually, it used to as well when running in typical external process isolation mode. These structured errors were only in use when the code was running in the same process mode.

I considered implementing structured errors but didn't see enough benefit in that.

### What is the reason there are two types of workers, rather than one?

I used two different kinds of workers (one for preparation and one for execution) because I anticipate we will introduce resource limits on worker processes and compilation and execution have two different resource consumption profiles.

### Only UTF-8 paths supported

As far as I understand, we only support those. Besides, most of sane configured environments we target are configured to use UTF-8 paths.

### Why Artifacts are prefixed with a manually bumped number rather than to just rely on the version of wasmtime?

I took this path that is there because it seems that it is more manual, which is a good thing in this case IMO. When the version of wasmtime is bumped then it would trigger a test error which would draw attention of a developer, which hopefully would give a chance to react to something unexpected and doesn't seem to be too burdensome. 

Should it get annoying we can consider switching to automatic version sourcing.

### Why not `fork`?

In order to avoid this scheme with creating an unix socket and spawn a process, we theoretically could use `fork`. However, turns out that fork is really too sharp of a tool. The main concern is that the forked process starts with only one thread running. This in turn makes such an environment similar to a signal handler. Such an environment is very limited, due to basic runtime services such as memory allocations could lead to a dead lock. I am not sure if there is a way to resume all threads, but resuming all threads also doesn't seem to be a good idea in the worker processes.

This sounds like a can of worms so I didn't go there.


pepyakin:
- I just have checked and macOS seem to work just fine. 
- The test failure indicates that I broke the test service.

node/core/pvf/src/artifacts.rs
```diff
@@ -0,0 +1,288 @@
...
+
+	/// Returns the expected path to this artifact given the root of the cache.
+	pub fn path(&self, cache_path: &Path) -> PathBuf {
+		let file_name = format!("{}{}", Self::PREFIX, self.code_hash.to_string());
+		cache_path.join(file_name)
+	}
+}
+
+#[test]
+fn ensure_wasmtime_version() {
```
bkchr, line 92: I understand that this is a somewhat special test, but why not move it to the tests module?

bkchr: Const panics would be nice for this.

node/core/pvf/src/artifacts.rs
```diff
@@ -0,0 +1,288 @@
...
+	//
+	// If in doubt bump it. This will lead to removal of the existing artifacts in the on-disk cache
+	// and recompilation.
+}
+
+pub enum ArtifactState {
+	/// The artifact is ready to be used by the executor.
+	///
+	/// That means that the artifact should be accessible through the path obtained by the artifact
+	/// id under (unless, it was removed externally).
```
bkchr, line 108:
```suggestion
	/// id (unless, it was removed externally).
```

node/core/pvf/src/artifacts.rs
```diff
@@ -0,0 +1,288 @@
...
+	/// Insert an artifact with the given ID as "prepared".
+	#[cfg(test)]
+	pub fn insert_prepared(&mut self, artifact_id: ArtifactId, last_time_needed: SystemTime) {
+		always!(self
+			.artifacts
+			.insert(artifact_id, ArtifactState::Prepared { last_time_needed })
+			.is_none());
+	}
+
+	/// Remove and retrive the artifacts from the table that are older than the supplied Time-To-Live.
```
bkchr, line 177: We don't remove in here?

node/core/pvf/src/executor_intf.rs
```diff
@@ -0,0 +1,236 @@
...
+			.call(InvokeMethod::Export("validate_block"), params)?)
+	})?;
+
+	result
+}
+
+type HostFunctions = sp_io::SubstrateHostFunctions;
+
+/// The validation externalities that will panic on any storage related access. They just provide
+/// access to the parachain extension.
```
bkchr, line 85:
```suggestion
/// The validation externalities that will panic on any storage related access.
```

node/core/pvf/src/executor_intf.rs
```diff
@@ -0,0 +1,236 @@
...
+
+/// An implementation of `SpawnNamed` on top of a futures' thread pool.
+///
+/// This is a light handle meaning it will only clone the handle not create a new thread pool.
+#[derive(Clone)]
+pub(crate) struct TaskExecutor(futures::executor::ThreadPool);
+
+impl TaskExecutor {
+	pub(crate) fn new() -> Result<Self, String> {
+		futures::executor::ThreadPool::new()
```
bkchr, line 223: I know that you just copied this code, however I think we maybe should restrict this thread pool to 2 threads maybe. By default, it will use as many threads as there exist cores

node/core/pvf/src/lib.rs
```diff
@@ -0,0 +1,100 @@
...
+//! (b) Heads up. This request allows to signal that the given PVF may be needed soon and that it
+//!     should be prepared for execution.
+//!
+//! The preparation results are cached for some time after they either used or was signalled in heads up.
+//! All requests that depends on preparation of the same PVF are bundled together and will be executed
+//! as soon as the artifact is prepared.
+//!
+//! # Priority
+//!
+//! PVF execution requests can specify the [priority][`Priority`] with which the given request should
```
bkchr, line 39: > [priority][`Priority`]

Is this valid syntax? :D

node/core/pvf/bin/puppet_worker.rs
```diff
@@ -0,0 +1,42 @@
...
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with Polkadot.  If not, see <http://www.gnu.org/licenses/>.
+
+//! A simple binary that allows mocking the behavior of the real worker.
+
+use std::time::Duration;
+
+fn main() {
```
bkchr, line 21: Could we put this in some kind of macro or whatever? :P 

Cumulus currently uses the in-process capability for its integration test and it would be nice to create such a worker relative easy :D

node/core/pvf/src/execute/queue.rs
```diff
@@ -0,0 +1,329 @@
...
+
+			purge_dead(&mut self.workers).await;
+		}
+	}
+}
+
+async fn purge_dead(workers: &mut Workers) {
+	let mut to_remove = vec![];
+	for (worker, data) in workers.running.iter_mut() {
+		if let Poll::Ready(()) = futures::poll!(&mut data.handle) {
```
bkchr, line 157:
```suggestion
		if futures::poll!(&mut data.handle).is_ready() {
```

node/core/pvf/src/execute/worker.rs
```diff
@@ -0,0 +1,230 @@
...
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with Polkadot.  If not, see <http://www.gnu.org/licenses/>.
+
+use crate::{artifacts::Artifact, executor_intf::TaskExecutor, worker_common::{
+		IdleWorker, SpawnErr, WorkerHandle, bytes_to_path, framed_recv, framed_send, path_to_bytes,
+		spawn_with_program_path, worker_event_loop,
+	}};
```
bkchr, line 20: Formatting

node/core/pvf/src/worker_common.rs
```diff
@@ -0,0 +1,280 @@
...
+#[doc(hidden)]
+pub async fn spawn_with_program_path(
+	debug_id: &'static str,
+	program_path: &str,
+	extra_args: &'static [&'static str],
+	spawn_timeout_secs: u64,
+) -> Result<(IdleWorker, WorkerHandle), SpawnErr> {
+	with_transient_socket_path(debug_id, |socket_path| {
+		let socket_path = socket_path.to_owned();
+		let program_path = PathBuf::from(program_path);
```
bkchr, line 49: Why don't we pass this directly as `PathBuf`? 

At the main calling site, convert the path before into some string. Looks weird.

node/core/pvf/src/worker_common.rs
```diff
@@ -0,0 +1,280 @@
...
+pub fn bytes_to_path(bytes: &[u8]) -> Option<PathBuf> {
+	let str_buf = std::str::from_utf8(bytes).ok()?;
+
+	match PathBuf::from_str(&str_buf) {
+		Ok(path) => Some(path),
+		Err(never) => {
+			// std::convert::Infallible doesn't allow for irrefutable patterns for some reason.
+			match never {}
+		}
+	}
```
bkchr, line 263:
```suggestion
	PathBuf::from_str(&str_buf).ok()
```

?

bkchr (commented):
Did not yet reviewed all of it

node/core/pvf/src/worker_common.rs
pepyakin: Yeah I wanted to emphasize that parsing `PathBuf` doesn't fail and introduced this poor man's `Result::into_ok` (which is still nightly only). 

I don't mind changing that, but I would prefer `expect` here rather than `ok` tho

node/core/pvf/src/execute/worker.rs
pepyakin: I have no clue what happened here 🙈

node/core/pvf/src/lib.rs
pepyakin: Yeah I believe so, it's like Markdown but if the link destination is backticks then it will be resolved as an path.

node/core/pvf/Cargo.toml
```diff
@@ -0,0 +1,56 @@
...
+version = "0.1.0"
+authors = ["Parity Technologies <admin@parity.io>"]
+edition = "2018"
+
+[[bin]]
+name = "puppet_worker"
+path = "bin/puppet_worker.rs"
+
+[[test]]
+name = "it"
```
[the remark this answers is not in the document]
pepyakin: I am actually not sure: `it` may be a leftover or it may be as well needed. I will check that.

node/core/pvf/src/artifacts.rs
pepyakin: I approached this with `static_assertions` first and sadly discovered that it only works with numbers. You cannot unfortunately make string comparisons in the const context.

Yeah, this is to make it closer to the `ArtifactId`, don't mind moving it downstairs under the tests module though. Just ask

node/core/pvf/src/execute/queue.rs
pepyakin: I prefer the former tho : )

node/core/pvf/src/executor_intf.rs
pepyakin: Oh heh this thought also crossed my mind but I forgot then! I think it's a good thing to fix it here. I would pick a greater number than 2 though.

node/core/pvf/src/artifacts.rs
pepyakin: Ugh, the comment is right and removal was lost during a refactoring. Fixed it and modified the test.

node/core/pvf/Cargo.toml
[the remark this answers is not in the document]
pepyakin: It's a leftover

node/core/pvf/src/worker_common.rs
pepyakin: Ok, so I changed this, however not directly to `PathBuf` but `impl Into<PathBuf>`. One of the three call sites is actually a string.

node/core/pvf/src/worker_common.rs
bkchr: Why not just use `PathBuf::from`?

node/core/pvf/src/artifacts.rs
bkchr: Yeah, would be cleaner IMHO.

node/core/pvf/src/lib.rs
bkchr: Okay fine :)

I always thought that only `[`Something`](something::Something)` works

node/core/pvf/src/worker_common.rs
pepyakin: ah damn, of course!

node/service/src/lib.rs
```diff
@@ -654,8 +653,8 @@ pub fn new_full<RuntimeApi, Executor>(
 	is_collator: IsCollator,
 	grandpa_pause: Option<(u32, u32)>,
 	jaeger_agent: Option<std::net::SocketAddr>,
-	isolation_strategy: IsolationStrategy,
 	telemetry_worker_handle: Option<TelemetryWorkerHandle>,
+	program_path: Option<std::path::PathBuf>,
```
rphmeier, line 657: This is the path of the binary itself? Or is it a db-path or somtehing like that?

node/core/pvf/src/artifacts.rs
```diff
@@ -0,0 +1,296 @@
...
+	}
+
+	/// Returns the state of the given artifact by its ID.
+	pub fn artifact_state_mut(&mut self, artifact_id: &ArtifactId) -> Option<&mut ArtifactState> {
+		self.artifacts.get_mut(artifact_id)
+	}
+
+	/// Inform the table about the artifact with the given ID. The state will be set to "preparing".
+	pub fn insert_preparing(&mut self, artifact_id: ArtifactId) {
+		always!(self
```
rphmeier, line 147: Is this a type of `debug_assert!`? We should add `qed` proofs for all such assertions.

node/core/pvf/Cargo.toml
```diff
@@ -0,0 +1,53 @@
...
+authors = ["Parity Technologies <admin@parity.io>"]
+edition = "2018"
+
+[[bin]]
+name = "puppet_worker"
+path = "bin/puppet_worker.rs"
+
+[dependencies]
+always-assert = "0.1"
+async-std = { version = "1.8.0", features = ["attributes"] }
```
rphmeier, line 13: Is it OK to mix `async-std` and tokio?

node/core/pvf/src/error.rs
```diff
@@ -0,0 +1,56 @@
...
+	/// This also includes the errors reported by the preparation pipeline.
+	WorkerReportedError(String),
+	/// The worker has died during validation of a candidate. That may fall in one of the following
+	/// categories, which we cannot distinguish programmatically:
+	///
+	/// (a) Some sort of transient glitch caused the worker process to abort. An example would be that
+	///     the host machine ran out of free memory and the OOM killer started killing the processes,
+	///     and in order to save the parent it will "sacrifice child" first.
+	///
+	/// (b) The candidate trigerred a code path that has lead to the process death. For example,
```
rphmeier, line 41: typo: triggered

node/core/pvf/src/error.rs
```diff
@@ -0,0 +1,56 @@
...
+	///     the PVF found a way to consume unbounded amount of resources and then it either exceeded
+	///     an rlimit (if set) or, again, invited OOM killer. Another possibility is a bug in
+	///     wasmtime allowed the PVF to gain control over the execution worker.
+	///
+	/// We attribute such an event to an invalid candidate in either case.
```
rphmeier, line 46: This seems mostly fair but it is worrying. For instance, it will not be a problem in backing, but if it leads the validator to issue a dispute against the candidate which is then counteracted by all other validators, the validator will be slashed.

Ambiguous child process death should be very rare, though, no?

node/core/pvf/src/host.rs
```diff
@@ -0,0 +1,1039 @@
...
+}
+
+/// Configuration for the validation host.
+pub struct Config {
+	/// The root directory where the prepared artifacts can be stored.
+	pub cache_path: PathBuf,
+	/// The path to the program that can be used to spawn the prepare workers.
+	pub prepare_worker_program_path: PathBuf,
+	/// The time alloted for a prepare worker to spawn and report to the host.
+	pub prepare_worker_spawn_timeout: u64,
```
rphmeier, line 114: It would be easier to read if the var name held the time

node/core/pvf/src/host.rs
```diff
@@ -0,0 +1,1039 @@
...
+	/// The maximum number of workers that can be spawned in the prepare pool for tasks with the
+	/// priority below crticial.
```
rphmeier, line 116: typo: critical

node/core/pvf/src/host.rs
```diff
@@ -0,0 +1,1039 @@
...
+	// Make sure that the task-futures are fused.
+	let mut prepare_queue = prepare_queue.fuse();
+	let mut prepare_pool = prepare_pool.fuse();
+	let mut execute_queue = execute_queue.fuse();
+	let mut sweeper = sweeper.fuse();
+
+	loop {
+		futures::select_biased! {
+			_ = prepare_queue => {
+				never!();
```
rphmeier, line 312: debug_asserts should have `qed` proofs

node/core/pvf/src/error.rs
burdges: Yes, I suppose validators must check rlimit, system memory, etc. before running and simply not in validator mode if they lack the resource, but..

As in https://github.com/paritytech/polkadot/pull/1656 we should worry if machines behave differently here, like an adversary triggers a slash because someone set their rlimit different from everyone else.  

We should start with quite low slashing for triggering an escalation unjustly.  We'll refund such slashes of course, except there exists some risk we might've trouble reproducing, making a big mess.

node/core/pvf/src/worker_common.rs
```diff
@@ -0,0 +1,272 @@
...
+
+	// Best effort to remove the socket file. Under normal circumstances the socket will be removed
+	// by the worker. We make sure that it is removed here, just in case a failed rendezvous.
+	let _ = async_std::fs::remove_file(socket_path).await;
+
+	result
+}
+
+pub fn tmpfile(prefix: &str) -> PathBuf {
+	use rand::distributions::Alphanumeric;
```
rphmeier, line 87: cool!

node/core/pvf/src/worker_common.rs
```diff
@@ -0,0 +1,272 @@
...
+			.take()
+			.expect("the process spawned with piped stdout should have the stdout handle");
+
+		Ok(WorkerHandle {
+			child,
+			stdout,
+			// We don't expect the bytes to be ever read. But in case we do, we should not use a buffer
+			// of a small size, because otherwise if the child process does return any data we will end up
+			// issuing a syscall for each byte. We also prefer not to do allocate that on the stack, since
+			// each poll the buffer will be allocated and initialized (and that's due poll_read takes &mut [u8]
```
rphmeier, line 195: makes sense

node/core/pvf/src/execute/worker.rs
```diff
@@ -0,0 +1,233 @@
...
+	worker_event_loop("execute", socket_path, |mut stream| async move {
+		let executor = TaskExecutor::new().map_err(|e| {
+			io::Error::new(
+				io::ErrorKind::Other,
+				format!("cannot create task executor: {}", e),
+			)
+		})?;
+		loop {
+			let (artifact_path, params) = recv_request(&mut stream).await?;
+			let artifact_bytes = async_std::fs::read(&artifact_path).await?;
```
rphmeier, line 180: > It pre-allocates a buffer based on the file size when available, so it is typically faster than manually opening a file and reading from i

I was checking whether this buffers, and indeed it does.

node/core/pvf/src/execute/worker.rs
```diff
@@ -0,0 +1,233 @@
...
+			let artifact = Artifact::deserialize(&artifact_bytes).map_err(|e| {
+				io::Error::new(
+					io::ErrorKind::Other,
+					format!("artifact deserialization error: {}", e),
+				)
+			})?;
```
rphmeier, line 186: Is it worth sending a response of some kind even if this fails, or will the host learn about the failure some other way?

node/core/pvf/src/prepare/queue.rs
```diff
@@ -0,0 +1,855 @@
...
+		self.queue_mut(prio).push_front(job);
+	}
+
+	fn is_empty(&self) -> bool {
+		self.background.is_empty() && self.normal.is_empty() && self.critical.is_empty()
+	}
+
+	fn next(&mut self) -> Option<Job> {
+		let mut check = |prio: Priority| self.queue_mut(prio).pop_front();
+		None.or_else(|| check(Priority::Critical))
```
rphmeier, line 132: `None` here is redundant, I suppose?

node/core/pvf/src/prepare/queue.rs
```diff
@@ -0,0 +1,855 @@
...
+		} => {
+			handle_amend(queue, priority, artifact_id).await?;
+		}
+	}
+	Ok(())
+}
+
+async fn handle_enqueue(queue: &mut Queue, priority: Priority, pvf: Pvf) -> Result<(), Fatal> {
+	let artifact_id = pvf.as_artifact_id();
+	if never!(queue.artifact_id_to_job.contains_key(&artifact_id)) {
```
rphmeier, line 231: Why is this a `never!`?

rphmeier (approved):
Generally looks good!

node/core/pvf/src/error.rs
rphmeier: Disputes should probably have asymmetric slashing. If you dispute against something and lose, you get slashed less than if you dispute for something and lose.

burdges: Yes exactly, initially way way less, but later maybe significant.

node/core/pvf/src/prepare/queue.rs
pepyakin: Yeah, just to make a chain of `or_else`. Does it look too weird?

node/core/pvf/src/host.rs
pepyakin: agreed. Or perhaps we can just plumb `Duration`

node/core/pvf/Cargo.toml
pepyakin: uh-oh!! 

I assumed the effective executor (i.e. the stuff that was spawned with ctx.spawn_blocking) is async-std, but I never actually checked. So you are saying it is tokio?

node/core/pvf/src/artifacts.rs
pepyakin: Yes which acts as a log in release.

Hm, I am not sure that the case. And I just run a search across the codebase and the status quo for `debug_assert` is that we don't. In fact, most of them go even without comments.

I can still add the proofs if you insist.

Also this is a somewhat a new situation since this is a fusion of debug_assert and log. Which makes me think that this should not contain proofs because they would look weird in logs.

node/service/src/lib.rs
pepyakin: The former one. Do you think the name is confusing?

node/core/pvf/src/prepare/queue.rs
pepyakin: I should leave a comment

pepyakin (commented):
Answers

node/core/pvf/src/prepare/queue.rs
rphmeier: Or add a `qed` proof

node/core/pvf/Cargo.toml
rphmeier: I don't know for sure 🙈

node/core/pvf/src/artifacts.rs
rphmeier: We had a conversation with Gav a couple days ago where he said to use the `qed` proofs for debug asserts, so I am basing my review on that

pepyakin: Gotcha, will add proper proofs.

node/core/pvf/Cargo.toml
pepyakin: Ok, this shouldn't be a problem. Apparently `async_std` should work on all executors.

node/service/src/lib.rs
ordian: Maybe `exe_path` is less ambiguous and in line with `std::evn::current_exe`?

node/core/pvf/src/worker_common.rs
```diff
@@ -0,0 +1,272 @@
...
+	let result = f(&socket_path).await;
+
+	// Best effort to remove the socket file. Under normal circumstances the socket will be removed
+	// by the worker. We make sure that it is removed here, just in case a failed rendezvous.
+	let _ = async_std::fs::remove_file(socket_path).await;
+
+	result
+}
+
+pub fn tmpfile(prefix: &str) -> PathBuf {
```
bkchr, line 86: What happens when the path already exists?

Why don't you just use the common crate for this? I think this maps down to some OS call?

node/core/pvf/src/prepare/worker.rs
```diff
@@ -0,0 +1,187 @@
...
+}
+
+/// The entrypoint that the spawned prepare worker should start with. The socket_path specifies
+/// the path to the socket used to communicate with the host.
+pub fn worker_entrypoint(socket_path: &str) {
+	worker_event_loop("prepare", socket_path, |mut stream| async move {
+		loop {
+			let code = recv_request(&mut stream).await?;
+
+			let artifact_bytes = prepare_artifact(&code).serialize();
```
bkchr, line 163: Maybe some logging here in the loop would be nice.

node/core/pvf/src/execute/worker.rs
```diff
@@ -0,0 +1,257 @@
...
+}
+
+async fn send_request(
+	stream: &mut UnixStream,
+	artifact_path: &Path,
+	validation_params: &[u8],
+) -> io::Result<()> {
+	framed_send(stream, path_to_bytes(artifact_path)).await?;
+	framed_send(stream, validation_params).await?;
+	Ok(())
```
bkchr, line 128:
```suggestion
	framed_send(stream, validation_params).await
```

node/core/pvf/src/execute/worker.rs
```diff
@@ -0,0 +1,257 @@
...
+			"execute pvf recv_request: non utf-8 artifact path".to_string(),
+		)
+	})?;
+	let params = framed_recv(stream).await?;
+	Ok((artifact_path, params))
+}
+
+async fn send_response(stream: &mut UnixStream, response: Response) -> io::Result<()> {
+	framed_send(stream, &response.encode()).await?;
+	Ok(())
```
bkchr, line 145:
```suggestion
	framed_send(stream, &response.encode()).await
```

node/core/pvf/src/execute/worker.rs
```diff
@@ -0,0 +1,257 @@
...
+
+async fn recv_response(stream: &mut UnixStream) -> io::Result<Response> {
+	let response_bytes = framed_recv(stream).await?;
+	let response = Response::decode(&mut &response_bytes[..]).map_err(|e| {
+		io::Error::new(
+			io::ErrorKind::Other,
+			format!("execute pvf recv_response: decode error: {:?}", e),
+		)
+	})?;
+	Ok(response)
```
bkchr, line 156:
```suggestion
	Response::decode(&mut &response_bytes[..]).map_err(|e| {
		io::Error::new(
			io::ErrorKind::Other,
			format!("execute pvf recv_response: decode error: {:?}", e),
		)
	})
```

node/core/pvf/src/worker_common.rs
```diff
@@ -0,0 +1,272 @@
...
+{
+	let err = async_std::task::block_on::<_, io::Result<()>>(async move {
+		let stream = UnixStream::connect(socket_path).await?;
+		let _ = async_std::fs::remove_file(socket_path).await;
+
+		event_loop(stream).await?;
+
+		Ok(())
+	})
+	.unwrap_err();
```
bkchr, line 119: Unwrap error? Hmm?

node/core/pvf/src/prepare/worker.rs
```diff
@@ -0,0 +1,187 @@
...
+	/// this could've happen because the worker hadn't finished the work until the given deadline.
+	///
+	/// Note that in this case the artifact file is written (unless there was an error writing the
+	/// the artifact).
+	///
+	/// This doesn't return an idle worker instance, thus this worker is no longer usable.
+	DidntMakeIt,
+}
+
+pub async fn start_work(
```
bkchr, line 68: Some docs would be nice.

node/core/pvf/src/prepare/pool.rs
```diff
@@ -0,0 +1,336 @@
...
+					never!();
+				}
+			} else {
+				// That's a relatively normal situation since the queue may send `start_work` and
+				// before receiving it the pool would report that the worker died.
+			}
+		}
+		ToPool::Kill(worker) => {
+			// It may be absent if it were previously already removed by `purge_dead`.
+			let _ = spawned.remove(worker);
```
bkchr, line 220: How does that kill the worker?

node/core/pvf/src/host.rs
```diff
@@ -0,0 +1,1049 @@
...
+	Future, FutureExt, SinkExt, StreamExt,
+	channel::{mpsc, oneshot},
+};
+
+/// An alias to not spell the type for the oneshot sender for the PVF execution result.
+pub(crate) type ResultSender = oneshot::Sender<Result<ValidationResult, ValidationError>>;
+
+/// A handle to the async process serving the validation host requests.
+pub struct ValidationHost {
+	to_host_tx: Mutex<mpsc::Sender<ToHost>>,
```
bkchr, line 48: Why do we need the mutex?

node/core/pvf/src/host.rs
```diff
@@ -0,0 +1,1049 @@
...
+	to_sweeper_tx: mpsc::Sender<PathBuf>,
+
+	awaiting_prepare: AwaitingPrepare,
+}
+
+#[derive(Debug)]
+struct Fatal;
+
+async fn run(
+	Inner {
```
bkchr, line 271: So this only exists to have named parameters? But why do we pass 4 further parameters below?

node/core/pvf/src/host.rs
```diff
@@ -0,0 +1,1049 @@
...
+	let mut from_prepare_queue_rx = from_prepare_queue_rx.fuse();
+
+	// Make sure that the task-futures are fused.
+	let mut prepare_queue = prepare_queue.fuse();
+	let mut prepare_pool = prepare_pool.fuse();
+	let mut execute_queue = execute_queue.fuse();
+	let mut sweeper = sweeper.fuse();
+
+	loop {
+		futures::select_biased! {
```
bkchr, line 310: Why biased?

node/core/pvf/src/prepare/queue.rs
```diff
@@ -0,0 +1,891 @@
...
+		self.queue_mut(prio).push_front(job);
+	}
+
+	fn is_empty(&self) -> bool {
+		self.background.is_empty() && self.normal.is_empty() && self.critical.is_empty()
+	}
+
+	fn next(&mut self) -> Option<Job> {
+		let mut check = |prio: Priority| self.queue_mut(prio).pop_front();
+		None.or_else(|| check(Priority::Critical))
```
bkchr, line 132:
```suggestion
		check(Priority::Critical)
```

Why not just this?

bkchr (approved):
The general design look really good. I'm not sure that I already understood everything 100%, but it looks really good ;)

node/core/pvf/src/host.rs
```diff
@@ -0,0 +1,1049 @@
...
+async fn handle_cleanup_pulse(
+	cache_path: &Path,
+	sweeper_tx: &mut mpsc::Sender<PathBuf>,
+	artifacts: &mut Artifacts,
+	artifact_ttl: Duration,
+) -> Result<(), Fatal> {
+	let to_remove = artifacts.prune(artifact_ttl);
+	for artifact_id in to_remove {
+		let artifact_path = artifact_id.path(cache_path);
+		sweeper_tx.send(artifact_path).await.map_err(|_| Fatal)?;
```
ordian, line 591: the error is unreachable because `sweeper_task` only stops when the sender is dropped

node/core/pvf/src/prepare/pool.rs
pepyakin: Through `Drop`. The value type is `WorkerData` which owns a `worker_common::WorkerHandle` which owns `async_process::Child`, which is configured to kill the process on drop.

node/core/pvf/src/pvf.rs
```diff
@@ -0,0 +1,56 @@
...
+	fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+		write!(f, "Pvf {{ code, code_hash: {:?} }}", self.code_hash)
+	}
+}
+
+impl Pvf {
+	/// Returns an instance of the PVF out of the given PVF code.
+	pub fn from_code(code: Vec<u8>) -> Self {
+		let code = Arc::new(code);
+		let code_hash = keccak_256(&code).into();
```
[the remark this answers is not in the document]
pepyakin: Do you @bkchr / @rphmeier think there should be other hash function be used?

node/core/pvf/src/worker_common.rs
```diff
@@ -0,0 +1,272 @@
...
+	//
+	// However, should be there reports we can incorporate such a crate here.
+	path.to_str().expect("non-UTF-8 path").as_bytes()
+}
+
+/// Interprets the given bytes as a path. Returns `None` if the given bytes do not constitute a
+/// a proper utf-8 string.
+pub fn bytes_to_path(bytes: &[u8]) -> Option<PathBuf> {
+	let str_buf = std::str::from_utf8(bytes).ok()?;
+	Some(PathBuf::from(str_buf))
```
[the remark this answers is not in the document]
pepyakin: Don't mind yeah

node/core/pvf/src/priority.rs
```diff
@@ -0,0 +1,46 @@
...
+}
+
+impl Priority {
+	/// Returns `true` if `self` is `Crticial`
+	pub fn is_critical(&self) -> bool {
+		*self == Priority::Critical
+	}
+
+	/// Returns `true` if `self` is `Background`
+	pub fn is_background(&self) -> bool {
```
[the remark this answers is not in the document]
pepyakin: Sure, np

node/core/pvf/src/execute/queue.rs
```diff
@@ -0,0 +1,332 @@
...
+			// a resolved future means that the worker has terminated. Weed it out.
+			to_remove.push(worker);
+		}
+	}
+	for w in to_remove {
+		let _ = workers.running.remove(w);
+	}
+}
+
+fn handle_to_queue(queue: &mut Queue, to_queue: ToQueue) {
```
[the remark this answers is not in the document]
pepyakin: yeah there were some reasons to use another approach and I didn't find enough motivation to switch the existing code )

node/core/pvf/src/execute/queue.rs
```diff
@@ -0,0 +1,332 @@
...
+			.iter()
+			.find_map(|d| if d.1.idle.is_some() { Some(d.0) } else { None })
+	}
+
+	fn claim_idle(&mut self, worker: Worker) -> IdleWorker {
+		let data = self
+			.running
+			.get_mut(worker)
+			.expect("the worker doesn't exist");
+		data.idle.take().expect("the worker has no idle token")
```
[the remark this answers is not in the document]
pepyakin: ok, will fix

node/core/pvf/src/prepare/queue.rs
pepyakin: If I knew it would attract so much attention I wouldn't have used it 🙈

node/core/pvf/src/worker_common.rs
pepyakin: Yeah, initially I just didn't want to add yet another dependency, but since then the crate changed a lot. I will change it to tempfile.

node/core/pvf/src/prepare/worker.rs
pepyakin: Yeah. Initially I assumed there will be no logging for the worker processes, but turned out that the legacy executor enabled them.

node/core/pvf/src/host.rs
pepyakin: again, historical reasons )

I had to extract these task-futures just because I didn't want to name them, so passing them as parameters allowed me to use `impl Future`.

node/core/pvf/src/host.rs
pepyakin: Good point, forgot to left a comment. The same reason as in the prep queue: just to enable deterministic testing.

node/core/pvf/src/host.rs
pepyakin: Yeah, and you can say the same about the other uses of `Fatal`, with the exception that here it is trivially visible. Not sure if expect would be more appropriate here.

parachain/test-parachains/adder/collator/Cargo.toml
```diff
@@ -28,6 +32,11 @@ sp-core = { git = "https://github.com/paritytech/substrate", branch = "master" }
 sc-authority-discovery = { git = "https://github.com/paritytech/substrate", branch = "master" }
 sc-service = { git = "https://github.com/paritytech/substrate", branch = "master" }
 
+# This one is tricky. Even though it is not used directly by the collator, we still need it for the
+# `puppet_worker` binary, which is required for the integration test. However, this shouldn't be
+# a big problem since it is used transitively, at least when built with `real-overseer`.
+polkadot-node-core-pvf = { path = "../../../../node/core/pvf", optional = true }
```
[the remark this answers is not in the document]
pepyakin: Yeah, each time we need to create a puppet worker. The alternative where you want to test PVF execution is to avoid using the whole validation host with external processes and all that jazz, but that would require you to depend on `polkadot-node-core-pvf` as well. That would be a dev-dependency though.

node/service/src/lib.rs
pepyakin: `exe` and `program`  sounds synonymous for me. I suspect the problem may be that it is not immediately obvious what program/executable we are talking about here.

node/core/pvf/src/host.rs
pepyakin: Just to not require `&mut` on methods of `ValidationHost`.

node/core/pvf/src/worker_common.rs
pepyakin: oh, this is a remnant of previous version where it could only return with an error.

pepyakin: ah, it's fine. the future's result is `Never`. I'll leave comment

node/core/pvf/src/host.rs
bkchr: But why? xD

node/core/pvf/src/pvf.rs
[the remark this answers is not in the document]
bkchr: We use `Blake2` everywhere else. I'm just curious why you want to use another one here.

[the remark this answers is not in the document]
pepyakin: Yeah, makes sense, will change to blake2

node/core/pvf/src/host.rs
pepyakin: Changed it to be &mut and no Mutex.

node/core/pvf/src/host.rs
ordian: probably fine as is, just an observation

node/core/pvf/src/artifacts.rs
```diff
@@ -0,0 +1,312 @@
...
+		// Make sure that the cache path directory and all it's parents are created.
+		let _ = async_std::fs::create_dir_all(cache_path).await;
+
+		let artifacts = match scan_for_known_artifacts(cache_path).await {
+			Ok(a) => a,
+			Err(err) => {
+				tracing::warn!(
+					target: LOG_TARGET,
+					"unable to seed the artifacts in memory cache: {:?}. Starting with a clean one",
+					err,
```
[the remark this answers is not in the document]
pepyakin: not sure. If we bump onto unrecognized files we remove them. The only thing that can go wrong is listing the directory which I doubt could be solved by deleting the directory.

parachain/test-parachains/adder/collator/Cargo.toml
```diff
@@ -28,6 +32,11 @@ sp-core = { git = "https://github.com/paritytech/substrate", branch = "master" }
 sc-authority-discovery = { git = "https://github.com/paritytech/substrate", branch = "master" }
 sc-service = { git = "https://github.com/paritytech/substrate", branch = "master" }
 
+# This one is tricky. Even though it is not used directly by the collator, we still need it for the
+# `puppet_worker` binary, which is required for the integration test. However, this shouldn't be
+# a big problem since it is used transitively, at least when built with `real-overseer`.
```
ordian, line 37:
```suggestion
```

parachain/test-parachains/adder/collator/Cargo.toml
```diff
@@ -36,4 +45,6 @@ substrate-test-utils = { git = "https://github.com/paritytech/substrate", branch
 sc-service = { git = "https://github.com/paritytech/substrate", branch = "master" }
 sp-keyring = { git = "https://github.com/paritytech/substrate", branch = "master" }
 
+polkadot-node-core-pvf = { path = "../../../../node/core/pvf" }
+
```
ordian, line 49:
```suggestion
```

ordian, line 49:
```suggestion
```

ordian:
(sorry for introducing more merge conflicts)

pepyakin (commented):
_

parachain/test-parachains/adder/collator/Cargo.toml
ordian: we still don't need it both as a dev-dep and a regular one?

pepyakin: ah right, we only need to have it as a regular dep.

ordian:
bot merge

PR #2714, opened 2021-03-25
Simnet triggers and images

TriplEight:
- add building binaries for `simnet` on `master` and schedules
- separate jobs for `simnet` binaries and images
- image file for a dummy collator
- return status from the triggered Simnet pipeline
- chore: fixes for small bugs (`mdbook` silently warns about errors, a mistake in testnet triggering)

What could be better:
- `build-linux-release` and `build-real-overseer` are now building separately but will run in parallel on `master` and schedules. And the latter could profit from reusing cache, but the job that would run them together would look ugly.
- `trigger-fail-guard:` is the workaround for [Gitlab's bug](https://gitlab.com/gitlab-org/gitlab/-/issues/326137). 
   TLDR; triggered job does not send the `job event` status to GH.


.gitlab-ci.yml
```diff
@@ -336,23 +404,24 @@ deploy-polkasync-kusama:
...
     POLKADOT_CI_COMMIT_NAME:       "${CI_COMMIT_REF_NAME}"
-    POLKADOT_CI_COMMIT_REF:        "${CI_COMMIT_REF}"
+    POLKADOT_CI_COMMIT_REF:        "${CI_COMMIT_SHORT_SHA}"
   allow_failure:                   true
   trigger:                         "parity/infrastructure/parity-testnet"
 
 trigger-simnet:
   stage:                           deploy
-  <<:                              *rules-test
+  # DAG from `build-adder-collator` job
```
coriolinus, line 417: I'm not sure what this comment means. If DAG means "directed acyclic graph" here, then the sentence doesn't make a lot of sense. If it means something else, then I don't know the acronym. Could be explained better.

coriolinus (approved):
Nits:

- `roadmap/` changes don't appear to have anything to do with the rest of the PR
- PR title should probably not be "WIP" anymore

Other than that, the GHA config and `collator.Dockerfile` look good to me, once the rest of CI is green.

.gitlab-ci.yml
TriplEight: It is a directed acyclic graph that starts on the mentioned job being dependent on whether if it included in the pipeline or not. Ok I will explain it a bit broader.

TriplEight:
The PR is ready, but I want to wait for Simnet to become green before this merge.

TriplEight:
`roadmap/` ones are just nits I've noticed doing this PR.

PR #2718, opened 2021-03-25
Add block number to activated leaves and associated fixes

rphmeier:
1. Block import notifications can come for forks against old blocks. Therefore more recently importedleaves were not necessarily more recent. This lead to an error in view computation.
2. `live_block_hash` used to query state could move backwards significantly in availability recovery

In BABE block height isn't as important as "number of Primary digests" but block height is a close proxy for it.

### DIFF
--- a/node/network/availability-recovery/src/lib.rs
+++ b/node/network/availability-recovery/src/lib.rs
@@ -522,8 +522,8 @@ async fn handle_signal(
 		OverseerSignal::Conclude => Ok(true),
 		OverseerSignal::ActiveLeaves(ActiveLeavesUpdate { activated, .. }) => {
 			// if activated is non-empty, set state.live_block_hash to the first block in Activated.
-			if let Some(hash) = activated.get(0) {
-				state.live_block_hash = hash.0;
+			if let Some(a) = activated.get(0) {
+				state.live_block_hash = a.hash;
 			}
 
 			Ok(false)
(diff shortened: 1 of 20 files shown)


node/network/availability-recovery/src/lib.rs
```diff
@@ -521,9 +521,11 @@ async fn handle_signal(
 	match signal {
 		OverseerSignal::Conclude => Ok(true),
 		OverseerSignal::ActiveLeaves(ActiveLeavesUpdate { activated, .. }) => {
-			// if activated is non-empty, set state.live_block_hash to the first block in Activated.
-			if let Some(hash) = activated.get(0) {
-				state.live_block_hash = hash.0;
+			// if activated is non-empty, set state.live_block_hash to the highest block in `activated`
+			for activated in activated {
+				if activated.number > state.live_block_hash.0 {
+					state.live_block_hash = (activated.number, activated.hash)
```
eskimor, line 527: Keeping the name `live_block_hash` is slightly confusing. Maybe rename it to just `live_block`?

node/network/bridge/src/lib.rs
```diff
@@ -1626,4 +1654,52 @@ mod tests {
...
+							hash: h,
+							number: i as _,
+							span: Arc::new(jaeger::Span::Disabled),
+						}).rev().collect(),
+						deactivated: Default::default(),
+					}
+				))
+			).await;
+
+			let view_heads = (MAX_VIEW_HEADS * 3 - MAX_VIEW_HEADS .. MAX_VIEW_HEADS * 3).rev()
```
eskimor, line 1686: Isn't MAX_VIEW_HEADS * 3 = MAX_VIEW_HEADS = MAX_VIEW_HEADS *2 ?

eskimor (approved):

PR #2723, opened 2021-03-26
Fix Crowdloan Withdraw Requirements

shawntabrizi:
Two fixes here in the condition checking of a crowdloan end:

1. Crowdloans could be griefed by sending some small amount of funds, making its free balance different than its raised balance. This solves that by allowing a crowdfund account to have more funds than necessary to refund all users. In the end, any extra funds are sent to the OrphanedFunds handler.

2. If a crowdloan has a starting lease period which has already started, then that crowdloan cannot win any auctions, so we should consider it completed. Before, we were checking the last lease period.

### DIFF
--- a/runtime/common/src/crowdloan.rs
+++ b/runtime/common/src/crowdloan.rs
@@ -605,7 +605,7 @@ impl<T: Config> Module<T> {
 		fund: &FundInfo<T::AccountId, BalanceOf<T>, T::BlockNumber, LeasePeriodOf<T>>
 	) -> DispatchResult {
 			// `fund.end` can represent the end of a failed crowdsale or the beginning of retirement
-			// If the current lease period is the first slot they are trying to bid for, then it is already too late
+			// If the current lease period is past the first slot they are trying to bid for, then it is already too late
 			// to win the bid.
 			let current_lease_period = T::Auctioneer::lease_period_index();
 			ensure!(now >= fund.end || current_lease_period > fund.first_slot, Error::<T>::FundNotEnded);
--- a/runtime/common/src/crowdloan.rs
+++ b/runtime/common/src/crowdloan.rs
@@ -608,7 +608,7 @@ impl<T: Config> Module<T> {
 			// If the current lease period is the first slot they are trying to bid for, then it is already too late
 			// to win the bid.
 			let current_lease_period = T::Auctioneer::lease_period_index();
-			ensure!(now >= fund.end || current_lease_period >= fund.first_slot, Error::<T>::FundNotEnded);
+			ensure!(now >= fund.end || current_lease_period > fund.first_slot, Error::<T>::FundNotEnded);
 			// free balance must greater than or equal amount raised, otherwise funds are being used
 			// and a bid or lease must be active.
 			ensure!(CurrencyOf::<T>::free_balance(&fund_account) >= fund.raised, Error::<T>::BidOrLeaseActive);
--- a/runtime/common/src/crowdloan.rs
+++ b/runtime/common/src/crowdloan.rs
@@ -596,11 +596,9 @@ impl<T: Config> Module<T> {
 	}
 
 	/// This function checks all conditions which would qualify a crowdloan has ended.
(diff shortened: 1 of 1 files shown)


runtime/common/src/crowdloan.rs
```diff
@@ -601,6 +594,29 @@ impl<T: Config> Module<T> {
 	pub fn crowdloan_kill(index: TrieIndex) -> child::KillChildStorageResult {
 		child::kill_storage(&Self::id_from_index(index), Some(T::RemoveKeysLimit::get()))
 	}
+
+	/// This function checks all conditions which would qualify a crowdloan has ended.
+	/// * If we have reached the `fund.end` block
+	///    * or
+	/// * If the first lease period the fund is trying to bid for has started already
+	///    * and
+	/// * If the fund has enough free funds to refund full raised amount.
```
coriolinus, line 603: The docs are ambiguous: is the condition like `(fund.end || lease_period_started) && enough_free_funds` or like `fund.end || (lease_period_started && enough_free_funds)`?

shawntabrizi: The former. I will try to improve

runtime/common/src/crowdloan.rs
```diff
@@ -601,6 +594,27 @@ impl<T: Config> Module<T> {
 	pub fn crowdloan_kill(index: TrieIndex) -> child::KillChildStorageResult {
 		child::kill_storage(&Self::id_from_index(index), Some(T::RemoveKeysLimit::get()))
 	}
+
+	/// This function checks all conditions which would qualify a crowdloan has ended.
+	/// * If we have reached the `fund.end` block OR the first lease period the fund is
+	///   trying to bid for has started already.
+	/// * And, if the fund has enough free funds to refund full raised amount.
+	fn ensure_crowdloan_ended(
```
shawntabrizi, line 602: This refactoring is in anticipation of me writing another function using this same logic, and not wanting to repeat it twice.

shawntabrizi: The function would allow for automatic refunds, so that users do not need to make lots of separate calls.

gavofyork (approved):

runtime/common/src/crowdloan.rs
```diff
@@ -601,6 +594,27 @@ impl<T: Config> Module<T> {
...
+		now: T::BlockNumber,
+		fund_account: &T::AccountId,
+		fund: &FundInfo<T::AccountId, BalanceOf<T>, T::BlockNumber, LeasePeriodOf<T>>
+	) -> DispatchResult {
+			// `fund.end` can represent the end of a failed crowdsale or the beginning of retirement
+			// If the current lease period is the first slot they are trying to bid for, then it is already too late
+			// to win the bid.
+			let current_lease_period = T::Auctioneer::lease_period_index();
+			ensure!(now >= fund.end || current_lease_period >= fund.first_slot, Error::<T>::FundNotEnded);
```
shawntabrizi, line 611:
```suggestion
			ensure!(now >= fund.end || current_lease_period > fund.first_slot, Error::<T>::FundNotEnded);
```

runtime/common/src/crowdloan.rs
```diff
@@ -601,6 +594,27 @@ impl<T: Config> Module<T> {
...
+	/// * If we have reached the `fund.end` block OR the first lease period the fund is
+	///   trying to bid for has started already.
+	/// * And, if the fund has enough free funds to refund full raised amount.
+	fn ensure_crowdloan_ended(
+		now: T::BlockNumber,
+		fund_account: &T::AccountId,
+		fund: &FundInfo<T::AccountId, BalanceOf<T>, T::BlockNumber, LeasePeriodOf<T>>
+	) -> DispatchResult {
+			// `fund.end` can represent the end of a failed crowdsale or the beginning of retirement
+			// If the current lease period is the first slot they are trying to bid for, then it is already too late
```
shawntabrizi, line 608:
```suggestion
			// If the current lease period is past the first slot they are trying to bid for, then it is already too late
```

PR #2724, opened 2021-03-26
Add some tests to demonstrate an estimate of the nominator limit

kianenigma:
```
can support 80665 nominators to yield a weight of 1474882615000
```


runtime/kusama/src/lib.rs
```diff
@@ -1481,4 +1481,29 @@ mod test_fees {
 		test_with_multiplier(Multiplier::saturating_from_rational(1, 1_000_000u128));
 		test_with_multiplier(Multiplier::saturating_from_rational(1, 1_000_000_000u128));
 	}
+
+	#[test]
+	fn nominator_limit() {
+		// starting point of the nominators.
```
kianenigma, line 1487:
```suggestion
		use pallet_election_provider_multi_phase::WeightInfo;
		// starting point of the nominators.
```

bkchr (approved):

shawntabrizi (approved):
This seems fine, but i feel that a test written like this just gets forgotten about.

We would more ideally put a target we want in `all_voters`, for example 50_000 voters.

Then do the loop which searches for the max, and then assert `active` is greater than our target.

If it is not, the test will fail, and you will write a bunch of nice comments in this test that we should talk about whats what.

kianenigma:
Agree, I thought of such scenarios as well, but I would be spinning the 50_000 out of thin air, and I am worried that some people will not like it once it fails. But I prefer the fact that a test without an assertion is, as you said, bound to be forgotten. Will add an estimate. 

FYI we do the same for payout costs as well and assert that a full payout of maximum nominators' weight never takes more than the maximum block weight.

runtime/polkadot/src/lib.rs
```diff
@@ -1510,4 +1510,29 @@ mod test_fees {
...
+			)
+		};
+
+		let mut active = 1;
+		while weight_with(active) <= OffchainSolutionWeightLimit::get() || active == all_voters {
+			active += 1;
+		}
+
+		println!("can support {} nominators to yield a weight of {}", active, weight_with(active));
+	}
```
shawntabrizi, line 1537:
```suggestion
	fn nominator_limit() {
		// starting point of the nominators.
		let target_voters: u32 = 50_000;

		// assuming we want around 5k candidates and 1k active validators. (March 31, 2021)
		let all_targets: u32 = 5_000;
		let desired: u32 = 1_000;
		let weight_with = |active| {
			<Runtime as pallet_election_provider_multi_phase::Config>::WeightInfo::submit_unsigned(
				target_voters.max(active),
				all_targets,
				active,
				desired,
			)
		};

		let mut active = target_voters;
		while weight_with(active) <= OffchainSolutionWeightLimit::get() || active == target_voters {
			active += 1;
		}
		
		println!("can support {} nominators to yield a weight of {}", active, weight_with(active));
		assert!(active > target_voters, "we need to reevaluate the weight of the election system");
	}
```

shawntabrizi:
> I would be spinning the 50_000 out of thin air

This shouldn't be true, it should be based on what we see on the network. Rough calculation stuff.

Better it fails in a CI than it fails on the network

runtime/polkadot/src/lib.rs
```diff
@@ -1514,25 +1514,26 @@ mod test_fees {
...
+		let target_voters: u32 = 50_000;
 
-		// assuming we want around 5k candidates and 1k active validators.
+		// assuming we want around 5k candidates and 1k active validators. (March 31, 2021)
 		let all_targets: u32 = 5_000;
 		let desired: u32 = 1_000;
 		let weight_with = |active| {
 			<Runtime as pallet_election_provider_multi_phase::Config>::WeightInfo::submit_unsigned(
-				all_voters.max(active),
+				target_voters.max(active),
```
shawntabrizi, line 1525: i guess this max isnt needed anymore with the changes

kianenigma:
bot merge

kianenigma:
bot merge

PR #2726, opened 2021-03-26
Call NetworkService::add_known_address before sending a request

tomaka:
Requires https://github.com/paritytech/substrate/pull/8467 to be merged first. As is, this PR doesn't compile because of the missing function.

See the comment I added to `OutgoingRequest`.

Basically, you can't just call `send_request(OutgoingRequest::Peer(some_peer_id))` and expect it to magically work, as `sc-network` might have to establish a TCP connection but might not be aware of any address for that peer.

This PR changes `Recipient::Authority` to pass the addresses from `sc-authority-discovery` to `sc-network`.

I've also removed the `peer_id_from_multiaddr` function, because it does the same as `sc-network::config::parse_addr`.


rphmeier (approved):

node/network/bridge/src/network.rs
```diff
@@ -235,16 +236,28 @@ impl Network for Arc<NetworkService<Block, Hash>> {
...
+				// to be very quick.
+				for addr in authority_discovery
+					.get_addresses_by_authority_id(authority).await
+					.into_iter().flat_map(|list| list.into_iter())
+				{
+					let (peer_id, addr) = match parse_addr(addr) {
+						Ok(v) => v,
+						Err(_) => continue,
+					};
+					NetworkService::add_known_address(
```
eskimor, line 251: Two questions:
1. Why do we only need this when we want to connect to authorities?
2. Don't we keep a DHT to resolve `PeerId`s to addresses? I mean if we have them cached locally, it would be stupid to not use them, but it should still work if we did not - right?

eskimor (approved):

rphmeier:
bot merge

PR #2728, opened 2021-03-26
Introduce `add_memo` for Crowdloans

shawntabrizi:
This PR introduces an additional extrinsic where a user can place some additional "memo" to a crowdloan contribution they have made.

The main reason to use this is to specify some other account that should receive the token distribution for a crowdloan you participated in.

cc @joshorndorff @xlc @jacogr

### DIFF
--- a/runtime/common/src/crowdloan.rs
+++ b/runtime/common/src/crowdloan.rs
@@ -247,6 +247,8 @@ decl_event! {
 		HandleBidResult(ParaId, DispatchResult),
 		/// The configuration to a crowdloan has been edited. [fund_index]
 		Edited(ParaId),
+		/// A memo has been updated. [who, fund_index, memo]
+		MemoUpdated(AccountId, ParaId, Vec<u8>),
 	}
 }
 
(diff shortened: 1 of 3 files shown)


JoshOrndorff:
This approach satisfies Moonbeam's needs :+1: 

Out of curiosity, what's the advantage of having a separate extrinsic as opposed to including the memo with the original contribution?

cc @girazoki

shawntabrizi:
Every contribution would require the memo. Also in terms of separation of concerns, we have written tests for and audited code as it was. Creating a new function isolates new code and how it can interact with the system.

tolak:
Nice job shawn,  I think some community also need this features. There  is a scenario that two third party community contribute to one parachain crowdloan, the community may distribute their own community token to user who participated this auction, then they can add memo to identify which user belong to their community that can get community token reward.

Besides, is it possible to include memo as an argument in extrinsic **contribute** and put it into event **RawEvent::Contributed** rather than save it into child trie? Or maybe need a query extrinsic to get the memo of a contribution.

kianenigma (approved):

tolak:
@shawntabrizi Maybe for the client who wants to catch memo just like **ParaId**, **Account** and **Amount** by event without query the child trie. I am not sure if we already have interface for client to get the memo of a contribution.

shawntabrizi:
@tolak is this a hypothetical request, or you actually need this?

rphmeier (approved):

shawntabrizi:
bot merge

PR #2731, opened 2021-03-27
Handle Onboarding in Current Lease Period

shawntabrizi:
This PR introduces two features which can help us handle parachains that should onboard in the current lease period:

* When we call `lease_out`, if the current lease period index is the same as the first slot they won, try to onboard them automatically.
* If there are any issues onboarding a parachain with an active lease during the current lease period, anyone can call `trigger_onboard` to resolve that and trigger an onboarding.

### DIFF
--- a/runtime/common/src/slots.rs
+++ b/runtime/common/src/slots.rs
@@ -447,7 +447,7 @@ mod tests {
 	use sp_core::H256;
 	use sp_runtime::traits::{BlakeTwo256, IdentityLookup};
 	use frame_support::{
-		parameter_types, assert_ok,
+		parameter_types, assert_ok, assert_noop,
 		traits::{OnInitialize, OnFinalize}
 	};
 	use pallet_balances;
@@ -771,6 +771,7 @@ mod tests {
 			run_to_block(1);
 
 			assert_ok!(TestRegistrar::<Test>::register(1, ParaId::from(1), Default::default(), Default::default()));
+			assert_ok!(TestRegistrar::<Test>::register(1, ParaId::from(2), Default::default(), Default::default()));
 
 			run_to_block(20);
 			assert_eq!(Slots::lease_period_index(), 2);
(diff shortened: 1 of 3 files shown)


runtime/common/src/slots.rs
```diff
@@ -189,6 +191,30 @@ decl_module! {
...
+		/// Origin must be signed, but can be called by anyone.
+		#[weight = T::WeightInfo::trigger_onboard()]
+		fn trigger_onboard(origin, para: ParaId) {
+			let _ = ensure_signed(origin)?;
+			let leases = Leases::<T>::get(para);
+			match leases.first() {
+				// If the first element in leases is present, then it has a lease!
+				// We can try to onboard it.
+				Some(Some(_lease_info)) => {
+					T::Registrar::make_parachain(para)?
```
gavofyork, line 210: should this have a `?` at the end?

gavofyork: and shouldn't we be checking that it's still a plain old parathread?

runtime/common/src/slots.rs
```diff
@@ -376,6 +403,14 @@ impl<T: Config> Leaser for Module<T> {
 			}
 
 			let reserved = maybe_additional.unwrap_or_default();
+
+			// Check if current lease period is same as period begin, and onboard them directly.
+			// This will allow us to support onboarding new parachains in the middle of a lease period.
+			if current_lease_period == period_begin {
+				let res = T::Registrar::make_parachain(para);
+				debug_assert!(res.is_ok());
```
gavofyork, line 411: let's always have a proof with any (debug) assertions.

gavofyork (approved):
Seems reasonable.

runtime/common/src/slots.rs
shawntabrizi: `make_parachain` itself will make sure it is a parathread and any other conditions required to allow `make_parachain` be valid. This is why we have the `?` because `make_parachain` could fail, and we will propagate up that error.

runtime/common/src/slots.rs
shawntabrizi: hmm, actually Slots does not _guarantee_ that they are a para and that `make_parachain` will succeed.

The `debug_assert` is more to ensure that all tests take into account that this should be successful in the tests, and if not, maybe the test itself is bad.

Am I using this kind of pattern wrong?

shawntabrizi:
bot merge

PR #2735, opened 2021-03-27
Max notification size -> 100k.

eskimor:
Can be merged once we have PoV distribution on request response and also runtime fetching on request response.


tomaka (approved):
This looks a bit low to me, but if you're confident that we'll never reach it, then +1

eskimor:
Low ... :-) I was thinking of reducing it further. @rphmeier can you think of any notifications that could not fit in that size, now that we requestified everything?

PR #2739, opened 2021-03-28
Check Intersecting Range on Para, not Bidder

shawntabrizi:
To optimize winning calculation, we do not allow non-intersecting bids for parachain slots.

However, the code was checking the bidder, not the parachain. A bidder should not have this requirement.


shawntabrizi:
superseded by https://github.com/paritytech/polkadot/pull/2741

PR #2745, opened 2021-03-28
Add Many Sanity Checks to Crowdloans

shawntabrizi:
When creating a crowdloan:

* Check that the end date is before the "last possible win date". i.e. if the crowdloan is for lease periods 1 - 4, ensure the end date is at most the beginning of lease period 2. Once we get to lease period 2, we will never have an auction for lease period 1.
* Check the first slot in a crowdloan is at least the current lease period index. For the same reasons, it does not make sense to create a crowdloan for a lease period in the past.

When contributing to a crowdloan:

* Check that the crowdloan is bidding for a valid lease period, i.e. the lease period has not passed already.
* Check the crowdloan account has not already won, i.e. it has some deposit held in the Leaser.

### DIFF
--- a/runtime/common/src/crowdloan.rs
+++ b/runtime/common/src/crowdloan.rs
@@ -1518,7 +1518,7 @@ mod benchmarking {
 			let cap = BalanceOf::<T>::max_value();
 			let first_period = 0u32.into();
 			let last_period = 3u32.into();
-			let end = T::Leaser::lease_period();
+			let end = T::Auctioneer::lease_period();
 
 			let caller: T::AccountId = whitelisted_caller();
 			let head_data = T::Registrar::worst_head_data();
@@ -1594,7 +1594,7 @@ mod benchmarking {
 			let cap = BalanceOf::<T>::max_value();
 			let first_period = 0u32.into();
 			let last_period = 3u32.into();
-			let end = T::Leaser::lease_period();
+			let end = T::Auctioneer::lease_period();
 
 			let caller: T::AccountId = whitelisted_caller();
 			let head_data = T::Registrar::worst_head_data();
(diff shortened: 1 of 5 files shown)


runtime/common/src/crowdloan.rs
```diff
@@ -150,6 +150,9 @@ pub trait Config: frame_system::Config {
 		LeasePeriod=Self::BlockNumber,
 	>;
 
+	/// The type representing the leasing system.
+	type Leaser: Leaser<AccountId=Self::AccountId, LeasePeriod=Self::BlockNumber>;
```
kianenigma, line 154: FYI `type Leaser` in `auctions.rs` has a unrelated doc.

kianenigma (approved):

Az111z (approved):

runtime/common/src/crowdloan.rs
shawntabrizi: I fixed the doc in this PR?

shawntabrizi:
cc @jacogr

shawntabrizi:
bot merge

shawntabrizi:
bot merge

runtime/common/src/crowdloan.rs
```diff
@@ -254,14 +245,18 @@ decl_event! {
...
-		/// The first slot needs to at least be less than 3 `max_value`.
-		FirstSlotTooFarInFuture,
-		/// Last slot must be greater than first slot.
-		LastSlotBeforeFirstSlot,
-		/// The last slot cannot be more then 3 slots after the first slot.
-		LastSlotTooFarInFuture,
+		/// The current lease period is more than the first lease period.
+		FirstPeriodInPast,
+		/// The first lease period needs to at least be less than 3 `max_value`.
+		FirstPeriodTooFarInFuture,
```
xlc, line 251: This is just overflow?

shawntabrizi: What do you mean here? That this error should be replaced with "Overflow" error?

xlc: Yes. Otherwise the description is confusion and I was looking for some max_value constant in trait Config

PR #2752, opened 2021-03-29
use unbounded when notifying distribution of new blocks

rphmeier:
This is the same way we break other cycles in approval-voting -> approval-distribution.


ordian (approved):

rphmeier:
bot merge

PR #2757, opened 2021-03-29
send assignments even when we have an approval

rphmeier:
Previously if we were unifying with a peer and we had an approval for a given validator, we'd only send the approval and not the assignment. We need to send the assignment as well, and this PR fixes that and adds a test.


ordian (approved):

rphmeier:
bot merge

PR #2764, opened 2021-03-30
Remove already triggered migrations

expenses:


runtime/kusama/src/lib.rs
```diff
@@ -1076,7 +1064,7 @@ pub type Executive = frame_executive::Executive<
 	frame_system::ChainContext<Runtime>,
 	Runtime,
 	AllPallets,
-	(BabeEpochConfigMigrations, KillOffchainPhragmenStorageTest),
+	(KillOffchainPhragmenStorageTest),
```
kianenigma, line 1067: Can you remove this

runtime/polkadot/src/lib.rs
```diff
@@ -1068,7 +1059,7 @@ pub type Executive = frame_executive::Executive<
 	frame_system::ChainContext<Runtime>,
 	Runtime,
 	AllPallets,
-	(BabeEpochConfigMigrations, FixPolkadotCouncilVotersDeposit),
+	(FixPolkadotCouncilVotersDeposit),
```
kianenigma, line 1062: and this as well 😂  all of these are already in v30 branch.

runtime/kusama/src/lib.rs
expenses: Remove `KillOffchainPhragmenStorageTest`?

kianenigma:
from v30 

https://github.com/paritytech/polkadot/blob/v0.8.30-rc1/runtime/polkadot/src/lib.rs#L1068
https://github.com/paritytech/polkadot/blob/v0.8.30-rc1/runtime/kusama/src/lib.rs#L1076

@s3krit if we need rc2, we backport some changes right? these will stay in the release anyway?

runtime/kusama/src/lib.rs
kianenigma: Yes

kianenigma (approved):
Should remove the definition of `FixPolkadotCouncilVotersDeposit` and `KillOffchainPhragmenStorageTest`  as well, as they don't serve any purpose any more.

expenses:
Probably best to wait for another review so I don't merge this and break things.

s3krit:
> @s3krit if we need rc2, we backport some changes right? these will stay in the release anyway?

We do have an rc2 now ;) But yes, only relevant changes get backported so this is fine.

expenses:
What should I do with this? Merge when ready?

kianenigma:
Yes.

expenses:
Okedoke

expenses:
bot merge

expenses:
bot merge

expenses:
Okay whatever

PR #2765, opened 2021-03-30
companion for substrate/pull/8044

kianenigma:
Companion to https://github.com/paritytech/substrate/pull/8044

- [x] Not yet tested for migrations.


Az111z (approved):

Az111z (approved):

thiolliere:
note: renaming the pallet makes the stored version outdated, the version is stored using the name of the pallet, thus we probably want to clean the version stored at the old name, and we have to be careful the renamed pallet is considered at its latest version (because no version is stored at its new name).

thiolliere (approved):
looks good to me

We need to clean the old version stored at `twox128(b"ElectionsPhragmen")++twox128(PALLET_VERSION_STORAGE_KEY_POSTFIX)`.

And be sure we that we don't use the version in the next PRs until next upgrade.

PR #2777, opened 2021-03-30
Add Root Functions to Parachains System

shawntabrizi:
These Root functions may provide useful for keeping the Parachains system up and running.

### DIFF
--- a/runtime/parachains/src/paras.rs
+++ b/runtime/parachains/src/paras.rs
@@ -345,6 +345,8 @@ decl_event! {
 		CodeUpgradeScheduled(ParaId),
 		/// A new head has been noted for a Para. \[para_id\]
 		NewHeadNoted(ParaId),
+		/// A para has been queued to execute pending actions. \[para_id\]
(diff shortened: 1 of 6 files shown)


runtime/parachains/src/paras.rs
```diff
@@ -332,10 +335,58 @@ decl_error! {
...
 		type Error = Error<T>;
+
+		fn deposit_event() = default;
+
+		#[weight = 0]
+		fn force_set_current_code(origin, para: ParaId, new_code: Vec<u8>) {
+			ensure_root(origin)?;
+			let validation_code = ValidationCode(new_code);
+			<Self as Store>::CurrentCode::insert(&para, validation_code);
+			Self::deposit_event(Event::CurrentCodeUpdated(para));
```
shawntabrizi, line 363: @rphmeier is it important that i put the existing code into the `PastCode` storage?

What if there is data already there, I should overwrite it?

rphmeier:
bot merge

PR #2782, opened 2021-03-31
Cost analysis of a council election full block

kianenigma:
Current value is `166_000 DOLLARS`. With the current value of DOT, this is more than a million. But I expect the DOT to DOLLARS ratio to be tuned soon, therefore let's assume DOLLARS is actually 1USD. 

In this case, I think it is a bit too low. Current `VotingBaseBond` is `20 * DOLLARS + ...` only for the storage item. I would say we increase this to 50 or 100 DOLLARS to reach a safety net of half a million DOLLARS at least.


shawntabrizi:
Just as a note, I think we want to deprecate the use of "DOLLAR" in our code base since it implies things which you mention clearly do not apply above.

It might make sense to come up with another unit of measurement. Maybe # of Big Macs?

https://www.economist.com/big-mac-index

bkchr:
An attack on Polkadot costs 500 Big macs :D

shawntabrizi (approved):

shawntabrizi:
@kianenigma some unused warning

Az111z (approved):

kianenigma:
> Just as a note, I think we want to deprecate the use of "DOLLAR" in our code base since it implies things which you mention clearly do not apply above.
> 
> It might make sense to come up with another unit of measurement. Maybe # of Big Macs?
> 
> https://www.economist.com/big-mac-index

I haven't had time to look into this yet, but the current measure is also not too bad. We've never changed these constants in the past, so I am a bit worried about the consequences there. But, assuming that they will have no consequence, and we can easily update them, I think this model is easy and simple enough. 

For example, the governance bodies can re-evaluate this once per quarter based on the mean value of DOT, and just change this one number. 

I'm curious to see what happens with #2451.

runtime/polkadot/src/lib.rs
```diff
@@ -1512,6 +1512,38 @@ mod test_fees {
...
+		let cost = voters as Balance * (VotingBondBase::get() + 16 * VotingBondFactor::get());
+		let cost_dollars = cost / DOLLARS;
+		println!(
+			"can support {} voters in a single block for council elections; total bond {}",
+			voters,
+			cost_dollars,
+		);
+		assert!(cost_dollars > 150_000); // DOLLAR ~ new DOT ~ 10e10
+	}
+
```
kianenigma, line 1546:
```suggestion

#[test]
```

runtime/polkadot/src/lib.rs
```diff
@@ -1512,6 +1512,39 @@ mod test_fees {
...
+#[test]
```
kianenigma, line 1547:
```suggestion
	#[test]
```

kianenigma:
bot merge

bkchr:
@kianenigma you need to merge master

kianenigma:
bot merge

kianenigma:
> @kianenigma you need to merge master

@bkchr you need to approve

bkchr (approved):

bkchr:
> > @kianenigma you need to merge master
> 
> @bkchr you need to approve 

Done 😜

PR #2785, opened 2021-03-31
Add Root Functions to HRMP

shawntabrizi:
To help test HRMP more quickly, these utility functions for Root can be used to speed up the processing or cleanup of HRMP channels.

### DIFF
--- a/runtime/parachains/src/hrmp.rs
+++ b/runtime/parachains/src/hrmp.rs
@@ -409,6 +409,14 @@ decl_module! {
 			Self::deposit_event(Event::ChannelClosed(origin, channel_id));
 			Ok(())
 		}
+
+		/// Force process hrmp open and close requests.
+
+		/// This extrinsic triggers the cleanup of all the HRMP storage items that
+		/// a para may have. Normally this happens once per session, but this allows
+		/// you to trigger the cleanup immediately for a specific parachain.
+		#[weight = 0]
+		pub fn force_clean_hrmp()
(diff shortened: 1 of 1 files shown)


pepyakin (approved):
This module relies on the fact that everything is changed on the session change boundary. If it's violated I am pretty sure there will be glitches. I would prefer to induce a forced session change if possible.

I also would suggest to clearly indicate in the docs that these functions are meant to be used for testing.

shawntabrizi:
@pepyakin will there actually be glitches in calling any of these functions?

The goal should be not to have any such problems.

Forcing a session change is a much bigger thing for the chain. We want to only be able to speed up the HRMP registration process.

pepyakin:
One glitch I can come up with right off the bat is invalidation of a parablock. 

E.g. a trivial case a parablock will be rejected if the para tries to send a message to a channel that becomes closed when parablock gets backed. This will be only one parablock just because we check this info before preparing each parablock. In future I think we should cache this until the next session change.

A worse category of scenarios, is when you do such changes between the candidate is backed and is included. That would at the very least leave junk on chain. But I also am afraid that unforeseen stuff may also lurk.

Even if the session change may be considered to have a bigger effect, it will be more closer to the production, since that would be similar to time travel, whereas doing only a partial change may introduce some unwanted side effects.

That said, I think opening a channel is safer than removing a para or closing channels.

gavofyork (approved):

Issue #2487, opened 2021-02-20
Availability Recovery: handle self-requests better


rphmeier:
Availability recovery isn't accounting for the fact that our own validator index can be selected and tries to request over the network which fails. Three ways to account for this:
  1. Be aware of which validator index is our own and simply don't select it
  2. Attempt to answer all requests issued by the `Interaction` via the availability store first.
  3. pre-load all chunks we have (usually 1) from the availability store and don't include the indices for any of those in the list of validators to request from.

PR #2792, opened 2021-04-01
Avoid querying the local validator in availability recovery

rphmeier:
Closes https://github.com/paritytech/polkadot/issues/2487

### DIFF
--- a/node/network/availability-recovery/src/lib.rs
+++ b/node/network/availability-recovery/src/lib.rs
@@ -334,6 +334,34 @@ impl RequestChunksPhase {
 		params: &InteractionParams,
 		sender: &mut impl SubsystemSender,
 	) -> Result<AvailableData, RecoveryError> {
+		// First query the store for any chunks we've got.
+		{
+			let (tx, rx) = oneshot::channel();
+			sender.send_message(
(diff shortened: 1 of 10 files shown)


node/network/availability-recovery/src/lib.rs
```diff
@@ -334,6 +334,34 @@ impl RequestChunksPhase {
...
+		{
+			let (tx, rx) = oneshot::channel();
+			sender.send_message(
+				AvailabilityStoreMessage::QueryAllChunks(params.candidate_hash, tx).into()
+			).await;
+
+			match rx.await {
+				Ok(chunks) => {
+					// This should either be length 1 or 0. If we had the whole data,
+					// we wouldn't have reached this stage.
```
drahnr, line 347:
```suggestion
					// we wouldn't have reached this stage.
					debug_assert!(chunks.len() <= 1, "Not all of the data is available.");
```

ordian (approved):

drahnr (approved):

node/network/availability-recovery/src/lib.rs
rphmeier: this defies the 'panic: prove or remove' rule. we handle this case just fine

PR #2793, opened 2021-04-01
Companion for Decouple Staking and Election - Part 3: Signed Phase

coriolinus:
https://github.com/paritytech/substrate/pull/7910


kianenigma:
needs update.

runtime/kusama/src/lib.rs
```diff
@@ -354,6 +354,12 @@ parameter_types! {
 	pub const SignedPhase: u32 = 0;
 	pub const UnsignedPhase: u32 = EPOCH_DURATION_IN_SLOTS / 4;
 
+	// signed config
+	pub const SignedMaxSubmissions: u32 = 0;
```
kianenigma, line 358: ?

runtime/polkadot/src/lib.rs
```diff
@@ -334,6 +334,12 @@ parameter_types! {
 	pub const SignedPhase: u32 = 0;
 	pub const UnsignedPhase: u32 = EPOCH_DURATION_IN_SLOTS / 4;
 
+	// signed config
+	pub const SignedMaxSubmissions: u32 = 0;
```
kianenigma, line 338: ?

runtime/polkadot/src/lib.rs
```diff
@@ -334,6 +334,12 @@ parameter_types! {
...
+	pub const SignedRewardBase: Balance = 1 * DOLLARS;
```
kianenigma, line 339: These number should be consulted.

kianenigma (changes requested):
Needs another more diligent update.

runtime/kusama/src/lib.rs
```diff
@@ -377,6 +383,14 @@ impl pallet_election_provider_multi_phase::Config for Runtime {
 	type Event = Event;
 	type Currency = Balances;
 	type UnsignedPhase = UnsignedPhase;
+	type SignedMaxSubmissions = SignedMaxSubmissions;
+	type SignedRewardBase = SignedRewardBase;
+	type SignedDepositBase = SignedDepositBase;
+	type SignedDepositByte = SignedDepositByte;
+	type SignedDepositWeight = ();
+	type SignedMaxWeight = Self::MinerMaxWeight;
+	type SlashHandler = (); // burn slashes
```
kianenigma, line 392: Same, could use more thought. My initial values here are all placeholders.

coriolinus:
@kianenigma Aside from the `SlashHandler`, all those values have now been changed.

kianenigma:
> @kianenigma Aside from the `SlashHandler`, all those values have now been changed.

feedback given in other mediums; needs more tweaks.

coriolinus:
@kianenigma I worked out the function we walked about for computing an appropriate value for the signed rewards, which would pay back 1.5x the submission fee. It's implemented [here](https://github.com/paritytech/polkadot/blob/b19f7e1dcaae9a042615c2698d569b9809571704/runtime/common/src/elections.rs#L47-L61). However, it doesn't work, because it needs to be a `const fn` to be legal within the `parameter_types` macro, but it can't be a `const fn` because it depends on trait functions. I therefore reverted it.

kianenigma:
> @kianenigma I worked out the function we walked about for computing an appropriate value for the signed rewards, which would pay back 1.5x the submission fee. It's implemented [here](https://github.com/paritytech/polkadot/blob/b19f7e1dcaae9a042615c2698d569b9809571704/runtime/common/src/elections.rs#L47-L61). However, it doesn't work, because it needs to be a `const fn` to be legal within the `parameter_types` macro, but it can't be a `const fn` because it depends on trait functions. I therefore reverted it.

remove `const` from `parameter_types` and it will accept non const values as well.

coriolinus:
:facepalm: Welp, I should have thought of that. For some reason I thought that `parameter_types` required `const` values, despite counterexamples just a few lines away.

Ok, as of ff228bc, `SignedRewardBase` is 1.5x the submission fee for the extrinsic. This is an interesting value, because it means that for more than 1 miner, the expected value of any particular submission goes negative. (The miners will get their deposit back, but they'll still have to pay the fee to submit the solution.) It'll generate a small profit for us as long as we're the only people running a signed solution miner, and anyone else considering running a miner is likely to notice that on average, a miner in a pool of 2 will receive a reward of 0.75x the submission fee. 

We can work out the economics of decentralization later. For now, I hope that this is a sufficient starting point for this PR.

runtime/common/src/elections.rs
```diff
@@ -0,0 +1,61 @@
...
+///
+/// Assumes that the signed submission queue is full.
+pub fn fee_for_submit_call<T, WeightToFee, WeightInfo>(multiplier: Perbill) -> WeightToFee::Balance
+where
+	T: pallet_election_provider_multi_phase::Config,
+	WeightToFee: WeightToFeePolynomial,
+	WeightInfo: pallet_election_provider_multi_phase::WeightInfo,
+{
+	let expected_weight = WeightInfo::submit(T::SignedMaxSubmissions::get());
+	multiplier * WeightToFee::calc(&expected_weight)
```
kianenigma, line 60: this might lead to negative turnout to run the bot, since your discounting the length fee (which in this case is probably more dominant than the weight fee)(I also forgot about legth fee earlier).

kianenigma: > maybe what you should do is run your polkadot branch and submit a solution to it via the miner, to both obtain this and do an end-to-end test

might still help, but not a blocker since this is too conservattive and thus safe.

kianenigma (approved):

runtime/polkadot/src/lib.rs
```diff
@@ -328,11 +329,23 @@ impl pallet_session::historical::Config for Runtime {
...
+	pub const SignedPhase: u32 = EPOCH_DURATION_IN_SLOTS / 4;
 	pub const UnsignedPhase: u32 = EPOCH_DURATION_IN_SLOTS / 4;
 
-	// fallback: run election on-chain.
+	// signed config
+	pub const SignedMaxSubmissions: u32 = 16;
+	pub const SignedDepositBase: Balance = deposit(1, 0);
+	// A typical solution currently occupies 120kb. This formula is currently adjusted such that a
+	// typical solution will spend approximately equal amounts on the base and per-byte deposits.
+	pub const SignedDepositByte: Balance = deposit(1, 0) / (120 * 1024 * 1024);
```
kianenigma, line 341: isn't this one too many 1024s?

kianenigma: I like the idea, basically *`deposit(1, 0)` per 120kb*

kianenigma: and since the 120kb is changing all the time, they could be a bit more fixed, like `deposit(1, 0) per 50kb` for both chains. Note that `fn deposit` of kusama is already adjusted for the lower entry barrier.

coriolinus: 1. Good spot! Yeah, I was thinking in mb for some reason.
2. Adjusted all the chains' rates to a base deposit per 50kb as suggested: https://github.com/paritytech/polkadot/pull/2793/commits/f5b6dd5d66d95fab425370702c3e296565641fda.

PR #2800, opened 2021-04-01
remove approval checking training wheels

rphmeier:
This replaces the approval voting diagnostic finality rule with a voting rule which actually votes based on the result of the approval voting subsystem.


ordian (approved):
:tada: :bike:

PR #2807, opened 2021-04-02
feat: Add poke to Crowdloan

### LINKED ISSUE #2801: Add `poke` to Crowdloan
A crowdloan won't bid in an auction unless it is part of the `NewRaise` Vec by the ending period of the auction.

For it to be in that `NewRaise` vec, someone needs to contribute to that fund. But it is possible that no one else wants to contribute or that the cap for the fund was reached.

In this case, we should have a simple function `fn poke(fund_id)`, which will check if the fund has non-zero raised, and then put it in the `NewRaise` vec if it is not already there.

This way, a fund can make sure that it will get its bid in, even if a contribution does not occur.

- [ ] Write `poke` function
- [ ] Tests
- [ ] Benchmarks

stanly-johnson:
Resolves #2801 

cc @shawntabrizi


runtime/common/src/crowdloan.rs
```diff
@@ -589,6 +595,19 @@ decl_module! {
 				T::DbWeight::get().reads(1)
 			}
 		}
+
+		/// Poke the fund into NewRaise
+		///
+		/// Origin must be Signed, and the fund has non-zero raise.
+		#[weight = T::WeightInfo::poke()]
+		pub fn poke(origin, index: ParaId) {
+			let who = ensure_signed(origin)?;
```
shawntabrizi, line 604:
```suggestion
			let _ = ensure_signed(origin)?;
```

runtime/common/src/crowdloan.rs
```diff
@@ -589,6 +595,19 @@ decl_module! {
...
+			let fund = Self::funds(index).ok_or(Error::<T>::InvalidParaId)?;
+			ensure!(!(fund.raised.is_zero()), Error::<T>::NoContributions);
```
shawntabrizi, line 606:
```suggestion
			ensure!(!fund.raised.is_zero(), Error::<T>::NoContributions);
```

runtime/common/src/crowdloan.rs
```diff
@@ -589,6 +595,19 @@ decl_module! {
...
+			ensure!(!(NewRaise::get().contains(&index)), Error::<T>::AlreadyInNewRaise);
```
shawntabrizi, line 607:
```suggestion
			ensure!(!NewRaise::get().contains(&index), Error::<T>::AlreadyInNewRaise);
```

runtime/common/src/crowdloan.rs
```diff
@@ -589,6 +595,19 @@ decl_module! {
...
+			NewRaise::append(index);
+			Self::deposit_event(RawEvent::AddedToNewRaise(who, index));
```
shawntabrizi, line 609: No need to include `who`

runtime/common/src/crowdloan.rs
```diff
@@ -1631,6 +1673,27 @@ mod benchmarking {
...
+			CurrencyOf::<T>::make_free_balance_be(&caller, BalanceOf::<T>::max_value());
+			Crowdloan::<T>::create(
+				RawOrigin::Signed(caller.clone()).into(),
+				para_id, cap, first_period, last_period, end, Some(verifier.clone()),
+			)?;
+			contribute_fund::<T>(&caller, para_id);
+			frame_system::Pallet::<T>::set_block_number(T::BlockNumber::max_value());
+		}: _(RawOrigin::Signed(caller.clone()), para_id)
+		verify {
+			assert_last_event::<T>(RawEvent::AddedToNewRaise(caller,para_id).into())
```
shawntabrizi, line 1694: did you check that this actually works?

```
cargo test -p polkadot-runtime-common --features runtime-benchmarks
```

I don't think you have done the right things to empty new raise here.

I think you might want to add a `NewRaise::remove()` here just to simulate if the `NewRaise` is empty.

shawntabrizi:
@stanly-johnson looks really close! Just a few nit picks

shawntabrizi (approved):

PR #2808, opened 2021-04-02
Relax Origin Checks in Registrar, Add Lock to Registration

shawntabrizi:
This PR relaxes the origin checks required in the Registrar pallet.

In places where the Para was required to make a call (deregister / swap), now the Para Manager can also make a call on behalf of that para as long as the para is not locked.

The para becomes locked as it upgrades to a parachain, and will remain locked for the rest of its lifetime until it is offboarded fully.

cc @jacogr 

Also fixes an issue where a para could not be deregistered if it had no lifecycle. Such a state should not normally be possible, but can happen when sudo calls are involved.


runtime/common/src/paras_registrar.rs
```diff
@@ -324,6 +321,26 @@ impl<T: Config> Registrar for Module<T> {
 }
 
 impl<T: Config> Module<T> {
+	/// Ensure the origin is one of Root, the `para` owner, or the `para` itself.
+	fn ensure_root_para_or_owner(origin: <T as frame_system::Config>::Origin, id: ParaId) -> DispatchResult {
+		// Check if root...
+		ensure_root(origin.clone())
+			.or_else(|_| -> DispatchResult {
+				// Else check if para chain origin...
```
shawntabrizi, line 329:
```suggestion
				// Else check if para origin...
```

Az111z (approved):

gavofyork (approved):

xlc (commented):
where is `remove_lock` getting called?

PR #2815, opened 2021-04-04
XCM v1

gavofyork:
A number of changes and renames to the XCM framework, with the centrepiece being a safer, more ergonomic `MultiAsset` where matching is a lot less verbose/tedious and the underlying code is simpler. Additionally, `Assets` has been largely rewritten, optimising a number of operations and removing a lot of superfluous code.

This fixes a number of issues with the existing `MultiAsset`:
- No type safety over where masks should be allowed
- Unspecified/incorrect behaviour where multiple wildcards are used in asset arithmetic
- Overly complicated matching and construction code needed

It replaces the large, unwieldy and overly general `MultiAsset` with:
- `MultiAsset`: A single asset type, either a single non-fungible instance or a non-zero amount of a fungible.
- `MultiAssets`: A vec of `MultiAsset`, sorted, with no duplicate non-fungibles or repetitions of the same fungible.
- `WildMultiAsset`: A wildcard which describes classes of assets; either `All` assets or all of a particular identified class of fungibles or non-fungibles.
- `MultiAssetFilter`: A means of filtering a collection of assets. Takes the form of either a `WildMultiAsset` value or a `MultiAssets` value. Unlike with the old `Vec<MultiAsset>` paradigm, the two cannot be combined - it's *either* one wildcard *or* a list of definite assets.

A number of conversions are provided to allow ease of use:
- `u128::into()` and `AssetInstance::into()` will convert into a `Fungibility`, as will anything which itself will convert into an `AssetInstance`.
- `Junction::into()`, `MultiLocation::into()` and `Vec::<u8>::into()` will convert into a `AssetId`.
- `()`, `[u8; 4]`, `[u8; 8]`, `[u8; 16]`, `[u8; 32]` and `Vec<u8>` will convert into an `AssetInstance`. A `u128` will *NOT* convert and must be stated explicitly with `Index(u128)`, since otherwise this collides with the `u128::into()` of `Fungibility`.
- `(AssetId, Fungibility)::into()` will convert into a `MultiAsset` or `MultiAssets`; `AssetId` and `Fungibility` can both be optionally substituted with a type that converts into them, allowing for e.g. `(Junction, u128)::into()`.
- `Vec<MultiAsset>::into()` will convert into `MultiAssets` and `MultiAssetFilter`
- `MultiAsset::into()` and `MultiAssets::into()` will convert into `MultiAssetFilter`
- `(AssetId, WildFungibility)::into()` will convert into a `WildMultiAsset` and `MultiAssetFilter`; `AssetId` can be optionally substituted with a type that converts into it.

## Compatibility

The `v0` API remains in place and unchanged with one minor exception (`AssetInstance::Index` moving from a struct item to a tuple item). XCM messages may be converted between `v0::Xcm` API and `v1::Xcm` API with `.try_into()`. `VersionedXcm` supports both.

Automatic backwards compatibility is provided for incoming messages, so runtimes using this code will be able to interpret and execute `v0::Xcm` messages sent from older runtimes.

Messages sent as part of executing an XCM operation (e.g. `DepositReserveAsset` sending the `ReserveAssetDeposited` message) will be authored under the `v1::Xcm` API. A new trait `WrapVersion` is provided to determine which version they will eventually be encoded with prior to sending based on the destination. Chains concerned about destination chains being unable to interpret `v1` messages  should implement this trait to ensure that such destintions have messages first translated to `v0::Xcm` for encoding. A `ForceV0` implementation is given which does a blanket `v0` translation for all destinations, however it is not recommended for general use since `v0` does not support all `v1` features and messsages using them will fail to be encoded and sent.

### Compatibility `v0` -> `v1` (receiving messages from a legacy chain)

- `BuyExecution::fees` is required to be a single, definite (non-wild) asset. `fees: vec![MultiAsset::All]`, a typical pattern in the past, has no equivalent in `v1`. This is due to an internal change in `xcm_executer` and cannot easily be worked around. If the `v0` chain has `vec![MultiAsset::All]` hardcoded for the `fees` parameter, it will need to be upgraded to specify fees explicitly and with only a single, definite asset. #3578 fixes this usage in `pallet_xcm` and all chains should be upgrades to use it ASAP.
- `Vec<MultiAsset>` may contain at most one wildcard, and if it does, it must be the only element.
- `MultiAsset::AllFungible` and `MultiAsset::AllNonFungible` have no equivalent in `v1` and messages using them will result in failure.

### Compatibility `v1` -> `v0` (sending messages to a legacy chain)

- `BuyExecution::orders` has no equivalent in `v0` and messages where it is not empty will fail translation.

## Migrating from `v0::Xcm` API

- Replace `xcm::v0` for `xcm::v1`.
- For ease of use you may want to use the prelude declaration: `use xcm::v1::prelude::*;`.
- `xcm::latest` may be used in order to track the latest XCM version. This is not generally recommended outside of the Polkadot codebase.

### `MultiAsset` & `MultiLocation`

- `MultiLocation::Null` becomes `MultiLocation::Here`
- `Vec<MultiAsset>` as a pattern is deprecated in favour of `MultiAssets` (which does not support wildcards) or `MultiAssetFilter` (which does).
- `MultiAsset::ConcreteFungible { id: MultiLocation::Null, amount: 100 }` becomes `MultiAsset { id: Concrete(Here), fun: Fungible(100) }` or just `(Here, 100).into()`.
- `MultiAsset::ConcreteNonFungible { id: MultiLocation::X1(Junction::Parent), instance: AssetInstance::Blob(vec![0]) }` becomes `MultiAsset { id: Concrete(X1(Parent)), fun: NonFungible(Blob(vec![0])) }` or just `(Parent, vec![0]).into()`.
- `vec![MultiAsset::All]` becomes `Wild(All)` or `All.into()` (both of type `MultiAssetFilter`).
- `vec![MultiAsset::None]` becomes `Definite(vec![])` (of type `MultiAssetFilter`) or just `vec![].into()` (of type `MultiAssets` or `MultiAssetFilter`).
- `vec![MultiAsset::AllConcreteFungible { id: Null }]` becomes `Wild(AllOf(fun: WildFungible, id: Here))` or just `(Here, WildFungible).into()`

- `MultiAsset::AllFungibles` and `MultiAsset::AllNonFungibles` have been removed and can no longer be used.
- A `MultiAsset` must strictly identify an asset: there is no `MultiAsset` value equivalent to the old `MultiAsset::None`. *Fungible assets may not have a zero amount*. To allow for an asset value which could include "no asset", use `MultiAssets` (if more than one asset type is allowable) or `Option<MultiAsset>` (otherwise).

### XCM

Within XCM, the once ubiquitous `Vec<MultiAsset>` has now been changed to be one of:
- `MultiAssets` in the `Xcm` items, since these describe definite (lists of) assets rather than filters and thus cannot be wild.
- `MultiAsset` in the `Order::BuyExecution` item, since it describes fees and we choose not to support fee payment in multiple asset types.
- `MultiAssetFilter` in all other `Order` items, since these select/filter assets from the Holding Register.

`DepositAsset` and `DepositReserveAsset` now both have an additional `max_assets` field, which states the maximum number of unique assets/asset instances which may be deposited from the Holding Register. This is trivial to provide in the case of a definite set of assets, but requires some knowledge of the expected value of the contents of Holding if wildcards are used.

There are some additional renames to `Xcm`:
- `ReserveAssetDeposit` becomes `ReserveAssetDeposited`
- `TeleportAsset` becomes `ReceiveTeleportedAsset`
- `TransferAsset::dest` becomes `TransferAsset::beneficiary`

And to `Order`:
- `Null` becomes `Noop`
- `DepositAsset::dest` becomes `DepositAsset::beneficiary`
- `TransferAsset::dest` becomes `TransferAsset::beneficiary`

### Other changes

- `FixedRateOfConcreteFungible` has been deprecated - use `FixedRateOfFungible` instead.
- `pallet_xcm::Config` now has a new type field `LocationInverter`; this should be equivalent to the field of the same name in `xcm_executor::Config`.
- `AssetInstance::Index ( index: N }` becomes `AssetInstance::Index(N)` (NOTE: This is also the case in `v0`)

## TODO:
- [x] Implement all the same functionality as `OldMultiAsset`.
- Convert all constructor and matching code over in:
  - [x] Polkadot codebase
  - [x] XCM Builder tests
  - [x] Cumulus codebase (https://github.com/paritytech/cumulus/pull/561)
- Optimize:
  - [x] Optimize `subsume_assets`
  - [x] Optimize `checked_sub`
- [x] Remove `OldMultiAsset`


apopiak (commented):
:+1: for the MultiAsset/WildMultiAsset/MultiAssetFilter distinction

xcm/src/v0/multiasset.rs
```diff
@@ -0,0 +1,383 @@
...
+				_ => (),
+			}
+		}
+		false
+	}
+}
+
+
+
+/// A `Vec` of `MultiAsset`s. There may be no duplicate fungible items in here and when decoding, they must be sorted.
```
apopiak, line 187: Why not use a `BTreeSet` if you want to disallow duplication? (And it would also be sorted, right?)

gavofyork: That's what `Assets` is for. This is meant to be lightweight.

xcm/src/v0/multiasset.rs
```diff
@@ -0,0 +1,383 @@
...
+	/// Prepend a `MultiLocation` to any concrete asset items, giving it a new root location.
+	pub fn reanchor(&mut self, prepend: &MultiLocation) -> Result<(), ()> {
+		self.0.iter_mut().try_for_each(|i| i.reanchor(prepend))
+	}
+}
+
+
+
+
+
```
apopiak, line 267:
```suggestion

```
reduce whitespace?

xcm/src/v0/multiasset.rs
```diff
@@ -0,0 +1,383 @@
...
+impl<A: Into<AssetId>, B: Into<WildFungibility>> From<(A, B)> for WildMultiAsset {
+	fn from((id, fun): (A, B)) -> WildMultiAsset {
+		WildMultiAsset::AllOf { fun: fun.into(), id: id.into() }
+	}
+}
+
+
+
+
```
apopiak, line 316:
```suggestion

```
reduce whitespace?

xcm/src/v0/multiasset.rs
```diff
@@ -0,0 +1,383 @@
...
+/// Classification of whether an asset is fungible or not, along with an optional amount or instance.
+#[derive(Copy, Clone, Eq, PartialEq, Ord, PartialOrd, Debug, Encode, Decode)]
+pub enum WildFungibility {
+	Fungible,
+	NonFungible,
+}
+
+#[derive(Clone, Eq, PartialEq, Ord, PartialOrd, Debug, Encode, Decode)]
```
apopiak, line 275:
```suggestion
/// A wildcard representing a set of assets.
#[derive(Clone, Eq, PartialEq, Ord, PartialOrd, Debug, Encode, Decode)]
```

xcm/src/v0/multiasset.rs
```diff
@@ -0,0 +1,383 @@
...
+	fn from((id, fun): (A, B)) -> WildMultiAsset {
+		WildMultiAsset::AllOf { fun: fun.into(), id: id.into() }
+	}
+}
+
+
+
+
+/// `MultiAsset` collection, either `MultiAssets` or a single wildcard. Note: vectors of wildcards
+/// whose encoding is supported in XCM v0 are unsupported in this implementation and will result in a decode error.
```
apopiak, line 318:
```suggestion
/// `MultiAsset` collection, either `MultiAssets` or a single wildcard.
/// 
/// Note: Vectors of wildcards whose encoding is supported in XCM v0 are unsupported
/// in this implementation and will result in a decode error.
```

xcm/src/v0/order.rs
```diff
@@ -29,7 +29,7 @@ use super::{MultiAsset, MultiLocation, Xcm};
 pub enum Order<Call> {
 	/// Do nothing. Not generally used.
 	#[codec(index = 0)]
-	Null,
+	Noop,
```
apopiak, line 32: :+1:

xcm/src/v0/order.rs
```diff
@@ -39,33 +39,34 @@ pub enum Order<Call> {
 	///
 	/// Errors:
 	#[codec(index = 1)]
-	DepositAsset { assets: Vec<MultiAsset>, dest: MultiLocation },
+	// TODO: https://github.com/paritytech/polkadot/issues/3547 introduce `, max_assets: u32`
+	DepositAsset { assets: MultiAssetFilter, dest: MultiLocation },
```
apopiak, line 43:
```suggestion
	DepositAsset { assets: MultiAssetFilter, beneficiary: MultiLocation },
```
Would (something like) this make sense to avoid leading people to think it might trigger another XCM?

gavofyork: perhaps, but maybe in a different PR.

shaunxw (commented):
XCM related libs should be able to handle both `v0` and `v1` IMO. Relay chain and parachains in the network won't have the same schedule to update XCM versions, if only one version is supported, the incompatibility could lead to execution failures on message destinations and fund loss.

xcm/src/lib.rs
```diff
@@ -23,29 +23,64 @@
...
+			VersionedXcm::V1(x) => Ok(x),
+			_ => Err(()),
+		}
+	}
 }
 
 pub mod opaque {
-	pub mod v0 {
+	pub mod v1 {
 		// Everything from v0
```
4meta5, line 76:
```suggestion
		// Everything from v1
```

xcm/src/lib.rs
```diff
@@ -23,29 +23,64 @@
...
-		pub use crate::v0::*;
+		pub use crate::v1::*;
 		// Then override with the opaque types in v0
```
4meta5, line 78:
```suggestion
		// Then override with the opaque types in v1
```

xcm/src/lib.rs
```diff
@@ -55,11 +90,45 @@ pub mod opaque {
 /// A versioned multi-location, a relative location of a cross-consensus system identifier.
 #[derive(Clone, Eq, PartialEq, Encode, Decode, Debug)]
 pub enum VersionedMultiLocation {
-	V0(v0::MultiLocation),
+	V0(v1::MultiLocation),
```
4meta5, line 93: why isn't this `V0(Unsupported)` like for `VersionedMultiAsset`

xcm/src/v1/mod.rs
```diff
@@ -122,27 +137,27 @@ pub enum Xcm<Call> {
...
+	/// been placed into the Holding Register.
 	///
-	/// - `assets`: The asset(s) that are minted into holding.
-	/// - `effects`: The order(s) to execute on the holding account.
+	/// - `assets`: The asset(s) that are minted into the Holding Register.
+	/// - `effects`: The order(s) to execute on the Holding Register.
 	///
-	/// Safety: `origin` must be trusted to have irrevocably destroyed the `assets` prior as a consequence of
-	/// sending this message.
+	/// Safety: `origin` must be trusted to have irrevocably destroyed the corresponding`assets` prior as a consequence
```
4meta5, line 151:
```suggestion
	/// Safety: `origin` must be trusted to have irrevocably destroyed the corresponding `assets` prior as a consequence
```

xcm/src/v1/multiasset.rs
```diff
@@ -0,0 +1,484 @@
...
+	MultiLocation::{self, X1},
+};
+use alloc::{vec, vec::Vec};
+use core::cmp::Ordering;
+use parity_scale_codec::{self as codec, Decode, Encode};
+
+/// A general identifier for an instance of a non-fungible asset class.
+#[derive(Clone, Eq, PartialEq, Ord, PartialOrd, Encode, Decode, Debug)]
+pub enum AssetInstance {
+	/// Undefined - used if the NFA class has only one instance.
```
4meta5, line 37:
```suggestion
	/// Undefined - used if the non-fungible asset class has only one instance.
```
unless the NFA acronym is defined somewhere easy to find

xcm/src/v1/multiasset.rs
```diff
@@ -0,0 +1,484 @@
...
+	fn from(_: ()) -> Self {
+		Self::Undefined
+	}
+}
+
+/*impl From<u128> for AssetInstance {
+	fn from(x: u128) -> Self {
+		Self::Index(x)
+	}
+}*/
```
4meta5, line 70: intentionally commented out?

xcm/src/v1/multiasset.rs
```diff
@@ -0,0 +1,484 @@
...
+	pub fn get(&self, index: usize) -> Option<&MultiAsset> {
+		self.0.get(index)
+	}
+}
+/// Classification of whether an asset is fungible or not, along with an optional amount or instance.
+#[derive(Copy, Clone, Eq, PartialEq, Ord, PartialOrd, Debug, Encode, Decode)]
+pub enum WildFungibility {
+	Fungible,
+	NonFungible,
+}
```
4meta5, line 389:
```suggestion
/// Classification of whether an asset is fungible or not.
#[derive(Copy, Clone, Eq, PartialEq, Ord, PartialOrd, Debug, Encode, Decode)]
pub enum WildFungibility {
	Fungible,
	NonFungible,
}
```
where is optional amount or instance?

xcm/src/v1/multiasset.rs
```diff
@@ -0,0 +1,484 @@
...
+
+/// A wildcard representing a set of assets.
+#[derive(Clone, Eq, PartialEq, Ord, PartialOrd, Debug, Encode, Decode)]
+pub enum WildMultiAsset {
+	/// All assets in the holding register, up to `usize` individual assets (different instances of non-fungibles could
+	/// as separate assets).
```
4meta5, line 395:
```suggestion
	/// be separate assets).
```

xcm/src/v1/order.rs
```diff
@@ -107,20 +122,32 @@ pub enum Order<Call> {
...
 	},
 
-	/// Pay for the execution of some XCM with up to `weight` picoseconds of execution time, paying for this with
-	/// up to `fees` from the holding account.
+	/// Pay for the execution of some XCM `instructions` and `orders` with up to `weight` picoseconds of execution time,
+	/// paying for this with up to `fees` from the Holding Register.
 	///
+	/// - `fees`: The asset(s) to remove from holding to pay for fees.
+	/// - `weight`: The amount of weight to purchase; this should be at least the shallow weight of `effects` and `xcm`.
+	/// - `debt`: The amount of weight-debt already incurred to be pay off; this should be equal to the unpaid weight of
```
4meta5, line 133:
```suggestion
	/// - `debt`: The amount of weight-debt already incurred to be paid off; this should be equal to the unpaid weight of
```

xcm/xcm-executor/src/assets.rs
```diff
@@ -311,308 +196,298 @@ impl Assets {
...
+		self.general_take(mask, false, usize::max_value())
+	}
+
+	/// Consumes `self` and returns its original value excluding `asset` iff it contains at least `asset`.
+	pub fn checked_sub(mut self, asset: MultiAsset) -> Result<Assets, Assets> {
+		match asset.fun {
+			Fungible(amount) => {
+				let remove = if let Some(balance) = self.fungible.get_mut(&asset.id) {
+					if *balance >= amount {
+						*balance -= amount;
```
4meta5, line 349:
```suggestion
						*balance.saturating_sub(amount);
```

xcm/xcm-executor/src/assets.rs
```diff
@@ -311,308 +196,298 @@ impl Assets {
...
+				if !saturate {
+					self.ensure_contains(&assets)?;
+				}
+				for asset in assets.drain().into_iter() {
+					match asset {
+						MultiAsset { fun: Fungible(amount), id } => {
+							let (remove, amount) = match self.fungible.get_mut(&id) {
+								Some(self_amount) => {
+									let amount = amount.min(*self_amount);
+									*self_amount -= amount;
```
4meta5, line 289:
```suggestion
									*self_amount.saturating_sub(amount);
```

xcm/xcm-executor/src/assets.rs
```diff
@@ -311,308 +196,298 @@ impl Assets {
...
-		MultiAsset::AbstractFungible { id: vec![id], amount }
+		(vec![id], amount).into()
 	}
 	#[allow(non_snake_case)]
-	fn ANF(class: u8, instance_id: u128) -> MultiAsset {
-		MultiAsset::AbstractNonFungible {
-			class: vec![class],
-			instance: AssetInstance::Index { id: instance_id },
-		}
+	fn ANF(class: u8, instance_id: u8) -> MultiAsset {
```
4meta5, line 435:
```suggestion
  // Abstract NonFungible constructor
	fn ANF(class: u8, instance_id: u8) -> MultiAsset {
```

xcm/xcm-executor/src/assets.rs
```diff
@@ -311,308 +196,298 @@ impl Assets {
...
 	}
 }
 
 #[cfg(test)]
 mod tests {
 	use super::*;
+	use xcm::v1::prelude::*;
+	use MultiLocation::Here;
 	#[allow(non_snake_case)]
 	fn AF(id: u8, amount: u128) -> MultiAsset {
```
4meta5, line 431:
```suggestion
  // Abstract Fungible constructor
	fn AF(id: u8, amount: u128) -> MultiAsset {
```

xcm/xcm-executor/src/assets.rs
```diff
@@ -311,308 +196,298 @@ impl Assets {
...
-	fn ANF(class: u8, instance_id: u128) -> MultiAsset {
-		MultiAsset::AbstractNonFungible {
-			class: vec![class],
-			instance: AssetInstance::Index { id: instance_id },
-		}
+	fn ANF(class: u8, instance_id: u8) -> MultiAsset {
+		(vec![class], vec![instance_id]).into()
 	}
 	#[allow(non_snake_case)]
 	fn CF(amount: u128) -> MultiAsset {
```
4meta5, line 439:
```suggestion
  /// Concrete Fungible constructor
	fn CF(amount: u128) -> MultiAsset {
```

xcm/xcm-executor/src/assets.rs
```diff
@@ -311,308 +196,298 @@ impl Assets {
...
-		MultiAsset::ConcreteFungible { id: MultiLocation::Null, amount }
+		(Here, amount).into()
 	}
 	#[allow(non_snake_case)]
-	fn CNF(instance_id: u128) -> MultiAsset {
-		MultiAsset::ConcreteNonFungible {
-			class: MultiLocation::Null,
-			instance: AssetInstance::Index { id: instance_id },
-		}
+	fn CNF(instance_id: u8) -> MultiAsset {
```
4meta5, line 443:
```suggestion
  /// Concrete non-fungible constructor
	fn CNF(instance_id: u8) -> MultiAsset {
```

gavofyork:
> XCM related libs should be able to handle both `v0` and `v1` IMO. Relay chain and parachains in the network won't have the same schedule to update XCM versions, if only one version is supported, the incompatibility could lead to execution failures on message destinations and fund loss.

Yes, and this will be the case once Polkadot hits 1.0 and parachains with XCMP launch. But given that parachains are not meaningfully interoperating yet and that we are in Polkadot's pre-launch `v0` we had no plans to keep `v0` backward compatible.

xcm/src/lib.rs
gavofyork: Because `MultiLocation` did not change.

xcm/src/v1/multiasset.rs
```diff
@@ -0,0 +1,484 @@
...
+		Self::Undefined
+	}
+}
+
+/*impl From<u128> for AssetInstance {
+	fn from(x: u128) -> Self {
+		Self::Index(x)
+	}
+}*/
+
```
gavofyork, line 71:
```suggestion
```

xcm/src/v1/mod.rs
```diff
@@ -0,0 +1,334 @@
...
+
+// Substrate is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with Cumulus.  If not, see <http://www.gnu.org/licenses/>.
+
+//! Version 0 of the Cross-Consensus Message format data structures.
```
4meta5, line 17:
```suggestion
//! Version 1 of the Cross-Consensus Message format data structures.
```

xcm/src/v1/order.rs
```diff
@@ -0,0 +1,183 @@
...
+
+// Substrate is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with Cumulus.  If not, see <http://www.gnu.org/licenses/>.
+
+//! Version 0 of the Cross-Consensus Message format data structures.
```
4meta5, line 17:
```suggestion
//! Version 1 of the Cross-Consensus Message format data structures.
```

runtime/kusama/src/lib.rs
```diff
@@ -1317,69 +1311,19 @@ pub type LocalOriginToLocation = (
...
 impl pallet_xcm::Config for Runtime {
 	type Event = Event;
 	type SendXcmOrigin = xcm_builder::EnsureXcmOrigin<Origin, LocalOriginToLocation>;
 	type XcmRouter = XcmRouter;
 	// Anyone can execute XCM messages locally...
 	type ExecuteXcmOrigin = xcm_builder::EnsureXcmOrigin<Origin, LocalOriginToLocation>;
-	// ...but they must match our filter, which requires them to be a simple withdraw + teleport.
-	type XcmExecuteFilter = OnlyWithdrawTeleportForAccounts;
+	// ...but they must match our filter, which rejects all.
+	type XcmExecuteFilter = ();
```
shawntabrizi, line 1321: so we are disabling transfers temporarily to statemine?

shawntabrizi, line 1321: so the plan is to disable any xcm until both the relay chain and statemine are updated?

gavofyork: Actually, I just saw no point maintaining the ability to manually execute transfers when there are dispatchables which do it for you. We can open up this filter as needed in the future.

xcm/pallet-xcm/src/lib.rs
```diff
@@ -210,29 +219,38 @@ pub mod pallet {
...
-			fees.reanchor(&inv_dest).map_err(|_| Error::<T>::CannotReanchor)?;
+			let fees = assets
+				.get(fee_asset_item as usize)
+				.ok_or(Error::<T>::Empty)?
+				.clone()
+				.reanchored(&inv_dest)
+				.map_err(|_| Error::<T>::CannotReanchor)?;
+			let max_assets = assets.len() as u32;
+			assets.sort();
+			let assets = assets.into();
```
shawntabrizi, line 239:
```suggestion
			let assets = assets.into();
```

xcm/src/v0/order.rs
```diff
@@ -152,3 +153,54 @@ impl<Call> Order<Call> {
...
+				assets: assets.into(),
+				dest: dest.into(),
+				effects: effects
+					.into_iter()
+					.map(Order::<()>::try_from)
+					.collect::<result::Result<_, _>>()?,
+			},
+			Order1::QueryHolding { query_id, dest, assets } =>
+				QueryHolding { query_id, dest: dest.into(), assets: assets.into() },
+			Order1::BuyExecution { fees, weight, debt, halt_on_error, instructions, orders } => {
```
shawntabrizi, line 194:
```suggestion
			Order1::BuyExecution { fees, weight, debt, halt_on_error, orders, instructions } => {
```

xcm/xcm-executor/src/assets.rs
```diff
@@ -64,232 +35,146 @@ pub struct Assets {
...
-			self.saturating_subsume(asset)
+	pub fn subsume_assets(&mut self, mut assets: Assets) {
+		let mut f_iter = assets.fungible.iter_mut();
+		let mut g_iter = self.fungible.iter_mut();
+		if let (Some(mut f), Some(mut g)) = (f_iter.next(), g_iter.next()) {
+			loop {
+				if f.0 == g.0 {
+					// keys are equal. in this case, we add `self`'s balance for the asset onto `assets`, balance, knowing
+					// that the `append` operation which follows will clobber `self`'s value and only use `assets`'s.
+					*f.1 += *g.1;
```
shawntabrizi, line 137:
```suggestion
					(*f.1).saturating_accrue(*g.1);
```

xcm/xcm-executor/src/assets.rs
```diff
@@ -64,232 +35,146 @@ pub struct Assets {
...
+
+	/// A borrowing iterator over all assets.
 	pub fn assets_iter<'a>(&'a self) -> impl Iterator<Item = MultiAsset> + 'a {
-		let fungible = self.fungible_assets_iter();
-		let non_fungible = self.non_fungible_assets_iter();
-		fungible.chain(non_fungible)
+		self.fungible_assets_iter().chain(self.non_fungible_assets_iter())
 	}
 
 	/// Mutate `self` to contain all given `assets`, saturating if necessary.
```
shawntabrizi, line 128:
```suggestion
	/// Mutate `self` to contain all given `assets`, saturating if necessary.
	///
	/// NOTE: [`Assets`] are always sorted, allowing us to optimize this function from `O(n^2)` to `O(n)`.
```

xcm/xcm-executor/src/assets.rs
```diff
@@ -311,308 +196,302 @@ impl Assets {
...
-									result.saturating_subsume(asset)
-								}
-							},
-							_ => (),
+			}
+		}
+		return Ok(())
+	}
+
+	/// Mutates `self` to its original value less `mask` and returns `true`.
```
shawntabrizi, line 226: doesnt return a bool?

shawntabrizi: Also note about `limit` and fungible order is probably relevant

xcm/xcm-executor/src/assets.rs
```diff
@@ -311,308 +196,302 @@ impl Assets {
...
 		}
-		result
 	}
 
-	/// Take all possible assets up to `assets` from `self`, mutating `self` and returning the
-	/// assets taken.
+	/// Return the assets in `self`, but (asset-wise) of no greater value than `assets`.
 	///
-	/// Wildcards work.
+	/// Result is undefined if `assets` includes elements which match to the same asset more than once.
```
shawntabrizi, line 373: `assets` -> `mask`?

xcm/src/v1/junction.rs
```diff
@@ -41,10 +41,7 @@ pub enum BodyId {
 	Named(Vec<u8>),
 	/// An indexed body.
 	// TODO: parity-scale-codec#262: Change to be a tuple.
```
shawntabrizi, line 43: https://github.com/paritytech/parity-scale-codec/issues/262

This is completed

xcm/src/v1/multiasset.rs
```diff
@@ -0,0 +1,575 @@
...
+			All => true,
+		}
+	}
+
+	/// Prepend a `MultiLocation` to any concrete asset components, giving it a new root location.
+	pub fn reanchor(&mut self, prepend: &MultiLocation) -> Result<(), ()> {
+		use WildMultiAsset::*;
+		match self {
+			AllOf { ref mut id, .. } => id.reanchor(prepend).map_err(|_| ()),
+			_ => Ok(()),
```
shawntabrizi, line 498:
```suggestion
			All => Ok(()),
```

shawntabrizi (approved):
lgtm

xcm/src/v1/multiasset.rs
```diff
@@ -0,0 +1,575 @@
...
+	pub fn from_sorted_and_deduplicated_skip_checks(r: Vec<MultiAsset>) -> Self {
+		Self(r)
+	}
+
+	/// Add some asset onto the list. This is quite a laborious operation since it maintains the ordering.
+	pub fn push(&mut self, a: MultiAsset) {
+		if let Fungibility::Fungible(ref amount) = a.fun {
+			for asset in self.0.iter_mut().filter(|x| x.id == a.id) {
+				if let Fungibility::Fungible(ref mut balance) = asset.fun {
+					*balance += *amount;
```
shawntabrizi, line 389:
```suggestion
					*balance.saturating_accrue(*amount);
```

xcm/src/v1/multiasset.rs
```diff
@@ -0,0 +1,575 @@
...
+	/// In release mode, this skips any checks to ensure that `r` is correct, making it a negligible-cost operation.
+	/// Generally though you should avoid using it unless you have a strict proof that `r` is valid.
+	///
+	/// In test mode, this checks anyway and panics on fail.
+	#[cfg(not(test))]
+	pub fn from_sorted_and_deduplicated_skip_checks(r: Vec<MultiAsset>) -> Self {
+		Self(r)
+	}
+
+	/// Add some asset onto the list. This is quite a laborious operation since it maintains the ordering.
```
shawntabrizi, line 384:
```suggestion
	/// Add some asset onto the list, saturating. This is quite a laborious operation since it maintains the ordering.
```

xcm/src/v1/multiasset.rs
```diff
@@ -0,0 +1,575 @@
...
+
+	/// Returns true if `self` is a super-set of the given `inner`.
+	pub fn contains(&self, inner: &MultiAsset) -> bool {
+		self.0.iter().any(|i| i.contains(inner))
+	}
+
+	/// Consume `self` and return the inner vec.
+	pub fn drain(self) -> Vec<MultiAsset> {
+		self.0
+	}
```
apopiak, line 411: late nitpick: I would expect an iterator from `drain` based the Substrate function of the same name
this looks more like `into_inner` to me

xcm/pallet-xcm/src/lib.rs
```diff
@@ -153,21 +154,28 @@ pub mod pallet {
 			origin: OriginFor<T>,
 			dest: MultiLocation,
 			beneficiary: MultiLocation,
-			assets: Vec<MultiAsset>,
+			assets: MultiAssets,
+			fee_asset_item: u32,
```
xlc, line 158: This is a `transaction_version` breaking change. Maybe need some additional label so frontend developers can be aware about this?

shawntabrizi: done

shaunxw: I think polkadot.js would need updates to include new types and type changes introduced in this PR. cc @jacogr

PR #2837, opened 2021-04-07
Update srtool

s3krit:
Switches to chevdor's latest version of srtool, and fixes an issue with uploading the WASM (issue can be seen in the latest pre-release of v0.8.30-rc2).


.github/workflows/publish-draft-release.yml
```diff
@@ -7,13 +7,26 @@ on:
...
+          echo "::set-output name=nightly::$(rustc +nightly --version)"
+
   build-runtimes:
     runs-on: ubuntu-latest
     strategy:
       matrix:
         runtime: ['polkadot', 'kusama']
     container:
-      image: paritytech/srtool:nightly-2020-10-27
+      image: chevdor/srtool:nightly-2021-03-15
```
gabreal, line 29: according to our [docker images maintenance policy](https://github.com/paritytech/devops/wiki/DevOps-Internal:-Best-Practices:-Docker-images-maintenance-policy) this image should be build in parity ci infrastructure.

TriplEight: @chevdor are you OK to ~nightly build and test `srtool` on our infrastructure?

TriplEight (changes requested):

.github/workflows/publish-draft-release.yml
chevdor: Sure thing. Westend does not support the new `on-chain-release-build` feature that is now the default for Kusama and Polkadot. So I need to check whether I add some magic to the feature detection in `srtool` (currently solely based on the version number) or if we bypass the detection and force `BUILD_OPTS` to be empty.

Refs. : 
- https://gitlab.com/chevdor/srtool/-/commit/accf5528a42aff64ee6b0919aa3a16716418fcbd#f6b38ed32919f0ac464cf607efca22951eac98d6_110_78
- https://gitlab.com/chevdor/srtool/-/commit/accf5528a42aff64ee6b0919aa3a16716418fcbd#f6b38ed32919f0ac464cf607efca22951eac98d6_110_78

s3krit: I actually came up with a better way for us (Parity) to handle this, without needing to maintain some Parity-specific fork. I've created a small repo [here](https://gitlab.parity.io/parity/infrastructure/ci_cd/srtool-ci/) with a `.gitlab-ci.yml` that is in charge of building and pushing srtool to *our* Dockerhub, and it has srtool as a git submodule. Whenever we need to update srtool, we can review the changes, and bump the submodule.

TriplEight (approved):
LGTM.
@s3krit why not just add an `srtool` image CI into our [scripts](https://github.com/paritytech/scripts/tree/master/dockerfiles) repo which is a central point for this kind of images? It also has scanning and schedules.

chevdor:
Please note the following points which can be annoying for CI but is nicer for users: a given srtool image tag is NOT immutable.

In other words, `chevdor/srtool:nightly-2021-03-15` from today *may* not be the same `chevdor/srtool:nightly-2021-03-15` than last week. The rustc version is garanteed to be but not the content of the image.
I try to avoid that but there is currently no garanty.

There are a few options to ensure having the latest image & version:
- pull the most recent anyway from teh docker hub
- call `srtool version`. There are 2 important versions: the `rustc` version and the build script version

The latest rustc version is the tag of the image. It is also at https://gitlab.com/chevdor/srtool/-/blob/master/RUSTC_VERSION

The latest build script version is available at https://gitlab.com/chevdor/srtool/-/blob/master/VERSION and this is also the latest tag in the https://gitlab.com/chevdor/srtool repo.

Here is a typical `srtool version` output:
```
$ srtool version
{
  "name": "srtool",
  "version": "0.9.10",
  "rustc": "nightly-2021-03-15"
}
```
Or the CI friendlier version (_c_ompact=single line + _M_onochrome): 
```
$ srtool version -cM
{"name":"srtool","version":"0.9.10","rustc":"nightly-2021-03-15"}
```

TriplEight:
@chevdor I don't get how the date-bind tag (`nightly-2021-03-15 `) once saved in the image can be mutable.

chevdor:
There are cases when the build script (defining the version of srtool) needs to be upgraded.
That may happen at times when the rustc version should remain the same.
That means that your `srtool:nightly-2021-xx-yy` from last month may not be the same than the `srtool:nightly-2021-xx-yy` you would get today from a docker pull.

TriplEight:
>  That may happen at times when the rustc version should remain the same.

@chevdor so you're rebuilding the same date-bind tag in case if you need to change the build script? So it kinda defeats the purpose of the date tag (although I understand it's not a date but a Rust version).

I recommend [looking into](https://github.com/paritytech/scripts/blob/master/.gitlab-ci.yml#L324) our concept of testing the CI image against the target repository after the nightly update has happened.

chevdor:
Yes, I know how this may sound but this is done to ensure ease of use on the user's side.
The way this is currently done, a user is guaranteed to get deterministic builds with a given tag.

Using tag like `srtool:<nightly-x-y-z_VERSION` where `VERSION` is the srtool version would be possible but highly inconvenient for the users. 

In periods when I don't need to update the build script, this is not really an issue as I usually take the chance to bump up the rustc version. In case of, like recently, patched versions, the user only needs to delete the image and pull it again. 

Using `srtool version` it would be actually easy to add our own tag:
```
TAG=nightly-2021-03-15; IMG=chevdor/srtool; VERSION=`srtool version -cMr | tr -d '[:space:]' | jq -r .version`; docker tag $IMG:$TAG $IMG:$TAG\_$VERSION
````

That ends up in an image tagged as mentioned above and this image would not change over time.

s3krit:
I'm going to opt to keep our srtool-ci repo as a separate repo, rather than adding it to the `scripts` repo - due to the use of a submodule, I would prefer to keep this complexity out of the scripts repo, and we don't need to make use of schedules or anything like that *yet*. This keep srtool-ci as a separate, distinct repository which only needs to receive updates at the time of a release (when we would want to ensure we are running some latest version of srtool that other people building the newly-proposed runtime are likely to also be running). I'll likely bake that bumping into automation at some point too, as well as adding a note to the changelogs indicating which commit of srtool we built with.

That being said, that conversation, and the conversation of how specifically we build srtool for use in our release process, are out of scope for *this particular* PR, which is strictly around bumping the version of `srtool` used, and fixing the path issue. I'll go ahead and merge this, and any further discussion for the nature of `paritytech/srtool` can be done in an issue in the CI/CD repo or somethiung. Cheers for the reviews, all! ^^

PR #2846, opened 2021-04-07
Introduce plurality XCM locations

gavofyork:
An additional `Junction` type for integrating origins such as the council into XCM without having to reduce them to `AccountId`s.

### DIFF
--- a/xcm/src/v0/junction.rs
+++ b/xcm/src/v0/junction.rs
@@ -32,6 +32,41 @@ pub enum NetworkId {
 	Kusama,
 }
 
+/// An identifier of a pluralistic body.
+#[derive(Clone, Eq, PartialEq, Ord, PartialOrd, Encode, Decode, Debug)]
+pub enum BodyId {
+	/// The only body in its context.
+	Unit,
+	/// A named body.
+	Named(Vec<u8>),
+	/// An indexed body.
+	// TODO: parity-scale-codec#262: Change to be a tuple.
+	Index { #[codec(compact)] id: u32 },
+	/// The unambiguous executive body (for Polkadot, this would be the Polkadot council).
+	Executive,
+	/// The unambiguous technical body (for Polkadot, this would be the Technical Committee).
+	Technical,
+	/// The unambiguous legislative body (for Polkadot, this could be considered the opinion of a majority of
+	/// lock-voters).
+	Legislative,
+	/// The unambiguous judicial body (this doesn't exist on Polkadot, but if it were to get a "grand oracle", it
+	/// may be considered as that).
+	Judicial,
+}
+
+/// A part of a pluralistic body.
+#[derive(Clone, Eq, PartialEq, Ord, PartialOrd, Encode, Decode, Debug)]
+pub enum BodyPart {
+	/// The body's declaration, under whatever means it decides.
+	Voice,
+	/// A given number of members of the body.
+	Members { #[codec(compact)] count: u32 },
+	/// No less than the given proportion of members of the body.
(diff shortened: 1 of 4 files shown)


xcm/src/v0/junction.rs
```diff
@@ -85,21 +120,28 @@ pub enum Junction {
 	///
 	/// Not currently used except as a fallback when deriving ancestry.
 	OnlyChild,
+	/// A pluralistic body existing within consensus.
+	///
+	/// Typical to be used to represent a governance origin of a chain, but could in principle be used to represent
+	/// things such as multisigs also.
+	Plurality { id: BodyId, part: BodyPart },
```
xlc, line 127: what will be the `id` if this is used to represent a multisig or some custom dynamic collective? `BodyId::Named(EncodedMutiSigAddressOrCollectiveId)`?

gavofyork: If the multisig has an address then it is probably best to simply use it as a multisig and not expose the plurality aspect. In that case you'd just collapse it to a `AccountId32` or whatever.

This is specifically for the case where you want to expose the nature of a collective over XCM. Right now I don't see much use beyond exposing major governance bodies between chains. If you wanted to expose more, then you'll need to come up with your own naming (or make suggestions for additional entries in the `BodyId` enum).

shawntabrizi (approved):

rphmeier (approved):

xcm/src/v0/junction.rs
```diff
@@ -32,6 +32,41 @@ pub enum NetworkId {
...
+	/// The unambiguous judicial body (this doesn't exist on Polkadot, but if it were to get a "grand oracle", it
+	/// may be considered as that).
+	Judicial,
+}
+
+/// A part of a pluralistic body.
+#[derive(Clone, Eq, PartialEq, Ord, PartialOrd, Encode, Decode, Debug)]
+pub enum BodyPart {
+	/// The body's declaration, under whatever means it decides.
+	Voice,
```
apopiak, line 61: @gavofyork Can you elaborate on this? Does this mean it could be any of [one member, a number of members, half the body, the full body]? What's the mapping to the e.g. the collective or democracy origins.

xcm/src/v0/junction.rs
```diff
@@ -32,6 +32,41 @@ pub enum NetworkId {
...
+	/// A given number of members of the body.
+	Members { #[codec(compact)] count: u32 },
+	/// No less than the given proportion of members of the body.
+	AtLeastProportion { #[codec(compact)] nom: u32, #[codec(compact)] denom: u32 },
+	/// More than than the given proportion of members of the body.
+	MoreThanProportion { #[codec(compact)] nom: u32, #[codec(compact)] denom: u32 },
```
apopiak, line 67: Why split `Proportion { nom: u32, denom: u32 }` into these two? Won't the receiving side have to interpret the numbers anyway?

xcm/src/v0/junction.rs
gavofyork: Collective wouldn't have a mapping since there's no unambiguous way for it to make a declaration.

Democracy would be the already established way - in Polkadot Relay chain, this would be a lock-vote with the AQB determined by council support.

xcm/src/v0/junction.rs
gavofyork: Because it allows for accurate representation of known origins.

apopiak: Can you point to them? I only see this one: https://github.com/paritytech/substrate/blob/282d57c0745b530fe7a9ebaffcd6ac36c09d0554/frame/collective/src/lib.rs#L160

PR #2854, opened 2021-04-08
Add parachains modules to Westend and Kusama runtimes

rphmeier:
TODO:
  - [x] Add pallet indices
  - [x] Migration to initialize `HostConfiguration`
  - [x] Set configuration values for slots, auctions to non-placeholder values (auctions, crowdloans no longer a part of this PR)
  - [x] Update implementation of `ParachainHost` API to use the `runtime_api_impl` module
  - [x] benchmarks
  - [x] Call filters to prevent registration & swapping of parathreads


runtime/kusama/src/lib.rs
```diff
@@ -1036,15 +1198,54 @@ impl pallet_babe::migrations::BabePalletPrefix for Runtime {
...
+			max_code_size: MaxCodeSize::get(),
+			max_head_data_size: MaxHeadSize::get(),
+			max_upward_queue_count: 10,
+			max_upward_queue_size: 50 * 1024,
+			max_upward_message_size: 50 * 1024,
+			max_upward_message_num_per_candidate: 10,
+			hrmp_max_message_num_per_candidate: 10,
+			validation_upgrade_frequency: 1 * DAYS,
+			validation_upgrade_delay: EPOCH_DURATION_IN_BLOCKS,
+			max_pov_size: 5 * 1024 * 1024,
```
eskimor, line 1214: We should use the constant `MAX_POV_SIZE` here to maintain consistency.

runtime/westend/src/lib.rs
```diff
@@ -748,6 +732,128 @@ impl pallet_proxy::Config for Runtime {
...
+	type Randomness = pallet_babe::RandomnessFromOneEpochAgo<Runtime>;
+	type ForceOrigin = EnsureRoot<AccountId>;
+}
+
+impl paras_sudo_wrapper::Config for Runtime {}
+
+parameter_types! {
+	pub const ParaDeposit: Balance = 5 * DOLLARS;
+	pub const DataDepositPerByte: Balance = deposit(0, 1);
+	pub const MaxCodeSize: u32 = 5 * 1024 * 1024; // 10 MB
```
eskimor, line 779: PR #2803 introduces a constant for this as well [here](https://github.com/paritytech/polkadot/blob/0b51e892ac89883dddf83e977971b02179746cb0/primitives/src/v1.rs#L181) which we should use once merged.

runtime/kusama/src/lib.rs
```diff
@@ -1036,15 +1198,54 @@ impl pallet_babe::migrations::BabePalletPrefix for Runtime {
...
-				allowed_slots: babe_primitives::AllowedSlots::PrimaryAndSecondaryPlainSlots,
-				..BABE_GENESIS_EPOCH_CONFIG
-			}
-		)
+		let config = parachains_configuration::HostConfiguration {
+			max_code_size: MaxCodeSize::get(),
+			max_head_data_size: MaxHeadSize::get(),
+			max_upward_queue_count: 10,
+			max_upward_queue_size: 50 * 1024,
+			max_upward_message_size: 50 * 1024,
```
eskimor, line 1209: 2.6 GB ....? Or is queue_size in bytes and this means we can only store 1 maximal sized message? The queue_count above could be a hint that it is the latter.

runtime/kusama/src/lib.rs
```diff
@@ -1036,15 +1198,54 @@ impl pallet_babe::migrations::BabePalletPrefix for Runtime {
...
+			hrmp_max_parachain_outbound_channels: 10,
+			hrmp_max_parathread_outbound_channels: 0,
+			hrmp_open_request_ttl: 2,
+			hrmp_sender_deposit: deposit(4, 128),
+			hrmp_recipient_deposit: deposit(4, 128),
+			hrmp_channel_max_capacity: 1000,
+			hrmp_channel_max_total_size: 100 * 1024,
+			hrmp_max_parachain_inbound_channels: 10,
+			hrmp_max_parathread_inbound_channels: 0,
+			hrmp_channel_max_message_size: 100 * 1024,
```
eskimor, line 1226: We likely don't want a single message eat up all the available space?

runtime/westend/src/lib.rs
```diff
@@ -816,12 +941,54 @@ impl pallet_babe::migrations::BabePalletPrefix for Runtime {
...
+			chain_availability_period: 1 * MINUTES,
+			thread_availability_period: 1 * MINUTES,
+			scheduling_lookahead: 1,
+			max_validators_per_core: Some(5),
+			max_validators: Some(200),
+			dispute_period: 6,
+			dispute_post_conclusion_acceptance_period: 1 * HOURS,
+			dispute_max_spam_slots: 2,
+			dispute_conclusion_by_time_out_period: 1 * HOURS,
+			no_show_slots: 2,
```
burdges, line 983: Are we doing okay with only two here?  We briefly discussed this but then I think other changed superseded anything considered there,  I still like tow if nothing else problematic.

runtime/kusama/src/lib.rs
eskimor: even if it is the latter, we likely don't want a single message to occupy all our capacity?

runtime/kusama/src/lib.rs
```diff
@@ -956,6 +980,126 @@ impl pallet_proxy::Config for Runtime {
...
+
+impl parachains_scheduler::Config for Runtime {}
+
+impl parachains_initializer::Config for Runtime {
+	type Randomness = pallet_babe::RandomnessFromOneEpochAgo<Runtime>;
+	type ForceOrigin = EnsureRoot<AccountId>;
+}
+
+parameter_types! {
+	pub const ParaDeposit: Balance = 5 * DOLLARS;
```
rphmeier, line 1023: TODO

runtime/kusama/src/lib.rs
```diff
@@ -956,6 +980,126 @@ impl pallet_proxy::Config for Runtime {
...
+	type Event = Event;
+	type Currency = Balances;
+	type Registrar = Registrar;
+	type LeasePeriod = LeasePeriod;
+	type WeightInfo = slots::TestWeightInfo;
+}
+
+parameter_types! {
+	pub const KusamaLocation: MultiLocation = MultiLocation::Null;
+	pub const KusamaNetwork: NetworkId = NetworkId::Polkadot;
```
rphmeier, line 1055: TODO

runtime/kusama/src/lib.rs
```diff
@@ -1036,15 +1198,54 @@ impl pallet_babe::migrations::BabePalletPrefix for Runtime {
...
+			validation_upgrade_frequency: 1 * DAYS,
+			validation_upgrade_delay: EPOCH_DURATION_IN_BLOCKS,
+			max_pov_size: 5 * 1024 * 1024,
+			max_downward_message_size: 50 * 1024,
+			preferred_dispatchable_upward_messages_step_weight: 0,
+			hrmp_max_parachain_outbound_channels: 10,
+			hrmp_max_parathread_outbound_channels: 0,
+			hrmp_open_request_ttl: 2,
+			hrmp_sender_deposit: deposit(4, 128),
+			hrmp_recipient_deposit: deposit(4, 128),
```
rphmeier, line 1221: TODO

runtime/kusama/src/lib.rs
```diff
@@ -956,6 +980,126 @@ impl pallet_proxy::Config for Runtime {
...
+
+type LocalOriginConverter = (
+	SovereignSignedViaLocation<LocationConverter, Origin>,
+	ChildParachainAsNative<parachains_origin::Origin, Origin>,
+	SignedAccountId32AsNative<KusamaNetwork, Origin>,
+	ChildSystemParachainAsSuperuser<ParaId, Origin>,
+);
+
+parameter_types! {
+	pub const BaseXcmWeight: Weight = 100_000;
```
rphmeier, line 1084: 1_000_000_000

runtime/kusama/src/lib.rs
```diff
@@ -956,6 +980,126 @@ impl pallet_proxy::Config for Runtime {
...
+pub struct XcmConfig;
+impl xcm_executor::Config for XcmConfig {
+	type Call = Call;
+	type XcmSender = xcm_sender::RelayChainXcmSender<Runtime>;
+	type AssetTransactor = LocalAssetTransactor;
+	type OriginConverter = LocalOriginConverter;
+	type IsReserve = ();
+	type IsTeleporter = ();
+	type LocationInverter = LocationInverter<Ancestry>;
+	type Barrier = ();
```
rphmeier, line 1097: TODO: set

rphmeier: `AllowPaidFromParachains`, probably

runtime/westend/src/lib.rs
```diff
@@ -816,12 +941,54 @@ impl pallet_babe::migrations::BabePalletPrefix for Runtime {
...
+			n_delay_tranches: 40,
```
burdges, line 984: I've never added enough to https://github.com/w3f/research-internal/issues/515 to explain this number, but it's sensitive.. too high and too many no shows bog us down.. too low and we're not secure..

I worked out `n_active_validators / n_delay_tranches = 2.25` once, although I should redo that computation, so `n_delay_tranches = 89` sounds good for kusama right now.

We'll want a separate number which is the number of tranches after which we give up and declare the candidate shitty but not yet invalid, and fork the relay chain to dump it.

runtime/kusama/src/lib.rs
```diff
@@ -1036,15 +1198,54 @@ impl pallet_babe::migrations::BabePalletPrefix for Runtime {
...
+			thread_availability_period: 1 * MINUTES,
+			scheduling_lookahead: 1,
+			max_validators_per_core: Some(5),
+			max_validators: Some(200),
+			dispute_period: 6,
+			dispute_post_conclusion_acceptance_period: 1 * HOURS,
+			dispute_max_spam_slots: 2,
+			dispute_conclusion_by_time_out_period: 1 * HOURS,
+			no_show_slots: 2,
+			n_delay_tranches: 40,
```
burdges, line 1241: `n_delay_tranches = 89` here.  https://github.com/paritytech/polkadot/pull/2854/files#r610163186

runtime/westend/src/lib.rs
```diff
@@ -816,12 +941,54 @@ impl pallet_babe::migrations::BabePalletPrefix for Runtime {
...
+			zeroth_delay_tranche_width: 0,
+			needed_approvals: 15,
+			relay_vrf_modulo_samples: 1,
```
burdges, line 987: We want roughly `needed_approvals * n_parachains = (relay_vrf_modulo_samples + 1 / n_delay_tranches) * n_active_validators = relay_vrf_modulo_samples * n_active_validators + 2.25`

Although the `+` could disappear if we changed `zeroth_delay_tranche_width = 0` to mean delay VRF criteria start in tranche 1, but maybe not worth the trouble.

runtime/westend/src/lib.rs
rphmeier: I'd aim for high rather than low as:
  * vast majority of nodes are not malicious, better safe than sorry
  * our test networks perform quite well with even hundreds of validators on approval-checking

runtime/kusama/src/lib.rs
```diff
@@ -956,6 +983,126 @@ impl pallet_proxy::Config for Runtime {
...
+impl paras_registrar::Config for Runtime {
+	type Event = Event;
+	type Origin = Origin;
+	type Currency = Balances;
+	type OnSwap = Slots;
+	type ParaDeposit = ParaDeposit;
+	type DataDepositPerByte = DataDepositPerByte;
+	type MaxCodeSize = MaxCodeSize;
+	type MaxHeadSize = MaxHeadSize;
+	type WeightInfo = paras_registrar::TestWeightInfo;
```
shawntabrizi, line 1040: TODO update

runtime/kusama/src/lib.rs
```diff
@@ -956,6 +983,126 @@ impl pallet_proxy::Config for Runtime {
...
+	// 6 weeks
+	pub const LeasePeriod: BlockNumber = 6 * 7 * DAYS;
+}
+
+impl slots::Config for Runtime {
+	type Event = Event;
+	type Currency = Balances;
+	type Registrar = Registrar;
+	type LeasePeriod = LeasePeriod;
+	type WeightInfo = slots::TestWeightInfo;
```
shawntabrizi, line 1053: TODO update

runtime/westend/src/lib.rs
```diff
@@ -741,6 +735,128 @@ impl pallet_proxy::Config for Runtime {
...
+impl paras_registrar::Config for Runtime {
+	type Event = Event;
+	type Origin = Origin;
+	type Currency = Balances;
+	type OnSwap = Slots;
+	type ParaDeposit = ParaDeposit;
+	type DataDepositPerByte = DataDepositPerByte;
+	type MaxCodeSize = MaxCodeSize;
+	type MaxHeadSize = MaxHeadSize;
+	type WeightInfo = paras_registrar::TestWeightInfo;
```
shawntabrizi, line 795: TODO: update

runtime/westend/src/lib.rs
```diff
@@ -741,6 +735,128 @@ impl pallet_proxy::Config for Runtime {
...
+parameter_types! {
+	pub const LeasePeriod: BlockNumber = 28 * DAYS;
+}
+
+impl slots::Config for Runtime {
+	type Event = Event;
+	type Currency = Balances;
+	type Registrar = Registrar;
+	type LeasePeriod = LeasePeriod;
+	type WeightInfo = slots::TestWeightInfo;
```
shawntabrizi, line 807: TODO: Update

runtime/kusama/src/lib.rs
rphmeier: > Or is queue_size in bytes and this means we can only store 1 maximal sized message? 

This is it

> we likely don't want a single message to occupy all our capacity?

why not? I figure if a parachain wants to use all their capacity in a single message then they ought to be able to

shawntabrizi:
For some reason the benchmarking bot isnt working?

runtime/kusama/src/lib.rs
```diff
@@ -956,6 +983,126 @@ impl pallet_proxy::Config for Runtime {
...
+	type ParaDeposit = ParaDeposit;
+	type DataDepositPerByte = DataDepositPerByte;
+	type MaxCodeSize = MaxCodeSize;
+	type MaxHeadSize = MaxHeadSize;
+	type WeightInfo = paras_registrar::TestWeightInfo;
+}
+
+parameter_types! {
+	// 6 weeks
+	pub const LeasePeriod: BlockNumber = 6 * 7 * DAYS;
```
shawntabrizi, line 1045: Use weeks

runtime/westend/src/lib.rs
burdges: I'm fine with high initially, especially before I've written anything concrete.  

If we add 5 validators per tranche, aka per no show, then an adversary with 20% could censor many candidates.  We also risk that too many existing validators run on insufficient hardware or connections, and could no show by accedent, but this requires other techniques I guess.

runtime/westend/src/lib.rs
rphmeier: I thought we'd settled on 2, but I could see bumping it up to 3

MrBond34 (approved):

runtime/kusama/src/lib.rs
```diff
@@ -27,16 +27,41 @@ use sp_core::u32_trait::{_1, _2, _3, _5};
...
+use runtime_parachains::inclusion as parachains_inclusion;
+use runtime_parachains::paras_inherent as parachains_paras_inherent;
+use runtime_parachains::initializer as parachains_initializer;
+use runtime_parachains::session_info as parachains_session_info;
+use runtime_parachains::paras as parachains_paras;
+use runtime_parachains::dmp as parachains_dmp;
+use runtime_parachains::ump as parachains_ump;
+use runtime_parachains::hrmp as parachains_hrmp;
+use runtime_parachains::scheduler as parachains_scheduler;
+use runtime_parachains::reward_points as parachains_reward_points;
```
gavofyork, line 53: could also just be `use runtime_parachains as parachains;`. it's fewer lines and you get the slightly more semantic `parachains::origin` rather than `parachains_origin`.

runtime/kusama/src/lib.rs
```diff
@@ -117,11 +142,14 @@ pub fn native_version() -> NativeVersion {
 	}
 }
 
-/// Avoid processing transactions from slots and parachain registrar.
+/// Avoid processing transactions from slots and parachain registrar except by root.
 pub struct BaseFilter;
 impl Filter<Call> for BaseFilter {
-	fn filter(_: &Call) -> bool {
-		true
+	fn filter(c: &Call) -> bool {
```
gavofyork, line 148: could use `match_type!` macro.

runtime/kusama/src/lib.rs
```diff
@@ -956,6 +984,150 @@ impl pallet_proxy::Config for Runtime {
...
+
+impl parachains_scheduler::Config for Runtime {}
+
+impl parachains_initializer::Config for Runtime {
+	type Randomness = pallet_babe::RandomnessFromOneEpochAgo<Runtime>;
+	type ForceOrigin = EnsureRoot<AccountId>;
+}
+
+parameter_types! {
+	pub const ParaDeposit: Balance = deposit(10, MAX_CODE_SIZE + MAX_HEAD_SIZE);
```
gavofyork, line 1032: should also account for the fact that messages can be sent to a parathread over DMP and HRMP.

runtime/kusama/src/lib.rs
```diff
@@ -956,6 +984,150 @@ impl pallet_proxy::Config for Runtime {
...
+	type WeightInfo = weights::runtime_common_slots::WeightInfo<Runtime>;
+}
+
+parameter_types! {
+	pub const KusamaLocation: MultiLocation = MultiLocation::Null;
+	pub const KusamaNetwork: NetworkId = NetworkId::Kusama;
+	pub const Ancestry: MultiLocation = MultiLocation::Null;
+}
+
+pub type LocationConverter = (
```
gavofyork, line 1068:
```suggestion
pub type LocationIntoAccountId = (
```

runtime/kusama/src/lib.rs
```diff
@@ -956,6 +984,150 @@ impl pallet_proxy::Config for Runtime {
...
+);
+
+pub type LocalAssetTransactor =
+	XcmCurrencyAdapter<
+		// Use this currency:
+		Balances,
+		// Use this currency when it is a fungible asset matching the given location or name:
+		IsConcrete<KusamaLocation>,
+		// We can convert the MultiLocations with our converter above:
+		LocationConverter,
```
gavofyork, line 1080:
```suggestion
		LocationIntoAccountId,
```

runtime/kusama/src/lib.rs
```diff
@@ -956,6 +984,150 @@ impl pallet_proxy::Config for Runtime {
...
+		// Our chain's account ID type (we can't get away without mentioning it explicitly):
+		AccountId,
+	>;
+
+type LocalOriginConverter = (
+	SovereignSignedViaLocation<LocationConverter, Origin>,
```
gavofyork, line 1086:
```suggestion
	SovereignSignedViaLocation<LocationIntoAccountId, Origin>,
```

runtime/kusama/src/lib.rs
```diff
@@ -956,6 +984,150 @@ impl pallet_proxy::Config for Runtime {
...
+	type Call = Call;
+	type XcmSender = XcmRouter;
+	type AssetTransactor = LocalAssetTransactor;
+	type OriginConverter = LocalOriginConverter;
+	type IsReserve = ();
+	type IsTeleporter = ();
+	type LocationInverter = LocationInverter<Ancestry>;
+	type Barrier = Barrier;
+	type Weigher = FixedWeightBounds<BaseXcmWeight, Call>;
+	type Trader = FixedRateOfConcreteFungible<KusamaFee>;
```
gavofyork, line 1127:
```suggestion
	type Trader = UsingComponents<WeightToFee, RocLocation, AccountId, Balances, ToAuthor<Runtime>>;
```

gavofyork (approved):
Aside from the comments appears ok.

node/service/src/chain_spec.rs
```diff
@@ -449,7 +449,6 @@ fn westend_staging_testnet_config_genesis(wasm_binary: &[u8]) -> westend::Genesi
 		pallet_sudo: westend::SudoConfig {
 			key: endowed_accounts[0].clone(),
 		},
-		//TODO: Don't use default.
```
gavofyork: Using the default here seems as good as anything else.

runtime/kusama/src/constants.rs
```diff
@@ -18,30 +18,22 @@
 pub mod currency {
 	use primitives::v0::Balance;
 
-	pub const DOTS: Balance = 1_000_000_000_000;
-	pub const DOLLARS: Balance = DOTS / 300;
-	pub const CENTS: Balance = DOLLARS / 100;
+	pub const UNITS: Balance = 1_000_000_000_000;
```
gavofyork, line 21: Removed tehe questionable `DOLLARS` to leave the (slightly more vague) `CENTS` and `GRAND`.

shawntabrizi: Not everywhere though? (still in Polkadot, Rococo, and Test-Runtime)

runtime/kusama/src/lib.rs
```diff
@@ -962,6 +987,168 @@ impl pallet_proxy::Config for Runtime {
...
+/// The means that we convert an the XCM message origin location into a local dispatch origin.
+type LocalOriginConverter = (
+	// A `Signed` origin of the sovereign account that the original location controls.
+	SovereignSignedViaLocation<SovereignAccountOf, Origin>,
+	// A child parachain, natively expressed, has the `Parachain` origin.
+	ChildParachainAsNative<parachains_origin::Origin, Origin>,
+	// The AccountId32 location type can be expressed natively as a `Signed` origin.
+	SignedAccountId32AsNative<KusamaNetwork, Origin>,
+	// A system child parachain, expressed as a Superuser, converts to the `Root` origin.
+	ChildSystemParachainAsSuperuser<ParaId, Origin>,
```
shawntabrizi, line 1111: Do we want this right now?

We seems we will not use it with statemint, so imo better to exclude it for now.

runtime/polkadot/src/constants.rs
```diff
@@ -18,8 +18,8 @@
 pub mod currency {
 	use primitives::v0::Balance;
 
-	pub const DOTS: Balance = 1_000_000_000_000;
-	pub const DOLLARS: Balance = DOTS / 100;       // 10_000_000_000
+	pub const UNITS: Balance = 1_000_000_000_000;
+	pub const DOLLARS: Balance = UNITS / 100;       // 10_000_000_000
```
shawntabrizi, line 22:
```suggestion
	pub const UNITS: Balance = 10_000_000_000;
	pub const DOLLARS: Balance = UNITS;       // 10_000_000_000
```

Wouldn't something like this make sense because of redenomination?

runtime/westend/src/lib.rs
```diff
@@ -690,6 +738,149 @@ impl pallet_proxy::Config for Runtime {
...
+		LocationConverter,
+		// Our chain's account ID type (we can't get away without mentioning it explicitly):
+		AccountId,
+	>;
+
+type LocalOriginConverter = (
+	SovereignSignedViaLocation<LocationConverter, Origin>,
+	ChildParachainAsNative<parachains_origin::Origin, Origin>,
+	SignedAccountId32AsNative<WestendNetwork, Origin>,
+	ChildSystemParachainAsSuperuser<ParaId, Origin>,
```
shawntabrizi, line 845: also can be removed here for now

runtime/westend/src/lib.rs
```diff
@@ -690,6 +738,149 @@ impl pallet_proxy::Config for Runtime {
...
+);
+
+parameter_types! {
+	pub const BaseXcmWeight: Weight = 100_000;
```
shawntabrizi, line 849: intentionally lower than kusama?

shawntabrizi (approved):
i left comments

runtime/kusama/src/lib.rs
gavofyork: Statemint won't be a system parachain

runtime/westend/src/lib.rs
gavofyork: i think it's best to leave in - the whole point of system parachains is that they can do this.

runtime/westend/src/lib.rs
gavofyork: It should be a bit lower since there's no need to be so conservative on what is purely a testnet. Perhaps this value is a bit too low though.

gavofyork:
```suggestion
	pub const BaseXcmWeight: Weight = 10_000_000;
```

PR #2856, opened 2021-04-08
bump spec versions across the board

rphmeier:
All our builds are failing until this is done.

@s3krit not sure if this is the way we're 'supposed' to do it but it'd be nice to get this in and then follow up to adjust to how you'd want it to be


ordian (approved):

bkchr (approved):

bkchr:
bot merge

PR #2860, opened 2021-04-08
no backing job early exit if not validator

rphmeier:
with the MaxValidators parameter set, this meant that only validators participating in parachain consensus would be able to author relay-chain blocks containing backed candidates


ordian (approved):
would be cool to have a test, but that's nonblocking

rphmeier:
bot merge

PR #2862, opened 2021-04-08
Change Auctions to Eight Lease Periods Per Slot

shawntabrizi:
This PR changes the Auctions and Crowdloan system to use a `SlotRange` which has 8 lease periods per slot.

Nothing philosophically changes in the behavior of these pallets, just more options.


rphmeier (approved):

emostov:
Is there anyway I can either derive or query for `SlotRange::SLOT_RANGE_COUNT` offchain?  

I was thinking of using something like the equivalent of `Winning::<T>::get(offset).len()`, but if no bids have been made in the auction then it is `None`.

jacogr:
I would suggest exposing the slot range as a constant - that way it is accessible externally (and even better, can be detected)

shawntabrizi:
yup done :)

rphmeier:
bot merge

jacogr (approved):

shawntabrizi:
bot merge

shawntabrizi:
bot merge

PR #2863, opened 2021-04-09
set groups correctly even if not validator

rphmeier:
shows what I get for being hasty last time. i added a test this time, which would have caught this issue last time (#2860).

### DIFF
--- a/node/core/backing/src/lib.rs
+++ b/node/core/backing/src/lib.rs
@@ -1186,18 +1186,17 @@ impl util::JobTrait for CandidateBackingJob {
 			let n_cores = cores.len();
 
 			let mut assignment = None;
-			if let Some(ref validator_index) = validator.as_ref().map(|v| v.index()) {
-				for (idx, core) in cores.into_iter().enumerate() {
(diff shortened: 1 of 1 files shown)


node/core/backing/src/lib.rs
```diff
@@ -1186,18 +1186,17 @@ impl util::JobTrait for CandidateBackingJob {
...
-				for (idx, core) in cores.into_iter().enumerate() {
-					// Ignore prospective assignments on occupied cores for the time being.
-					if let CoreState::Scheduled(scheduled) = core {
-						let core_index = CoreIndex(idx as _);
-						let group_index = group_rotation_info.group_for_core(core_index, n_cores);
-						if let Some(g) = validator_groups.get(group_index.0 as usize) {
-							if g.contains(&validator_index) {
-								assignment = Some((scheduled.para_id, scheduled.collator));
-							}
-							groups.insert(scheduled.para_id, g.clone());
```
rphmeier, line 1199: This line is crucial and was not being hit when not a validator.

ordian (approved):

Issue #2841, opened 2021-04-07
Tracking issue for XCM Revamp


gavofyork:
Follow-up issue: #4118

### Priorities
- [x] Harden the `execute_xcm` function against recursion-bombs (https://github.com/paritytech/polkadot/pull/3555, https://github.com/paritytech/polkadot/pull/3586)
- [x] Type-safe wildcard management in `MultiAsset` (https://github.com/paritytech/polkadot/issues/3168).
- [x] Implement proper weight pricing for the queuing and dispatch - it should integrate into the chain's usual transaction weight pricing. (https://github.com/paritytech/polkadot/pull/2896)
- [x] Don't just burn fees - send them somewhere, and test that we get them! (https://github.com/paritytech/polkadot/pull/2896)
- [x] Avoid using `burn` & `mint` in teleport - instead transfer into and out of a special reserved "teleported" account. (https://github.com/paritytech/polkadot/pull/3007)
- [x] Introduce a `reserve_asset_transfer` extrinsic and execution filter enabling it (https://github.com/paritytech/polkadot/pull/3010)
- [x] Add a `register_id` function to paras_registrar (https://github.com/paritytech/polkadot/pull/3008)
- [x] #3168: Typesafe `MultiAsset` wildcards. (https://github.com/paritytech/polkadot/pull/2815)
- [x] #3547: `DepositAsset` and `DepositReserveAsset` should have an extra parameter limiting the number of distinct assets to deposit. (https://github.com/paritytech/polkadot/pull/2815)
- [x] Introduce internal `Vec<Order>` field for `BuyExecution`. (https://github.com/paritytech/polkadot/pull/2815)
- [x] Update XCM format document to reflect new format (https://github.com/paritytech/xcm-format/issues/23).

### Substantial features 
- [x] General `ResponseMap` of `(QueryId, Origin)` -> `EitherOr<Dispatch, Option<Response>>` which allows free execution of the message and then either stores the result or calls a dispatch - in this case it could go overweight be up to the weight of whatever dispatch is in there. It should be added on and checked before executing any further XCMs. (https://github.com/paritytech/polkadot/pull/3629)
- [x] Prepaid outcome reporting mechanism for success/failure. (https://github.com/paritytech/polkadot/pull/3629)
- [x] New instruction `RefundSurplusWeight` which trades surplius weight in and places it back in holding. (https://github.com/paritytech/polkadot/pull/3629)
[...]

PR #2865, opened 2021-04-09
XCM Revamp Continued

gavofyork:
Tracking issue: https://github.com/paritytech/polkadot/issues/2841

This batch of work is mostly to allow governance VMP messages.

- Add `pallet-xcm`, this is a pallet which contains dispatchables to send and execute XCM messages. Also contains a MultiLocation `Origin` and a corresponding `EnsureOrigin` implementation to recognise it as well as a body-majority.
- Hardening of UMP message pocessing:
  - Alter `process_upward_message` in order to take the acceptable amount of weight to spend processing the message, and return an Option to indicate whether the weight limit was reached.
  - Limit the amount of weight that can be processed for UMP dispatch and provide a trapdoor for the first message to help ensure it keeps processing in unexpectedly heavy situations.
- Various refactorings.
- Add `XcmPallet` to Rococo runtime.

TODO in another PR:
- Add a collective pallet to Rococo runtime.


runtime/parachains/src/ump.rs
```diff
@@ -148,6 +144,9 @@ impl fmt::Debug for AcceptanceCheckErr {
 pub trait Config: frame_system::Config + configuration::Config {
 	/// A place where all received upward messages are funneled.
 	type UmpSink: UmpSink;
+
+	/// The factor by which the weight limit it multiplied for the first UMP message to execute with.
+	type FirstMessageFactorPercent: Get<Weight>;
```
shawntabrizi, line 149: Why not use perbill?

runtime/parachains/src/ump.rs
```diff
@@ -337,12 +336,21 @@ impl<T: Config> Module<T> {
 				// if so - bail.
 				break;
 			}
+			let max_weight = if used_weight_so_far == 0 {
+				// we increase the amount of weight that we're allowed to use on the first message to try to prevent
+				// the possibility of blockage of the queue.
```
shawntabrizi, line 341: this scenario is not quite clear to me yet

runtime/parachains/src/ump.rs
```diff
@@ -572,10 +580,10 @@ pub(crate) mod mock_sink {
 						}
 					};
 					assert_eq!(expected_origin, actual_origin);
-					assert_eq!(expected_msg, actual_msg);
-					mock_weight
+					assert_eq!(expected_msg, &actual_msg[..]);
+					Some(mock_weight)
 				}
-				None => 0,
+				None => None,
```
shawntabrizi, line 586: maybe a useless nit, but this could be `.map()` rather than match now

xcm/pallet-xcm/src/lib.rs
```diff
@@ -0,0 +1,185 @@
...
+		/// An event is deposited indicating whether `msg` could be executed completely or only
+		/// partially.
+		///
+		/// No more than `max_weight` will be used in its attempted execution. If this is less than the
+		/// maximum amount of weight that the message could take to be executed, then no execution
+		/// attempt will be made.
+		///
+		/// NOTE: A successful return to this does *not* imply that the `msg` was executed successfully
+		/// to completion; only that *some* of it was executed.
+		#[pallet::weight(1_000 + max_weight)]
```
shawntabrizi, line 103:
```suggestion
		#[pallet::weight(1_000.saturating_add(max_weight))]
```

My understanding is this calculation can still be triggered by an external user who does not match this origin in the transaction queue.

xcm/pallet-xcm/src/lib.rs
```diff
@@ -0,0 +1,185 @@
...
+					XcmError::CannotReachDestination(..) => Error::<T>::Unreachable,
+					_ => Error::<T>::SendFailure,
+				})?;
+			Ok(())
+		}
+
+		/// Execute an XCM message from a local, signed, origin.
+		///
+		/// Returns `DispatchError` if execution was not possible. Otherwise returns the amount of
+		/// weight that was used in its attempted execution.
```
kianenigma, line 92:
```suggestion
		/// Returns `DispatchError` if execution was not possible. Otherwise returns the amount of
		/// weight that was used in its attempted execution.
```

How does it return this with `DispatchResult`?

runtime/parachains/src/ump.rs
gavofyork: No huge reason but didn't see a huge need to pull in another type for a simple parameter.

xcm/pallet-xcm/src/lib.rs
```diff
@@ -0,0 +1,185 @@
...
+		SendFailure,
+	}
+
+	#[pallet::hooks]
+	impl<T: Config> Hooks<BlockNumberFor<T>> for Pallet<T> {}
+
+	#[pallet::call]
+	impl<T: Config> Pallet<T> {
+		#[pallet::weight(1_000)]
+		fn send(origin: OriginFor<T>, dest: MultiLocation, message: Xcm<()>) -> DispatchResult {
```
kianenigma, line 79: umm.. so I am pretty new to this stuff, but don't we eventually want to send an xcm `Transact` to trigger an update in a parachain? then, wouldn't we need an actual `Call` rather than `()`?

runtime/parachains/src/ump.rs
gavofyork: The original UMP dispatcher guaraunteed one dispatch per block minimum with the weight limitation being a soft cap. This kept things running the case of a single heavy message. Now, we don't quite guarantee that, but the soft-cap gets lifted to this hard-cap for the first message per block to try to mitigate any blockage.

xcm/pallet-xcm/src/lib.rs
gavofyork: it doesn't - old docs.

xcm/pallet-xcm/src/lib.rs
gavofyork: The type there doesn't bound what `Transact` can represent, only what we expect to decode into locally. Since this is being sent to another chain, we don't need to bound it with anything sensible since we know it won't be interpreted locally. If we did bound to our `Call` it wouldn't do anything bad, but wouldn't make much sense either unless we happened to know that the remote chain had the same `Call` type as the local.

xcm/src/v0/traits.rs
```diff
@@ -121,21 +122,38 @@ impl Outcome {
...
+	type Call = C;
 	fn execute_xcm(_origin: MultiLocation, _message: Xcm<C>, _weight_limit: Weight) -> Outcome {
 		Outcome::Error(Error::Unimplemented)
 	}
 }
 
+/// Utility for sending an XCM message.
+///
+/// These can be amalgamted in tuples to form sophisticated routing systems.
 pub trait SendXcm {
```
kianenigma, line 139: I was reading this few hours ago and this was a fundamental part of the code without docs, thanks for filling 👍

xcm/pallet-xcm/src/lib.rs
kianenigma: So only for decoding -- gotcha.

shawntabrizi (approved):

kianenigma (approved):

PR #2872, opened 2021-04-09
Rococo allows ROC to be teleported from Tick, Trick and Track

gavofyork:
Also allows unpaid messages from them.


kianenigma (approved):

kianenigma (commented):
Could use more doc and nicer imports, but I guess that's not the criteria for now.

PR #2880, opened 2021-04-13
Burnin for #8609

tomaka:
https://github.com/paritytech/substrate/pull/8609


tomaka:
As predicted, connectivity dropped and number of incoming connection errors spiked, then everything went back to normal after all nodes have updated their DHT entries:

image

tomaka:
Around 24 hours later, we still have around 5 incoming errors/sec, but everything else is back to normal.

Here's the libp2p task CPU consumption:

image

It's not clear how much CPU overhead there is, but it looks negligible.

PR #2885, opened 2021-04-14
CI: optimizations

TriplEight:
Publish fewer binaries and images, no need to push them all to the official sources.
- change the bucket for binaries from master and schedules
- change container registry for the images from master and schedules

Optimize the binary build.
- instead of having a parallel job that just builds a binary, we reuse the binary built by `cargo test --release`
  I'm pretty sure it should be similar, however, binaries from builds are a bit lighter (185M) than the ones from tests. 
  As an alternative, running a binary build after the test adds a ~ 1:40 overhead.
  @bkchr how do you think?


.gitlab-ci.yml
```diff
@@ -340,9 +319,8 @@ publish-s3-release:                &publish-s3
     BUCKET:                        "releases.parity.io"
     PREFIX:                        "polkadot/${ARCH}-${DOCKER_OS}"
   rules:
+    # FIXME: decide where/when to publish these releases
     - if: $CI_PIPELINE_SOURCE == "schedule"
-    - if: $CI_COMMIT_REF_NAME == "master"
```
TriplEight, line 344: @s3krit wanted to draw your attention here. @lovelaced too, AFAIR you use the released binaries in some testnet deployments.
It's clear that publishing to releases.parity.io every master commit is a bit too much, schedules (nightly master run) do not make much sense either. @s3krit you mentioned you want to publish releases on your own as well.
Shall we publish i.e. just nightlies to some another bucket?

lovelaced: I don't think we really need to publish nightlies anywhere. If we need binaries to deploy on nodes now they go through the burn-in process which creates its own binaries (is that right @haikoschol ? or does the burn-in get the binaries through this process?) Either way they should stop going into the releases bucket.

s3krit: I think we *may as well* push nightly builds, although to a separate bucket. I'm building up the bucket + AWS AMI stuff right now - might if I commit to this branch with the new bucket name, etc?

haikoschol: > If we need binaries to deploy on nodes now they go through the burn-in process which creates its own binaries (is that right @haikoschol ? or does the burn-in get the binaries through this process?)

The burn-in automation usually downloads binaries from [CI job artifact links](https://gitlab.parity.io/parity/polkadot/-/jobs/888293/artifacts/raw/artifacts/polkadot).

The daily deployment to `canary-sentry-ew4-0` and `canary-validator-ew4-0` uses https://releases.parity.io/polkadot/x86_64-debian:stretch/master/polkadot

TriplEight:
The required `gitlab-test-linux-stable` status will be changed to `gitlab-test-build-linux-stable` after this PR.

rcny (approved):

PR #2886, opened 2021-04-14
Expose paras config in rococo runtime

JoshOrndorff:
The `paras` pallet has a genesis configuration that allows having parachains pre-registered at the time of relay chain genesis. 

Before this PR, this configuration was missed from the rococo runtime which means the genesis config couldn't be used in practice. This PR adds the necessary line to the runtime, and updates the `chain_spec.rs` file accordingly.

This change will be useful for parachain teams looking to test their parchains without waiting two sessions. This improvement can also be reflected in Polkadot Launch as described in https://github.com/paritytech/polkadot-launch/issues/66

cc @4meta5 @joelamouche

### DIFF
--- a/runtime/rococo/src/lib.rs
+++ b/runtime/rococo/src/lib.rs
@@ -214,7 +214,7 @@ construct_runtime! {
 		Inclusion: parachains_inclusion::{Pallet, Call, Storage, Event<T>},
 		ParasInherent: parachains_paras_inherent::{Pallet, Call, Storage, Inherent},
 		Scheduler: parachains_scheduler::{Pallet, Call, Storage},
-		Paras: parachains_paras::{Pallet, Call, Storage, Event},
+		Paras: parachains_paras::{Pallet, Call, Storage, Event, Config<T>},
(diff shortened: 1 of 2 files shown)


runtime/rococo/src/lib.rs
```diff
@@ -110,7 +110,7 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
 	spec_name: create_runtime_str!("rococo"),
 	impl_name: create_runtime_str!("parity-rococo-v1.5"),
 	authoring_version: 0,
-	spec_version: 231,
+	spec_version: 232,
```
bkchr, line 113:
```suggestion
	spec_version: 231,
```

bkchr (approved):

runtime/rococo/src/lib.rs
JoshOrndorff: Should I bump impl version instead? Or just not bump anything?

bkchr: Nothing changed, nothing needs to be bumped

PR #2897, opened 2021-04-16
Infrastructure improvements

eskimor:
This PR does mostly two things (which I need right now and will be generally useful):

1. I moved the runtime module from availability-distribution  to subsystem-util, so it can be used by other subsystems as well.
2. We now expose on `PeerConnected` events if a peer is a validator node. We have been already gathering this information for all peers that connect, we might as well make it easily accessible.

The latter makes it easy for subsystems to treat validator nodes differently than normal peers, e.g. give them better priority and similar things. Now that we just connect to all validators by means of `gossip-support`, this is also needed for easily sending a message to a particular validator via `SendValidationMessage` (which works with `PeerId`s).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6096,6 +6096,7 @@ dependencies = [
  "futures 0.3.13",
  "futures-timer 3.0.2",
  "log",
+ "lru",
  "metered-channel",
  "parity-scale-codec",
  "parking_lot 0.11.1",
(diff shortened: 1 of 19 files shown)


ordian (approved):
Thanks, looks good!

eskimor:
Thanks for such a quick review on a Friday night ;-)

PR #2900, opened 2021-04-19
Fix bench bot

kianenigma:
that whole `is_polkadot` was wrong, I looked around and it seems like we only have `is_xxx` for chains other than polkadot and `else`, it is polkadot. It makes sense given the name of the repo :P


bkchr (approved):

shawntabrizi (approved):

shawntabrizi:
bot merge

shawntabrizi:
bot merge

PR #2901, opened 2021-04-19
Disable logging in runtime when enabling `runtime-benchmarks`

athei:
Same PR as https://github.com/paritytech/substrate/pull/8638

This is different from the substrate PR in that here the logs are automatically disabled whenever the `runtime-benchmarks` feature is supplied. On substrate this is not possible because of tests which parse log output.


runtime/common/Cargo.toml
```diff
@@ -102,6 +102,7 @@ std = [
 	"xcm/std",
 ]
 runtime-benchmarks = [
+	"sp-api/disable-logging",
```
kianenigma, line 105: Don't we need to enable this in other places as well?

kianenigma (approved):

runtime/common/Cargo.toml
athei: The common runtime is a dependency of all other runtimes. AFAIK this is the only place where we need to at it. It does not seem to be applicable to bridge and xcm runtimes because those don't depend on `sp-api`. Maybe someone else can step in with more knowledge about those runtimes.

kianenigma: Sounds like we could have some kind of test to make sure we don't make this mistake again.

athei: You mean having logging enabled during benchmarking by accident?

kianenigma: yes, but it is probably a PITA

shawntabrizi (approved):

shawntabrizi:
Do we need to re-run benchmarks?

athei:
Every benchmark which was ran since https://github.com/paritytech/substrate/pull/8128 was merged is potentially flawed when it emits a substantial amount of log messages.

That said, for `pallet_contracts` only one benchmark was affected.

bkchr (commented):
It would be nice if we could try to enable the `disable-logging` feature for `sp-api` in the benchmarking crate, to not having the requirement to add this everywhere

shawntabrizi:
@bkchr can we do this by simply adding `sp-api` to the cargo.toml of the benchmarking stuff?

bkchr:
Yeah, this is what I wanted to express with my message xD

bkchr:
I tested it and it doesn't work...

Cargo feature resolution is just the best piece of software I know... :(

runtime/common/Cargo.toml
athei: I have no idea on how to construct such a test other than some benchmark which will look for regressions. And something like that would be really nice to have anyways. After each benchmarks of the contracts pallet I need to manually look through a ton of benchmarks and scan for regressions. Would be nice to have this automated.

athei:
> I tested it and it doesn't work...
> 
> Cargo feature resolution is just the best piece of software I know... :(

I can confirm that. So we just keep this PR the way it is.

athei:
Superseded by https://github.com/paritytech/substrate/pull/8655

PR #2908, opened 2021-04-20
Companion for substrate#8643

tomaka:
https://github.com/paritytech/substrate/pull/8643

This is technically speaking not a companion, because Polkaster master still compiles with the Substrate version in that PR.
However it makes sense to merge this PR at the same time as the Substrate one.

### DIFF
--- a/node/service/src/lib.rs
+++ b/node/service/src/lib.rs
@@ -714,6 +714,7 @@ pub fn new_full<RuntimeApi, Executor>(
 	let prometheus_registry = config.prometheus_registry().cloned();
 
 	let shared_voter_state = rpc_setup;
+	let auth_disc_publish_non_global_ips = config.network.allow_non_globals_in_dht;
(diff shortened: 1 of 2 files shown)


node/service/src/lib.rs
```diff
@@ -809,7 +810,7 @@ pub fn new_full<RuntimeApi, Executor>(
 	let spawner = task_manager.spawn_handle();
 	let active_leaves = active_leaves(&select_chain, &*client)?;
 
-	let authority_discovery_service = if role.is_authority() || is_collator.is_collator() {
+	let authority_discovery_service = if true || role.is_authority() || is_collator.is_collator() {
```
tomaka, line 813: Ooops

```suggestion
	let authority_discovery_service = if role.is_authority() || is_collator.is_collator() {
```

ordian (approved):

PR #2914, opened 2021-04-21
runtime: prepare Westend runtime for BEEFY

andresilva:


runtime/westend/src/lib.rs
```diff
@@ -91,7 +91,11 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
 	spec_name: create_runtime_str!("westend"),
 	impl_name: create_runtime_str!("parity-westend"),
 	authoring_version: 2,
-	spec_version: 51,
+	// On westend we might deploy multiple runtimes without actually doing a
+	// client release. The first two digits reflect the minor version of the
+	// upcoming client software while the last two can be incremented multiple
+	// times as we go through different runtimes.
+	spec_version: 3100,
```
tomusdrw, line 98: Can you elaborate on this? Why can't we simply increment `spec_version` now and with every "minor" version?

tomusdrw: Nvm, just saw the internal discussion. Long story short - this will allow us to keep spec versioning of multiple chains match release numbers.

andresilva: I reverted back to v51 and added a note to the wiki page explaining the process.

andresilva:
@jacogr I think there shouldn't be anything breaking here. We're deploying the same changes that we had already deployed on Rococo. But we want to be good coworkers and give you a heads up :)

jacogr:
All good. After the 50-that-was-supposed-to-have-beefy I seem to have already added anything >= 51 with Beefy keys. ...

jacogr (approved):

andresilva:
bot merge

PR #2916, opened 2021-04-21
runtime: remove mmr and beefy from westend runtime

andresilva:
Plan aborted, BEEFY and MMR won't go into Westend.


tomusdrw (approved):

tomusdrw (approved):

andresilva:
bot merge

PR #2917, opened 2021-04-21
Update Bridges Subtree

HCastano:
This PR is a follow up to #2515 where the `parity-bridges-common` repo was added a a Git
Subtree.

Unknown to us at the time, squash merging that PR ended up putting us in a situation
where we were unable up update the bridges subtree via `git subtree pull`. This is
because we were already working with a squashed version of the bridge history (since we
added the repo through `git subtree add ... --squash`).

Note that this time we should **not** squash merge this PR so that in the future we can
easily update the bridges subtree using `git subtree pull`


Cargo.toml
```diff
@@ -26,8 +26,10 @@ tempfile = "3.2.0"
 
 [workspace]
 members = [
-	"bridges/primitives/kusama",
-	"bridges/primitives/polkadot",
+    "bridges/primitives/chain-kusama",
```
ordian, line 29: nice attempt to sneak in spaces ;)

tomusdrw (approved):
LGTM! To anyone hitting the green button at the end: please make sure to **NOT squash-merge** this, but rather use regular merge as indicated in the description.

ordian:
how are we supposed to merge it?
image

HCastano:
> how are we supposed to merge it?
> image

So a squash-merge won't work for the above stated reasons, a rebase-merge won't work due to conflicts, and a regular merge is disabled.

So we either need someone with admin access to (temporarily) disable regular merges, or merge manually and push to `master`

PR #2918, opened 2021-04-21
Introduce Wococo testnet

andresilva:
Merge after #2916.

This is a temporary testnet meant as an initial test ground for BEEFY and MMR. This testnet will be using the exact same runtime as Rococo (its evil twin 😈) just with a different chain spec for bootstrapping (different bootnodes, different network protocol id, etc.). The changes in this PR are meant to minimize the amount of boilerplate needed to spin this up.


andresilva:
The list of bootnodes in `wococo.json` still needs to be populated before merging.

andresilva:
cc @jacogr Still not sure about naming, etc. But this is the way we're going to move forward (launch a clone of rococo).

node/service/src/chain_spec.rs
```diff
@@ -112,6 +112,11 @@ pub fn rococo_config() -> Result<PolkadotChainSpec, String> {
 	PolkadotChainSpec::from_json_bytes(&include_bytes!("../res/rococo.json")[..])
 }
 
+/// This is a temporary testnet that uses the same runtime as rococo.
+pub fn wococo_config() -> Result<PolkadotChainSpec, String> {
+	PolkadotChainSpec::from_json_bytes(&include_bytes!("../res/wococo.json")[..])
```
tomusdrw, line 117: What exactly did you change to have different genesis hash?

tomusdrw (commented):
Note it currently includes #2916. Looks good to me, but we need to make sure the genesis hash is different and there are no networking issues between networks.
Also: are we planning to deploy parachains to Wococo (CC @lovelaced? )?

node/service/src/chain_spec.rs
andresilva: It is built from a later version of substrate which by itself should generate a different WASM blob, and that get includes into storage as part of genesis. But if I diff `rococo.json` and `wococo.json` there are more storage entries that have different values.

```
❯ ./target/release/polkadot --chain wococo --tmp 
[...]
2021-04-21 21:34:57  💾 Database: RocksDb at /tmp/substrategSXl1c/chains/wococo/db
2021-04-21 21:34:57  ⛓  Native runtime: rococo-231 (parity-rococo-v1.5-0.tx0.au0)
2021-04-21 21:34:57  🔨 Initializing Genesis block/state (state: 0x3ea4…a8ba, header-hash: 0x89f5…694e)
```

```
❯ ./target/release/polkadot --chain rococo --tmp
[...]
2021-04-21 21:35:10  💾 Database: RocksDb at /tmp/substrateavbCDJ/chains/rococo_v1_5/db
2021-04-21 21:35:10  ⛓  Native runtime: rococo-231 (parity-rococo-v1.5-0.tx0.au0)
2021-04-21 21:35:10  🔨 Initializing Genesis block/state (state: 0x63b1…064d, header-hash: 0x1611…a215)
```

Additionally they also have a different `protocolId` in the chain spec, which gets used to separate the networks (and also a different token name but that's just metadata).

tomusdrw:
bot merge

PR #2926, opened 2021-04-23
Update Kusama Nominations to 24 

kianenigma:


wpank (approved):

burdges:
As an aside, we could likely make people pay for more nominations, perhaps when validators sweep rewards to their validators, but definitely when they create the nominations.  I've no idea if doing so warrants the extra complexity.

PR #2927, opened 2021-04-23
Send statements to own backing group first

eskimor:
Also introduced generalized error handling to make code sharing among subsystems easier and more ergonomic.


node/subsystem-util/src/error_handling.rs
```diff
@@ -0,0 +1,201 @@
...
+///		#[error("Error while accessing runtime information")]
+///		Runtime(#[from] #[source] runtime::NonFatal),
+/// }
+/// ```
+/// Then mostly use `Error` in functions, you may also use `NonFatal` and `Fatal` directly in
+/// functions that strictly only fail non fatal or fatal respectively, as `Fatal` and `NonFatal`
+/// can automatically converted into the above defined `Error`.
+/// ```
+#[derive(Debug, Error)]
+pub enum PolkaErr<E, F>
```
rphmeier, line 109: I would prefer to call this something else

node/network/statement-distribution/src/lib.rs
```diff
@@ -907,17 +924,39 @@ async fn circulate_statement(
...
+	let min_size = std::cmp::max(peers_to_send.len(), MIN_GOSSIP_PEERS);
+	// Make set full:
+	let needed_peers = min_size as i64 - priority_peers.len() as i64;
+	if needed_peers > 0 {
+		peers_to_send.truncate(needed_peers as usize);
+		priority_peers.append(&mut peers_to_send);
+	}
+	peers_to_send = priority_peers;
+	// We must not have duplicates:
+	debug_assert!(peers_to_send.len() == peers_to_send.clone().into_iter().collect::<HashSet<_>>().len());
```
rphmeier, line 959: add a `qed` here

node/network/statement-distribution/src/lib.rs
```diff
@@ -907,17 +924,39 @@ async fn circulate_statement(
...
+	// Avoid duplicates:
+	let priority_set: HashSet<&PeerId> = priority_peers.iter().collect();
+	peers_to_send.retain(|p| !priority_set.contains(p));
+
+	let mut peers_to_send =
+		util::choose_random_sqrt_subset(peers_to_send, MIN_GOSSIP_PEERS);
+	// We don't want to use less peers, than we would without any priority peers:
+	let min_size = std::cmp::max(peers_to_send.len(), MIN_GOSSIP_PEERS);
+	// Make set full:
+	let needed_peers = min_size as i64 - priority_peers.len() as i64;
```
rphmeier, line 952: Could `priority_peers.len()` not be greater than `min_size` here?

node/subsystem-util/src/error_handling.rs
eskimor: Haha, me too :-) . I wanted a short, not (too) overloaded, but still descriptive name. I'll try to come up with something better.

node/network/statement-distribution/src/lib.rs
```diff
@@ -1500,53 +1547,62 @@ impl StatementDistribution {
 	async fn run(
 		self,
 		mut ctx: impl SubsystemContext<Message = StatementDistributionMessage>,
-	) -> SubsystemResult<()> {
+	) -> std::result::Result<(), Fatal> {
 		let mut peers: HashMap<PeerId, PeerData> = HashMap::new();
+		let mut authorities: HashMap<AuthorityDiscoveryId, PeerId> = HashMap::new();
 		let mut active_heads: HashMap<Hash, ActiveHeadData> = HashMap::new();
+
+		let mut runtime = Runtime::new(self.keystore.clone());
```
rphmeier, line 1555: FWIW, I really don't like that this is called 'Runtime' since that has a separate meaning and this is really just a session info cache + keystore thing

node/network/statement-distribution/src/lib.rs
eskimor: It can (that's even tested in the test suite), but is fine. The if below will do the right thing in that case.

node/network/statement-distribution/src/lib.rs
```diff
@@ -1751,45 +1790,12 @@ impl StatementDistribution {
...
-								// Lacking this bookkeeping might make us behave funny, although
-								// not in any slashable way. But we shouldn't take down the node
-								// on what are likely spurious runtime API errors.
-								return Ok(false)
-							}
-						}
-					};
+					let session_index = runtime.get_session_index(ctx, relay_parent).await?;
+					let info = runtime.get_session_info_by_index(ctx, relay_parent, session_index).await?;
+					let session_info = &info.session_info;
```
rphmeier, line 1795: It's not clear to me where the `&mut authorities` is actually written into. I thought it'd be somewhere around here, but I only see the place where it's read from below.

node/network/statement-distribution/src/lib.rs
```diff
@@ -1447,15 +1488,21 @@ async fn handle_network_update(
 			peers.insert(peer, PeerData {
 				view: Default::default(),
 				view_knowledge: Default::default(),
+				maybe_authority: maybe_authority.clone(),
 			});
+			if let Some(authority) = maybe_authority {
+				authorities.insert(authority, peer);
```
rphmeier, line 1494: ahh, ok, it's here.

rphmeier (approved):

node/network/statement-distribution/src/lib.rs
eskimor: It is written on network update (`PeerConnected` events).

node/network/statement-distribution/src/lib.rs
rphmeier: I see - the cast to `i64` prevents underflow

node/network/statement-distribution/src/lib.rs
eskimor: (and disconnect of course).

node/network/statement-distribution/src/lib.rs
eskimor: I'll go with `RuntimeInfo` then, as it is meant as a convenient and efficient API for accessing runtime information. (E.g. also core status and such)

PR #2928, opened 2021-04-23
Add Dev Config for Rococo and Wococo

HCastano:
While working on the bridges integration I found it useful to have the option to run
single node networks for Rococo and Wococo instead of dealing with multiple nodes.

This PR adds `--chain=rococo-dev` and `--chain=wococo-dev` options to facilitate this.


node/service/src/chain_spec.rs
```diff
@@ -1560,6 +1569,49 @@ pub fn westend_development_config() -> Result<WestendChainSpec, String> {
...
+		None,
+		Some(DEFAULT_PROTOCOL_ID),
+		None,
+		Default::default(),
+	))
+}
+
+/// Wococo development config (single validator Alice)
+pub fn wococo_development_config() -> Result<RococoChainSpec, String> {
+	const WOCOCO_DEV_PROTOCOL_ID: &str = "woco";
```
HCastano, line 1595: I added this since having both a Rococo-Dev and Wococo-Dev node running locally would cause them to peer with each other - open to other suggestions for avoiding this

bkchr (approved):

PR #2929, opened 2021-04-23
Add validation_code_by_hash request in RuntimeApiRequest

thiolliere:
Based on https://github.com/paritytech/polkadot/pull/2907

There is currently no usage of this runtime API in the client, but for consistency and future usage we can add it.

PR should be in reviewable state, but better to open it once https://github.com/paritytech/polkadot/pull/2907 is merged (or if people request it I can rebase it on master)


thiolliere:
this is not required for now and might be outdated when we move all code offchain, thus I close

PR #2940, opened 2021-04-27
change junction parachain id from named field to unnamed field

thiolliere:
part of https://github.com/paritytech/polkadot/issues/2841

cumulus changes: https://github.com/paritytech/cumulus/pull/412

### DIFF
--- a/xcm/src/v0/junction.rs
+++ b/xcm/src/v0/junction.rs
@@ -94,9 +94,6 @@ pub enum Junction {
 	/// An indexed parachain belonging to and operated by the context.
 	///
 	/// Generally used when the context is a Polkadot Relay-chain.
-	///
-	/// There is also `Parachain` which can be used in tests to avoid the faffy `{ id: ... }` syntax. Production
-	/// code should use this.
 	Parachain(#[codec(compact)] u32),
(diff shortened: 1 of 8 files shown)


xcm/src/v0/junction.rs
```diff
@@ -94,11 +94,7 @@ pub enum Junction {
 	/// An indexed parachain belonging to and operated by the context.
 	///
 	/// Generally used when the context is a Polkadot Relay-chain.
-	///
-	/// There is also `Parachain` which can be used in tests to avoid the faffy `{ id: ... }` syntax. Production
-	/// code should use this.
```
thiolliere, line 99: I can't understand what this sentence means, I don't see `Parachain` types to be used in tests.
But I expect that it is not needed now that we fixed the syntax.

thiolliere:
CI fails because of implementers guide, should be fixed by https://github.com/paritytech/polkadot/pull/2941

xcm/src/v0/junction.rs
gavofyork: Yeah - that's a stale comment.

gavofyork (approved):

PR #2947, opened 2021-04-28
Disputes runtime

rphmeier:
This PR implements the core of the disputes runtime as per the implementers' guide.

The disputes module handles the import of conflicting statements, and provides slashing and rewarding interfaces for runtimes to fill out. In practice we will plug disputes into slashing/rewarding. 

A dispute is unconfirmed until 1/3+ of validators have participated. Validators have a limit on how many unconfirmed disputes they are allowed to participate in. This prevents spam from colluding malicious validators disputing each other over non-existent parachain blocks. Disputes that time out while unconfirmed should lead to a minor punishment of the participating validators, making spam of the chain expensive to perform.

Disputes concerning parablocks included in the local chain are always confirmed.

A dispute is considered concluded either when it times out unconfirmed or when concluding either for or against the candidate. It is mildly punishable to be against a valid candidate, and very punishable to be for an invalid candidate. If participants act on both sides of the dispute, it's possible for the dispute to conclude in both directions. However, acting on both sides of the dispute raises the likelihood of being slashed substantially, so there is no incentive to do so.

This adds the disputes to all runtimes, but does not combine it with any slashing or rewarding logic. That is, the node-side code for engaging with disputes is not implemented, and the result of any dispute activated on any network right now will be a no-op until the code is combined with the staking system. Thus there is no downside to integrating the module immediately.


runtime/parachains/src/disputes.rs
```diff
@@ -0,0 +1,726 @@
...
+
+decl_event! {
+	pub enum Event {
+		/// A dispute has been initiated. The boolean is true if the dispute is local. \[para_id\]
+		DisputeInitiated(ParaId, CandidateHash, bool),
+		/// A dispute has concluded for or against a candidate. The boolean is true if the candidate
+		/// is deemed valid (for) and false if invalid (against).
+		DisputeConcluded(ParaId, CandidateHash, bool),
+		/// A dispute has timed out due to insufficient participation.
+		DisputeTimedOut(ParaId, CandidateHash),
```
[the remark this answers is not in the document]
drahnr: Governance mode, you can not say for sure what should be done, so do nothing and let gov take over.

runtime/parachains/src/disputes.rs
```diff
@@ -0,0 +1,726 @@
...
+
+/// Punishment hooks for disputes.
+pub trait PunishValidators {
+	/// Punish a series of validators who were for an invalid parablock. This is expected to be a major
+	/// punishment.
+	fn punish_for_invalid(session: SessionIndex, validators: impl IntoIterator<Item=ValidatorIndex>);
+
+	/// Punish a series of validators who were against a valid parablock. This is expected to be a minor
+	/// punishment.
+	fn punish_against_valid(session: SessionIndex, validators: impl IntoIterator<Item=ValidatorIndex>);
```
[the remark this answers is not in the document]
burdges: No.  It merely causes escalation, which wastes everyone's time.  I believe the usual slashing system suffices here, including correlation magnification, so enough parallel spam disputes causes 100% slashes, but one lone wrong dispute sounds like 1% or so.  We'd expect the dispute originates from a bug so we'd refund it after forensics.  

Invalid parablocks being accepted otoh violates soundness directly.  We remove all stake supporting an invalid parablocks because our security rests upon gambler's ruin, so immediate 100% slashes that bypass the slash scaling system.

runtime/parachains/src/disputes.rs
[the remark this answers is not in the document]
burdges: We'll slash everyone who should've participated in availability a tiny amount.  Interestingly, if we worried about disputes spam preventing real disputes, then we'll want correlated timeouts to increase up to 100%, but..  

Afaik, If we're confident the disputes spam blocking cannot obstruct real disputes because either honest parties votes in the spam dispute, which slashes someone adversarial, and slashes them lots due to the correlated slashing on bad disputes. 

We'd ideally spell this reasoning out more clearly I guess, but I felt Rob's spam design aligns nicely.

roadmap/implementers-guide/src/runtime/disputes.md
```diff
@@ -42,7 +42,7 @@ Included: double_map (SessionIndex, CandidateHash) -> Option<BlockNumber>,
 // fewer than `byzantine_threshold + 1` validators.
 //
 // The i'th entry of the vector corresponds to the i'th validator in the session.
```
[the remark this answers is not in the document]
thiolliere: This is just a storage detail.

in decl_storage if one use `Option<something>` then the API will return None when no value is stored in the trie.
if one use `something` directly then API will return the default value of the type when no value is stored in the trie.

from the code I can see that no value stored in trie is equivalent to a vec with all validator having `0`:
```rust
			let mut spam_slots: Vec<u32> = SpamSlots::<T>::get(&set.session)
				.unwrap_or_else(|| vec![0; n_validators]);
```

runtime/parachains/src/disputes.rs
```diff
@@ -0,0 +1,726 @@
...
+	}
+}
+
+decl_event! {
+	pub enum Event {
+		/// A dispute has been initiated. The boolean is true if the dispute is local. \[para_id\]
+		DisputeInitiated(ParaId, CandidateHash, bool),
+		/// A dispute has concluded for or against a candidate. The boolean is true if the candidate
+		/// is deemed valid (for) and false if invalid (against).
+		DisputeConcluded(ParaId, CandidateHash, bool),
```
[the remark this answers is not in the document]
thiolliere: yes I updated

runtime/parachains/src/disputes.rs
```diff
@@ -0,0 +1,726 @@
...
+			slash_against,
+			slash_for,
+			new_participants: self.new_participants,
+			new_flags: post_flags - pre_flags,
+		}
+	}
+}
+
+impl<T: Config> Module<T> {
+	/// Called by the iniitalizer to initialize the disputes module.
```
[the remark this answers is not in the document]
thiolliere: fixed

runtime/parachains/src/disputes.rs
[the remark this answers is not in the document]
rphmeier: If a dispute times out, it means that the nodes are unable to participate and recover availability. A dispute initiation is a claim that the data is available, so we apply a minor punishment to all nodes which participated in the dispute.

runtime/parachains/src/disputes.rs
```diff
@@ -0,0 +1,752 @@
...
+			let submitted = targets.len();
+			targets.dedup();
+
+			ensure!(submitted == targets.len(), Error::<T>::DuplicateDisputeStatementSets);
+		}
+
+		let mut fresh = Vec::with_capacity(statement_sets.len());
+		for statement_set in statement_sets {
+			let dispute_target = (statement_set.session, statement_set.candidate_hash);
+			if Self::provide_dispute_data(&config, statement_set)? {
```
thiolliere, line 496: in case this fail, a dispatch error is returned, but the previous successful call modified the storage.

I don't know exactly in which context provide_multi_dispute_data is used, but this contradict the usual rule of `check first then writes`, and in case of failure, the fresh written statement are not returned by the function.

So I guess it is a mistake

runtime/parachains/src/disputes.rs
```diff
@@ -0,0 +1,752 @@
...
+				// a valid candidate, according to 2/3. Punish those on the 'against' side.
+				T::PunishValidators::punish_against_valid(
+					set.session,
+					summary.state.validators_against.iter_ones().map(|i| ValidatorIndex(i as _)),
+				);
+			}
+
+			if summary.new_flags.contains(DisputeStateFlags::AGAINST_SUPERMAJORITY) {
+				// an invalid candidate, according to 2/3. Punish those on the 'for' side.
+				T::PunishValidators::punish_against_valid(
```
thiolliere, line 618:
```suggestion
				T::PunishValidators::punish_for_invalid(
```

roadmap/implementers-guide/src/types/disputes.md
```diff
@@ -33,8 +33,8 @@ Kinds of dispute statements. Each of these can be combined with a candidate hash
 ```rust
 enum ValidDisputeStatementKind {
     Explicit,
-    BackingSeconded,
-    BackingValid,
+    BackingSeconded(Hash),
```
drahnr, line 36: I'd expect this to be used with key value store, where the key is the hash, otherwise all other variants would need the `Hash` as well?

roadmap/implementers-guide/src/runtime/disputes.md
```diff
@@ -75,18 +74,18 @@ Frozen: bool,
...
+      1. increment `SpamSlots` for each validator in the `DisputeStatementSet` which is not already in the `DisputeState`. Initialize the `SpamSlots` to a zeroed vector first, if necessary. do not increment `SpamSlots` if the candidate is local.
+      1. If the value for any spam slot exceeds `config.dispute_max_spam_slots`, return false.
   1. If the overlap of the validators in the `DisputeStatementSet` and those already present in the `DisputeState` is at least `byzantine_threshold + 1`, the `DisputeState` has fewer than `byzantine_threshold + 1` validators, and the candidate is not present in the `Included` map, decrement `SpamSlots` for each validator in the `DisputeState`.
-  1. Import all statements into the dispute. This should fail if any statements are duplicate; if the corresponding bit for the corresponding validator is set in the dispute already.
-  1. If `concluded_at` is `None`, reward all statements slightly less.
+  1. Import all statements into the dispute. This should fail if any statements are duplicate or if the corresponding bit for the corresponding validator is set in the dispute already.
```
drahnr, line 80: Iirc, voting twice for a validator in opposing ways is not an issue, since that would impose slashing itself for sure.

roadmap/implementers-guide/src/runtime/disputes.md
```diff
@@ -75,18 +74,18 @@ Frozen: bool,
   1. If there is no dispute under `Disputes`, create a new `DisputeState` with blank bitfields.
   1. If `concluded_at` is `Some`, and is `concluded_at + config.post_conclusion_acceptance_period < now`, return false.
   1. If the overlap of the validators in the `DisputeStatementSet` and those already present in the `DisputeState` is fewer in number than `byzantine_threshold + 1` and the candidate is not present in the `Included` map
```
drahnr, line 76: Does _overlap_ mean _intersection_ or _union_ here? Either way I don't think either works perfectly.

Assuming the first message is less than the threshold, in the intersection case, the second could be a disjoint set and hence be empty and marked as spam.

Assuming the union, and `q`-size small subsets (`total / q`) it's marking only the first `q-1` spammers as spammers.

Could elaborate a bit on the intended spam mechanics?

runtime/parachains/src/disputes.rs
rphmeier: Because the `enter` inherent is required for block validity in Polkadot (see `on_finalize` implementation), it is OK for us to bypass the typical requirement of not changing storage before failing. That allows us to make the parachains implementation more efficient.

roadmap/implementers-guide/src/types/disputes.md
rphmeier: I don't know what a key-value store has to do with this, but in the case of `BackingSeconded` and `BackingValid`, the hash is the 'inclusion parent' or relay-parent referenced by the `SigningContext`. neither explicit dispute statements nor approval votes require the `SigningContext` as they have no need to be pinned to any specific point in the chain.

roadmap/implementers-guide/src/runtime/disputes.md
rphmeier: Voting twice in opposing ways doesn't necessarily need to lead to a slash as long as it's -EV. The way that disputes are designed now ensures that it is -EV as it raises the likelihood of ending up against a supermajority.

roadmap/implementers-guide/src/runtime/disputes.md
```diff
@@ -65,7 +65,6 @@ Frozen: bool,
 ## Routines
 
 * `provide_multi_dispute_data(MultiDisputeStatementSet) -> Vec<(SessionIndex, Hash)>`:
-  1. Fail if any disputes in the set are duplicate or concluded before the `config.dispute_post_conclusion_acceptance_period` window relative to now.
```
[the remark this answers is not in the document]
rphmeier: In the implementation of `provide_multi_dispute_data`, we fail if there are any two statement _sets_ which refer to the same candidate in the same session. That is independent of the statements within those actual sets. The block author is meant to include not only his own vote, but also other votes that he has witnessed, which are not already present on-chain.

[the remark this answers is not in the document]
thiolliere: ah got it

runtime/parachains/src/disputes.rs
```diff
@@ -0,0 +1,1977 @@
...
+						b,
+						vec![(&0, v0.public()), (&1, v1.public())],
+						Some(vec![(&0, v0.public()), (&1, v1.public())]),
+					))
+				}
+			);
+
+			let candidate_hash = CandidateHash(sp_core::H256::repeat_byte(1));
+			let stmts = vec![
+				// // TODO TODO: this test is currently failing due to fail after writes.
```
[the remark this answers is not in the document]
thiolliere: I removed the test and modify the doc a bit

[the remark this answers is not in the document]
rphmeier: great, so those TODOs can be removed?

roadmap/implementers-guide/src/runtime/disputes.md
drahnr: That wording clarifies it 👍

runtime/parachains/src/disputes.rs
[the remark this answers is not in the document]
thiolliere: Yes it is removed with https://github.com/paritytech/polkadot/pull/2947/commits/7570236316b22b2d39ee59676b307c2ed0f873a0
I added a bit of verbose documentation on the function, feel free to shorten it, but seems good to have it mentioned quickly.

runtime/parachains/src/disputes.rs
```diff
@@ -0,0 +1,1963 @@
...
+	{
+		type Event: From<Event> + IsType<<Self as frame_system::Config>::Event>;
+		type RewardValidators: RewardValidators;
+		type PunishValidators: PunishValidators;
+	}
+
+	#[pallet::pallet]
+	pub struct Pallet<T>(_);
+
+	/// The last pruneed session, if any. All data stored by this module
```
andresilva, line 114:
```suggestion
	/// The last pruned session, if any. All data stored by this module
```

runtime/parachains/src/disputes.rs
```diff
@@ -0,0 +1,1963 @@
...
+		if validator.0 as usize >= self.new_participants.len() {
+			return Err(VoteImportError::ValidatorIndexOutOfBounds);
+		}
+
+		bits.set(validator.0 as usize, true);
+
+		// New participants tracks those which didn't appear on either
+		// side of the dispute until now. So we check the other side
+		// and checked the first side before.
+		if other_bits.get(validator.0 as usize).map_or(false, |b| !*b) {
```
andresilva, line 327: This would be an equivocation though right?

runtime/parachains/src/disputes.rs
```diff
@@ -0,0 +1,1963 @@
...
+
+// The supermajority threshold of validators which is required to
+// conclude a dispute.
+fn supermajority_threshold(n: usize) -> usize {
+	n - byzantine_threshold(n)
+}
+
+bitflags::bitflags! {
+	#[derive(Default)]
+	struct DisputeStateFlags: u8 {
```
andresilva, line 204: Can't this just be an enum instead? All states are mutually exclusive, you can't have a super majority without having a confirmation first.

runtime/parachains/src/disputes.rs
```diff
@@ -0,0 +1,1963 @@
...
+	/// Called by the initalizer to initialize the disputes module.
+	pub(crate) fn initializer_initialize(now: T::BlockNumber) -> Weight {
+		let config = <configuration::Module<T>>::config();
+
+		let mut weight = 0;
+		for (session_index, candidate_hash, mut dispute) in <Disputes<T>>::iter() {
+			weight += T::DbWeight::get().reads_writes(1, 0);
+
+			if dispute.concluded_at.is_none()
+				&& dispute.start + config.dispute_conclusion_by_time_out_period < now
```
andresilva, line 419:
```suggestion
				&& dispute.start + config.dispute_conclusion_by_time_out_period >= now
```

runtime/parachains/src/disputes.rs
```diff
@@ -0,0 +1,1963 @@
...
+
+fn has_supermajority_against<BlockNumber>(dispute: &DisputeState<BlockNumber>) -> bool {
+	let supermajority_threshold = supermajority_threshold(dispute.validators_against.len());
+	dispute.validators_against.count_ones() >= supermajority_threshold
+}
+
+// If the dispute had not enough validators to confirm, decrement spam slots for all the participating
+// validators.
+//
+// returns the set of participating validators as a bitvec.
```
andresilva, line 742:
```suggestion
// Returns the set of participating validators as a bitvec.
```

runtime/parachains/src/disputes.rs
```diff
@@ -0,0 +1,1963 @@
...
+					};
+
+					// also reduce spam slots for all validators involved, if the dispute was unconfirmed.
+					// this does open us up to more spam, but only for validators who are willing
+					// to be punished more.
+					//
+					// it would be unexpected for any change here to occur when the dispute has not concluded
+					// in time, as a dispute guaranteed to have at least one honest participant should
+					// conclude quickly.
+					let participating = decrement_spam(spam_slots, &dispute);
```
andresilva, line 448: Shouldn't we also be slashing the validators who didn't participate at all?

runtime/parachains/src/disputes.rs
```diff
@@ -0,0 +1,1963 @@
...
+				.unwrap_or_else(|| vec![0; n_validators]);
+
+			for (validator_index, spam_slot_change) in summary.spam_slot_changes {
+				let spam_slot = spam_slots.get_mut(validator_index.0 as usize)
+					.expect("index is in-bounds, as checked above; qed");
+
+				match spam_slot_change {
+					SpamSlotChange::Inc => {
+						ensure!(
+							*spam_slot < config.dispute_max_spam_slots,
```
andresilva, line 615: What should the client do here? Retry and not include the spam statement?

runtime/parachains/src/disputes.rs
```diff
@@ -0,0 +1,1963 @@
...
+				Self::deposit_event(Event::DisputeConcluded(
+					set.candidate_hash,
+					DisputeResult::Valid,
+				));
+			}
+
+			// It is possible, although unexpected, for a dispute to conclude twice.
+			// This would require f+1 validators to vote in both directions.
+			// A dispute cannot conclude more than once in each direction.
+
```
andresilva, line 648: For this to happen we need to have had equivocations in the dispute resolution protocol as well. Wondering whether we should make this more explicit to the client.

runtime/parachains/src/disputes.rs
```diff
@@ -0,0 +1,1963 @@
...
+				summary.slash_for,
+			);
+		}
+
+		<Disputes<T>>::insert(&set.session, &set.candidate_hash, &summary.state);
+
+		// Freeze if just concluded against some local candidate
+		if summary.new_flags.contains(DisputeStateFlags::AGAINST_SUPERMAJORITY) {
+			if let Some(revert_to) = <Included<T>>::get(&set.session, &set.candidate_hash) {
+				Self::revert_and_freeze(revert_to);
```
andresilva, line 683: It's unlikely but possible that two different disputes would resolve at this block, each one asking to revert to different blocks. I assume the expected behavior is that the client reverts to the earliest one, would it make sense to deduplicate in the runtime?

runtime/parachains/src/disputes.rs
```diff
@@ -0,0 +1,1963 @@
...
+	}
+
+	pub(crate) fn note_included(session: SessionIndex, candidate_hash: CandidateHash, included_in: T::BlockNumber) {
+		if included_in.is_zero() { return }
+
+		let revert_to = included_in - One::one();
+
+		<Included<T>>::insert(&session, &candidate_hash, revert_to);
+
+		// If we just included a block locally which has a live dispute, decrement spam slots
```
andresilva, line 702: In principle the client shouldn't be doing this in the first place right?

andresilva (changes requested):

runtime/parachains/src/disputes.rs
andresilva: Except when we have `f + 1` equivocators, in which case we might have majority for and against.

runtime/parachains/src/disputes.rs
burdges: We could replace ImOnline by GRANDPA by asking that validators vote on who they saw enough GRANDPA votes from.

It's messy doing anything similar here because they could always wait until the winning side emerges.  If we do refund for bugs, and do a good job of identifying bugs, then slashing rewards make non-participation unattractive, and modifying your code around this stuff becomes costly.

runtime/parachains/src/disputes.rs
burdges: Yes we need reversions to merge.

runtime/parachains/src/disputes.rs
rphmeier: It shouldn't, but it might happen especially at the exact point of reversion. If there's a bad candidate C which is included at a block B+1', and then we revert back to B, there's the possibility that the transplanted dispute and the candidate both get included. That leads to the new chain also needing to be abandoned, of course.

What I want honest nodes to do is avoid importing availability bitfields on stale leaves, so this scenario won't happen unless the node authoring the first block after a revert is malicious. And this code is to defend against such nodes

runtime/parachains/src/disputes.rs
rphmeier: Right, good point.

runtime/parachains/src/disputes.rs
rphmeier: The client should not include anything beyond the maximum spam slots. I wrote some docs in the provisioner page of the guide to describe that process

runtime/parachains/src/disputes.rs
rphmeier: @andresilva not really. if 2 validators cook up a dispute on something that doesn't exist, we shouldn't slash everybody as a result

runtime/parachains/src/disputes.rs
rphmeier: I added some code for merging reversions. FWIW, after the first reversion, the chain is already likely to be abandoned and any prior dispute should be transplanted, leading to yet another reversion.

runtime/parachains/src/disputes.rs
rphmeier: I don't think we need to, because it's -EV for validators to equivocate anyway.

runtime/parachains/src/disputes.rs
rphmeier: Yeah.

andresilva:
Added consensus log and event for reversion, and decoupled dispute handling to be able to disable it when constructing runtimes.

primitives/src/v1/mod.rs
```diff
@@ -1149,6 +1179,19 @@ pub struct InherentData<HDR: HeaderT = Header> {
 	pub parent_header: HDR,
 }
 
+/// The maximum number of validators `f` which may safely be faulty.
+///
+/// The total number of validators is `n = 3f + e` where `e in { 1, 2, 3 }`.
+pub fn byzantine_threshold(n: usize) -> usize {
```
drahnr, line 1185: :+1: so the re-impl can stop now :)

andresilva:
Should be ready for a final review and merge.

runtime/parachains/src/disputes.rs
```diff
@@ -0,0 +1,2080 @@
...
+		/// instead revert to the block at the given height which is the last
+		/// known valid block in this chain.
+		Revert(T::BlockNumber),
+	}
+
+	#[pallet::hooks]
+	impl<T: Config> Hooks<BlockNumberFor<T>> for Pallet<T> {}
+
+	#[pallet::call]
+	impl<T: Config> Pallet<T> {}
```
shawntabrizi, line 296: you should be able to omit these now

runtime/parachains/src/disputes.rs
```diff
@@ -0,0 +1,2080 @@
...
+			let to_prune = if let Some(last_pruned) = last_pruned {
+				*last_pruned + 1 ..= pruning_target
+			} else {
+				pruning_target ..= pruning_target
+			};
+
+			for to_prune in to_prune {
+				<Disputes<T>>::remove_prefix(to_prune, None);
+				<Included<T>>::remove_prefix(to_prune, None);
+				SpamSlots::<T>::remove(to_prune);
```
shawntabrizi, line 601: This kind of stuff is very expensive and needs to be limited.

dont take the `None` option here for granted, you should provide an upper bound to how many items can be deleted in one go, and then provide methods for users to continue to clean up the data.

rphmeier: What do you think would be a sensible upper bound to do per block initialization? I don't think we should defer this to users without introducing an economic incentive for them to do so

rphmeier: We spoke off the thread, but https://github.com/paritytech/polkadot/issues/3469 will address this in near future. For now it's fine.

rphmeier:
bot merge

rphmeier:
bot merge

PR #2952, opened 2021-04-28
Required weight is returned in case it exceeds limit.

gavofyork:


xcm/src/v0/multi_location.rs
```diff
@@ -578,7 +578,7 @@ impl MultiLocation {
 		let self_parents = self.parent_count();
 		let prefix_rest = prefix.len() - prefix.parent_count();
 		let skipped = self_parents.min(prefix_rest);
-		if self.len() + prefix.len() - 2 * skipped > 4 {
+		if self.len() + prefix.len() - 2 * skipped > 8 {
```
gavofyork, line 581: Minor fix since we now have 8 items in `MultiLocation`, not 4.

shawntabrizi (approved):

xcm/src/v0/multi_location.rs
shawntabrizi: We will always need to remember to bump this number?

pepyakin (approved):

PR #2953, opened 2021-04-28
Allow normal accounts to execute XCM for withdraw+teleport

gavofyork:
Companion: https://github.com/paritytech/cumulus/pull/418

### DIFF
--- a/runtime/rococo/src/lib.rs
+++ b/runtime/rococo/src/lib.rs
@@ -86,7 +86,7 @@ pub use pallet_balances::Call as BalancesCall;
 
 use polkadot_parachain::primitives::Id as ParaId;
 
-use xcm::v0::{MultiLocation, NetworkId, BodyId};
+use xcm::v0::{Xcm, MultiLocation, NetworkId, BodyId};
 use xcm_executor::XcmExecutor;
 use xcm_builder::{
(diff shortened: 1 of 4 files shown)


shawntabrizi (approved):

riusricardo (approved):

xlc:
any plan to allow reserve based transfer of ROC to non system parachain?

runtime/rococo/src/lib.rs
```diff
@@ -664,12 +664,61 @@ pub type LocalOriginToLocation = (
...
+				//       dest: AccountId32,
+				//     } ]
+				//   } ]
+				// }
+				matches!(msg, WithdrawAsset { ref assets, ref effects }
+					if assets.len() == 1
+					&& matches!(assets[0], ConcreteFungible { id: Null, .. })
+					&& effects.len() == 2
+					&& matches!(effects[0], BuyExecution { .. })
+					&& matches!(effects[1], InitiateTeleport { ref assets, dest: X1(Parachain(..)), ref effects }
```
apopiak, line 696: late dumb question: doesn't `Index` panic in case the length is less than 2 here?

xlc: `&& effects.len() == 2`

PR #2420, opened 2021-02-11
Introduce PerPeerSet utility that allows to segrate based on PeerSet

pepyakin:

PR #2954, opened 2021-04-29
First step towards creating a set of Collator Fitness Metrics

Lldenaurois:
This PullRequest implements a modified peer acceptance/rejection policy.

There are two parts to this PR:

1. Maintain Historic Fitness Metrics for collators where we keep track of the last, total and avg occurrence of undesired behavior
2. Manage incoming connections and evict peers in order to ensure that we are actively sampling collators from the current or next parachain assignment uniformly at random from suitable collators attempting to connect

In order to achieve this, I have introduced an additional abstraction around `HashMap<PeerId, PeerData>`, called PeerSlots.

`PeerSlots` keeps track of the `PeerData`, the `FitnessMetrics` for all historical `CollatorId` as well as a list of stale peers to evict at the next incoming connection. The goal of this PR is to begin to introduce a randomized algorithm to manage incoming connections, such that we can ensure that we are uniformly randomly selecting collators for the next parachain assignment. Note that it is not possible to select collators at random without applying some type of stateful iterative algorithm, since the total number of collators attempting to connect to our node needn't be known in advance. Therefore, I have implemented a modified Reservoir sampling algorithm to ensure that at any given point in time we can ensure that we have sampled our incoming connections uniformly at random.

Currently, all peers are disconnected after we handle our view change. This behavior is unchanged; that is after every parachain assignment we move the old peers to the list of stale peers and then evict all stale peers. However, I believe it may be beneficial to only disconnect a single peer whenever a new peer is attempting to connect to the validator node.

The current eviction policy will always evict stale_peers and will evict additional peers if the total number of connections still exceeds the desired number of connected peers.

Of the connected peers not in the stale peers list, we prefer to disconnect peers with a higher weighted average FitnessMetric. That is, unknown or newly connected peers will have a weighted average FitnessMetric of 0 and will therefore only be evicted with low probability until they become stale (in which case they will get evicted after the assignment). For every negative metric, i.e. UnknownMessage, SignatureError, NetworkError, WrongParachain, etc, we keep track of every occurrence and will prefer to evict peers with a higher total sum of metric avg (average time between occurrences of negative event).

Currently, determining which peers to evict if the list of stale peers is insufficient to keep the number of connections below the reservoir size is implemented naively. There are performance optimizations that we can explore if need be.

Finally, whenever a peer submit a Declaration, we determine whether or not to disconnect that peer based on a k/n reservoir sampling algorithm. This ensures that if the number of collators attempting to connect for the next parachain assignment exceeds the total number of connections we would like to manage, we can ensure that we uniformly-randomly select incoming connections via this randomized algorithm. Implementation-specific details are included in-line (see `fn sample_connection()`).


node/network/collator-protocol/src/lib.rs
```diff
@@ -164,3 +164,129 @@ where
 		NetworkBridgeMessage::ReportPeer(peer, rep),
 	)).await;
 }
+
+pub(crate) mod fitness_metrics {
+    use super::{CollatorId, CollatorProtocolMessage, SubsystemContext, modify_reputation, PeerId};
```
bkchr, line 169: Please use tabs everywhere ;)

Lldenaurois: Yes, I'm working on this PR right now. I will adjust everything once it's ready.

node/network/collator-protocol/src/lib.rs
```diff
@@ -43,124 +43,285 @@ const LOG_TARGET: &'static str = "parachain::collator-protocol";
...
+                FitnessEvent::NetworkError => 2,
+                FitnessEvent::Timeout => 3,
+                FitnessEvent::SigError => 4,
+                FitnessEvent::ParaError => 5,
+                FitnessEvent::Superfluous => 6,
+                FitnessEvent::ReportBad => 7,
+                FitnessEvent::NotifyGood => 8,
+            }
+        }
+    }
```
ordian, line 224: FWIW, we used to have a `repr(usize)` enum, but opted for a more type-safe API
https://github.com/paritytech/polkadot/pull/2420/files

node/network/collator-protocol/src/lib.rs
```diff
@@ -43,124 +43,285 @@ const LOG_TARGET: &'static str = "parachain::collator-protocol";
...
-	Subsystem(#[from] SubsystemError),
-	#[error(transparent)]
-	Oneshot(#[from] oneshot::Canceled),
-	#[error(transparent)]
-	RuntimeApi(#[from] RuntimeApiError),
-	#[error(transparent)]
-	UtilError(#[from] util::Error),
-	#[error(transparent)]
-	Prometheus(#[from] prometheus::PrometheusError),
+    #[error(transparent)]
```
rphmeier, line 46: indentation: looks like tabs got replaced with spaces

ordian (commented):
first pass

node/network/collator-protocol/src/peer_slots.rs
```diff
@@ -0,0 +1,639 @@
...
+impl CollatorFitnessMetric {
+	pub fn insert_event(&mut self, event: FitnessEvent) {
+		(&mut self[event]).insert_event()
+	}
+
+	// Compute collator's fitness value.
+	// Note that this is a naive sum of all metric averages.
+	// In the future we will expand on this functionality.
+	pub fn compute_fitness(&self) -> f64 {
+		self.0.iter().map(|a| a.avg).sum::<f64>()
```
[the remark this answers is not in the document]
Lldenaurois: This is an oversight on my part. We discussed that we would only track Cost rather than handling benefit and cost in the weighted average.

node/network/collator-protocol/src/peer_slots.rs
```diff
@@ -0,0 +1,639 @@
...
+
+	pub fn insert_event(&mut self) {
+		let mut cumulative = if self.avg == 0f64 {
+			0f64
+		} else {
+			self.total as f64 / self.avg
+		};
+		self.total += 1;
+		cumulative += self.last.elapsed().as_secs() as f64;
+		self.avg = self.total as f64 / cumulative;
```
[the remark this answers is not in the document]
Lldenaurois: You're right. If the time_elapsed() since last is 0, then this will still be 0.

I'm debating whether it's necessary to keep track of the average independently. Instead we should probably sum the time and the total and then divide when we compute the weighted average.

I will adjust

node/network/collator-protocol/src/peer_slots.rs
```diff
@@ -0,0 +1,639 @@
...
+	COST_REQUEST_TIMED_OUT,
+	COST_INVALID_SIGNATURE,
+	COST_WRONG_PARA,
+	COST_UNNEEDED_COLLATOR,
+	COST_REPORT_BAD,
+	BENEFIT_NOTIFY_GOOD,
+];
+
+#[derive(Copy, Clone)]
+#[repr(usize)]
```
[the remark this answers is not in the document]
Lldenaurois: I was thinking of doing that as well. Glad you mentioned it.

node/network/collator-protocol/src/peer_slots.rs
```diff
@@ -0,0 +1,639 @@
+use super::{CollatorId, CollatorProtocolMessage, PeerId, SubsystemContext, LOG_TARGET};
```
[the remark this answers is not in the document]
Lldenaurois: That's a great suggestion! Will do.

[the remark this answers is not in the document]
rphmeier: And please add docs on all public items, including module docs

node/network/collator-protocol/src/peer_slots.rs
```diff
@@ -0,0 +1,639 @@
...
+			peer_data: HashMap::new(),
+			active_paras: ActiveParas::default(),
+			stale_peers: Vec::new(),
+		}
+	}
+}
+
+impl std::fmt::Debug for PeerSlots {
+	fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
+		write!(f, "Eladed")
```
rphmeier, line 453: ?

node/network/collator-protocol/src/peer_slots.rs
```diff
@@ -0,0 +1,639 @@
...
+	active_paras: &mut ActiveParas,
+	keystore: &SyncCryptoStorePtr,
+	new_relay_parents: impl IntoIterator<Item = Hash>,
+	old_relay_parents: impl IntoIterator<Item = Hash>,
+) {
+	active_paras.assign_incoming(sender, keystore, new_relay_parents).await;
+	active_paras.remove_outgoing(old_relay_parents);
+}
+
+fn peers_to_evict(peer_data: &Reservoir, fitness: &mut CollatorFitness, num: usize, rng: &mut ThreadRng) -> Vec<PeerId> {
```
rphmeier, line 496: I suggest either pass a generic Rng in, or call `thread_rng` internally and remove the parameter. Generic RNG may be better for unit-testing

rphmeier:
ready for review? checkmarks are still red

roadmap/implementers-guide/src/node/collators/collator-protocol.md
```diff
@@ -98,7 +98,11 @@ digraph G {
...
+When peers attempt to connect to a validator node, they are required to present a PrePoV to the validator node as part of the handshake of the network connection within substrate. This will provide all information regarding that collator's identity, i.e. their public key (`CollatorId`), the specific para ID they intend to collate as well as a witness to the PrePVF's `validate` function. Upon successful verification the collator may begin to send advertisements of collations. The validator maintains historic state for every `CollatorId` they have come across in a persistent data structure which allows keep track of the total and average occurence (per unit time) of undesired behavior for this collator. Examples of undesired behavior include but are not limited to submitting advertisements for collations that are on a relay-parent outside of our view or for a different para than the one they have declared in the handshake of their network connection.
```
rphmeier, line 101: 1. `PrePov` is not necessarily part of the handshake although I suspect it should be at least an optional element so we can support restrictive collator selection algorithms that are entirely determined by the validation Wasm
2. The guide should not refer to Substrate directly; our aim is to keep it high-level and agnostic. Potentially as a reference document to be used by other client teams which are not using Substrate.

node/network/collator-protocol/src/validator_side/mod.rs
```diff
@@ -575,11 +279,13 @@ async fn handle_peer_view_change(
...
-
-	peer_data.update_view(view);
-	state.requested_collations
-		.retain(|(rp, _, pid), _| pid != &peer_id || !peer_data.has_advertised(&rp));
+	if let Some(peer_data) = state.peer_slots.peer_data.get_mut(&peer_id) {
+		peer_data.update_view(view);
+		state.requested_collations
+			.retain(|(rp, _, pid), _| pid != &peer_id || !peer_data.has_advertised(&rp));
+	} else {
+		unimplemented!{}
```
rphmeier, line 287: unimplemented here needs to be implemented

node/network/collator-protocol/src/validator_side/mod.rs
```diff
@@ -439,6 +438,26 @@ impl ActiveParas {
...
 
+#[derive(Debug, Clone, Hash, Eq, PartialEq)]
+struct PendingCollation {
+	relay_parent: Hash,
+	para_id: ParaId,
+	peer_id: PeerId,
+}
+
+impl PendingCollation {
+	pub fn new(relay_parent: Hash, para_id: &ParaId, peer_id: &PeerId) -> Self {
```
rphmeier, line 449: Does this need to be `pub` since the struct is private?

node/network/collator-protocol/src/validator_side/mod.rs
```diff
@@ -612,13 +620,13 @@ where
 		);
 		return;
 	}
-
-	if state.requested_collations.contains_key(&(relay_parent, para_id.clone(), peer_id.clone())) {
+	let pending_collation = PendingCollation { relay_parent, para_id, peer_id };
```
rphmeier, line 623: This is about as terse as the `new` constructor which exists, so I suggest we remove the `new` entirely

node/network/collator-protocol/Cargo.toml
```diff
@@ -30,3 +32,6 @@ sp-core = { git = "https://github.com/paritytech/substrate", branch = "master",
 sp-keyring = { git = "https://github.com/paritytech/substrate", branch = "master" }
 
 polkadot-subsystem-testhelpers = { package = "polkadot-node-subsystem-test-helpers", path = "../../subsystem-test-helpers" }
+
+[features]
+sampling = []
```
ordian, line 37: What's the plan with the feature? We probably don't want to have two binaries for validators depending on whether we want sampling or not. If we want to test this, we can roll it out on some rococo validators to see how they are behaving. And remove the feature altogether.

node/network/collator-protocol/src/validator_side/mod.rs
```diff
@@ -1,3 +1,4 @@
+#![allow(dead_code)]
```
ordian, line 1: I assume this is due to the feature above? IMO we should remove this.

node/network/collator-protocol/src/validator_side/peer_slots.rs
```diff
@@ -0,0 +1,1046 @@
...
+		self.cumulative += elapsed;
+		self.last = std::time::Instant::now();
+	}
+
+	pub fn elapsed(&self) -> Duration {
+		self.last.elapsed()
+	}
+
+	pub fn fitness(&self) -> f64 {
+		if self.total != 0 {
```
ordian, line 136: I'd check here `self.cumulitive.as_nanos() != 0` just to be sure, in case elapsed event is zero for some reason.

node/network/collator-protocol/Cargo.toml
rphmeier: I suggested the feature-flag, so it doesn't go into a release until we've tested it thoroughly on Rococo

ordian: My suggestion is instead not to merge it until we've tested it thoroughly. The binary can be built from the branch using CI and deployed for testing on Rococo.

node/network/collator-protocol/src/validator_side/peer_slots.rs
```diff
@@ -0,0 +1,1046 @@
...
+use polkadot_primitives::v1::{Hash, Id as ParaId};
+use polkadot_subsystem::{messages::{AllMessages, NetworkBridgeMessage}, SubsystemSender};
+
+// CACHE_SIZE determines the size (in number of entries) of in-memory cache maintained for the
+// purposes of determining a collator's fitness. Metrics for collators that do not fit in the
+// Cache are stored in a database.
+const CACHE_SIZE: usize = 1_000;
+// RESERVOIR SIZE determines the upper bound on the number of collator connections
+#[cfg(not(test))]
+pub(super) const RESERVOIR_SIZE: usize = 128;
```
ordian, line 39: this should be hooked into https://github.com/paritytech/polkadot/blob/8be63cd653e6ee73e320a3aaa6cfacb5f13b5c57/node/network/protocol/src/peer_set.rs#L75

alternatively, we can make the latter limit much higher

node/network/collator-protocol/src/validator_side/mod.rs
```diff
@@ -853,23 +667,21 @@ async fn handle_our_view_change(
...
 		// declare.
-		if let Some(para_id) = peer_data.collating_para() {
-			if !state.active_paras.is_current_or_next(para_id) {
-				disconnect_peer(ctx, peer_id.clone()).await;
-			}
-		}
+		disconnect_peer(ctx, peer_id).await;
 	}
 
+	state.evicted_peers.clear();
```
ordian, line 682: why do we clear evicted peers on our view change which happens on every new relay block import?

ordian:
Before we merge this PR, we need to improve the fitness metric calculation. Currently, it doesn't account for valid collations, only undesired behavior. Also it doesn't differentiate between different types of misbehavior, e.g. out-of-view collation vs invalid signature. Maybe we can utilize the existing reputation system to do a weighted calculation.

Lldenaurois:
@ordian Agreed.

Regarding positive reinforcement: Jeff and Rob suggested that positive reinforcement is probably not useful as we really only want to penalize bad collators and keep all things equal for the rest.

You are correct that in this diff the fitness is just the sum of values across all metrics; I wasn't sure how to weigh one metric vs the other. I think your suggestion is the right one, we should use the existing system as much as possible to validate this approach.

rphmeier:
Closing due to stale. Let's revive later on!

PR #2962, opened 2021-04-29
refactor overseer into proc-macro based pattern

drahnr:
The branch name is historical, this expanded into a refactoring of the `Overseer` into a pattern to actually allow simpler spawning of modified nodes with subsystems that exhibit modified (malicious and/or broken) behavior besides future compile time checks based on subsystem.

This is stage one and lies the groundwork for further checks:

* [x] generate the overseer as it is currently
* [x] provide a builder pattern for overseer construction
* [x] Avoid `AllMessages` and `.into()` in any subsystem, always send the subsystem specific types directly
* [x] make the overseer macro idempotent and eliminate all leaked types from the current impl

Next steps are covered in #3427

---

Notes to reviewers:

* You will find two traits called `SubsystemContext`, where one binds 3 associated types, the other is a wrapper around it. This was meant to simplifify the trait bounds across all subsystem's generic type `Context`. In retrospective I am not sure which is better: `Subsystem<Message=X> + overseer::Subsystem<Message=X>` vs `overseer::Subsystem<Message=X, AllMessages=AllMessages, Signal=OverseerSignal, Error=SubsystemError>`.

* Most usages of conversions into `AllMessages` around `fn send_message(..)` were removed, instead almost all instances now do the conversion implicitly which removes quite a bit of clutter and will be necessary for #3427 anyways.

* I moved the `#[cfg(test)] mod tests;` back at the end of the file. It caused a whole lot of duplicate definitions and it's just clearer to have one spot to look for the test module, and that's at the bottom of a file. External or not is imho secondary to that.

* Most changes are boring trait bounds, so I'd recommend on focusing on the [node/overseer/src/lib.rs](https://github.com/paritytech/polkadot/pull/2962/files#diff-95bf7f2e804b78a7af9831aa8142a52a51506f19d2dbe088949341fac607c699) and more generally on `node/{overseer,subsystem}/*` where the most relevant changes are.


doc/testing.md
```diff
@@ -0,0 +1,158 @@
...
+
+### Unit testing
+
+Boring, small scale correctness tests of individual functions.
+
+### Integration tests
+
+One particular subsystem (subsystem under test) interacts with a
+mocked overseer that is made to assert incoming and outgoing messages
+of the subsystem under test.
```
rphmeier, line 17: This sounds kind of like what we do with our unit tests of subsystems. Could you clarify the difference?

doc/testing.md
```diff
@@ -0,0 +1,158 @@
...
+The main goals are is to allow creating a test node which
+exhibits a certain behaviour by utilizing a subset of _wrapped_ or _replaced_ subsystems easily.
+The runtime must not matter at all for these tests and should be simplistic.
+The execution must be fast, this mostly means to assure a close to zero network latency as
+well as shorting the block time and epoch times down to a few `100ms` and a few dozend blocks per epoch.
+
+### Approach
+
+`AllSubsystems` is an intermediate mocking type. As such it is a prime target for modification.
+`AllSubsystemsGen` is a proc-macro that should be extended as needed for per subsystem specific
```
rphmeier, line 128: I think proc-macro would also be interesting, for the purpose of extracting the `Overseer` to be a more general utility, which can be used by other projects.

doc/testing.md
drahnr: It pretty much is. There is little consistency between the subsystems and some generalization with addition helper macros might reduce the boiler plate a bit. That's the only thing left for this point and is certainly of low priority.

doc/testing.md
drahnr: Yeah, that'd be another option and certainly more useful. For the testing story, this won't make much of a difference impl wise.

doc/testing.md
rphmeier: So you make a distinction between integration test and system tests. It seems like this section is redundant

drahnr: They are two distinct things:

a `integration test` is (in our case) the test of exactly one individual subsystem w/ mocked surroundings (overseer, other subsystems) where as a `system test` is the test of a node w/ partially real, partially mocked subsystems in conjunction w/ other subsystems. I guess some additional bike-shedding is needed to make the distinction cleaner.

Cargo.lock
```diff
@@ -38,9 +38,9 @@ dependencies = [
 
 [[package]]
 name = "aes"
-version = "0.4.0"
+version = "0.5.0"
```
ordian, line 41: please revert `cargo update`

node/overseer/overseer-gen/tests/ui/gen.rs
```diff
@@ -0,0 +1,80 @@
...
+	fn builder() -> OooohBuilder {
+		OooohBuilder::default()
+	}
+}
+
+struct OooohBuilder<Ctx, Sub0, Sub1> {
+	sub0: ::std::option::Option<Sub0>,
+	yyy: ::std::option::Option<Sub1>,
+	spawner: ::std::option::Option<spawner>,
+	metrics: ::std::option::Option<metrics>,
```
rphmeier, line 22: prefix `::` is not needed

node/overseer/overseer-gen/tests/ui/ok-02.rs
```diff
@@ -0,0 +1,26 @@
...
+#[overlord(Wrapper)]
+struct Overseer {
+	#[subsystem(X)]
+	sub0: AwesomeSubSys,
+}
+
+struct Spawner;
+
+fn main() {
+	let overseer = Overseer::<Spawner, SubSystems<X>>::builder()
```
rphmeier, line 21: From an API perspective, this `SubSystems` appears from nowhere and is confusing

behavior-tests/Cargo.toml
```diff
@@ -0,0 +1,19 @@
+[package]
+name = "behavior-tests"
+version = "0.1.0"
+authors = ["Bernhard Schuster <bernhard@ahoi.io>"]
```
rphmeier, line 4: `admin@parity.io`

node/overseer/overseer-gen/tests/ui/ok-02.rs
drahnr: This is the last instance and it is already flattened away. The gen.rs is to be removed.

node/overseer/overseer-gen/tests/ui/gen.rs
drahnr: Technically it is, `std` could refer to a in-scope module and hence the wrong `Option` type could be used.

node/overseer/overseer-gen/tests/ui/ok-02.rs
drahnr: Yanked!

drahnr:
After quite a few iterations this is the initial implementation with these TODOs for a follow up PR (issues to be created):

* [ ] refactor the whole `NetworkMessage` enum and conversions - there is already the `Event` type which should be the single point of entry for all non-subsystem generated messages or the bridge should be a subsystem like all others
* [ ] improve error handling messages in a few places
* [ ] support `where` clauses on generics for the overseer that are unrelated to the subsystems
* [ ] improved documentation

node/overseer/overseer-gen/proc-macro/src/impl_builder.rs
```diff
@@ -0,0 +1,339 @@
+use quote::quote;
```
ordian, line 1: nit: some of the files are missing the license header

node/subsystem-types/src/errors.rs
```diff
@@ -77,3 +80,70 @@ impl std::fmt::Display for RecoveryError {
...
+}
+
+// impl AnnotateErrorOrigin for SubsystemError {
+// 	fn with_origin(self, origin: &'static str) -> Self {
+// 		Self::FromOrigin {
+// 			origin,
+// 			source: Box::new(self),
+// 		}
+// 	}
+// }
```
ordian, line 139: why do we need a trait for this?

node/subsystem-types/src/messages/network_bridge_event.rs
```diff
@@ -70,6 +70,8 @@ impl<M> NetworkBridgeEvent<M> {
 		where T: 'a + Clone, &'a T: TryFrom<&'a M, Error = WrongVariant>
 	{
 		Ok(match *self {
+			NetworkBridgeEvent::PeerMessage(ref peer, ref msg)
+				=> NetworkBridgeEvent::PeerMessage(peer.clone(), <&'a T>::try_from(msg)?.clone()),
```
ordian, line 74: post-merge remnant

ordian (commented):
small nits from a quick glance

Cargo.toml
```diff
@@ -23,6 +23,12 @@ tempfile = "3.2.0"
 
 [workspace]
 members = [
+	"bridges/primitives/chain-kusama",
+	"bridges/primitives/chain-polkadot",
+	"bridges/primitives/chain-rococo",
+	"bridges/primitives/chain-westend",
+	"bridges/primitives/runtime",
```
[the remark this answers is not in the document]
drahnr: Yeah, merging was a two day job really.. so I have to revisit those changes.

node/network/bitfield-distribution/src/lib.rs
```diff
@@ -838,3 +839,6 @@ impl metrics::Metrics for Metrics {
 	}
 }
 
+
+#[cfg(test)]
+mod tests;
```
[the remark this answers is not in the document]
drahnr: Right now actually most of them are at the bottom, and it seems a majority of people actually prefer the test module to be at the end of the file.

node/subsystem-types/src/errors.rs
drahnr: Not necessarily. It just seemed illogical to restrict it to the `SubsystemError` (or now `OverseerError` when it would be both viable for user defined as well as the proc macro internally generated error type. TBD

node/network/bitfield-distribution/src/lib.rs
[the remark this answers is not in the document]
drahnr: I am still leaning towards undoing the previous move-up, `mod tests {}` usually are at the bottom, and a `mod tests;` would indicate, there are tests just elesewhere at the expected location.

node/overseer/src/lib.rs
```diff
@@ -61,70 +61,89 @@
...
+	FromOverseer,
+	ToOverseer,
+	MapSubsystem,
+};
+pub use polkadot_overseer_gen as gen;
 
-	fn map_subsystem(&self, sub: T) -> Self::Output;
+/// Boiler plate reduction, `Signal` and `AllMessages` are fixed for one overseer.
+/// So the only variation for external entities is `M`.
+pub trait SubsystemCtx<M> : SubsystemContext<Message = M, Signal = OverseerSignal, AllMessages = AllMessages> {
```
eskimor, line 137: Looks indeed useful, but does not seem to be used at all - why?

node/overseer/src/lib.rs
```diff
@@ -534,798 +331,104 @@ pub async fn forward_events<P: BlockchainEvents<Block>>(
...
-	signals: metered::MeteredReceiver<OverseerSignal>,
-	messages: SubsystemIncomingMessages<M>,
-	to_subsystems: OverseerSubsystemSender,
-	to_overseer: metered::UnboundedMeteredSender<ToOverseer>,
-	signals_received: SignalsReceived,
-	pending_incoming: Option<(usize, M)>,
-	metrics: Metrics,
-}
+/// The `Overseer` itself.
+#[overlord(
```
eskimor, line 335: Nice naming :laughing:

node/core/approval-voting/src/lib.rs
```diff
@@ -723,13 +725,15 @@ enum Action {
...
+async fn run<Context>(
+	mut ctx: Context,
 	mut subsystem: ApprovalVotingSubsystem,
 	clock: Box<dyn Clock + Send + Sync>,
 	assignment_criteria: Box<dyn AssignmentCriteria + Send + Sync>,
 ) -> SubsystemResult<()>
-	where C: SubsystemContext<Message = ApprovalVotingMessage>
+where
+	Context: SubsystemContext<Message = ApprovalVotingMessage>,
+	Context: overseer::SubsystemContext<Message = ApprovalVotingMessage>,
```
eskimor, line 736: I don't quite get why this seemingly redundant trait bound is needed.

eskimor (commented):
Looks good overall.

node/overseer/src/lib.rs
drahnr: Huh, that should not exist anymore. This also did not proof viable. There are some usage patterns that want to omit `type Message` for usage in multiple subsystem, which can not be achieved with generics.

node/core/approval-voting/src/lib.rs
drahnr: It's this trade-off:

`Subsystem<Message=X> + overseer::Subsystem<Message=X>`

OR

`overseer::Subsystem<Message=X, AllMessages=AllMessages, Signal=OverseerSignal, Error=SubsystemError>`

since the other types need to be nailed down.

ordian: I personally like the latter, because it's less magic even though it's more verbose. Is it possible to provide defaults fot the associated types except for the `Message`?

drahnr: They are unstable unfortunately, tried that as well.

eskimor: Weird, isn't one of those traits a super trait of the other?

node/core/approval-voting/src/lib.rs
```diff
@@ -850,8 +854,7 @@ async fn run<C>(
 //
 // returns `true` if any of the actions was a `Conclude` command.
 async fn handle_actions(
-	ctx: &mut impl SubsystemContext,
-	state: &mut State<impl DBReader>,
+	ctx: &mut (impl SubsystemContext<Message = ApprovalVotingMessage> + overseer::SubsystemContext<Message = ApprovalVotingMessage>),	state: &mut State<impl DBReader>,
```
drahnr, line 857:
```suggestion
	ctx: &mut (impl SubsystemContext<Message = ApprovalVotingMessage> + overseer::SubsystemContext<Message = ApprovalVotingMessage>),
	state: &mut State<impl DBReader>,
```

node/overseer/src/lib.rs
eskimor: :bulb:

eskimor: Wait, can those usages not just pick a generic `M`?

node/core/approval-voting/src/lib.rs
ordian: Is it possible to use generics with defaults for `AllMessages, Signal, Error` as a workaround?

drahnr: Yes, but that's not enough since `trait X: Y` and using `Context: X` does not resolve the `fn`s of `Y`.

Alt: Reversing the dependency `trait Y: X` and forwarding all `fn`s is impossible due to duplicate symbols in the final type and one has to use fully qualified syntax, which is an order of magnitude worse.

drahnr: > Is it possible to use generics with defaults for `AllMessages, Signal, Error` as a workaround?

You mean using smth. like `trait overseer::Subsystem<AllMessages=AllMessages, Signal=OverseerSignal, Error=SubsystemError> {  type Message; type Sender;  /* .. */}`.

Technically yes, but `SubsystemError` and the ident `AllMessages` are user provided. So that's generated code and at that point we're back to moving the traits into the proc-macro / `quote!{}` which was something I wanted to avoid

drahnr: But maybe for theose two traits it might be the better solution, in order to avoid all the clutter across all subsystems.

eskimor: I see :cry: . A simple `as_ctx()` method to expose the super trait would work, I guess. But not sure that's better.

ordian: > > Is it possible to use generics with defaults for `AllMessages, Signal, Error` as a workaround?
> 
> You mean using smth. like `trait overseer::Subsystem<AllMessages=AllMessages, Signal=OverseerSignal, Error=SubsystemError> { type Message; type Sender; /* .. */}`.
> 
> Technically yes, but `SubsystemError` and the ident `AllMessages` are user provided. So that's generated code and at that point we're back to moving the traits into the proc-macro / `quote!{}` which was something I wanted to avoid

Yes, that what I was thinking. Maybe introduce yet another trait :grimacing:  close to user provided structs
```rust
trait RealSubsystemContext {
  type Message;
  ...
}
impl<T: RealSubsystemContext> for overseer::SubsystemContext {
  types AllMessages = AllMessages;
  ...
}
```

node/overseer/overseer-gen/Cargo.toml
```diff
@@ -0,0 +1,26 @@
...
+async-trait = "0.1"
+thiserror = "1"
+metered = { package = "metered-channel", path = "../../metered-channel" }
+polkadot-overseer-gen-proc-macro = { path = "./proc-macro" }
+polkadot-node-network-protocol = { path = "../../network/protocol"}
+
+# trait SpawnNamed
+sp-core = { git = "https://github.com/paritytech/substrate", branch = "master" }
+futures-timer = "3.0.2"
+tokio-util = { version = "0.6.7", features = [ "time" ] }
```
ordian, line 20: what's tokio-util used for?

drahnr: Hmm, I think that's copy paste. I'll add some `#[deny(unused_dependencies)]`

node/overseer/overseer-gen/src/lib.rs
```diff
@@ -0,0 +1,519 @@
...
+}
+
+/// An error type that describes faults that may happen
+///
+/// These are:
+///   * Channels being closed
+///   * Subsystems dying when they are not expected to
+///   * Subsystems not dying when they are told to die
+///   * etc.
+// FIXME XXX make generic over the source error of FromOrigin
```
ordian, line 249: is this tracked in #3427? what would be the benefit of that change?

node/overseer/src/lib.rs
```diff
@@ -1421,490 +533,118 @@ where
...
+			.chain_api(all_subsystems.chain_api)
+			.collation_generation(all_subsystems.collation_generation)
+			.collator_protocol(all_subsystems.collator_protocol)
+			.approval_distribution(all_subsystems.approval_distribution)
+			.approval_voting(all_subsystems.approval_voting)
+			.gossip_support(all_subsystems.gossip_support)
+			.leaves(Vec::from_iter(
+				leaves.into_iter().map(|BlockInfo { hash, parent_hash: _, number }| (hash, number))
+			))
+			.known_leaves(LruCache::new(128))// FIXME verify this number
```
ordian, line 589: would be nice to bring the constant back
```rust
/// Store 2 days worth of blocks, not accounting for forks,
/// in the LRU cache. Assumes a 6-second block time.
const KNOWN_LEAVES_CACHE_SIZE: usize = 2 * 24 * 3600 / 6;

File: node/overseer/src/lib.rs [hunk 19]
```diff
@@ -1421,490 +533,118 @@ where
...
-				_ = self.running_subsystems.next() => {
-					if self.running_subsystems.is_empty() {
-						break;
-					}
-				},
-				_ = stop_delay => break,
-				complete => break,
-			}
-		}
+		::std::mem::drop(
```
ordian, line 642: what's mem::drop for?

node/overseer/src/lib.rs
```diff
@@ -534,798 +319,104 @@ pub async fn forward_events<P: BlockchainEvents<Block>>(
...
 
-struct SubsystemMeterReadouts {
-	bounded: metered::Readout,
-	unbounded: metered::Readout,
-	signals: metered::Readout,
-}
+	#[subsystem(no_dispatch, GossipSupportMessage)]
+	gossip_support: GossipSupport,
 
+	#[subsystem(no_dispatch, wip, DisputeCoordinatorMessage)]
```
ordian, line 383: `wip` :sunglasses:

drahnr: yeah, I realized that this was the case for a few subsystems, so "formalized" this into a annotation :shrug:

node/overseer/src/lib.rs
drahnr: Moving to a  `let _ = ..` for sake of idiomatics, I don't quite remember why I did it this way.

node/overseer/overseer-gen/src/lib.rs
drahnr: That particular one is obsolete, made sure to add references to all TODO tags in `node/overseer` and have a bullet point in #3427 chores section

node/overseer/overseer-gen/proc-macro/src/lib.rs
```diff
@@ -0,0 +1,82 @@
...
+
+use proc_macro2::TokenStream;
+use syn::{parse2, Result};
+
+mod impl_builder;
+mod impl_misc;
+mod impl_overseer;
+mod parse_attr;
+mod parse_struct;
+// mod impl_replace;
```
ordian, line 27: is it not used?

drahnr: Currently `AllSubsystems` exists. This on the other hand will replace the `AllSubsystemsGen` proc-macro and work directly on the `#[overlord(..)]` generated struct and replace subsystems there as needed.

I will remove that part and re-introduce when ready.

node/subsystem/src/lib.rs
```diff
@@ -14,366 +14,77 @@
...
-			}
-		});
-
-		SpawnedSubsystem {
-			name: "dummy-subsystem",
-			future,
-		}
-	}
+	type Message = <Self as polkadot_overseer::gen::SubsystemContext>::Message;
+	type Sender = <Self as polkadot_overseer::gen::SubsystemContext>::Sender;
```
ordian, line 89: have you tried instead do
```rust
trait SubsystemContext {
  type Message;
  ...
}
impl<T: SubsystemContext> overseer::SubsystemContext for T {
  types AllMessages = AllMessages;
  ...
}
```
Would it solve the problem of having to specify two traits?

node/core/approval-voting/src/lib.rs
drahnr: It looks simple, but it's a rather evasive problem. The last thing that I can think of is providing a `Deref` impl for `Subsystem` decaying into `overseer::Subsystem` which could work.

node/subsystem/src/lib.rs
drahnr: The issue will stay, since the the information of the associated types for which we impl the `overseer::SubsystemContext` will not propate to the `fn foo(ctx: impl SubsystemContext)` and hence we can not use concrete types there, which is what we want..

ordian (approved):
Haven't looked deeply into proc-macros, but great work overall! :muscle:

node/overseer/src/lib.rs
drahnr: Whoops, but I am tempted to leave it as is :sweat_smile:

node/subsystem/src/lib.rs
drahnr: That was indeed a variation I did not try before, but the `overseer::SubsystemContext` is not defined in the same crate, and hence `impl<T> X for T {}` will not work either:

```
error[E0210]: type parameter `T` must be used as the type parameter for some local type (e.g., `MyStruct<T>`)
   --> node/subsystem/src/lib.rs:110:6
    |
110 | impl<T> polkadot_overseer::gen::SubsystemContext for T
    |      ^ type parameter `T` must be used as the type parameter for some local type
    |
    = note: implementing a foreign trait is only possible if at least one of the types for which it is implemented is local
    = note: only traits defined in the current crate can be implemented for a type parameter

```

drahnr: As a conclusion there are two options:

* accept as is
* generate the traits with the proper types and avoid the clutter

node/core/chain-api/src/tests.rs
```diff
@@ -0,0 +1,304 @@
+use super::*;
```
bkchr, line 1: Missing license header.

node/core/chain-api/src/tests.rs
```diff
@@ -0,0 +1,304 @@
...
+
+use polkadot_primitives::v1::{Hash, BlockNumber, BlockId, Header};
+use polkadot_node_primitives::BlockWeight;
+use polkadot_node_subsystem_test_helpers::{make_subsystem_context, TestSubsystemContextHandle};
+use sp_blockchain::Info as BlockInfo;
+use sp_core::testing::TaskExecutor;
+
+#[derive(Clone)]
+struct TestClient {
+    blocks: BTreeMap<Hash, BlockNumber>,
```
bkchr, line 15: Spaces here and probably in every created file.

node/core/bitfield-signing/src/lib.rs
```diff
@@ -312,72 +312,4 @@ impl JobTrait for BitfieldSigningJob {
...
-							break
-						},
-						Err(e) => panic!("Failed: {:?}", e),
-					},
-				}
-			}
-		});
-	}
-}
+mod tests;
```
bkchr, line 315: Please move to the top as we do it everywhere.

node/core/bitfield-signing/src/tests.rs
```diff
@@ -0,0 +1,67 @@
+use super::*;
+use futures::{pin_mut, executor::block_on};
+use polkadot_primitives::v1::{CandidateHash, OccupiedCore};
+use polkadot_node_subsystem::messages::AllMessages;
+
+fn occupied_core(para_id: u32, candidate_hash: CandidateHash) -> CoreState {
+    CoreState::Occupied(OccupiedCore {
```
bkchr, line 7: Spaces

node/core/bitfield-signing/src/tests.rs
```diff
@@ -0,0 +1,67 @@
+use super::*;
```
bkchr, line 1: License header

node/overseer/overseer-gen/proc-macro/src/parse_struct.rs
```diff
@@ -0,0 +1,367 @@
...
+				}
+				let mut consumes_paths = Vec::with_capacity(attrs.len());
+				let attr_tokens = attr_tokens.clone();
+				let variant: SubSystemTag = syn::parse2(attr_tokens.clone())?;
+				if variant.consumes.len() != 1 {
+					return Err(Error::new(attr_tokens.span(), "Exactly one message can be consumed per subsystem."));
+				}
+				consumes_paths.extend(variant.consumes.into_iter());
+
+				if consumes_paths.is_empty() {
```
bkchr, line 296: That is already checked above.

node/overseer/overseer-gen/proc-macro/src/parse_struct.rs
```diff
@@ -0,0 +1,367 @@
...
+						err
+					});
+				}
+				let mut consumes_paths = Vec::with_capacity(attrs.len());
+				let attr_tokens = attr_tokens.clone();
+				let variant: SubSystemTag = syn::parse2(attr_tokens.clone())?;
+				if variant.consumes.len() != 1 {
+					return Err(Error::new(attr_tokens.span(), "Exactly one message can be consumed per subsystem."));
+				}
+				consumes_paths.extend(variant.consumes.into_iter());
```
bkchr, line 294: Why don't we extract the single value here directly?

node/overseer/overseer-gen/proc-macro/src/parse_struct.rs
```diff
@@ -0,0 +1,367 @@
...
+				subsystems.push(SubSysField {
+					name: ident,
+					generic,
+					consumes: consumes_paths[0].clone(),
+					no_dispatch: variant.no_dispatch,
+					wip: variant.wip,
+					blocking: variant.blocking,
+				});
+			} else {
+				let field_ty: Path = try_type_to_path(ty, ident.span())?;
```
bkchr, line 318:
```suggestion
				let field_ty = try_type_to_path(ty, ident.span())?;
```

node/overseer/overseer-gen/proc-macro/src/parse_struct.rs
```diff
@@ -0,0 +1,367 @@
...
+				let generic: bool =
+					if let Some(ident) = field_ty.get_ident() { baggage_generics.contains(ident) } else { false };
```
bkchr, line 320:
```suggestion
				let generic = field_ty.get_ident().map(|i| baggage_generics.contains(ident)).unwrap_or_default();
```

node/overseer/overseer-gen/src/lib.rs
```diff
@@ -0,0 +1,519 @@
...
+pub use std::time::Duration;
+use std::sync::atomic::{self, AtomicUsize};
+use std::sync::Arc;
+
+#[doc(hidden)]
+pub use futures_timer::Delay;
+#[doc(hidden)]
+pub use futures_util::stream::StreamExt;
+#[doc(hidden)]
+pub use futures_util::future::FutureExt;
```
bkchr, line 103: They are just re-exported by `future` as well.
```suggestion
```

node/overseer/overseer-gen/src/lib.rs
```diff
@@ -0,0 +1,519 @@
...
+pub use tracing;
+#[doc(hidden)]
+pub use metered;
+#[doc(hidden)]
+pub use sp_core::traits::SpawnNamed;
+
+#[doc(hidden)]
+pub use futures::{
+	self,
+	select,
```
bkchr, line 75:
```suggestion
	select,
	StreamExt,
	FutureExt,
```

node/overseer/overseer-gen/src/lib.rs
```diff
@@ -0,0 +1,519 @@
...
+	/// dedicated thread pool.
+	SpawnBlockingJob {
+		/// Name of the task to spawn which be shown in jaeger and tracing logs.
+		name: &'static str,
+		/// The future to execute.
+		s: BoxFuture<'static, ()>,
+	},
+}
+
+use std::fmt;
```
bkchr, line 132:
```suggestion
```

node/overseer/overseer-gen/src/lib.rs
```diff
@@ -0,0 +1,519 @@
...
+	},
+	channel::{mpsc, oneshot},
+};
+#[doc(hidden)]
+pub use std::pin::Pin;
+#[doc(hidden)]
+pub use async_trait::async_trait;
+
+#[doc(hidden)]
+pub use std::time::Duration;
```
bkchr, line 94: Why?

node/overseer/overseer-gen/src/lib.rs
```diff
@@ -0,0 +1,519 @@
...
+use std::sync::atomic::{self, AtomicUsize};
+use std::sync::Arc;
```
bkchr, line 96:
```suggestion
use std::sync{Arc, atomic::{self, AtomicUsize}};
```

node/service/src/lib.rs
```diff
@@ -56,7 +56,7 @@ pub use {
 	sp_authority_discovery::AuthorityDiscoveryApi,
 	sc_client_api::AuxStore,
 	polkadot_primitives::v1::ParachainHost,
-	polkadot_overseer::{Overseer, OverseerHandler},
+	polkadot_overseer::{Overseer, Handler},
```
bkchr, line 59: Why is this called handler and not handle?

node/overseer/overseer-gen/proc-macro/src/impl_builder.rs
```diff
@@ -0,0 +1,358 @@
...
+			}
+		})
+		.collect::<Vec<_>>();
+
+	let generics = quote! {
+		< S, #( #baggage_generic_ty, )* >
+	};
+	let where_clause = quote! {
+		where
+			S: ::polkadot_overseer_gen::SpawnNamed,
```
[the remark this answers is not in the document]
drahnr: My point being, we are not moving the `mod test { #[test] fn foo() {..}}` up, so why not have all the tests in one place - at the bottom. That's the rationale. After all it's not a "regular" module anyways.

We just recently started moving them up.

WDYT?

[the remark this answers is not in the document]
bkchr: And then I'm going down to search for the test and find a mod declaration and need to open the file anyway. While you could already spot this when opening the file (while most people probably already in their editor see the tests file).

And we also do this everywhere. In Substrate, in Polkadot. We don't need to start shuffling stuff now randomly around IMHO.

node/overseer/overseer-gen/src/lib.rs
drahnr: I'll move it up, but it's needed two lines below.

node/overseer/overseer-gen/proc-macro/src/impl_builder.rs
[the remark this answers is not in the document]
drahnr: If you find the `mod tests` decl it should be just one keystroke away to resolve it and jump to the file, right? Imho it's still easier to have a consistent location than searching for `mod tests` in between a dozen `use` and other `mod` statements. But I guess this boils down to pref, and if this is already consistent within substrate, then lets just stick to that.

node/subsystem/src/lib.rs
eskimor: I don't have a strong opinion. Less clutter is good, but if that's cumbersome to achieve, I think it is fine.

node/service/src/lib.rs
drahnr: Because that's what it does. While it is a handle, it does the handling of the overseer, and thus is a handler. But I see your point, renamed.

node/overseer/overseer-gen/proc-macro/src/impl_builder.rs
```diff
@@ -0,0 +1,358 @@
...
+			}
+		}
+
+		#[allow(missing_docs)]
+		struct Blocking;
+		impl TaskKind for Blocking {
+			fn launch_task<S: SpawnNamed>(spawner: &mut S, name: &'static str, future: BoxFuture<'static, ()>) {
+				spawner.spawn_blocking(name, future)
+			}
+		}
```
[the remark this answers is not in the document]
drahnr: It doesn't have to be and this works nicely without much `match`? This is a small scope impact and hence does not change compile time.

node/overseer/overseer-gen/proc-macro/src/impl_overseer.rs
```diff
@@ -0,0 +1,264 @@
...
+					match instance.tx_signal.send(signal).timeout(SIGNAL_TIMEOUT).await {
+						None => {
+							Err(#error_ty :: from(
+								::polkadot_overseer_gen::OverseerError::SubsystemStalled(instance.name)
+							))
+						}
+						Some(res) => {
+							let res = res.map_err(Into::into);
+							if res.is_ok() {
+								instance.signals_received += 1;
```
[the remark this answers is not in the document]
drahnr: I'd like to defer this to separate PR, this was present before the refactoring and should be done independently of this refactor.

node/overseer/overseer-gen/proc-macro/src/impl_dispatch.rs
```diff
@@ -0,0 +1,75 @@
...
+		.map(|ssf| ssf.consumes.clone())
+		.collect::<Vec<Path>>();
+
+	let mut ts = TokenStream::new();
+	if let Some(extern_network_ty) = &info.extern_network_ty.clone() {
+		ts.extend(quote! {
+			impl #message_wrapper {
+				/// Generated dispatch iterator generator.
+				pub fn dispatch_iter(extern_msg: #extern_network_ty) -> impl Iterator<Item=Self> + Send {
+					None
```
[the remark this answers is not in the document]
drahnr: This will only work with edition 2021 and onwards, until then `[].into_iter()` exhibits the old behavior.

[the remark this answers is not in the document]
drahnr: Using `array::IntoIter::new([..])`  around works.

node/overseer/overseer-gen/proc-macro/src/parse_attr.rs
```diff
@@ -0,0 +1,227 @@
...
+			message_channel_capacity,
+			extern_event_ty: event,
+			extern_signal_ty: signal,
+			extern_error_ty: error,
+			extern_network_ty: network,
+			outgoing_ty: outgoing,
+			message_wrapper,
+		})
+	}
+}
```
drahnr, line 227: @bkchr could you take a glance at this if there are further changes mandated? Thanks! Much appreciated.

drahnr:
bot merge

node/overseer/overseer-gen/proc-macro/src/parse_attr.rs
Lldenaurois: LGTM

node/subsystem-util/src/lib.rs
```diff
@@ -148,13 +179,14 @@ macro_rules! specialize_requests {
 		#[doc = "Request `"]
 		#[doc = $doc_name]
 		#[doc = "` from the runtime"]
-		pub async fn $func_name(
+		pub async fn $func_name (
```
Lldenaurois, line 182: nit: extra space

Lldenaurois (approved):
Made a few small suggestions. Overall, I think this is fine to merge as-is though

drahnr:
I'll take care of the nits in a follow up

PR #2963, opened 2021-04-30
More secure `Signed` implementation

### LINKED ISSUE #2944: Signed struct should not have `Encode` and `Decode` instances.
The [`Signed`](https://github.com/paritytech/polkadot/blob/3489696986829f9bbcaa9523be48b24c70ac3220/primitives/src/v0.rs#L900) type has `Encode` and `Decode` instances, but at the same time tries to ensure that signatures are always valid. E.g. `Signed::new` only allows creation if signature is valid, fields are private, .... Still we can transfer a `Signed` via the network and need to depend on discipline to actually call `check_signature` on the received data.

Besides the obvious security drawback, this design also leads to needless double checking of signatures in case we get a `Signed` via two code paths. One where we need to construct it via `new` and one where we receive it directly. The one that gets constructed will have its signature checked twice.

I would suggest to introduce a variant of `Signed` either via an additional type parameter or via an isomorphic struct, let's call it `UncheckedSigned`, with the following properties:

- `Signed` -> `UncheckedSigned` straight forward conversion.
- `UncheckedSigned` -> `Option<Signed>` .. only results in Some(Signed), when signature is valid
- Only `UncheckedSigned` has an `Encode`/`Decode` instance.

This way we never ever need to check signature on `Signed` as it is guaranteed to be valid and at the same time we are forced to check it if we receive untrusted data.

eskimor:
Fixes #2944 

Also:

- A bunch of cleanup
- Check signature on collator side on seconded notifications
- Couple of convenience functions
- Get rid of redundant signature checks (in backing for example) - we now have type level guarantee that signature has been already checked.
- Collator protocol: Better connection management - only drop connections we really no longer need.
- `CollatorProtocol` has changed as we now also transmit the relay parent for the signing context in order to actually check the signature.

_Important_: This PR changes the wire protocol for collations.


node/network/collator-protocol/src/error.rs
```diff
@@ -0,0 +1,94 @@
...
+
+use polkadot_node_subsystem_util::{Fault, runtime, unwrap_non_fatal};
+
+use crate::LOG_TARGET;
+
+/// General result.
+pub type Result<T> = std::result::Result<T, Error>;
+/// Result for non fatal only failures.
+#[allow(dead_code)]
+pub type NonFatalResult<T> = std::result::Result<T, NonFatal>;
```
bkchr, line 32: Where is this used?

node/network/collator-protocol/src/error.rs
```diff
@@ -0,0 +1,94 @@
...
+/// Fatal runtime errors.
+#[derive(Debug, Error)]
+pub enum Fatal {
+	/// Receiving subsystem message from overseer failed.
+	#[error("Receiving message from overseer failed")]
+	SubsystemReceive(#[source] SubsystemError),
+
+	/// Errors coming from runtime::Runtime.
+	#[error("Error while accessing runtime information")]
+	Runtime(#[from] #[source] runtime::Fatal),
```
bkchr, line 68: What is a fatal runtime error that should bring down this subsystem?

eskimor: For runtime communication, this is whenever the actual communication channel gets closed.

bkchr: What is runtime communication?

primitives/src/v0.rs
```diff
@@ -951,69 +950,172 @@ impl<Payload: EncodeAs<RealPayload>, RealPayload: Encode> Signed<Payload, RealPa
...
+		let r = UncheckedSigned::sign(keystore, payload, context, validator_index, key).await?;
+		Ok(r.map(Self))
+	}
 
-		let signature = match signature {
-			Some(sig) => sig.try_into().map_err(|_| KeystoreError::KeyNotSupported(ValidatorId::ID))?,
-			None => return Ok(None),
-		};
+	/// Try to convert from `UncheckedSigned` by checking the signature.
+	pub fn from_unchecked<H: Encode>(
```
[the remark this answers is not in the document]
eskimor: Because it needs additional data (signing context and `ValidatorId`) to do the "conversion". But the `try_from_unchecked` is more consistent - will change that.

node/network/collator-protocol/src/error.rs
eskimor: It is not, but I still want to have it as a general framework code. E.g. every subsystem has `Fatal` and `NonFatal` errors and a `NonFatalResult`, .... In the spirit of consistency I would like to leave it, although it is not used right now. I would definitely like to avoid for example, if it is needed at one point that people make up a new name.

bkchr: Then you should write something in the guide, but dead code isn't a good practice and if someone needs this, it is restored in seconds.

node/network/collator-protocol/src/error.rs
drahnr: `#[from]` is a superset of `#[cause]` so `#[cause]` is a bit redudant: https://docs.rs/thiserror/1.0.21/thiserror/

node/subsystem-util/src/runtime/mod.rs
```diff
@@ -72,7 +75,7 @@ pub struct ValidatorInfo {
 
 impl RuntimeInfo {
 	/// Create a new `RuntimeInfo` for convenient runtime fetches.
-	pub fn new(keystore: SyncCryptoStorePtr) -> Self {
+	pub fn new(keystore: Option<SyncCryptoStorePtr>) -> Self {
```
drahnr, line 78: I don't quite see where the keystore is ever `None` :see_no_evil:

primitives/src/v1/signed.rs
```diff
@@ -0,0 +1,282 @@
...
+	///
+	/// The signature is checked as part of the process.
+	#[cfg(feature = "std")]
+	pub fn new<H: Encode>(
+		payload: Payload,
+		validator_index: ValidatorIndex,
+		signature: ValidatorSignature,
+		context: &SigningContext<H>,
+		key: &ValidatorId,
+	) -> Option<Self> {
```
drahnr, line 68: nit: Isn't this rather a `Result` with a unit `struct SignatureVerificationError`?

node/network/statement-distribution/src/lib.rs
```diff
@@ -801,16 +781,17 @@ impl ActiveHeadData {
 fn check_statement_signature(
 	head: &ActiveHeadData,
 	relay_parent: Hash,
-	statement: &SignedFullStatement,
-) -> std::result::Result<(), ()> {
+	statement: UncheckedSignedFullStatement,
+) -> std::result::Result<SignedFullStatement, UncheckedSignedFullStatement> {
```
drahnr, line 785: :heart: this is nice!

drahnr (commented):
Generally this LGTM besides a few nits.

Iiuc this unifies the signature check into a type, and uses that to only verify it once and hence reduce the load.

The open question that is left is, if the path from the point of checking where it was to where it is now, if that opens up any. TBD

drahnr (approved):
Second sweep, LGTM

node/subsystem-util/src/runtime/mod.rs
eskimor: In the collator protocol (collator side) we don't need it, because we are not a validator.

primitives/src/v1/signed.rs
eskimor: It was this way before and I also think it is pretty obvious what `None` means in this case. To be honest you have a point, but if you do not object, I am going to be lazy and just merge if this is the only nit :grimacing:

PR #2970, opened 2021-05-03
Enable Pallet XCM for Kusama & Westend

gavofyork:
not-live because it doesn't touch Polkadot runtime.


shawntabrizi (approved):

runtime/kusama/src/lib.rs
```diff
@@ -1146,6 +1146,79 @@ impl xcm_executor::Config for XcmConfig {
...
+impl frame_support::traits::Contains<(MultiLocation, Xcm<Call>)> for OnlyWithdrawTeleportForAccounts {
+	fn contains((ref origin, ref msg): &(MultiLocation, Xcm<Call>)) -> bool {
+		use xcm::v0::{
+			Xcm::WithdrawAsset, Order::{BuyExecution, InitiateTeleport, DepositAsset},
+			MultiAsset::{All, ConcreteFungible}, Junction::{AccountId32, Plurality, Parachain},
+			MultiLocation::{Null, X1},
+		};
+		match origin {
+			// Root and council are are allowed to execute anything.
+			Null | X1(Plurality { .. }) => true,
```
shawntabrizi, line 1173: We don't do any verification on the `Plurality`. Seems fine now with only council, but could be a foot gun later.

gavofyork: i'm assumign that collectives in general don't need the same safeguards as random accounts (which are trivially sybillable).

shawntabrizi: actually we allow any 1 / N council origin to pass here?

shawntabrizi (changes requested):
I think 1 / N council here is given full access to XCM

gavofyork:
yeah. that's fine.

shawntabrizi (approved):
Note to make sure that statemint or any downstream parachain correctly checks the threshold of the plurality origin

cc @joepetrowski

andresilva (approved):

runtime/westend/src/lib.rs
```diff
@@ -878,6 +879,72 @@ impl xcm_executor::Config for XcmConfig {
...
+impl frame_support::traits::Contains<(MultiLocation, Xcm<Call>)> for OnlyWithdrawTeleportForAccounts {
+	fn contains((ref origin, ref msg): &(MultiLocation, Xcm<Call>)) -> bool {
+		use xcm::v0::{
+			Xcm::WithdrawAsset, Order::{BuyExecution, InitiateTeleport, DepositAsset},
+			MultiAsset::{All, ConcreteFungible}, Junction::{AccountId32, Parachain},
+			MultiLocation::{Null, X1},
+		};
+		match origin {
+			// Root is allowed to execute anything.
+			Null => true,
```
apopiak, line 899: It seems that this will mean we cannot test plurality XCMs on Westend?

andresilva: We don't have a council on Westend since we have sudo. Your point might still be relevant though.

apopiak:
> Note to make sure that statemint or any downstream parachain correctly checks the threshold of the plurality origin
> 
> cc @joepetrowski

It will check `is_majority` for now, could check other percentages later.

runtime/westend/src/lib.rs
gavofyork: We could introduce a council on Westend, but I think it's probably enough to test on Rococo.

rphmeier (approved):

andresilva:
bot merge

PR #2974, opened 2021-05-04
Change Statemint ID (not system para)

joepetrowski:
Statemint is not a system level para and thus requires an ID > 1000.


riusricardo (approved):
This will work after a Rococo reset or if we relaunch Statemint to take this new para Id.

kianenigma (approved):

rphmeier (approved):

PR #2960, opened 2021-04-29
Update Bridges Subtree

HCastano:
This PR bumps the `parity-bridges-common` Subtree. One notable thing this adds support
for is relaying headers between the Rococo and Wococo networks. However, a follow up
Polkadot PR is still required in order to add the pallets to Rococo runtime.

PR #2978, opened 2021-05-04
Bump Bridges Subtree

HCastano:
This PR supersedes #2960, which updated the `parity-bridges-common` subtree in
preparation for Wococo<>Rococo bridge support.

Turns out that if we squash-merge these subtree update PRs then the subtree history in
future updates won't be super accurate (but the diff will look correct). The history,
which is what you see in this commit for example 70985cee89da5ce098fa7d445d269646e6f2c3eb, will always start at the same 
commit instead of showing the changes between the last update and the current update.

In our case that means every subtree update's commit history will be from:

`f51fb59d Prefix in relay loops logs (#920) -> parity-bridges-common@HEAD`

Instead of:

`parity-bridges-common@HEAD-at-last-update -> parity-bridges-common@HEAD`

To re-emphasize, please **do not** squash-merge this - we need to do a regular merge 🙂


tomusdrw (approved):
Looks good, as mentioned in description please don't **squash**, but regular-merge this.

PR #2988, opened 2021-05-06
Make ParaId value of public paras non-discretionary

gavofyork:
Original `register` function left in to avoid transaction prototypes changing. Semantics altered to ensure that they're in line with the new rules.

### DIFF
--- a/parachain/src/primitives.rs
+++ b/parachain/src/primitives.rs
@@ -122,9 +122,13 @@ impl From<i32> for Id {
 }
 
 const USER_INDEX_START: u32 = 1000;
+const PUBLIC_INDEX_START: u32 = 2000;
 
 /// The ID of the first user (non-system) parachain.
 pub const LOWEST_USER_ID: Id = Id(USER_INDEX_START);
+
+/// The ID of the first publicly registerable parachain.
+pub const LOWEST_PUBLIC_ID: Id = Id(PUBLIC_INDEX_START);
 
 impl Id {
 	/// Create an `Id`.
--- a/parachain/src/primitives.rs
+++ b/parachain/src/primitives.rs
@@ -155,6 +155,14 @@ impl sp_std::ops::Add<u32> for Id {
 	}
 }
 
+impl sp_std::ops::Sub<u32> for Id {
+	type Output = Self;
+
+	fn sub(self, other: u32) -> Self {
+		Self(self.0 - other)
+	}
+}
+
 #[derive(Clone, Copy, Default, Encode, Decode, Eq, PartialEq, Ord, PartialOrd, RuntimeDebug)]
 pub struct Sibling(pub Id);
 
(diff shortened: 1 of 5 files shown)


runtime/common/src/paras_registrar.rs
```diff
@@ -185,7 +188,15 @@ decl_module! {
...
-			ensure!(id >= LOWEST_USER_ID, Error::<T>::InvalidParaId);
+			let valid_id = NextFreeParaId::mutate(|id| {
+				if *id < LOWEST_USER_ID {
+					*id = LOWEST_USER_ID
+				}
+				let result = *id;
+				*id = *id + 1;
+				result
+			});
+			ensure!(id == valid_id, Error::<T>::InvalidParaId);
```
ordian, line 199: If I read this correctly, this bumps `NextFreeParaId` slot even if the `ParaId` is occupied.
This should be fixed by having `ensure(id == max(LOWEST_USER_ID, NextFreeParaId::get() + 1))` above or something like that.

Also it's not clear to me what's the purpose of this function given there is newly introduced `register_next`? Are we worried about breaking changes? Or to give a try to occupy a given `ParaId`?

gavofyork: As I wrote already:

image

gavofyork: > If I read this correctly, this bumps NextFreeParaId slot even if the ParaId is occupied.

Fixed in latest logic.

runtime/common/src/paras_registrar.rs
```diff
@@ -271,6 +283,30 @@ decl_module! {
...
+		///
+		/// This function will queue the new Para Id to be a parathread.
+		/// Using the Slots pallet, a parathread can then be upgraded to get a
+		/// parachain slot.
+		///
+		/// This function must be called by a signed origin.
+		///
+		/// The origin must pay a deposit for the registration information,
+		/// including the genesis information and validation code. ParaId
+		/// must be greater than or equal to 1000.
```
bkchr, line 297:
```suggestion
		/// including the genesis information and validation code. The ParaId
		/// will be set to the next available id.
```

Do we maybe want to dispatch an event with the assigned paraid?

bkchr (commented):
I assume we don't care about reclaiming of ParaIds for now?

rphmeier (approved):

shawntabrizi (approved):

shawntabrizi:
will fix the tests on this one real quick

parachain/src/primitives.rs
```diff
@@ -122,10 +122,14 @@ impl From<i32> for Id {
 }
 
 const USER_INDEX_START: u32 = 1000;
+const PUBLIC_INDEX_START: u32 = 2000;
```
xlc, line 125: what's the difference between user index & public index?
does this means 1000 - 2000 are reserved for common good chain only? that can only be registered with root?

gavofyork: Correct. System chains (0..=999) can have their transactions executed with root. Anything 1000 and above are user-level and cannot.

Within the user-level ID space, we begin the public IDs at 2000 to prevent clashing with IDs used for common-good chains.

gavofyork:
bot merge

PR #2995, opened 2021-05-07
Dispatchable for teleporting assets

gavofyork:
Added a dispatchable to make it a little easier to dispatch an asset by removing the need for a local weight payment in XCM.

### DIFF
--- a/node/network/protocol/src/peer_set.rs
+++ b/node/network/protocol/src/peer_set.rs
@@ -57,7 +57,7 @@ impl PeerSet {
 				max_notification_size,
 				set_config: sc_network::config::SetConfig {
 					// we allow full nodes to connect to validators for gossip
-					// to ensure any `MIN_GOSSIP_PEERS` always include reserved peers®
+					// to ensure any `MIN_GOSSIP_PEERS` always include reserved peers
 					// we limit the amount of non-reserved slots to be less
 					// than `MIN_GOSSIP_PEERS` in total
 					in_peers: super::MIN_GOSSIP_PEERS as u32 / 2 - 1,
--- a/node/network/protocol/src/peer_set.rs
+++ b/node/network/protocol/src/peer_set.rs
@@ -57,7 +57,7 @@ impl PeerSet {
 				max_notification_size,
 				set_config: sc_network::config::SetConfig {
 					// we allow full nodes to connect to validators for gossip
-					// to ensure any `MIN_GOSSIP_PEERS` always include reserved peers
+					// to ensure any `MIN_GOSSIP_PEERS` always include reserved peers®
 					// we limit the amount of non-reserved slots to be less
 					// than `MIN_GOSSIP_PEERS` in total
 					in_peers: super::MIN_GOSSIP_PEERS as u32 / 2 - 1,
--- a/xcm/pallet-xcm/src/lib.rs
+++ b/xcm/pallet-xcm/src/lib.rs
@@ -18,7 +18,7 @@
 
 #![cfg_attr(not(feature = "std"), no_std)]
 
-use sp_std::{prelude::*, marker::PhantomData, convert::TryInto, boxed::Box};
+use sp_std::{prelude::*, marker::PhantomData, convert::TryInto, boxed::Box, vec};
 use codec::{Encode, Decode};
 use xcm::v0::prelude::*;
 use xcm_executor::traits::ConvertOrigin;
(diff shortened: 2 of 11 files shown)


xcm/xcm-executor/src/lib.rs
```diff
@@ -60,7 +65,7 @@ impl<Config: config::Config> ExecuteXcm<Config::Call> for XcmExecutor<Config> {
 			return Outcome::Error(XcmError::WeightLimitReached(maximum_weight));
 		}
 		let mut trader = Config::Trader::new();
-		let result = Self::do_execute_xcm(origin, true, message, &mut 0, Some(shallow_weight), &mut trader);
+		let result = Self::do_execute_xcm(origin, true, message, &mut weight_credit, Some(shallow_weight), &mut trader);
```
apopiak, line 68: Can you elaborate on why these need to be mutable?

gavofyork: Not sure what you mean. This API already existed.

xcm/src/v0/traits.rs
```diff
@@ -126,14 +126,37 @@ impl Outcome {
 	}
 }
 
+/// Type of XCM message executor.
```
kianenigma, line 129: Why is there both a generic and an associated `Call`?

gavofyork: Removed.

xcm/src/v0/traits.rs
```diff
@@ -126,14 +126,36 @@ impl Outcome {
...
 pub trait ExecuteXcm<Call> {
-	type Call;
-	fn execute_xcm(origin: MultiLocation, message: Xcm<Call>, weight_limit: Weight) -> Outcome;
+//	/// The `Call` type for the XCM. Usually just the native `Call` of thw chain. Can also be `()`.
+//	type Call;
```
kianenigma, line 132: stale doc?

kianenigma (approved):
looks good to me.

xcm/pallet-xcm/src/lib.rs
```diff
@@ -94,6 +103,66 @@ pub mod pallet {
...
+		///   `dest` side.
+		/// - `dest_weight`: Equal to the total weight on `dest` of the XCM message
+		///   `Teleport { assets, effects: [ BuyExecution{..}, DepositAsset{..} ] }`.
+		#[pallet::weight({
+			let mut message = Xcm::WithdrawAsset {
+				assets: assets.clone(),
+				effects: sp_std::vec![ InitiateTeleport {
+					assets: sp_std::vec![ All ],
+					dest: dest.clone(),
+					effects: sp_std::vec![],
```
kianenigma, line 123: looking at a few lines down below, `vec` must be in scope

kianenigma: ahh it must be because this block is expanded elsewhere by the macro.

xcm/xcm-executor/src/traits/weight.rs
```diff
@@ -43,6 +43,18 @@ pub trait WeightBounds<Call> {
...
+
+	/// Return the total weight for executing `message`.
+	fn weight(message: &mut Xcm<Call>) -> Result<Weight, ()> {
+		Self::shallow(message)?.checked_add(Self::deep(message)?).ok_or(())
+	}
+}
+
+/// A means of getting approximate weight consumption for a given destination message executor and a
+/// message.
+pub trait UniversalWeigher {
```
kianenigma, line 55: I read this from my phone so not 100% sure, but I can't find this trait being used.

node/network/protocol/src/peer_set.rs
```diff
@@ -57,7 +57,7 @@ impl PeerSet {
 				max_notification_size,
 				set_config: sc_network::config::SetConfig {
 					// we allow full nodes to connect to validators for gossip
-					// to ensure any `MIN_GOSSIP_PEERS` always include reserved peers
+					// to ensure any `MIN_GOSSIP_PEERS` always include reserved peers®
```
shawntabrizi, line 60:
```suggestion
					// to ensure any `MIN_GOSSIP_PEERS` always include reserved peers
```

xcm/src/v0/traits.rs
gavofyork, line 132:
```suggestion
```

xcm/xcm-executor/src/traits/weight.rs
gavofyork: Yeah, it's not referenced yet, but I expect it'll be useful in the future. Can remove if it offends.

PR #3003, opened 2021-05-10
Support opening HRMP channels in genesis

pepyakin:
This change allows to list channels that should be opened at the genesis. This should be useful for integration testing.

### DIFF
--- a/runtime/parachains/src/hrmp.rs
+++ b/runtime/parachains/src/hrmp.rs
@@ -299,6 +299,51 @@ decl_storage! {
 		///   block number.
 		HrmpChannelDigests: map hasher(twox_64_concat) ParaId => Vec<(T::BlockNumber, Vec<ParaId>)>;
 	}
+	add_extra_genesis {
+		/// Preopen the given HRMP channels.
+		///
+		/// The values in the tuple corresponds to `(sender, recipient, max_capacity, max_message_size)`,
+		/// i.e. similar to `init_open_channel`. In fact, the initialization is performed as if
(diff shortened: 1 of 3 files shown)


runtime/parachains/src/hrmp.rs
```diff
@@ -299,6 +299,51 @@ decl_storage! {
...
+
+#[cfg(feature = "std")]
+fn initialize_storage<T: Config>(preopen_hrmp_channels: &[(ParaId, ParaId, u32, u32)]) {
+	let host_config = configuration::Module::<T>::config();
+	for &(sender, recipient, max_capacity, max_message_size) in preopen_hrmp_channels {
+		if let Err(err) = preopen_hrmp_channel::<T>(sender, recipient, max_capacity, max_message_size) {
+			panic!("failed to initialize the genesis storage: {:?}", err);
+		}
+	}
+	<Module<T>>::process_hrmp_open_channel_requests(&host_config);
```
shawntabrizi, line 329: Is this still queued for a future session change?

shawntabrizi (approved):

runtime/parachains/src/hrmp.rs
pepyakin: No, this processes pending HRMP channel requests. Those are added by `preopen_hrmp_channel` above. By the block number 1 all channels should be ready to accept messages.

shawntabrizi:
@pepyakin you need to merge here too?

pepyakin:
Ah, yes, of course.

PR #3008, opened 2021-05-11
Extrinsic for reserving a parachain ID

gavofyork:


runtime/common/src/paras_registrar.rs
```diff
@@ -435,20 +454,32 @@ impl<T: Config> Module<T> {
...
-		ensure!(!Paras::<T>::contains_key(id), Error::<T>::AlreadyRegistered);
+		let deposited = if let Some(para_data) = Paras::<T>::get(id) {
+			ensure!(para_data.manager == who, Error::<T>::NotOwner);
+			ensure!(!para_data.locked, Error::<T>::ParaLocked);
+			para_data.deposit
+		} else {
+			ensure!(!ensure_reserved, Error::<T>::NotReserved);
+			Default::default()
+		};
 		ensure!(paras::Module::<T>::lifecycle(id).is_none(), Error::<T>::AlreadyRegistered);
```
shawntabrizi, line 467: We might be able to loosen this restriction and allow re-registration too.

Will look into it as a follow up

PR #3010, opened 2021-05-11
Introduce reserve-asset transfer extrinsic

gavofyork:
CC @xlc


runtime/kusama/src/lib.rs
```diff
@@ -1265,6 +1265,7 @@ impl pallet_xcm::Config for Runtime {
 	type XcmExecuteFilter = OnlyWithdrawTeleportForAccounts;
```
xlc, line 1265: do we need to update this?

gavofyork: The extrinsic will still work without any update. I decided to leave it alone for now since with the extrinsic it's unnecessary to add a filter.

xlc (approved):

shawntabrizi (approved):

PR #3014, opened 2021-05-12
Enable Para Registration on Kusama

shawntabrizi:
Removes the call filters which would prevent external users from calling the Registrar Pallet.

We still disable swaps, since the parathread story is not complete yet.

### DIFF
--- a/runtime/westend/src/lib.rs
+++ b/runtime/westend/src/lib.rs
@@ -146,7 +146,7 @@ pub fn native_version() -> NativeVersion {
 /// Allow everything.
 pub struct BaseFilter;
 impl Filter<Call> for BaseFilter {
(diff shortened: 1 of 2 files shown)


gavofyork (approved):

runtime/westend/src/lib.rs
```diff
@@ -143,14 +143,11 @@ pub fn native_version() -> NativeVersion {
 	}
 }
 
-/// Avoid processing transactions from slots and parachain registrar except by root.
+/// Allow everything.
 pub struct BaseFilter;
 impl Filter<Call> for BaseFilter {
 	fn filter(c: &Call) -> bool {
```
gavofyork, line 149:
```suggestion
	fn filter(_: &Call) -> bool {
```

PR #3022, opened 2021-05-13
primitives: export the current slot number storage key

andresilva:
Required for https://github.com/paritytech/cumulus/pull/437.

### DIFF
--- a/primitives/src/v1/mod.rs
+++ b/primitives/src/v1/mod.rs
@@ -57,6 +57,7 @@ pub use crate::v0::{ValidatorPair, CollatorPair};
 
 pub use sp_staking::SessionIndex;
 pub use sp_authority_discovery::AuthorityId as AuthorityDiscoveryId;
+pub use sp_consensus_slots::Slot;
 
 /// Signed data.
 mod signed;
@@ -81,6 +82,12 @@ pub mod well_known_keys {
 	//
 	//     <Hrmp as Store>::HrmpEgressChannelsIndex::prefix_hash();
 	//
+
+	/// The current slot number.
+	///
+	/// The storage entry should be accessed as a `Slot` encoded value.
+	pub const CURRENT_SLOT: &[u8] =
+		&hex!["1cb6f36e027abb2091cfb5110ab5087f06155b3cd9a8c9e5e9a23fd5dc13a5ed"];
 
 	/// The currently active host configuration.
 	///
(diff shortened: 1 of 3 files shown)


bkchr (approved):

primitives/src/v1/mod.rs
```diff
@@ -82,6 +83,12 @@ pub mod well_known_keys {
 	//     <Hrmp as Store>::HrmpEgressChannelsIndex::prefix_hash();
 	//
 
+	/// The current slot number.
+	///
+	/// The storage entry should be accessed as a `Slot` encoded value.
+	pub const CURRENT_SLOT: &[u8] =
+		&hex!["1cb6f36e027abb2091cfb5110ab5087f06155b3cd9a8c9e5e9a23fd5dc13a5ed"];
```
pepyakin, line 90: Nit: This is a bit unfortunate that we don't have a test for this. 

The problem I see is that `sp_consensus_slot` (I assume that modules hosts the storage value) doesn't have any indication that the value is accessed externally and that seems fragile.

Maybe we could move the const to the pallet near to where the value lives, write the test there and make it a pub export, and here we just re-export it?

primitives/src/v1/mod.rs
```diff
@@ -57,6 +57,7 @@ pub use crate::v0::{ValidatorPair, CollatorPair};
 
 pub use sp_staking::SessionIndex;
 pub use sp_authority_discovery::AuthorityId as AuthorityDiscoveryId;
+pub use sp_consensus_slots::Slot;
```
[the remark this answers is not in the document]
andresilva: I just re-exported it so that we can use it in `cumulus` without depending on `sp-consensus-slots` (e.g. when we read from the storage key below). IMO if the storage key is important enough to have it exported, then it makes sense that the type of the value behind that key is also exported.

primitives/src/v1/mod.rs
andresilva: Yeah, I'll do that. This key is coming from BABE so I'll move this `const` to substrate, add a test, and then re-export it here as you suggested.

bkchr: No, please don't move this to substrate! This is highly polkadot specific, especially because it depends on the naming of the pallet in the runtime.

We can add a test and improve this stuff more in the future.

PR #3025, opened 2021-05-14
Grab stream of networking events earlier

tomaka:
As explained in comment, we should make sure to grab the stream of network events before the service actually finishes initializing.


ordian (approved):
Let's test this hypothesis. I don't see any harm in this change anyway.

bkchr (approved):

PR #3028, opened 2021-05-15
Tmp dir issue on docker

xlc:
When running within docker, for some reason the pvf worker have trouble to move the artifact from tmp dir to cache dir.

This demo that by not using tmp dir, the parachain validation works in docker.

This is a PoC and not intended for merging.

cc @pepyakin

### DIFF
--- a/node/core/pvf/src/prepare/worker.rs
+++ b/node/core/pvf/src/prepare/worker.rs
@@ -87,7 +87,18 @@ pub async fn start_work(
 		renice(pid, NICENESS_BACKGROUND);
 	}
 
-	if let Err(err) = send_request(&mut stream, code).await {
+	let mut parent = artifact_path.clone();
+	parent.pop();
+
+	let path_bytes = path_to_bytes(&parent);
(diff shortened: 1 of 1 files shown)


node/core/pvf/src/prepare/worker.rs
```diff
@@ -168,12 +179,56 @@ fn renice(pid: u32, niceness: i32) {
...
+
 /// The entrypoint that the spawned prepare worker should start with. The socket_path specifies
 /// the path to the socket used to communicate with the host.
 pub fn worker_entrypoint(socket_path: &str) {
 	worker_event_loop("prepare", socket_path, |mut stream| async move {
 		loop {
-			let code = recv_request(&mut stream).await?;
+			let payload = recv_request(&mut stream).await?;
+			let path_len = payload[0] as usize;
+			let path = &payload[1..path_len];
```
xlc, line 226: I have a off by one error here

xlc:
fixed by #3057

Issue #2882, opened 2021-04-14
Upgrade Pallets to FRAME v2


shawntabrizi:
There are a number of custom Polkadot pallets in this repo, and we could use help to migrate them to the new v2 macro.

The pallets which are already live need to check that they won't break, but new pallets like the onboarding systems should get migrated before they are put live on the network.

- [x] auctions (#3031)
- [x] claims (#3032)
- [x] crowdloans (#3036)
- [x] mmr (#3061)
- [x] registrar (#3063)
- [x] sudo wrapper (#3074)
- [x] purchase (#3075)
- [x] slots (#3218)
- [x] runtime > parachains (a lot in here)

shaunxw:
XCM sender doesn't need migration as it's not a pallet. Migrations in `runtime > common` are done.

ferrell-code:
Once the last three are merged all the parachains pallets will have been migrated.

With a quick grep I see that there is one small validator_manager.rs pallet in rococo and a few pallets in bridges. Not sure if these are going to be used in the long term, but most are now using the attribute macro :dragon_face:

shawntabrizi:
thanks @ferrell-code !

probably we want to migrate the bridge pallet stuff, but we can open a new issue for that and have the bridge team track it cc @HCastano @tomusdrw

HCastano:
Yeah, @ferrell-code feel free to open an issue and/or PRs in the [`parity-bridges-common`](https://github.com/paritytech/parity-bridges-common) repo

PR #3032, opened 2021-05-16
Update claims pallet to FRAME v2

shaunxw:
Part of #2882 

Converts the `Claims` pallet to the new pallet attribute macro introduced in #6877.

## ⚠️ Breaking Change ⚠️ 

So users of the `Claims` pallet must be careful about the name they used in `construct_runtime!`. Hence the `runtime-migration` label, which might not be needed depending on the configuration of the `Claims` pallet.

### DIFF
--- a/runtime/common/src/claims.rs
+++ b/runtime/common/src/claims.rs
@@ -14,43 +14,28 @@
 // You should have received a copy of the GNU General Public License
 // along with Substrate.  If not, see <http://www.gnu.org/licenses/>.
 
-//! Module to process claims from Ethereum addresses.
+//! Pallet to process claims from Ethereum addresses.
 
 use sp_std::{prelude::*, fmt::Debug};
 use sp_io::{hashing::keccak_256, crypto::secp256k1_ecdsa_recover};
-use frame_support::{
-	decl_event, decl_storage, decl_module, decl_error, ensure,
-	traits::{Currency, Get, VestingSchedule, EnsureOrigin, IsSubType},
-	weights::{Weight, Pays, DispatchClass},
-	pallet_prelude::DispatchResultWithPostInfo,
-};
-use frame_system::{ensure_signed, ensure_root, ensure_none};
+use frame_support::{ensure, traits::{Currency, Get, VestingSchedule, IsSubType}, weights::Weight};
 use parity_scale_codec::{Encode, Decode};
 #[cfg(feature = "std")]
 use serde::{self, Serialize, Deserialize, Serializer, Deserializer};
 #[cfg(feature = "std")]
 use sp_runtime::traits::Zero;
 use sp_runtime::{
-	traits::{CheckedSub, SignedExtension, DispatchInfoOf}, RuntimeDebug, DispatchResult,
+	traits::{CheckedSub, SignedExtension, DispatchInfoOf}, RuntimeDebug,
 	transaction_validity::{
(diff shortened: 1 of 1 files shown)


runtime/common/src/claims.rs
```diff
@@ -147,18 +132,38 @@ impl sp_std::fmt::Debug for EcdsaSignature {
...
+	pub struct Pallet<T>(_);
+
+	/// Configuration trait.
+	#[pallet::config]
+	pub trait Config: frame_system::Config {
+		/// The overarching event type.
+		type Event: From<Event<Self>> + IsType<<Self as frame_system::Config>::Event>;
+		type VestingSchedule: VestingSchedule<Self::AccountId, Moment=Self::BlockNumber>;
+		#[pallet::constant]
+		type Prefix: Get<&'static [u8]>;
```
shaunxw, line 152: Here introduced a metadata difference compared to FRAME v1 version. The constant type should be `&[u8]`, however there is no way to change the type in metadata(if I didn't missing any v2 new syntax). Need to wait for https://github.com/paritytech/substrate/issues/8827 being solved.

shawntabrizi: does this metadata change really matter? my guess is probably no?

shaunxw: The change will make `Prefix` type as `& 'static[u8]`. Not sure if polkadotjs handles this type, and if not, I think it would result a `createTypes` error(which would probably break apps). @jacogr could you please help to confirm?

shaunxw: I just checked CI. Yes there's a `createType` error, and polkadotjs API can't be initialized because of it https://gitlab.parity.io/parity/polkadot/-/jobs/928510

> 2021-05-16 15:57:58        REGISTRY: Unable to resolve type &'static[u8], it will fail on construction
> 2021-05-16 15:57:58        API/INIT: Error: FATAL: Unable to initialize the API: createType(&'static[u8]):: Cannot construct unknown type &'static[u8]

jacogr: https://github.com/polkadot-js/api/pull/3545

In general, and not getting in the way here, this one I have mixed feelings about -

1. It needs to be done, since we want all converted
2. It is absolutely a nightmare to test this end-to-end to ensure nothing has been broken (so if something does break, this becomes a support horror)
3. It is not very active (most have claimed), but when used it just need to work flawlessly

So all-in-all, needs to be done, but not quite seeing the overall benefit.
3.

runtime/common/src/claims.rs
```diff
@@ -147,18 +132,38 @@ impl sp_std::fmt::Debug for EcdsaSignature {
...
 
-decl_event!(
-	pub enum Event<T> where
-		Balance = BalanceOf<T>,
-		AccountId = <T as frame_system::Config>::AccountId
-	{
+#[frame_support::pallet]
+pub mod pallet {
+	use frame_support::{pallet_prelude::*, traits::EnsureOrigin, weights::{Pays, DispatchClass}};
+	use frame_system::{pallet_prelude::*, ensure_signed, ensure_root, ensure_none};
```
KiChjang, line 138: Importing just `pallet_prelude` should be enough:

```suggestion
	use frame_support::pallet_prelude::*;
	use frame_system::pallet_prelude::*;
```

shawntabrizi (approved):

KiChjang (approved):

shawntabrizi:
bot merge

PR #3036, opened 2021-05-17
Update crowdloan pallet to FRAME v2

shaunxw:
Part of #2882 

Converts the `Crowdloan` pallet to the new pallet attribute macro introduced in #6877.

## ⚠️ Breaking Change ⚠️ 

So users of the `Crowdloan` pallet must be careful about the name they used in `construct_runtime!`. Hence the `runtime-migration` label, which might not be needed depending on the configuration of the `Crowdloan` pallet.

### DIFF
--- a/runtime/common/src/crowdloan.rs
+++ b/runtime/common/src/crowdloan.rs
@@ -14,9 +14,9 @@
 // You should have received a copy of the GNU General Public License
 // along with Polkadot.  If not, see <http://www.gnu.org/licenses/>.
 
-//! # Parachain Crowdloaning module
+//! # Parachain Crowdloaning pallet
 //!
-//! The point of this module is to allow parachain projects to offer the ability to help fund a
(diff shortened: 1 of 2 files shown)


shaunxw:
This pallet is declared in Kusama and Westend runtime as `Crowdloan`: https://github.com/paritytech/polkadot/blob/ebbb354bc4521858c8bc56c9102e62955f41a333/runtime/kusama/src/lib.rs#L1445 https://github.com/paritytech/polkadot/blob/ebbb354bc4521858c8bc56c9102e62955f41a333/runtime/westend/src/lib.rs#L1047

Same with in `decl_storage`: https://github.com/paritytech/polkadot/blob/ebbb354bc4521858c8bc56c9102e62955f41a333/runtime/common/src/crowdloan.rs#L184

thiolliere (approved):
looks good to me, no need for migration indeed as name is same as previously hardcoded pallet prefix in decl_storage.

shawntabrizi:
bot merge

PR #3041, opened 2021-05-17
reword error: channel is _terminated and_ empty

drahnr:
The essential part of this error is the fact the channel is terminated, not that the channel is empty.


ordian (approved):
I'd say something like `Overseer is shutting down`, but that also works for me.

PR #3043, opened 2021-05-17
CI: remove unnecessary jobs while simnet is under work

TriplEight:


rcny:
Relates to #3004 and #3020.

rcny (approved):

PR #3049, opened 2021-05-18
Do not merge - debugging PR for Rococo

eskimor:


bkchr:
@eskimor do we still need this?

eskimor:
Quite likely yes - due to that moron who destroyed Rococo, I was never able to actually use this. Looks like I will be able today.

eskimor:
Closing in favour of #3173

PR #3057, opened 2021-05-19
Put WIP artifacts next to ready ones

### LINKED ISSUE #3044: PVF validation host doesn't work well with Docker
PVF validation host owns a cache of compiled artifacts. This cache is populated by preparation workers, sub-processes that perform some checks and compile the artifact.

Specifically, a preparation worker first creates a file in the temporary directory. Once compiled and wrote, the artifact is moved into its final location. We do this so that because other parts of 

The move is performed with `async_std::fs::rename` which, similarly to `std::fs::rename`, relies on [`rename`](https://man7.org/linux/man-pages/man2/rename.2.html)(2). 

From [`std::fs::rename`](https://doc.rust-lang.org/std/fs/fn.rename.html):

> This will not work if the new name is on a different mount point.

Almost always `/tmp` has a different mount point than the node's basedir. My tests and test run on Rococo showed it worked well. I assume this is related to some special handling of tmpfs. Regardless, there was a report of this not working on Docker. It manifested in empty files copied into the artifact caches.

pepyakin:
This PR makes it so that the artifacts are put next to the ready artifacts. This will ensure that the rename is within a single mounting point. Therefore the issues related to renaming/moving a file while promoting an artifact should be solved.

Fixes #3044

### DIFF
--- a/node/core/pvf/src/prepare/worker.rs
+++ b/node/core/pvf/src/prepare/worker.rs
@@ -19,7 +19,7 @@ use crate::{
 	artifacts::Artifact,
 	worker_common::{
 		IdleWorker, SpawnErr, WorkerHandle, bytes_to_path, framed_recv, framed_send, path_to_bytes,
-		spawn_with_program_path, tmpfile, worker_event_loop,
+		spawn_with_program_path, tmpfile_in, worker_event_loop,
 	},
 };
 use async_std::{
@@ -70,6 +70,7 @@ pub enum Outcome {
 pub async fn start_work(
 	worker: IdleWorker,
 	code: Arc<Vec<u8>>,
+	cache_path: &Path,
 	artifact_path: PathBuf,
 	background_priority: bool,
 ) -> Outcome {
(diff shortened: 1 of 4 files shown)


node/core/pvf/src/prepare/worker.rs
```diff
@@ -192,8 +293,8 @@ pub fn worker_entrypoint(socket_path: &str) {
 			);
 			async_std::fs::write(&dest, &artifact_bytes).await?;
 
-			// Communicate the results back to the host.
-			framed_send(&mut stream, &path_to_bytes(&dest)).await?;
+			// Return back a byte that signals finsihing the work.
```
bkchr, line 296:
```suggestion
			// Return back a byte that signals finishing the work.
```

node/core/pvf/src/prepare/worker.rs
```diff
@@ -87,63 +88,164 @@ pub async fn start_work(
...
 
-	let selected = futures::select! {
-		artifact_path_bytes = framed_recv(&mut stream).fuse() => {
-			match artifact_path_bytes {
-				Ok(bytes) => {
-					if let Some(tmp_path) = bytes_to_path(&bytes) {
-						async_std::fs::rename(tmp_path, &artifact_path)
+		let selected = futures::select! {
+			artifact_path_bytes = framed_recv(&mut stream).fuse() => {
+				match artifact_path_bytes {
```
bkchr, line 117:
```suggestion
			res = framed_recv(&mut stream).fuse() => {
				match res {
```

bkchr (approved):

PR #3058, opened 2021-05-19
Reserves for bids during ending period persist until auction done

gavofyork:


runtime/common/src/auctions.rs
```diff
@@ -494,9 +495,9 @@ impl<T: Config> Pallet<T> {
 						// It really should be reserved; there's not much we can do here on fail.
 						let err_amt = CurrencyOf::<T>::unreserve(&who, amount);
 						debug_assert!(err_amt.is_zero());
-
-						Self::deposit_event(Event::<T>::Unreserved(who, amount));
 					}
+
+					Self::deposit_event(Event::<T>::Unreserved(who, amount));
```
shawntabrizi, line 500: Why did this move outside of the if statement?

shawntabrizi (approved):

PR #3061, opened 2021-05-20
Update mmr pallet to FRAME v2

shaunxw:
Part of #2882 

Converts the `Mmr` pallet to the new pallet attribute macro introduced in #6877.

## ⚠️ Breaking Change ⚠️ 

So users of the `Mmr` pallet must be careful about the name they used in `construct_runtime!`. Hence the `runtime-migration` label, which might not be needed depending on the configuration of the `Mmr` pallet.

### DIFF
--- a/runtime/common/src/mmr.rs
+++ b/runtime/common/src/mmr.rs
@@ -20,10 +20,11 @@ use beefy_primitives::ValidatorSetId;
 use sp_core::H256;
 use sp_runtime::traits::Convert;
 use sp_std::prelude::*;
-use frame_support::{decl_module, decl_storage, RuntimeDebug};
+use frame_support::RuntimeDebug;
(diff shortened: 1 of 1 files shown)


KiChjang (approved):

shawntabrizi:
Pallet MMR does not seem to be used anywhere in Polkadot or Cumulus runtimes.

shawntabrizi (approved):

shawntabrizi:
bot merge

tomusdrw:
CC @musnit @denalimarsh @philipstanislaus when Rococo&Wococo is upgrade to this new format it might change the storage keys / storage proofs for you.

philipstanislaus:
Thanks for the note, also pinging @vgeddes and @Rizziepit here

PR #3065, opened 2021-05-20
New Rococo genesis

bkchr:


andresilva (approved):

andresilva:
Should we maybe update `protocolId` as well?

bkchr:
Why?

andresilva:
To avoid having peers from the existing rococo network try to connect to us. It doesn't make much of a difference since genesis hash won't match, but I think the DHT will also be merged.

bkchr:
We never have done this before, so I think it is fine do ignore it here as well :D

andresilva:
bot merge

PR #3074, opened 2021-05-22
Update para sudo wrapper pallet to FRAME v2

shaunxw:
Part of #2882 

Converts the `ParaSudoWrapper` pallet to the new pallet attribute macro introduced in #6877.

## ⚠️ Breaking Change ⚠️ 

So users of the `ParaSudoWrapper ` pallet must be careful about the name they used in `construct_runtime!`. Hence the `runtime-migration` label, which might not be needed depending on the configuration of the `ParaSudoWrapper ` pallet.

### DIFF
--- a/runtime/common/src/paras_sudo_wrapper.rs
+++ b/runtime/common/src/paras_sudo_wrapper.rs
@@ -16,13 +16,8 @@
 
 //! A simple wrapper allowing `Sudo` to call into `paras` routines.
 
-use sp_std::prelude::*;
-use frame_support::{
-	decl_error, decl_module, ensure,
-	dispatch::DispatchResult,
-	weights::DispatchClass,
-};
-use frame_system::ensure_root;
+use frame_support::{pallet_prelude::*, ensure, dispatch::DispatchResult, weights::DispatchClass};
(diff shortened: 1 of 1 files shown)


runtime/common/src/paras_sudo_wrapper.rs
```diff
@@ -16,29 +16,33 @@
...
 
-use sp_std::prelude::*;
-use frame_support::{
-	decl_error, decl_module, ensure,
-	dispatch::DispatchResult,
-	weights::DispatchClass,
-};
-use frame_system::ensure_root;
+use frame_support::{pallet_prelude::*, ensure, dispatch::DispatchResult, weights::DispatchClass};
+use frame_system::{pallet_prelude::*, ensure_root};
```
KiChjang, line 20: I think the rest of the imports are already included in `pallet_prelude`, so you can just do:

```suggestion
use frame_support::pallet_prelude::*;
use frame_system::pallet_prelude::*;
```

KiChjang (approved):

shawntabrizi (approved):
There is no decl storage in this pallet

shawntabrizi:
bot merge

PR #3075, opened 2021-05-22
Update purchase pallet to FRAME v2

shaunxw:
Part of #2882 

Converts the `Purchase` pallet to the new pallet attribute macro introduced in #6877.

## ⚠️ Breaking Change ⚠️ 

So users of the `Purchase` pallet must be careful about the name they used in `construct_runtime!`. Hence the `runtime-migration` label, which might not be needed depending on the configuration of the `Purchase` pallet.

### DIFF
--- a/runtime/common/src/purchase.rs
+++ b/runtime/common/src/purchase.rs
@@ -348,7 +348,7 @@ pub mod pallet {
 		#[pallet::weight(T::DbWeight::get().writes(1))]
 		pub(super) fn set_statement(origin: OriginFor<T>, statement: Vec<u8>) -> DispatchResult {
 			T::ConfigurationOrigin::ensure_origin(origin)?;
-			ensure!(statement.len() < T::MaxStatementLength::get() as usize, Error::<T>::InvalidStatement);
+			ensure!(statement.len() as u32 < T::MaxStatementLength::get(), Error::<T>::InvalidStatement);
 			// Possibly this is worse than having the caller account be the payment account?
 			Statement::<T>::set(statement);
 			Self::deposit_event(Event::<T>::StatementUpdated);
--- a/runtime/common/src/purchase.rs
+++ b/runtime/common/src/purchase.rs
@@ -14,38 +14,19 @@
 // You should have received a copy of the GNU General Public License
 // along with Substrate.  If not, see <http://www.gnu.org/licenses/>.
 
-//! Module to process purchase of DOTs.
+//! Pallet to process purchase of DOTs.
 
 use parity_scale_codec::{Encode, Decode};
 use sp_runtime::{Permill, RuntimeDebug, DispatchResult, DispatchError, AnySignature};
 use sp_runtime::traits::{Zero, CheckedAdd, Verify, Saturating};
-use frame_support::{decl_event, decl_storage, decl_module, decl_error, ensure};
+use frame_support::{pallet_prelude::*, ensure};
 use frame_support::traits::{
 	EnsureOrigin, Currency, ExistenceRequirement, VestingSchedule, Get
 };
-use frame_system::ensure_signed;
+use frame_system::{pallet_prelude::*, ensure_signed};
 use sp_core::sr25519;
 use sp_std::prelude::*;
-
-/// Configuration trait.
-pub trait Config: frame_system::Config {
-	/// The overarching event type.
-	type Event: From<Event<Self>> + Into<<Self as frame_system::Config>::Event>;
-	/// Balances Pallet
-	type Currency: Currency<Self::AccountId>;
-	/// Vesting Pallet
-	type VestingSchedule: VestingSchedule<Self::AccountId, Moment=Self::BlockNumber, Currency=Self::Currency>;
-	/// The origin allowed to set account status.
-	type ValidityOrigin: EnsureOrigin<Self::Origin>;
-	/// The origin allowed to make configurations to the pallet.
-	type ConfigurationOrigin: EnsureOrigin<Self::Origin>;
-	/// The maximum statement length for the statement users to sign when creating an account.
-	type MaxStatementLength: Get<usize>;
-	/// The amount of purchased locked DOTs that we will unlock for basic actions on the chain.
-	type UnlockedProportion: Get<Permill>;
-	/// The maximum amount of locked DOTs that we will unlock.
-	type MaxUnlocked: Get<BalanceOf<Self>>;
-}
+pub use pallet::*;
 
 type BalanceOf<T> = <<T as Config>::Currency as Currency<<T as frame_system::Config>::AccountId>>::Balance;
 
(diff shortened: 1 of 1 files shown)


runtime/common/src/purchase.rs
```diff
@@ -14,38 +14,19 @@
...
 // along with Substrate.  If not, see <http://www.gnu.org/licenses/>.
 
-//! Module to process purchase of DOTs.
+//! Pallet to process purchase of DOTs.
 
 use parity_scale_codec::{Encode, Decode};
 use sp_runtime::{Permill, RuntimeDebug, DispatchResult, DispatchError, AnySignature};
 use sp_runtime::traits::{Zero, CheckedAdd, Verify, Saturating};
-use frame_support::{decl_event, decl_storage, decl_module, decl_error, ensure};
+use frame_support::{pallet_prelude::*, ensure};
```
KiChjang, line 22:
```suggestion
use frame_support::pallet_prelude::*;
```

runtime/common/src/purchase.rs
```diff
@@ -14,38 +14,19 @@
...
 use frame_support::traits::{
 	EnsureOrigin, Currency, ExistenceRequirement, VestingSchedule, Get
 };
-use frame_system::ensure_signed;
+use frame_system::{pallet_prelude::*, ensure_signed};
```
KiChjang, line 26:
```suggestion
use frame_system::pallet_prelude::*;
```

runtime/common/src/purchase.rs
```diff
@@ -143,45 +163,44 @@ decl_error! {
...
-		Statement: Vec<u8>;
-		// The block where all locked dots will unlock.
-		UnlockBlock: T::BlockNumber;
-	}
-}
+	// A map of all participants in the DOT purchase process.
+	#[pallet::storage]
+	pub(super) type Accounts<T: Config> = StorageMap<
+		_,
+		Blake2_128Concat, T::AccountId,
```
KiChjang, line 171:
```suggestion
		Blake2_128Concat,
		T::AccountId,
```

runtime/common/src/purchase.rs
```diff
@@ -101,31 +82,70 @@ pub struct AccountStatus<Balance> {
...
+
+		/// The origin allowed to set account status.
+		type ValidityOrigin: EnsureOrigin<Self::Origin>;
+
+		/// The origin allowed to make configurations to the pallet.
+		type ConfigurationOrigin: EnsureOrigin<Self::Origin>;
+
+		/// The maximum statement length for the statement users to sign when creating an account.
+		#[pallet::constant]
+		type MaxStatementLength: Get<u32>;
```
KiChjang, line 112: Why is this changed into a `u32`?

runtime/common/src/purchase.rs
shaunxw: imho hasher and key type in same line makes `StorageMap`/`StorageDoubleMap` more readable: it's easier to tell which is the key part and the value part.

runtime/common/src/purchase.rs
shaunxw: `MaxStatementLength` constant is `u32`. Changing the type would make it more clean, or we have to use `pallet::extra_constants` to manually convert the type. There is no breaking change so imho it's reasonable. (example of changing length type in [Substrate migration PR](https://github.com/paritytech/substrate/pull/8762/files#diff-e44afa3f58035adbf7d19adf68eed43df626e55f7fba287641e4a4546d104b18R147))

KiChjang: Technically it is a breaking change, because `usize` on 32-bit systems is 32 bits long, but on 64-bit systems it is 64 bits long, while `u32` is fixed to be 32 bits. In other words, if the user has previously configured their `MaxStatementLength` to be larger than `u32::MAX`, they may run into issues when upgrading this pallet.

shaunxw: By no breaking change I mean purchase is not used in any runtime yet. If `MaxStatementLength` is configured to be larger `u32::MAX`, users will have trouble running on 32-bit system even without type change here.

KiChjang: Please do not dismiss as an ancillary concern. The issue isn't whether you can imagine cases of how it can or can't break, it's that you're changing the public API here, which IS a breaking change. I don't seem to see a good reason here other than the reluctance of using an extra attribute, unless I'm not hearing you correctly on what the difficulty is?

shaunxw: If you insist, of course, in strict definition, it IS a breaking change.

As expressed in the beginning of this conversion, I think it will make code more clean. I have no idea why you think it's a reluctance of doing more work using an extra attribute, or other difficulty. If that's the case, I won't bother to spend time to migrate any pallet.

KiChjang: If your cleanliness change makes a breaking API change, then I don't think it's worth it. This isn't to dismiss your work here; it's that public API changes involves a lot of other downstream changes, such as polkadot-js front-end modifications to support a different type, and others that both you and I didn't even think of at this moment in time.

Again I stress, do not dismiss public API changes as ancillary.

shaunxw: Please don't assume I "dismiss public API changes as ancillary".

In which way exactly this type change will do breaks, other than in theory it's a breaking change? For PolkadotJS there is no new type introduced, and there is no metadata change since the constant was `u32`. This associated type is only used "publicly" in unit tests, if I don't miss anything, what is the "others that both you and I didn't even think of at this moment in time"?

KiChjang: We're hosting this as FOSS, I don't think you can confidently say that there is absolutely 0 downstream users of this crate, especially when it's in the public domain.

shaunxw: I could argue more about "should we do technically breaking change like this", or "how much time I've spent to fix reasonable breaking changes of substrate/polkadot since 2018".

There's just too many off-topic things happening here and sorry I don't have all day for this.  I'm happy to revert it if any party say this is making their life difficult. Or feel free to close this PR. That's it.

shawntabrizi: This should be fine. This should be a u32 to begin with, and the value should never have exceeded `u32::max()`.

shawntabrizi (approved):

shawntabrizi:
This pallet is not used in any runtime, and thus no worries about any migration or breaking changes here. Thank you @shaunxw

runtime/common/src/purchase.rs
```diff
@@ -296,51 +320,57 @@ decl_module! {
...
 		/// Set the statement that must be signed for a user to participate on the DOT sale.
 		///
 		/// Origin must match the `ConfigurationOrigin`
-		#[weight = T::DbWeight::get().writes(1)]
-		fn set_statement(origin, statement: Vec<u8>) {
+		#[pallet::weight(T::DbWeight::get().writes(1))]
+		pub(super) fn set_statement(origin: OriginFor<T>, statement: Vec<u8>) -> DispatchResult {
 			T::ConfigurationOrigin::ensure_origin(origin)?;
-			ensure!(statement.len() < T::MaxStatementLength::get(), Error::<T>::InvalidStatement);
+			ensure!(statement.len() < T::MaxStatementLength::get() as usize, Error::<T>::InvalidStatement);
```
shawntabrizi, line 351:
```suggestion
			ensure!(statement.len() as u32 < T::MaxStatementLength::get(), Error::<T>::InvalidStatement);
```

KiChjang (approved):

runtime/common/src/purchase.rs
```diff
@@ -296,51 +320,57 @@ decl_module! {
...
 		/// Set the statement that must be signed for a user to participate on the DOT sale.
 		///
 		/// Origin must match the `ConfigurationOrigin`
-		#[weight = T::DbWeight::get().writes(1)]
-		fn set_statement(origin, statement: Vec<u8>) {
+		#[pallet::weight(T::DbWeight::get().writes(1))]
+		pub(super) fn set_statement(origin: OriginFor<T>, statement: Vec<u8>) -> DispatchResult {
 			T::ConfigurationOrigin::ensure_origin(origin)?;
-			ensure!(statement.len() < T::MaxStatementLength::get(), Error::<T>::InvalidStatement);
+			ensure!(statement.len() as u32 < T::MaxStatementLength::get(), Error::<T>::InvalidStatement);
```
shawntabrizi, line 351:
```suggestion
			ensure!((statement.len() as u32) < T::MaxStatementLength::get(), Error::<T>::InvalidStatement);
```

thiolliere (approved):
Thanks Shawnx, the check-runtime CI is failing, but the current change doesn't affect any runtime in polkadot thus it is a false-negative

thiolliere:
bot force merge

PR #3076, opened 2021-05-22
Remove memory limits

Neurone:
With these RAM limitations if polkadot takes more then 5.4Gb RAM it will be heavily slowed down and with more then 5.5Gb RAM it will be killed with an "out-of-memory" message. I don't think any fixed value is a good idea and I suggest to remove those parameters.

Also, consider the [current suggested standard hardware](https://wiki.polkadot.network/docs/en/maintain-guides-how-to-validate-polkadot#standard-hardware) for a validator - for example - is 64GB RAM minimum.

My rig on the mainnet stays well below those parameters - so 64GBs seem to me an overkill - but this is another topic :)


burdges:
We've some other limit for WASM instances somewhere?  We fix a memory cap for WASM instances for a particular chain and kill blocks that exceed this, but afaik this configuration should not be how that happens.

s3krit (approved):
Yep happy to remove the memory limits, they're probably no longer relevant. I forget why they were introduced in the first place. Thanks!

PR #3088, opened 2021-05-24
Update `FallbackStrategy` to `Nothing`

shawntabrizi:
cc @kianenigma 

This will prevent a fallback to an on chain election solution, which can cause issues with block production when too complex.

This will mean the old set will continue to be used when the election fails.

### DIFF
--- a/runtime/kusama/src/lib.rs
+++ b/runtime/kusama/src/lib.rs
@@ -345,7 +345,7 @@ parameter_types! {
 
 	// fallback: run election on-chain.
 	pub const Fallback: pallet_election_provider_multi_phase::FallbackStrategy =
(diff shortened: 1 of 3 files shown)


kianenigma:
Let's leave this open and see if it is needed for any immediate release, thanks!

kianenigma (approved):
The session pallet will use the old set, so this does what we intend it to do: https://github.com/paritytech/substrate/blob/4973e7e933dfdefcfc32b88eeb83309787fa9f5f/frame/session/src/lib.rs#L607

coriolinus (approved):

kianenigma:
blocked on https://github.com/paritytech/substrate/pull/8912

kianenigma:
bot merge

kianenigma:
goodbye onchain election.

PR #3114, opened 2021-05-26
Companion for PR 8913

pepyakin:
polkadot companion: https://github.com/paritytech/substrate/pull/8913


node/core/pvf/src/execute/worker.rs
```diff
@@ -246,7 +246,12 @@ async fn validate_using_artifact(
 
 	let validation_started_at = Instant::now();
 	let descriptor_bytes =
-		match crate::executor_intf::execute(compiled_artifact, params, spawner.clone()) {
+		match unsafe {
+			// SAFETY: this should be safe since the compiled artifact passed here comes from the
+			//         file created by the prepare workers. These files are obtained by calling
+			//         [`executor_intf::prepare`].
+			crate::executor_intf::execute(compiled_artifact, params, spawner.clone())
```
pepyakin, line 253: I am not sure if this is enough TBH. The data that comes here is from a file. The file can in theory be compromised or corrupted (e.g. due to hardware failure). 

However, there is also little sense in extending the `unsafe` blocks further. It's awkward and will probably lead marking all the worker_common machinery. Yet, it won't provide any additional benefit - we ultimately will walk into the cross process boundary. Ultimately, it's up to validation host to decide what file will be passed here.

I think we do not take into account the compromisation concern. We mostly assume that if the node is compromised the attacker can do way worse things anyway. The corruption is a bit more severe. However, I am not sure if it's likely: after all, to my knowledge, ELFs do not have checksums.

node/core/pvf/Cargo.toml
```diff
@@ -36,18 +36,3 @@ adder = { package = "test-parachain-adder", path = "../../../parachain/test-para
...
-# here then there will be linking errors like
-#
-#     `multiple definitions of `set_vmctx_memory`.
-#
-# or similar, because wasmtime exports these symbols and does not support multiple versions compiled
-# in at the same time.
-#
-# Another safeguard is a test `ensure_wasmtime_version` that will fail on each bump and prompt the
-# developer to correspondingly act upon the change.
-wasmtime-jit = "0.24"
```
pepyakin, line 53: This is now removed along with the test before because as per https://github.com/paritytech/polkadot/pull/3084 we no longer leave artifacts in the cache and remove them each start. That means we do not need to foresee a possibility of an old artifact getting into the new engine.

node/core/pvf/src/execute/worker.rs
bkchr: If you want to be 100% sure, you could hash them. However, I'm not sure that is really required.

bkchr (approved):

PR #3125, opened 2021-05-27
Explicit Matching Statements in `NonTransfer` Proxy Filter

shawntabrizi:
This PR switches the definition for the `NonTransfer` proxy to use an explicit `match` statement.

This will allow the compiler to tell us when there are pallets not being handled by the proxy filters.

Other filters can continue to use `matches!` presuming that the initial error with `NonTransfer` will prompt updating the all filters.

### DIFF
--- a/runtime/kusama/src/lib.rs
+++ b/runtime/kusama/src/lib.rs
@@ -945,57 +945,85 @@ impl InstanceFilter<Call> for ProxyType {
 	fn filter(&self, c: &Call) -> bool {
 		match self {
 			ProxyType::Any => true,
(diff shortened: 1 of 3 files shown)


shawntabrizi:
gav wasn't a fan of being so explicit

xlc (commented):
I think we need some macros to allow specify filters easier

shawntabrizi:
what kind of syntax are you thinking?

xlc:
I am not very sure, just the code should be represent what we want (enable this, disable that), instead of like this (use comment to describe what we want, and hope reviewer & future maintainer is able to handle it correctly).

PR #3130, opened 2021-05-28
Use proper host configuration everywhere

bkchr:


drahnr (approved):

node/service/src/chain_spec.rs
```diff
@@ -117,6 +117,50 @@ pub fn wococo_config() -> Result<PolkadotChainSpec, String> {
...
+		code_retention_period: 1200,
+		max_code_size: MAX_CODE_SIZE,
+		max_pov_size: MAX_POV_SIZE,
+		max_head_data_size: 32 * 1024,
+		group_rotation_frequency: 20,
+		chain_availability_period: 4,
+		thread_availability_period: 4,
+		max_upward_queue_count: 8,
+		max_upward_queue_size: 1024 * 1024,
+		max_downward_message_size: 1024,
```
apopiak, line 134: Would it be ok to change this?
Would be great if we can send sth. like e.g. a `set_storage` with a bunch of keys (e.g. 20).

bkchr: What do we use in Kusama?

PR #3133, opened 2021-05-28
fix approval-checking GRANDPA voting rule

rphmeier:
a `None` return value implies to vote on the best, not to vote on the base.

this explicitly changes the logic to vote on the base

NOTE: DO NOT RELEASE UNTIL THOROUGHLY TESTED ON ROCOCO

### DIFF
--- a/node/service/src/grandpa_support.rs
+++ b/node/service/src/grandpa_support.rs
@@ -71,6 +71,32 @@ impl ApprovalCheckingVotingRule {
 	}
 }
 
+#[derive(Debug, PartialEq)]
+enum ParachainVotingRuleTarget<H, N> {
+	/// Vote explicitly on the given hash.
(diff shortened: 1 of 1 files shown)


ordian (approved):
Seems to do what the comment says now, but it would be good to extract this logic into a function and add a test for future-proofing. Also seems a bit weird that we now always return `Some`.

rphmeier:
> Also seems a bit weird that we now always return Some.

I agree, but that is IMO a GRANDPA API issue - the approval-checking voting rule is really opinionated, so we have to be explicit in informing the caller of what we want to vote on.

ordian (approved):

andresilva:
We could return `None` when `ParachainVotingRuleTarget::Current` so that the voting rule doesn't filter anything. Looks good to me as is though.

PR #3135, opened 2021-05-28
Be more smart when loading the chain spec

bkchr:
Instead of relying on the name of the chain spec file, we now use the
`is_*` methods for this.

### DIFF
--- a/cli/src/command.rs
+++ b/cli/src/command.rs
@@ -95,20 +95,18 @@ impl SubstrateCli for Cli {
 			path => {
 				let path = std::path::PathBuf::from(path);
 
-				let starts_with = |prefix: &str| {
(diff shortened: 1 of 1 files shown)


ordian (approved):

arkpar:
Does it parse the json file twice now?

bkchr:
> Does it parse the json file twice now?

Yes, when it isn't `polkadot`. However, as this is only done on startup, I don't see any problems in regards to the performance?

arkpar:
Genesis state may potentially be huge. Initially there was an optimization not to parse genesis section if it is not needed. It is only needed on the first start after all.  But I'm not sure if the optimization is still there after so many refactorings.

Anyway, after the native runtime is removed, there should be no reason to have specialized chain spec here.

PR #3140, opened 2021-05-31
Companion for Substrate#8953

bkchr:
This removes the default `initialize_block` call before doing a runtime api request.


pepyakin (approved):
This seems straightforward

PR #3141, opened 2021-05-31
Staking Miner

kianenigma:
List of substrate PRs that need to be merged before this is usable: 

- #2793 https://github.com/paritytech/substrate/pull/7910#pullrequestreview-674371735
- https://github.com/paritytech/substrate/pull/8912/files and all of its child PRs
- #3088


runtime/polkadot/src/constants.rs
```diff
@@ -33,7 +33,7 @@ pub mod time {
 	use primitives::v0::{Moment, BlockNumber};
 	pub const MILLISECS_PER_BLOCK: Moment = 6000;
 	pub const SLOT_DURATION: Moment = MILLISECS_PER_BLOCK;
-	pub const EPOCH_DURATION_IN_SLOTS: BlockNumber = 4 * HOURS;
+	pub const EPOCH_DURATION_IN_SLOTS: BlockNumber = 4 * MINUTES;
```
shawntabrizi, line 36: note to DEF revert this before merge :)

utils/staking-miner/src/monitor.rs
```diff
@@ -0,0 +1,191 @@
...
+		};
+
+		log::info!(target: LOG_TARGET, "subscribing to {:?}", subscription_method);
+		let mut subscription: Subscription<Header> = client
+			.subscribe(&subscription_method, JsonRpcParams::NoParams, "unsubscribe")
+			.await
+			.unwrap();
+
+		loop {
+			let now = subscription.next().await.unwrap();
```
coriolinus, line 83:
```suggestion
		while let Some(now) = subscription.next().await {
```

Handle upstream connection closures more gracefully.

runtime/polkadot/src/lib.rs
```diff
@@ -76,6 +76,7 @@ pub use pallet_staking::StakerStatus;
 pub use sp_runtime::BuildStorage;
 pub use pallet_timestamp::Call as TimestampCall;
 pub use pallet_balances::Call as BalancesCall;
+pub use pallet_election_provider_multi_phase::Call as EPMCall;
```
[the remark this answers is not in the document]
kianenigma: correct, removed.

utils/staking-miner/src/dry_run.rs
```diff
@@ -0,0 +1,57 @@
...
+		if <EPM::Snapshot<T>>::exists() {
+			log::info!(target: LOG_TARGET, "snapshot already exists.");
+		} else {
+			log::info!(target: LOG_TARGET, "creating a fake snapshot now.");
+		}
+		<EPM::Pallet<T>>::create_snapshot().unwrap();
+	});
+}
+
+macro_rules! dry_run_cmd_for { ($runtime:tt) => { paste::paste! {
```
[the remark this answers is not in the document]
kianenigma: yeah works well with `ident` too, I think I was battling another issue and changed to `tt`, and forgot to revert.

utils/staking-miner/src/main.rs
```diff
@@ -0,0 +1,517 @@
...
+	Westend,
+}
+
+pub(crate) static mut RUNTIME: AnyRuntime = AnyRuntime::Polkadot;
+
+macro_rules! construct_runtime_prelude {
+	($runtime:ident, $npos:ty, $signed_extra:expr) => {
+		paste::paste! {
+				#[allow(unused_import)]
+				pub(crate) mod [<$runtime _runtime_exports>] {
```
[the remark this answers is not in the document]
kianenigma: ahh yeah I hated the extra indent due to paste! :D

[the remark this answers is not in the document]
kianenigma: err.. well although we still need to `paste!` for the function name.. will keep as-is for now then.

utils/staking-miner/src/main.rs
```diff
@@ -0,0 +1,517 @@
...
+						) -> UncheckedExtrinsic {
+							use codec::Encode as _;
+							use sp_core::Pair as _;
+							use sp_runtime::traits::StaticLookup as _;
+
+							let crate::Signer { account, pair, .. } = signer;
+
+							let local_call = EPMCall::<Runtime>::submit(raw_solution, witness);
+							// this one is a bit tricky.. for now I just "hope" it does not change.
+							let call = Call::ElectionProviderMultiPhase(local_call);
```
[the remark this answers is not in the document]
kianenigma: excellent.

utils/staking-miner/src/main.rs
```diff
@@ -0,0 +1,517 @@
...
+}
+
+impl MinerProfile {
+	/// Get the next miner profile to use, should this one fail.
+	fn next(self) -> Self {
+		use sp_runtime::Percent;
+		match self {
+			MinerProfile::WithBalancing(count) => {
+				if count > 0 {
+					MinerProfile::WithBalancing(count.saturating_sub(3))
```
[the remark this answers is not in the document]
kianenigma: yeah, it is a placeholder for now.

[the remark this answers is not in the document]
kianenigma: I am not sure if I want to use `MinerProfile `, it is mostly a placeholder for my ideas.

utils/staking-miner/src/main.rs
```diff
@@ -0,0 +1,517 @@
...
+
+/// Read the signer account's uri from the given `path`.
+async fn read_signer_uri<
+	P: AsRef<Path>,
+	T: frame_system::Config<AccountId = AccountId, Index = Index>,
+>(
+	path: P,
+	client: &WsClient,
+) -> Result<Signer, Error> {
+	let mut uri = std::fs::read_to_string(path).map_err(|_| Error::KeyFileNotFound)?;
```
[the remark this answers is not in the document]
kianenigma: nice, I agree. The only downside of `thiserror` is that it works nicely for errors that implement `std::Error`, while most substrate errors don't unfortunately. But still quite nie.

utils/staking-miner/src/main.rs
```diff
@@ -190,15 +210,42 @@ impl MinerProfile {
...
 enum Error {
-	KeyFileNotFound,
-	KeyFileCorrupt,
+	Io(#[from] std::io::Error),
+	Jsonrpsee(#[from] jsonrpsee_ws_client::Error),
+	Codec(#[from] codec::Error),
+	Crypto(sp_core::crypto::SecretStringError),
+	RemoteExternalities(&'static str),
+	PalletMiner(EPM::unsigned::MinerError),
+	PalletElection(EPM::ElectionError),
```
coriolinus, line 221:
```suggestion
	Crypto(#[from] sp_core::crypto::SecretStringError),
	RemoteExternalities(&'static str),
	PalletMiner(#[from] EPM::unsigned::MinerError),
	PalletElection(#[from] EPM::ElectionError),
```

You don't need the manual implementations for these errors.

Cargo.lock
```diff
@@ -14,27 +14,18 @@ dependencies = [
...
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "a55f82cfe485775d02112886f4169bde0c5894d75e79ead7eafe7e40a25e45f7"
-dependencies = [
- "gimli 0.23.0",
-]
-
-[[package]]
-name = "addr2line"
-version = "0.15.1"
+version = "0.15.2"
```
ordian, line 17: please revert `cargo update`

bridges/bin/millau/node/Cargo.toml
```diff
@@ -29,7 +29,7 @@ frame-benchmarking-cli = { git = "https://github.com/paritytech/substrate", bran
 node-inspect = { git = "https://github.com/paritytech/substrate", branch = "master" }
 pallet-transaction-payment-rpc = { git = "https://github.com/paritytech/substrate", branch = "master" }
 sc-basic-authorship = { git = "https://github.com/paritytech/substrate", branch = "master" }
-sc-cli = { git = "https://github.com/paritytech/substrate", branch = "master", features = ["wasmtime"] }
+sc-cli = { git = "https://github.com/paritytech/substrate", features = ["wasmtime"] , branch = "master" }
```
ordian, line 32: please revert unrelated `Cargo.toml` changes here and elsewhere

utils/staking-miner/alice
```diff
@@ -0,0 +1 @@
+bottom drive obey lake curtain smoke basket hold race lonely fit walk//Alice
```
ordian, line 1: is this intended to be committed?

runtime/westend/src/constants.rs
```diff
@@ -33,7 +33,7 @@ pub mod time {
 	use primitives::v0::{Moment, BlockNumber};
 	pub const MILLISECS_PER_BLOCK: Moment = 6000;
 	pub const SLOT_DURATION: Moment = MILLISECS_PER_BLOCK;
-	pub const EPOCH_DURATION_IN_SLOTS: BlockNumber = 1 * HOURS;
+	pub const EPOCH_DURATION_IN_SLOTS: BlockNumber = 1 * MINUTES;
```
ordian, line 36: if this is intended, it should probably go into separate PR as it is orthogonal to staking miner

runtime/westend/src/lib.rs
```diff
@@ -412,7 +413,7 @@ pallet_staking_reward_curve::build! {
 
 parameter_types! {
 	// Six sessions in an era (6 hours).
-	pub const SessionsPerEra: SessionIndex = 6;
+	pub const SessionsPerEra: SessionIndex = 2;
```
ordian, line 416: and this as well

ordian (commented):
please revert unrelated to the PR changes

utils/staking-miner/src/main.rs
```diff
@@ -0,0 +1,651 @@
...
+		let score = {
+			let min_staker = *winners.last().map(|(_, stake)| stake).unwrap();
+			let sum_stake = winners.iter().fold(0u128, |acc, (_, stake)| acc + stake);
+			let sum_squared = winners.iter().fold(0u128, |acc, (_, stake)| acc + stake);
+			[min_staker, sum_stake, sum_squared]
+		};
+
+		println!("mined a dpos-like solution with score = {:?}", score);
+
+		todo!();
```
ordian, line 513: todo ^^

runtime/polkadot/src/lib.rs
```diff
@@ -415,7 +416,7 @@ pallet_staking_reward_curve::build! {
 
 parameter_types! {
 	// Six sessions in an era (24 hours).
-	pub const SessionsPerEra: SessionIndex = 6;
+	pub const SessionsPerEra: SessionIndex = 1; // TODO: revert
```
shawntabrizi, line 419: TODO

utils/staking-miner/src/dry_run.rs
```diff
@@ -0,0 +1,95 @@
...
+		client: WsClient,
+		shared: SharedConfig,
+		config: DryRunConfig,
+		signer: Signer,
+	) -> Result<(), Error> {
+		use $crate::[<$runtime _runtime_exports>]::*;
+		let mut ext = crate::create_election_ext::<Runtime, Block>(shared.uri.clone(), config.at, true).await?;
+		force_create_snapshot::<Runtime>(&mut ext)?;
+		measure_snapshot_size::<Runtime>(&mut ext);
+		find_threshold::<Runtime>(&mut ext, 20_000);
```
kianenigma, line 71:
```suggestion
```

runtime/polkadot/src/lib.rs
kianenigma, line 419:
```suggestion
	pub const SessionsPerEra: SessionIndex = 6;
```

runtime/westend/src/lib.rs
kianenigma, line 416:
```suggestion
	pub const SessionsPerEra: SessionIndex = 6;
```

utils/staking-miner/alice
kianenigma: nope.

utils/staking-miner/src/main.rs
```diff
@@ -0,0 +1,651 @@
...
+					use $crate::westend_runtime_exports::*;
+					$($code)*
+				}
+			}
+		}
+	}
+}
+
+#[derive(codec::Encode, codec::Decode, Clone, Copy, Debug)]
+#[allow(unused)]
```
kianenigma, line 181: All of this is just some ideas... I might remove it later, but I think it can stay for now. It pretty much demonstrates what the miner should do in case of failuer.

utils/staking-miner/src/main.rs
kianenigma: I have to try again, I think I ran into an issue.

kianenigma: ah yeah so the inner errors must implement `StdError` apparently?

kianenigma:
@ordian all cargo changes are reverted, you were a bit too quick to review :D

utils/staking-miner/src/main.rs
```diff
@@ -0,0 +1,651 @@
...
+							);
+							extrinsic
+						}
+					}
+				}
+			}
+	};
+}
+
+// TODO: we might be able to use some code from the bridges repo here.
```
shawntabrizi, line 101: todo needs issue

utils/staking-miner/src/main.rs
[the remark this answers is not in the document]
shawntabrizi: Re-opening this since it seems to still exist

utils/staking-miner/src/main.rs
```diff
@@ -0,0 +1,651 @@
...
+		use sp_runtime::Percent;
+		match self {
+			MinerProfile::WithBalancing(count) => {
+				if count > 0 {
+					MinerProfile::WithBalancing(count.saturating_sub(3))
+				} else {
+					MinerProfile::JustSeqPhragmen
+				}
+			}
+			MinerProfile::JustSeqPhragmen => MinerProfile::TrimmedVoters(Percent::from_percent(90)),
```
shawntabrizi, line 206: is 90% also a known hard-coded thing?

and 10 percent below

kianenigma: This was just an idea, I removed it already.

utils/staking-miner/src/main.rs
coriolinus: Hmm, I thought that requirement triggered on `#[derive(thiserror::Error)]`, not on the `#[from]`. But yeah, error-handling is a lot better IMO when all of them implement `std::error::Error`.

utils/staking-miner/src/dry_run.rs
```diff
@@ -0,0 +1,97 @@
...
+			"Encoded Length: {:?}",
+			<EPM::Pallet<T>>::snapshot()
+				.expect("snapshot must exist before calling `measure_snapshot_size`")
+				.encode()
+				.len()
+		);
+	})
+}
+
+/// Fir the stake threshold to in order to have at most `count` voters.
```
coriolinus:
```suggestion
/// Find the stake threshold to in order to have at most `count` voters.
```

utils/staking-miner/src/dry_run.rs
```diff
@@ -0,0 +1,97 @@
...
+#[allow(unused)]
+fn find_threshold<T: EPM::Config>(ext: &mut Ext, count: usize) {
+	ext.execute_with(|| {
+		let mut voters = <EPM::Pallet<T>>::snapshot()
+			.expect("snapshot must exist before calling `measure_snapshot_size`")
+			.voters;
+		voters.sort_by_key(|v| v.1);
```
coriolinus, line 57:
```suggestion
		voters.sort_by_key(|(_voter, score, _targets)| std::cmp::Reverse(score));
```

Just makes it much clearer what's happening there, and also sorts `voters` in reverse without any extra work.

utils/staking-miner/src/dry_run.rs
```diff
@@ -0,0 +1,97 @@
...
+		println!("smallest allowed voter is {:?}", voters.into_iter().rev().take(count).last());
```
coriolinus, line 58: Two comments here:

- Handling this with iteration does some extra work that isn't really necessary, since we have the vector handy. (It would be lots of extra work, but `vec::Iter` is a `DoubleEndedIterator`.)
- Using `take` hides the difference between two cases: where we needed to truncate the voter list, and where we didn't. This matters because if we don't actually need to truncate the list, then we can't use the reported score as the current threshold; smaller voters would be allowed.

Assuming we reverse-sort `voters` as suggested above, we could do:

```suggestion
		match voters.get(count) {
			Some(threshold_voter) => println!("smallest allowed voter is {:?}", threshold_voter),
			None => {
				println!("requested truncation to {} voters but had only {}", count, voters.len());
				println!("smallest current voter: {:?}", voters.last());
			}
		}
```

utils/staking-miner/src/monitor.rs
```diff
@@ -0,0 +1,167 @@
...
+//! The monitor command.
+
+use crate::{params, prelude::*, rpc_helpers::*, Error, MonitorConfig, SharedConfig, Signer};
+use codec::Encode;
+use jsonrpsee_ws_client::{
+	traits::SubscriptionClient, v2::params::JsonRpcParams, Subscription, WsClient,
+};
+use sp_transaction_pool::TransactionStatus;
+
+/// Ensure that now is the singed phase.
```
coriolinus, line 26:
```suggestion
/// Ensure that now is the signed phase.
```

utils/staking-miner/src/monitor.rs
```diff
@@ -0,0 +1,167 @@
...
+			let (raw_solution, witness) = crate::mine_checked::<Runtime>(&mut ext, 50)?;
+			log::info!(target: LOG_TARGET, "mined solution with {:?}", &raw_solution.score);
+
+			let nonce = crate::get_account_info::<Runtime>(&client, &signer.account, Some(hash))
+				.await?
+				.map(|i| i.nonce)
+				.expect("signer account is checked to exist upon startup; it can only die if it \
+					transfers funds out of it, or get slashed. If it does not exist at this point, \
+					it is likely due to a bug, or the signer got slashed. Terminating."
+				);
```
coriolinus, line 115: This code, including the long expect, exists elsewhere in the codebase. Probably better to extract a function instead of copy-pasting it.

utils/staking-miner/src/monitor.rs
```diff
@@ -0,0 +1,167 @@
...
+			let bytes = sp_core::Bytes(extrinsic.encode());
+
+			let mut tx_subscription: Subscription<
+				TransactionStatus<<Block as BlockT>::Hash, <Block as BlockT>::Hash>
+			> = client
+				.subscribe(&"author_submitAndWatchExtrinsic", params! { bytes }, "author_unwatchExtrinsic")
+				.await
+				.unwrap();
+
+			let _success = while let Some(status_update) = tx_subscription.next().await.unwrap() {
```
coriolinus, line 134: Seems weird to have both `unwrap` and `while let` in the same line. Maybe

```suggestion
			let _success = while let Some(Some(status_update)) = tx_subscription.next().await {
```

utils/staking-miner/src/dry_run.rs
kianenigma, line 50:
```suggestion
/// Find the stake threshold in order to have at most `count` voters.
```

utils/staking-miner/src/main.rs
```diff
@@ -0,0 +1,517 @@
...
+			sp_core::crypto::set_default_ss58_version(
+				sp_core::crypto::Ss58AddressFormat::PolkadotAccount,
+			);
+			// safety: this program will always be single threaded, thus accessing global static is
+			// safe.
+			unsafe {
+				RUNTIME = AnyRuntime::Westend;
+			}
+		}
+		_ => panic!("unexpected chain: {:?}", chain),
```
coriolinus, line 473:
```suggestion
		_ => {
			eprintln!("unexpected chain: {:?}", chain);
			return;
		}
```

Panicing produces a bunch of other debug info that the user doesn't need or want in this case. We could also get fancy and `std::process::exit(1)`, but I'm not sure if we'd need a wrapper function in this case to ensure that everything gets dropped appropriately.

coriolinus (approved):
Nice!

utils/staking-miner/src/main.rs
kianenigma: I'll resolve this for now. Moving all our errors to implement `std::error:Error` might be a good substrate vision issue. Although, I am not sure if this is in `sp_std`, and if not, with the removal of the native runtime it will become kinda pointless.

utils/staking-miner/src/rpc_helpers.rs
```diff
@@ -0,0 +1,107 @@
...
+use jsonrpsee_ws_client::traits::Client;
+pub(crate) use jsonrpsee_ws_client::v2::params::JsonRpcParams;
+
+#[macro_export]
+macro_rules! params {
+	($($param:expr),*) => {
+		{
+			let mut __params = vec![];
+			$(
+				__params.push(serde_json::to_value($param).expect("json serialization infallible; qed."));
```
emostov, line 29: Why can expect param serialization to be infallibe?

utils/staking-miner/src/main.rs
```diff
@@ -0,0 +1,523 @@
...
+					why
+				);
+				std::thread::sleep_ms(2500);
+			}
+		}
+	};
+
+	let chain = rpc_helpers::rpc::<String>(&client, "system_chain", params! {})
+		.await
+		.expect("system_chain infallible; qed.");
```
emostov, line 444: Maybe ok for now, but from my experience these rpcs + websockets can fail at a negligible rate.

emostov: If we are attempting idempotent RPC calls we could build in some retry logic with exponential backoff

utils/staking-miner/src/monitor.rs
```diff
@@ -0,0 +1,164 @@
...
+			("chain_subscribeFinalizedHeads", "chain_unsubscribeFinalizedHeads")
+		};
+
+		log::info!(target: LOG_TARGET, "subscribing to {:?} / {:?}", sub, unsub);
+		let mut subscription: Subscription<Header> = client
+			.subscribe(&sub, JsonRpcParams::NoParams, &unsub)
+			.await
+			.unwrap();
+
+		while let Some(now) = subscription.next().await.unwrap() {
```
emostov, line 82: Is there a way to force a new header subscription here if this one breaks?
- crude solution could be to panic if the subscription breaks and use systemd to restart

Also, is there a time limit to how long a node will allow a header subscription? 
- crude solution could be to run this miner with a cronjob/systemdtimer that restarts say every hour

utils/staking-miner/src/main.rs
```diff
@@ -0,0 +1,523 @@
...
+	Builder::<B>::new()
+		.mode(Mode::Online(OnlineConfig {
+			transport: uri.into(),
+			at,
+			modules: if with_staking {
+				vec![
+					<T as frame_system::Config>::PalletInfo::name::<EPM::Pallet<T>>()
+						.expect("Pallet always has name; qed.")
+						.to_string(),
+					<T as frame_system::Config>::PalletInfo::name::<pallet_staking::Pallet<T>>()
```
emostov, line 306: If we are grabbing the snapshot I think we need to make the max rpc payload param on the node is raised

emostov: also would be guranteed failure on public rpc nodes (like the default uri) assumng the have not raise the max rpc payload param

utils/staking-miner/src/main.rs
```diff
@@ -0,0 +1,523 @@
...
+					raw_solution: EPM::RawSolution<EPM::CompactOf<Runtime>>,
+					witness: u32,
+					signer: crate::Signer,
+					nonce: crate::prelude::Index,
+					tip: crate::prelude::Balance,
+					era: sp_runtime::generic::Era,
+				) -> UncheckedExtrinsic {
+					use codec::Encode as _;
+					use sp_core::Pair as _;
+					use sp_runtime::traits::StaticLookup as _;
```
emostov, line 69: I don't understand this name aliasing, but trusting it is correct :P

utils/staking-miner/src/main.rs
```diff
@@ -0,0 +1,523 @@
...
+	nonce: Index,
+	tip: Balance,
+	era: sp_runtime::generic::Era,
+) -> polkadot_runtime_exports::SignedExtra {
+	use polkadot_runtime_exports::Runtime;
+	(
+		frame_system::CheckSpecVersion::<Runtime>::new(),
+		frame_system::CheckTxVersion::<Runtime>::new(),
+		frame_system::CheckGenesis::<Runtime>::new(),
+		frame_system::CheckMortality::<Runtime>::from(era),
```
emostov, line 109: I don't understand how the additional signed works on this. do we have acces to the `BlockHash` storage item that gets accessed here

https://docs.rs/frame-system/3.0.0/src/frame_system/extensions/check_mortality.rs.html#76-80

utils/staking-miner/src/main.rs
```diff
@@ -0,0 +1,523 @@
...
+	/// Provide a solution that can be submitted to the chian as an emergency response.
+	EmergencySolution,
+}
+
+#[derive(Debug, Clone, StructOpt)]
+struct MonitorConfig {
+	/// They type of event to listen to.
+	///
+	/// Typically, finalized is safer and there is no chance of anything going wrong, but it can be
+	/// slower. It is recommended if the duration of the signed phase is longer than the a
```
emostov, line 243: I think something got chopped off at the end of the doc comment

utils/staking-miner/src/main.rs
emostov: a note about the max rpc payload could be useful in this cli help or maybe higher level docs

utils/staking-miner/src/main.rs
```diff
@@ -0,0 +1,523 @@
...
+	P: AsRef<Path>,
+	T: frame_system::Config<AccountId = AccountId, Index = Index>,
+>(
+	path: P,
+	client: &WsClient,
+) -> Result<Signer, Error> {
+	let uri = std::fs::read_to_string(path)?;
+
+	// trim any trailing garbage.
+	let uri = uri.trim_end();
```
emostov, line 407: Maybe not worth it but a micro improvement might be `uri.trim_end().to_string().trim_start()`

utils/staking-miner/src/main.rs
```diff
@@ -0,0 +1,523 @@
...
+	at: Option<Hash>,
+}
+
+#[derive(Debug, Clone, StructOpt)]
+struct SharedConfig {
+	/// The ws node to connect to.
+	#[structopt(long, default_value = DEFAULT_URI)]
+	uri: String,
+
+	/// The file from which we read the account seed.
```
emostov, line 261:
```suggestion
	/// The file from which we read the account seed. WARNING: use discretion when
	/// when choosing an account to submit solutions from.
```
Basically just an obvious warning don't like use a stash account or some account where you would be upset if all funds where lost, particularly since this is the first iteration for public consumption

Also maybe just a general "here be dragons" type warning for users would be beneficial for the community

utils/staking-miner/src/main.rs
kianenigma: For now I only put retry on the initial startup, and for all the rest we assume there's never going to be a failure. If anything fails, I (for now) want to program to crash and restart, and notify us.

kianenigma: I'm quite open to improve this, but this way is just much much easier to implement.

utils/staking-miner/src/main.rs
kianenigma: yeah for sure, this tool needs that flag.

kianenigma: > also would be guranteed failure on public rpc nodes (like the default uri) assumng the have not raise the max rpc payload param

unfortunate, but true.

utils/staking-miner/src/main.rs
kianenigma: This is a pattern when you just want to use a trait's function (and thereby need to `use` it), but don't want the trait name itself to be in scope.

utils/staking-miner/src/main.rs
kianenigma: (you found your answer elsewhere.)

utils/staking-miner/src/main.rs
```diff
@@ -0,0 +1,523 @@
...
+	let uri = std::fs::read_to_string(path)?;
+
+	// trim any trailing garbage.
+	let uri = uri.trim_end();
+
+	let pair = Pair::from_string(&uri, None)?;
+	let account = T::AccountId::from(pair.public());
+	let _info =
+		get_account_info::<T>(&client, &account, None).await?.ok_or(Error::AccountDoesNotExists)?;
+	log::info!(target: LOG_TARGET, "loaded account {:?}, info: {:?}", &account, _info);
```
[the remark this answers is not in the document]
kianenigma: fair point, I think getting one account's information is fast enough + it is critical to ensure that the account exists (on sturtup) and this is the only way.

utils/staking-miner/src/main.rs
kianenigma: Great point, addressed.

runtime/polkadot/src/constants.rs
```diff
@@ -33,7 +33,7 @@ pub mod time {
 	use primitives::v0::{Moment, BlockNumber};
 	pub const MILLISECS_PER_BLOCK: Moment = 6000;
 	pub const SLOT_DURATION: Moment = MILLISECS_PER_BLOCK;
-	pub const EPOCH_DURATION_IN_SLOTS: BlockNumber = 4 * HOURS;
+	pub const EPOCH_DURATION_IN_SLOTS: BlockNumber = 2 * MINUTES;
```
emostov, line 36: Is this intentional?

runtime/polkadot/src/lib.rs
```diff
@@ -415,7 +420,7 @@ pallet_staking_reward_curve::build! {
 
 parameter_types! {
 	// Six sessions in an era (24 hours).
-	pub const SessionsPerEra: SessionIndex = 6;
+	pub const SessionsPerEra: SessionIndex = 1;
```
emostov, line 423: revert

utils/staking-miner/src/main.rs
emostov: For posterity sake, we explicitly load the system `BlockHash` storage item into the test externalities, and I assume the additional signed for check mortality then just uses that.

utils/staking-miner/src/rpc_helpers.rs
emostov: Given your other comments that we want it to fail so issues are clear I think this expect is totally fine

runtime/westend/src/lib.rs
```diff
@@ -100,6 +100,7 @@ pub use pallet_staking::StakerStatus;
 pub use sp_runtime::BuildStorage;
 pub use pallet_timestamp::Call as TimestampCall;
 pub use pallet_balances::Call as BalancesCall;
+pub use pallet_election_provider_multi_phase::Call as EPMCall;
```
emostov, line 103: Is this used? Same goes for polkadot + kusama runtimes

runtime/polkadot/src/constants.rs
kianenigma: yeah, I keep forgetting to revert them before push 🤦

runtime/westend/src/lib.rs
kianenigma: yes, in `create_uxt_$runtime` function.

utils/staking-miner/src/monitor.rs
kianenigma: honestly, one basic issue that I have is that I don't quite know when this kind of thing can even break 🤦

utils/staking-miner/src/emergency_solution.rs
```diff
@@ -0,0 +1,53 @@
...
+			assert!(EPM::Pallet::<Runtime>::current_phase().is_emergency());
+			// NOTE: this internally calls feasibility_check, but we just re-do it here as an easy way
+			// to get a `ReadySolution`.
+			let (raw_solution, _) = <EPM::Pallet<Runtime>>::mine_solution(50)?;
+			log::info!(target: LOG_TARGET, "mined solution with {:?}", &raw_solution.score);
+			let ready_solution = EPM::Pallet::<Runtime>::feasibility_check(raw_solution, EPM::ElectionCompute::Signed)?;
+			let encoded_ready = ready_solution.encode();
+			let encoded_support = ready_solution.supports.encode();
+			let mut solution_file = std::fs::File::create("solution.bin")?;
+			let mut supports_file = std::fs::File::create("solution.supports.bin")?;
```
emostov, line 41: i assume this works with no `./` or `/full/path/to/file`

utils/staking-miner/src/emergency_solution.rs
```diff
@@ -0,0 +1,53 @@
...
+			solution_file.write_all(&encoded_ready)?;
+			supports_file.write_all(&encoded_support)?;
+			log::info!(target: LOG_TARGET, "ReadySolution: size {:?} / score = {:?}", encoded_ready.len(), ready_solution.score);
+			log::trace!(target: LOG_TARGET, "Supports: {}", sp_core::hexdisplay::HexDisplay::from(&encoded_support));
+			Ok(())
+		})
+	}
+}}}
```
emostov, line 49: Nothing to change for now, but in the future would nice to have cli options for
- solution_file
- supports_file
- iters (min solution)

kianenigma: Iters has already been added right?

utils/staking-miner/src/main.rs
```diff
@@ -0,0 +1,492 @@
...
+// `construct_runtime_prelude` and macro's the import directly from it. A part of the code is also
+// still generic over `T`. My hope is to still make everything generic over a `Runtime`, but sadly
+// that is not currently possible as each runtime has its unique `Call`, and all Calls are not
+// sharing any generic trait. In other words, to create the `UncheckedExtrinsic` of each chain, you
+// need the concrete `Call` of that chain as well.
+#[macro_export]
+macro_rules! any_runtime {
+	($($code:tt)*) => {
+		unsafe {
+			match $crate::RUNTIME {
```
emostov, line 173: Not something to change for this draft, but in the future it would be nice to not have to use this `RUNTIME` and instead figure out a way to explicitly pass it in as an arg. Maybe even there is some builder model we could come up with?

utils/staking-miner/src/monitor.rs
```diff
@@ -0,0 +1,163 @@
...
+
+		log::info!(target: LOG_TARGET, "subscribing to {:?} / {:?}", sub, unsub);
+		let mut subscription: Subscription<Header> = client
+			.subscribe(&sub, JsonRpcParams::NoParams, &unsub)
+			.await
+			.unwrap();
+
+		while let Ok(Some(now)) = subscription.next().await {
+			let hash = now.hash();
+			log::debug!(target: LOG_TARGET, "new event at #{:?} ({:?})", now.number, hash);
```
emostov, line 86:
```suggestion
			log::debug!(target: LOG_TARGET, "new subscription event at #{:?} ({:?})", now.number, hash);
```
just not to be confused with a frame `Event`

utils/staking-miner/src/monitor.rs
```diff
@@ -0,0 +1,163 @@
...
+
+			if ensure_no_previous_solution::<Runtime, Block>(&mut ext, &signer.account).await.is_err() {
+				log::debug!(target: LOG_TARGET, "We already have a solution in this phase, skipping.");
+				continue;
+			}
+
+			let (raw_solution, witness) = crate::mine_unchecked::<Runtime>(&mut ext, config.iterations, true)?;
+			log::info!(target: LOG_TARGET, "mined solution with {:?}", &raw_solution.score);
+
+			let nonce = crate::get_account_info::<Runtime>(&client, &signer.account, Some(hash))
```
emostov: 1) would be nice if we could assume the nonce some way to speed up the critical path to solution submission. only thing i can think of is if we querried before this subscription loop, but that could probs run into other issues. 

2) consider using accountNextIndex rpc as I think they takes into account pending txs in the queue

emostov (approved):
For a first draft it looks good to me.

main feedback 
 1) There should be some clear warnings around loss of funds from the given URI. UPDATE: I see good warning in the docs at the top or `main.rs` - whoever is writing release notes may want to hoist those docs.
  2) With all the `expects` (especially the ones on RPC calls that are fallible even in ok conditions) we must rely on restarting the service. If we restart during the signed phase, the start up time is fairly important (we don't want to miss the entire signed phase and less important we don't want to get beaten by other miners). Given we are at least loading the EPM pallet storage I suspect this will take at min a few minutes. In the future it would be nice if we could store the chain state to file, and then check the file on restart (or even better not have to restart and keep it in memory :D)
  3) make sure to revert changes to polkadots `SessionsPerEra` and `EPOCH_DURATION_IN_SLOTS` if they where not intentional 
  (this may be out of date feedback now)

Review limitations:
- I did not review functions under `#[allow(unused)]` 
- My macro syntax understanding is primitive, so could easily have missed details within the macro impls

runtime/polkadot/src/lib.rs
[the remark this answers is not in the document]
kianenigma: No, it is used.

PR #3144, opened 2021-05-31
Increase PoV timeout slightly.

eskimor:
This is a slight increase of timeouts for PoV like data fetching. It is no longer calculated by how much it is supposed to take, but just set to a value that sounds sensible with regards to requirements.

### DIFF
--- a/node/network/protocol/src/request_response/mod.rs
+++ b/node/network/protocol/src/request_response/mod.rs
@@ -86,10 +86,10 @@ const DEFAULT_REQUEST_TIMEOUT: Duration = Duration::from_secs(3);
 /// peer set as well).
 const DEFAULT_REQUEST_TIMEOUT_CONNECTED: Duration = Duration::from_secs(1);
 
-/// Timeout for PoV like data, 2 times what it should take, assuming we can fully utilize the
-/// bandwidth. This amounts to two seconds right now.
-const POV_REQUEST_TIMEOUT_CONNECTED: Duration =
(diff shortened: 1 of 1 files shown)


rphmeier (commented):
What is the downside of the slightly larger timeout? Have we seen performance degradations due to the existing timeout? Some data from prometheus would be good to understand the motivation for the change

eskimor:
At least in the code path of PoV distribution there are not too many nodes to fetch the PoV from, so shorter timeouts and queue management are hardly useful. For the availability recovery code path, things are slightly different, but it should not matter too much either.

In hindsight it just makes more sense to set it to a value that seems acceptable, than to base it on calculations that won't hold anyways the moment we have forks.

For the motivation: I looked into that code, because you discovered some issues with failing requests recently and indeed a timeout smaller than one second might be tough at times, hence the slight increase. Longer term, we should really try to get rid of forks as much as possible and make PoVs smaller.

With forks and/or multiple collations the timing is tough. Running into a timeout at this stage does not really do any good - it would, if we prioritized some candidates/forks over others.

For availability recovery/fast path a shorter timeout might have benefits, but if we needed to optimize that I would work with queue management and back pressure as we did with statement distribution. This way we can fail early and don't need to wait for a timeout on overload.

rphmeier (approved):

rphmeier:
> Longer term, we should really try to get rid of forks as much as possible and make PoVs smaller.

Agreed!

PR #3147, opened 2021-05-31
Make script usable from both pipeline and localhost

radupopa2010:
I started by making simnet version an argument to the script, but then I wanted to make it also to trigger pipelines from localhost.

I tested locally and it works nicely. Just need to set the 2 tokens manually before.
```
./trigger_pipeline.sh --simnet-version=v4 \
                      --trgr-project=polkadot \
                      --trgr-ref=master \
                      --image-name=docker.io/paritypr/synth-wave \
                      --image-tag=master \
                      --collator-image-tag=master \
                      --ci-server-host=gitlab.parity.io \
                      --dwnstrm-id=332 \
                      --trigger-token="${TR_TOKEN}" \
                      --access-token="${PERS_TOKEN}"
```

Example of failing job when --simnet-version is not set
https://gitlab.parity.io/parity/polkadot/-/jobs/946213

Triggering works, here is pipeline:  https://gitlab.parity.io/parity/polkadot/-/pipelines/140165

We need now to include the fix in parachains test, when we run `npm run build` before running the test
to update to latest version of polkadotjs

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -388,5 +388,5 @@ trigger-simnet:
     DWNSTRM_ID:                    332
   script:
     # API trigger for a simnet job
-    - ./scripts/gitlab/trigger_pipeline.sh --simnet-version=v4
+    - ./scripts/gitlab/trigger_pipeline.sh --simnet-version=${SIMNET_REF}
   allow_failure:                   true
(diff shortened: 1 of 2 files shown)


TriplEight (approved):
LGTM, could've made some use of echoing some arguments, but can be done later.

PR #3158, opened 2021-06-02
Dispute distribution guide

eskimor:
Note: Guide PR needs update. The following things have changed:

There is no separate message for reporting unavailable candidates anymore, instead `ImportStatements` got a more informative back channel:

```rust
		ImportStatements {
		/// The hash of the candidate.
		/// The candidate receipt itself.
		candidate_receipt: CandidateReceipt,
		/// The session the candidate appears in.
		session: SessionIndex,
		/// Statements, with signatures checked, by validators participating in disputes.
		///
		/// The validator index passed alongside each statement should correspond to the index
		/// of the validator in the set.
		statements: Vec<(SignedDisputeStatement, ValidatorIndex)>,
		/// Inform the requester once we finished importing.
		///
		/// This is, we either discarded the votes, just record them because we
		/// casted our vote already or recovered availability for the candidate
		/// successfully.
		pending_confirmation: oneshot::Sender<ImportStatementsResult>
	},
```

with `ImportStatementsResult` looking like this:

```rust
pub enum ImportStatementsResult {
	/// Import was invalid (candidate was not available)  and the sending peer should get banned.
	InvalidImport,
	/// Import was valid and can be confirmed to peer.
	ValidImport
}
```

Details in implementation [PR](https://github.com/paritytech/polkadot/pull/3282).


roadmap/implementers-guide/src/node/disputes/dispute-distribution.md
```diff
@@ -10,48 +20,211 @@
...
+}
+
+struct InvalidVote {
+  /// The candidate being disputed.
+  candidate_hash: CandidateHash,
+  /// The voting validator.
+  validator_index: ValidatorIndex,
+  /// The session the candidate appears in.
+  candidate_session: SessionIndex,
+}
```
rphmeier, line 50: Even though it is unnecessary now, the `InvalidDisputeStatementKind` would be helpful for future-proofing, I think

eskimor: uh sure - I somehow missed that we have that.

roadmap/implementers-guide/src/node/disputes/dispute-distribution.md
```diff
@@ -1,3 +1,261 @@
...
+#### Disputes
+
+Protocol: "/polkadot/dispute/1"
+
+Request:
+
+```rust
+struct DisputeRequest {
+  // Either initiating invalid vote or our own (if we voted invalid).
+  invalid_vote: SignedV2<InvalidVote>,
```
rphmeier, line 37: What is `SignedV2`?

roadmap/implementers-guide/src/node/disputes/dispute-distribution.md
```diff
@@ -1,3 +1,261 @@
...
+need to check whether the disputing votes are somewhat current and not some
+stale ones.
+
+### Participating in a Dispute
+
+Upon receiving a `Dispute` message, a dispute distribution will trigger the
+import of the received votes via the dispute coordinator
+(`DisputeCoordinatorMessage::ImportStatements`). The dispute coordinator will
+take care of participating in that dispute if necessary. Once it is done, the
+coordinator will send a `DisputeDistributionMessage::SendDispute` message to dispute
```
rphmeier, line 128: So the intent is that this will occur in `DisputeCoordinatorSubsystem::IssueLocalStatement`?

roadmap/implementers-guide/src/node/disputes/dispute-distribution.md
```diff
@@ -1,3 +1,261 @@
...
+### Sending of messages
+
+Starting and participting in a dispute are pretty similar from the perspective
+of disptute distribution. Once we receive a `SendDispute` message we try to make
+sure to get the data out. We keep track of all the parachain validators that
+should see the message, which are all the parachain validators of the session
+where the dispute happened as they will want to participate in the dispute.  In
+addition we also need to get the votes out to all authorities of the current
+session (which might be the same or not). Those authorities will not
+participtate in the dispute, but need to see the statements so they can include
```
rphmeier, line 143:
```suggestion
participate in the dispute, but need to see the statements so they can include
```

roadmap/implementers-guide/src/node/disputes/dispute-distribution.md
```diff
@@ -1,3 +1,261 @@
...
+We keep track of connected parachain validators and authorities and will issue
+warnings in the logs if connected nodes are less than two thirds of the
+corresponding sets. We also only consider a message transmitted, once we
+received a confirmation message. If not we will keep retrying getting that
+message out as long as the dispute is deemed alive. To determine whether a
+dispute is still alive we will issue a
+`DisputeCoordinatorMessage::ActiveDisputes` message before each retry run. Once
+a dispute is no longer live, we will clean up the state coordingly.
+
+To cather with spam issues, we will in a first implementation only consider
```
rphmeier, line 155: i don't know the word 'cather', was this meant to be something else?

roadmap/implementers-guide/src/node/disputes/dispute-distribution.md
```diff
@@ -1,3 +1,261 @@
...
+disputes of already included data. Therefore only for candidates that are
```
rphmeier, line 156: This statement carries a lot of complexity - how do you know which data is already included?

rphmeier: I believe that we should already handle spam and then only participate in disputes where data is included already and thus available

roadmap/implementers-guide/src/node/disputes/dispute-distribution.md
```diff
@@ -1,3 +1,261 @@
...
+
+#### Vote Recovery
+
+Protocol: "/polkadot/vote-recovery/1"
+
+```rust
+struct IHaveVotesRequest {
+  candidate_hash: CandidateHash,
+  session: SessionIndex,
+  votes: VotesBitfield,
```
rphmeier, line 79: Does this convey the votes that the requester _has_ or _wants_?

roadmap/implementers-guide/src/node/disputes/dispute-distribution.md
```diff
@@ -1,3 +1,261 @@
...
+
+Response:
+
+```rust
+struct VotesResponse {
+  /// All votes we have, but the requester was missing.
+  missing: Vec<(DisputeStatement, ValidatorIndex, ValidatorSignature)>,
+  /// Any additional equivocating votes, we transmit those even if the sender
+  /// claims to have votes for that validator (as it might only have one).
+  equivocating: Vec<(DisputeStatement, ValidatorIndex, ValidatorSignature)>,
```
rphmeier, line 93: This may lead to a lot of extra data. The `IHaveVotesRequest` can contain two bitfields instead of 1 if necessary

roadmap/implementers-guide/src/types/overseer-protocol.md
```diff
@@ -424,6 +424,35 @@ enum DisputeParticipationMessage {
...
+
+Messages received by the [Dispute Distribution
+subsystem](../node/disputes/dispute-distribution.md). This subsystem is
+responsible of distributing explicit dispute statements.
+
+```rust
+enum DisputeDistributionMessage {
+
+  /// Tell dispute distribution to distribute an explicit dispute statement to
+  validators.
```
rphmeier, line 437:
```suggestion
  /// validators.
```

roadmap/implementers-guide/src/types/overseer-protocol.md
```diff
@@ -424,6 +424,35 @@ enum DisputeParticipationMessage {
...
+  SendDispute((ValidVote, InvalidVote)),
+
+  /// Ask DisputeDistribution to get votes we don't know about.
+  /// Fetched votes will be reported via `DisputeCoordinatorMessage::ImportStatements`
+  FetchMissingVotes {
+    candiate_hash: CandidateHash,
```
rphmeier, line 443:
```suggestion
    candidate_hash: CandidateHash,
```

roadmap/implementers-guide/src/types/overseer-protocol.md
```diff
@@ -424,6 +424,35 @@ enum DisputeParticipationMessage {
...
+    session: SessionIndex,
+    knownVotes: Bitfield,
```
rphmeier, line 445:
```suggestion
    known_votes: Bitfield,
```

roadmap/implementers-guide/src/types/overseer-protocol.md
```diff
@@ -424,6 +424,35 @@ enum DisputeParticipationMessage {
...
+    /// Optional validator to query from. `ValidatorIndex` as in the above
+    referenced session.
+    from_validator: Option<ValidatorIndex>,
+  }
+  /// Tell the subsystem that a candidate is not available. Dispute distribution
+  can punish peers distributing votes on unavailable hashes for example.
```
rphmeier, line 451:
```suggestion
  /// can punish peers distributing votes on unavailable hashes for example.
```

roadmap/implementers-guide/src/types/overseer-protocol.md
```diff
@@ -424,6 +424,35 @@ enum DisputeParticipationMessage {
...
+  SendDispute((ValidVote, InvalidVote)),
+
+  /// Ask DisputeDistribution to get votes we don't know about.
+  /// Fetched votes will be reported via `DisputeCoordinatorMessage::ImportStatements`
+  FetchMissingVotes {
+    candiate_hash: CandidateHash,
+    session: SessionIndex,
+    knownVotes: Bitfield,
+    /// Optional validator to query from. `ValidatorIndex` as in the above
+    referenced session.
```
rphmeier, line 447:
```suggestion
    /// referenced session.
```

roadmap/implementers-guide/src/node/disputes/dispute-distribution.md
```diff
@@ -1,3 +1,261 @@
...
+connections and issue warnings if we are not connected to a majority of
+validators. We should also keep track of failed sending attempts and log
+warnings accordingly. As disputes are rare and TCP is a reliable protocol,
+probably each failed attempt should trigger a warning in logs and also logged
+into some Prometheus metric.
+
+## Disputes for non included candidates
+
+If deemed necessary we can later on also support disputes for non included
+candidates, but disputes for those cases have totally different requirements.
```
rphmeier, line 242: It's not described anywhere in the document how a receiving node can tell whether a dispute is on an included or non-included candidate

rphmeier: The only way I am aware of is to iterate all heads of the chain, traversing backwards to the genesis, and look for `CandidateIncluded` events.

This has problems we've discussed at length before:
1. State pruning. Events for old blocks aren't necessarily available
2. Expense. It's an expensive computation
3. Reversions. When we revert some bad candidate we might begin to think that the candidate wasn't included anymore, when it's crucial that the dispute be transplanted onto other branches

roadmap/implementers-guide/src/node/disputes/dispute-distribution.md
eskimor: Has, although we could do either way - information is the same.

roadmap/implementers-guide/src/node/disputes/dispute-distribution.md
eskimor: Yeah, I was a bit torn here. It seems equivocations are not something that are expected to happen, since it results in a guaranteed slash. So I leaned towards increasing the response size in those cases, in favor of increasing the request size in all cases. Considering the possibility of a general malfunction, you have a point that increasing request size is likely the better idea.

roadmap/implementers-guide/src/node/disputes/dispute-distribution.md
eskimor: Not quite actually, will be fixed with real types on next commit.

roadmap/implementers-guide/src/node/disputes/dispute-distribution.md
eskimor: Sorry, I still mean by included, that it has been included somewhere - thus it is available. Changed it to say "available" instead.

roadmap/implementers-guide/src/node/disputes/dispute-distribution.md
```diff
@@ -29,7 +29,7 @@ This design should result in a protocol that is:
 
 #### Disputes
 
-Protocol: "/polkadot/dispute/1"
+Protocol: "/polkadot/send\_dispute/1"
```
rphmeier, line 32: is the escape necessary for `_`?

roadmap/implementers-guide/src/node/disputes/dispute-coordinator.md
```diff
@@ -104,8 +104,11 @@ Do nothing.
 ### On `DisputeCoordinatorMessage::IssueLocalStatement`
 
 * Deconstruct into parts `{ session_index, candidate_hash, candidate_receipt, is_valid }`.
-* Construct a [`DisputeStatement`][DisputeStatement] based on `Valid` or `Invalid`, depending on the parameterization of this routine. 
+* Construct a [`DisputeStatement`][DisputeStatement] based on `Valid` or `Invalid`, depending on the parameterization of this routine.
 * Sign the statement with each key in the `SessionInfo`'s list of parachain validation keys which is present in the keystore, except those whose indices appear in `voted_indices`. This will typically just be one key, but this does provide some future-proofing for situations where the same node may run on behalf multiple validators. At the time of writing, this is not a use-case we support as other subsystems do not invariably provide this guarantee.
+* Write statement to DB.
+* Send a `DisputeDistributionMessage::SendDispute` message to get the vote
```
rphmeier, line 110: 👍

roadmap/implementers-guide/src/node/disputes/dispute-distribution.md
```diff
@@ -1,3 +1,351 @@
...
+import of the received votes via the dispute coordinator
+(`DisputeCoordinatorMessage::ImportStatements`). The dispute coordinator will
+take care of participating in that dispute if necessary. Once it is done, the
+coordinator will send a `DisputeDistributionMessage::SendDispute` message to dispute
+distribution. From here, everything is the same as for starting a dispute,
+except that if the local node deemed the candidate valid, the `SendDispute`
+message will contain a valid vote signed by our node and will contain the
+initially received `Invalid` vote.
+
+Note, that we rely on the coordinator to check availability for spam protection
```
rphmeier, line 136: The current guide changes don't contain any modification to the coordinator that avoids disk usage / participation for spam.

I believe the goal was to provide a system of callbacks whereby distribution can learn whether the data was available and then decide to rate-limit the peer. Is that accurate? It seems that availability is the only thing that needs to be checked, not validity, since anything that is included anywhere is certainly available.

roadmap/implementers-guide/src/node/disputes/dispute-distribution.md
```diff
@@ -1,3 +1,351 @@
...
+(see below).
+In case the current node is only a potential block producer and does not
+actually need to recover availability (as it is not going to participate in the
+dispute), there is a potential optimization available: The coordinator could
+first just check whether we have our piece and only if we don't, try to recover
```
rphmeier, line 141: If the current node is a potential block producer or was not a parachain validator in the session of the dispute, then it won't have an availability piece at all. Am I misinterpreting the argument of this paragraph?

I do think it's accurate that having a chunk locally is a good proxy for general availability. However, not having a piece is not a proxy for unavailability. Even when the local node is a parachain validator, as only 2/3 of nodes need to sign off on availability. 

An attacker with f nodes can achieve availability by distributing chunks only to f+1 honest nodes. We need those f+1 as well as another f honest nodes to participate in the dispute so we can achieve supermajority and slash.

roadmap/implementers-guide/src/node/disputes/dispute-distribution.md
```diff
@@ -1,3 +1,351 @@
...
+availability. Our node having a piece would be proof enough of the
+data to be available and thus the dispute to not be spam.
+
+### Sending of messages
+
+Starting and participating in a dispute are pretty similar from the perspective
+of disptute distribution. Once we receive a `SendDispute` message we try to make
```
rphmeier, line 148:
```suggestion
of dispute distribution. Once we receive a `SendDispute` message we try to make
```

roadmap/implementers-guide/src/node/disputes/dispute-distribution.md
```diff
@@ -1,3 +1,351 @@
...
+sure to get the data out. We keep track of all the parachain validators that
+should see the message, which are all the parachain validators of the session
+where the dispute happened as they will want to participate in the dispute.  In
+addition we also need to get the votes out to all authorities of the current
+session (which might be the same or not and may change during the dispute).
+Those authorities will not participate in the dispute, but need to see the
+statements so they can include them in blocks.
+
+We keep track of connected parachain validators and authorities and will issue
+warnings in the logs if connected nodes are less than two thirds of the
```
rphmeier, line 158: What are validator operators meant to do in this situation?

roadmap/implementers-guide/src/node/disputes/dispute-distribution.md
```diff
@@ -1,3 +1,351 @@
...
+   queue of requests could get monopolized by those malicious requests and we
+   won't be able to import any valid disputes and we could run out of resources,
+   if we tried to process them all in parallel.
+
+For tackling 1, we make sure to not occupy resources before we don't know a
+candidate is available. So we will not record statements to disk until we
+recovered availability for the candidate or know by some other means that the
+dispute is legit.
+
+For 2, we will pick up on any dispute on restart, so assuming that any realistic
```
rphmeier, line 193: The dispute coordinator does not currently do this, but probably should.

roadmap/implementers-guide/src/node/disputes/dispute-distribution.md
```diff
@@ -1,3 +1,351 @@
...
+memory filling attack will take some time, we should be able to participate in a
+dispute under such attacks.
+
+For 3, full monopolization of the incoming queue should not be possible assuming
+substrate handles incoming requests in a somewhat fair way. Still we want some
+defense mechanisms, at the very least we need to make sure to not exhaust
+resources.
+
+The dispute coordinator will notify us
+via `DisputeDistributionMessage::ReportCandidateUnavailable` about unavailable
```
rphmeier, line 203: Dispute participation does this, not dispute coordinator, I assume.

roadmap/implementers-guide/src/node/disputes/dispute-distribution.md
```diff
@@ -1,3 +1,351 @@
...
+other hand we will also be very quick in processing them. A channel size of 100
+requests seems plenty and should be able to handle bursts adequately.
+
+### Node Startup
+
+On startup we need to check with the dispute coordinator for any ongoing
+disputes and assume we have not yet sent our statement for those. In case we
+find an explicit statement from ourselves via
+`DisputeCoordinatorMessage::QueryCandidateVotes` we will pretend to just have
+received a `SendDispute` message for that candidate.
```
rphmeier, line 268: 😎 

very glad you've addressed this.

roadmap/implementers-guide/src/node/disputes/dispute-distribution.md
```diff
@@ -1,3 +1,351 @@
...
+some of them might even be able to reconnect eventually.
+
+To mitigate those issues we will process dispute messages with a maximum
+parallelism `N`. We initiate import processes for up to `N` candidates in
+parallel. Once we reached `N` parallel requests we will start back pressuring on
+the incoming requests. This saves us from resource exhaustion.
+
+To reduce impact of malicious nodes further, we can keep track from which nodes the
+currently importing statements came from and will drop requests from nodes that
+already have imports in flight.
```
rphmeier, line 222: Ok, so the idea is that if a node tries to send K requests, that we'd begin processing 1 and then drop the K-1 other requests. An honest node is expected to retry, and in the meantime, we'll not accept requests from the other node. As we detect the data is not available, we drop the dispute and punish the bad peer. Does that sound accurate?

roadmap/implementers-guide/src/node/disputes/dispute-distribution.md
rphmeier: The guide page for participation should be updated so as to reflect the sending of this message.

rphmeier (approved):
Looks good modulo nits. I think the spam-resilience strategy makes sense and should protect against any practical attacks.

roadmap/implementers-guide/src/node/disputes/dispute-distribution.md
eskimor: I had spell check enabled ... damn it. Thanks!

roadmap/implementers-guide/src/node/disputes/dispute-distribution.md
eskimor: On a live network - check their internet connection, on a test network: find the bug. Like we had the case already a couple of times that we were not connected properly and we did only realize because of other issues that have been caused by this. Dispute distribution already tries its best - it will try to send requests, even if the receiver is not connected to us and it will keep trying.

I was just thinking about disputes and because they are so critical, I wanted to do whatever I can to ensure our messages gets out, but general connection warnings should go to gossip support already, I guess. And warnings when a dispute is already happening are a bit late - still useful though as a additional safety guard though. If validators become aware of a dispute that did not work out for some weird reasons/bugs/whatever we still have governance - better than nobody noticing.

roadmap/implementers-guide/src/node/disputes/dispute-distribution.md
eskimor: Yeah, I was thinking where this functionality should go and I decided to put it in distribution - with the following reasoning: The coordinator told distribution already to propagate the votes - it handed over that responsibility, so distribution should do whatever is needed to ensure it can do its job. Like it is literally the only job of distribution: Get that damn votes out. So the coordinator should not need to worry about it.

roadmap/implementers-guide/src/node/disputes/dispute-distribution.md
eskimor: Yeah sorry, I was to focused on distribution - obviously I need to update those other subsystems accordingly. Thanks!

roadmap/implementers-guide/src/node/disputes/dispute-distribution.md
eskimor: Almost. We receive a request from node A - we will start importing votes. We will now drop any requests from node A until the import finished. This even works for distributed attacks - we will drop requests of up to `N` nodes and after we discovered it was spam, we will disconnect - freeing up resources even more. Assuming substrate networking is sane (and it will not let a single source monopolize the incoming channel) this is also already a second level defense. So we really should be fine here. The most important thing (apart from preventing monopolization - which can only be handled effectively at the substrate level) is, that we won't participate and thus escalate in a spam dispute, as this would amplify the attackers resources by the power of all nodes.

roadmap/implementers-guide/src/node/disputes/dispute-distribution.md
eskimor: You are right - that was stupid. If the block producer had a piece, it would obviously be participating in the dispute. :roll_eyes:

roadmap/implementers-guide/src/node/disputes/dispute-distribution.md
eskimor: seems so. My editor highlighted it red otherwise and I checked markdown rules - it seems to be necessary. (Obviously there are different markdown implementations, so I am not 100% sure)

roadmap/implementers-guide/src/node/disputes/dispute-distribution.md
rphmeier: Yes, I agree, that makes sense. We should still alter the coordinator to participate in the local disputes in the case that it was in the process of participating when it shut down and thus did not issue a local statement yet. That is beyond the scope here, just writing it down for completeness.

rphmeier:
Follow-up changes should go in another PR. Thanks!

PR #3167, opened 2021-06-03
Fix teleport accounting and add some events

gavofyork:
Also makes a few renames to improve comprehensibility and readability.

### DIFF
--- a/runtime/parachains/src/ump.rs
+++ b/runtime/parachains/src/ump.rs
@@ -18,12 +18,11 @@ use crate::{
 	configuration::{self, HostConfiguration},
 	initializer,
 };
-use sp_std::{prelude::*, fmt, marker::PhantomData};
+use sp_std::{prelude::*, fmt, marker::PhantomData, convert::TryFrom};
 use sp_std::collections::{btree_map::BTreeMap, vec_deque::VecDeque};
 use frame_support::{decl_module, decl_event, decl_storage, StorageMap, StorageValue, weights::Weight, traits::Get};
 use primitives::v1::{Id as ParaId, UpwardMessage};
-
-const LOG_TARGET: &str = "runtime::ump-sink";
+use xcm::v0::Outcome;
 
 /// All upward messages coming from parachains will be funneled into an implementation of this trait.
 ///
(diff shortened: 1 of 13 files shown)


runtime/parachains/src/ump.rs
```diff
@@ -45,45 +44,56 @@ pub trait UmpSink {
...
-	fn process_upward_message(_: ParaId, _: &[u8], _: Weight) -> Option<Weight> {
-		Some(0)
+	fn process_upward_message(_: ParaId, _: &[u8], _: Weight) -> Result<Weight, (MessageId, Weight)> {
+		Ok(0)
 	}
 }
 
+/// Simple type used to identify messages for the purpose of reporting events. Secure if and only
+/// if the message content is unique.
+pub type MessageId = [u8; 32];
```
pepyakin, line 60: If we care about this invariant, should it be a unit type with its own constructor?

runtime/parachains/src/ump.rs
```diff
@@ -45,45 +44,56 @@ pub trait UmpSink {
...
-				VersionedXcm::V0(xcm_message) => {
-					let xcm_junction: Junction = Junction::Parachain(origin.into());
-					let xcm_location: MultiLocation = xcm_junction.into();
-					match XcmExecutor::execute_xcm(xcm_location, xcm_message, max_weight) {
-						Outcome::Complete(w) | Outcome::Incomplete(w, _) => Some(w),
-						Outcome::Error(XcmError::WeightLimitReached(..)) => None,
-						Outcome::Error(_) => Some(0),
+		use xcm::v0::{Xcm, Junction, MultiLocation, Error as XcmError};
+
+		let id = sp_io::hashing::blake2_256(&data[..]);
```
pepyakin, line 72: Because `decode` allows for unconsumed trailing data, this may lead to the situation where there are two values `x` and `y` of type `VersionedXcm` and `x == y` but `id(x) ≠ id(y)`. Not sure if that's a problem, but `MessageId` docs may imply that it is.

runtime/westend/src/lib.rs
```diff
@@ -849,6 +850,12 @@ parameter_types! {
 	pub CheckAccount: AccountId = XcmPallet::check_account();
 }
 
+#[test]
+fn print_check_account() {
+	use sp_core::crypto::Ss58Codec;
+	panic!("{}", XcmPallet::check_account().to_ss58check());
+}
```
pepyakin, line 857: I guess this wasn't meant to be committed.

xcm/xcm-executor/src/lib.rs
```diff
@@ -246,10 +246,12 @@ impl<Config: config::Config> XcmExecutor<Config> {
 				Config::XcmSender::send_xcm(reserve, Xcm::WithdrawAsset { assets, effects })?;
 			}
 			Order::InitiateTeleport { assets, dest, effects} => {
-				for asset in assets.iter() {
-					Config::AssetTransactor::check_out(&origin, asset);
+				// We must do this first in order to resolve wildcards.
+				let assets = holding.saturating_take(assets);
+				for asset in assets.assets_iter() {
+					Config::AssetTransactor::check_out(&origin, &asset);
```
shawntabrizi, line 252: This is the only logical change?

(and adding events)

shawntabrizi (approved):

runtime/parachains/src/ump.rs
shawntabrizi: id is 256 bit which will always fit into [u8; 32], so not sure what you mean here actually

pepyakin: Both `u8::decode(&mut &[69])` and `u8::decode(&mut &[69, 0])` return `69` as a result. `blake2_256(&[69])` and `blake2_256(&[69, 0])` obviously return different results.

gavofyork: Not really - the docs mean that the same `MessageId` must imply the same message. They don't imply the reverse.

pepyakin (approved):

PR #3187, opened 2021-06-05
Fix historical validation code runtime API

rphmeier:
The previous implementation was giving code based on the minimum relay-parent number needed to apply the code upgrade, which caused issues when there was a gap between the signal of the code upgrade and the actual application of it.

Here's an example:

A code upgrade signaled at block 10 is then applied at block 20. The parablock included at block 20 has a context of 18 in practice. When fetching historical validation code at block 20 with a context of 18, the `historical_validation_code` runtime API would return the new code as opposed to the old code.

This PR changes the implementation to focus on the block where the code is activated as opposed to the minimal block where the code could be activated.

I think the previous implementation was intended to be robust across forks, but upon closer inspection it's not possible to do so without losing information about the particular chain you're looking at.

A long-term fix would be to do https://github.com/paritytech/polkadot/issues/3188


pepyakin (approved):
This LGTM, but I must admit that I am not fluent at this module and had to reread the rest of it to understand this change.

PR #3195, opened 2021-06-08
bump spec_version

s3krit:
kusama 9040 has been deployed on chain which means any subsequent runtime changes require a new spec_version. v0.9.4 isn't out the door yet so we just bump rc version.


ordian (approved):

s3krit:
bot merge

PR #3198, opened 2021-06-08
update pvf heap page limit

xlc:
This is required to prevent pvf worker panic during runtime upgrade

cc @bkchr @pepyakin


pepyakin (approved):

bkchr (approved):

xlc:
Should this be medium or high priority? We need validators to be updated with this fix to prevent parachain validation issue.

PR #3202, opened 2021-06-10
Add Support For Crowdloan Credits

shawntabrizi:
This PR introduces a way for one acount to credit another user a balance to be used specifically for crowdloans.

The creditor will transfer some balance into a system level account managed by the Crowdloan pallet which will hold the balance.

Upon successful transfer, that credit will be then be assigned to the specified target user, who can then call `contribute_credit`, which will allow users to spend their funds from that system account just like making a normal crowdloan contribution.

Nothing else changes.

Basically, this is a way to transfer funds to a user that can ONLY be used for crowdloan contributions.

NOTE: A user can turn these credits into regular free balance by contributing them to a crowdloan which ends unsuccessfully.

If we want to prevent this, some more thought will be needed.

### DIFF
--- a/runtime/common/src/crowdloan.rs
+++ b/runtime/common/src/crowdloan.rs
@@ -53,9 +53,9 @@ use frame_support::{
 	ensure, Identity, PalletId,
 	storage::{child, ChildTriePrefixIterator},
 	traits::{
-		Currency, ReservableCurrency, Get, ExistenceRequirement::AllowDeath
+		Currency, ReservableCurrency, Get, ExistenceRequirement::{AllowDeath, KeepAlive},
 	},
-	pallet_prelude::Weight,
+	pallet_prelude::{Weight, DispatchResult},
 };
 use sp_runtime::{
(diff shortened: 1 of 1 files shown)


runtime/common/src/crowdloan.rs
```diff
@@ -707,6 +689,89 @@ impl<T: Config> Pallet<T> {
...
+			Credits::<T>::try_mutate(&who, |credit| -> DispatchResult {
+				ensure!(*credit <= value, Error::<T>::NotEnoughCredit);
+				let credit_account = Self::credit_account_id();
+				CurrencyOf::<T>::transfer(&credit_account, &fund_account, value, AllowDeath)?;
+				credit.saturating_reduce(value);
+				Ok(())
+			})?;
+		} else {
+			CurrencyOf::<T>::transfer(&who, &fund_account, value, AllowDeath)?;
+		}
```
shawntabrizi, line 737: This is the only logic changed in the contribute function

PR #3237, opened 2021-06-14
Allow Council to Use Scheduler

shawntabrizi:
This PR allows any council motion with 1/2 of the members or greater to use the scheduler, which can be helpful for setting things up in advance.


gavofyork (approved):

shawntabrizi:
Bot merge

PR #3244, opened 2021-06-14
[burnin] double allocator limit. 

kianenigma:


kianenigma:
This is added on top of the 9.3 release.

PR #3247, opened 2021-06-14
Better metrics for approval voting

rphmeier:
1. Add failure modes for the approvals produced metric. We're observing that the amount of approvals produced is far below the number of assignments when they should be 1:1. This change should help us see what's happening to these approvals.
2. Change the assignments produced metric to a histogram, which records the tranche of the local assignment. If things are working correctly we should only be producing assignments for tranche 0 and maybe tranches 1-3 when there are no-shows. This change will let us see if we're being overeager in assignments which is putting unnecessary load on the network.


ordian (approved):

drahnr (approved):

rphmeier:
bot merge

burdges:
We actually should see more assignments that approval if (1) anyone is no-showing but then actually turning up or (2) revealing their low tranche assignment late, so like a slow clock.  It's really the tranche that matters here.

PR #3264, opened 2021-06-16
Extract and test count_no_shows method for approval voting

Lldenaurois:
This commit extracts no_show computation into a pure function so that it can be
extensively unit tested.

This commit extracts no_show computation into a pure function so that it can be extensively unit tested. So far the tests behave mostly as I would expect, though I plan on further expanding them to better test the interactions of the `Tick` parameters.

The main thing that currently sticks out is the `next_no_show` calculation, which I suspect is adding an extra `clock_drift` factor. For example, in the current test harness, a validator assigned at tick 20 is counted as a no_show when drifted_tick_now is also 20. However, a validator assigned at tick 21 (which would be counted as a no_show at a drifted_tick_now of 21) is reported as a next_no_show of 31 (21 + clock_drift). If this is the correct behavior I can continue to add unit tests on this interaction, but wanted to clarify before proceeding.

### DIFF
--- a/node/core/approval-voting/src/approval_checking.rs
+++ b/node/core/approval-voting/src/approval_checking.rs
@@ -325,6 +325,12 @@ fn count_no_shows(
 			let is_no_show = !has_approved && no_show_at <= drifted_tick_now;
 
 			if !is_no_show && !has_approved {
+				// When doing the comparison above, no_show_at and drifted_tick_now are calculated
+				// with the clock_drift removed. The reason for adding back the clock_drift in
+				// computing next_no_show is so that the scheduler knows the deadline at which
+				// *this node* should observe whether or not the validator is a no show. Recall
+				// that when the when drifted_tick_now is computed during that subsequent wake up,
+				// the clock drift will be removed again to do the comparison above.
 				next_no_show = super::min_prefer_some(
 					next_no_show,
 					Some(no_show_at + clock_drift),
--- a/node/core/approval-voting/src/approval_checking.rs
+++ b/node/core/approval-voting/src/approval_checking.rs
@@ -297,6 +297,42 @@ fn filled_tranche_iterator<'a>(
 	pre.chain(approval_entries_filled).chain(post)
 }
 
+/// Computes the number of no_show validators in a set of assignments given the relevant approvals
+/// and tick parameters. This method also returns the next tick at which a no_show will occur
+/// amongst the set of validators that have not submitted an approval.
+///
+/// If the returned `next_no_show` is not None, there are two possible cases for the value of
+/// based on the earliest assignment `tick` of a non-approving, yet-to-be-no-show validator:
+///  - if `tick` <= `clock_drift`: the value will always be `clock_drift` + `no_show_duration`.
+///  - if `tick` >  `clock_drift`: the value is equal to `tick` + `no_show_duration`.
+fn count_no_shows(
+	assignments: &[(ValidatorIndex, Tick)],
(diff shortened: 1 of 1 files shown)


node/core/approval-voting/src/approval_checking.rs
```diff
@@ -938,6 +958,145 @@ mod tests {
...
+                approvals: vec![1, 2, 3],
+                clock_drift: 10,
+                no_show_duration: 10,
+                drifted_tick_now: 20,
+                exp_no_shows: 0,
+                exp_next_no_show: None,
+            },
+        ];
+
+        for test in tests {
```
rphmeier, line 1083: style nit: I'd prefer each test to get its own `#[test]` function which uses a helper function as the body of this loop

rphmeier:
>  For example, in the current test harness, a validator assigned at tick 20 is counted as a no_show when drifted_tick_now is also 20. However, a validator assigned at tick 21 (which would be counted as a no_show at a drifted_tick_now of 21) is reported as a next_no_show of 31 (21 + clock_drift). If this is the correct behavior I can continue to add unit tests on this interaction, but wanted to clarify before proceeding.

Yes, I think this is correct. The reasoning is that the `next_no_show` is measured in ticks (time), not tranches, so the scheduler can schedule a wake-up at the next no-show. With `clock_drift = 10, drifted_tick_now = 20`, the absolute tick is 30 and so a wakeup will be scheduled 1 tick from now at tick 31 to check for the no-show.

node/core/approval-voting/src/approval_checking.rs
Lldenaurois: Good to know, will make not of that going forward.

Lldenaurois: @rphmeier force pushed the change to create each test case individually. Should be good to go.

burdges:
Yes, we want being no-show to indicate that you'll likely not show within any sane time period, like your power when off or you got DoS.  If your assignment was late then your approval vote will likely be late too, so we measure no shows in wall time since we learned about your assignment.

node/core/approval-voting/src/approval_checking.rs
```diff
@@ -993,6 +1016,205 @@ mod tests {
...
+
+	#[test]
+	fn count_no_shows_validator_index_out_of_approvals_range_is_no_show() {
+		test_count_no_shows(NoShowTest {
+			assignments: vec![(ValidatorIndex(1000), 20)],
+			approvals: vec![],
+			clock_drift: 10,
+			no_show_duration: 10,
+			drifted_tick_now: 20,
+			exp_no_shows: 1, // TODO(ladi): should this be counted or ignored?
```
Lldenaurois, line 1201: Should validator indexes past the length of the approvals vector be ignored or handled?

rphmeier:
> I do think that a comment explaining logic in the API would be useful.

Please add a commit adding a comment to the `+ clock_drift` line which explains the logic. Remembering why that's correct was not obvious at all so it'll be good to have it there for posterity

node/core/approval-voting/src/approval_checking.rs
rphmeier: Should be ignored; Garbage-in-Garbage-out

Lldenaurois:
@rphmeier Added comment and modified logic such that we ignore indexes past the length of the approval vector.

Should be ready to go.

rphmeier (approved):

rphmeier:
bot merge

PR #3269, opened 2021-06-16
XCM `canonicalize` + `prepend_with` fix

shawntabrizi:
This introduces a new function `canonicalize` which will convert an XCM into its canonical/normalized form.

I would have preferred the name `normalize` but it seems this is the standard naming in Rust: https://doc.rust-lang.org/std/fs/fn.canonicalize.html

Using this function, we repair a bug with `prepend_with`, where an assumption about the number of elements we skip when forming the new object was incorrect.


xcm/src/v0/multi_location.rs
```diff
@@ -541,10 +541,37 @@ impl MultiLocation {
...
 
+	/// This function ensures a multi-junction is in it's canonicalized/normalized form, removing
+	/// any internal `[Non-Parent, Parent]` combinations.
+	pub fn canonicalize(&mut self) {
+		let mut normalized = MultiLocation::Null;
+		let mut iter = self.iter().peekable();
+		// We build up the the new normalized path by taking items from the original multi-location.
+		// When the next item we would add is `Parent`, we instead remove the last item assuming
+		// it is non-parent.
+		const EXPECT_MESSAGE: &'static str = "normalized path must be less than or equal to the original path length";
```
gavofyork, line 552: not a strict proof.

xcm/src/v0/multi_location.rs
```diff
@@ -541,10 +541,37 @@ impl MultiLocation {
 		}
 	}
 
+	/// This function ensures a multi-junction is in it's canonicalized/normalized form, removing
+	/// any internal `[Non-Parent, Parent]` combinations.
+	pub fn canonicalize(&mut self) {
+		let mut normalized = MultiLocation::Null;
+		let mut iter = self.iter().peekable();
```
gavofyork, line 548: why `.peekable()`?

shawntabrizi:
```suggestion
		let mut iter = self.iter();
```

xcm/src/v0/multi_location.rs
```diff
@@ -541,10 +541,40 @@ impl MultiLocation {
...
+	/// This function ensures a multi-junction is in it's canonicalized/normalized form, removing
+	/// any internal `[Non-Parent, Parent]` combinations.
+	pub fn canonicalize(&mut self) {
+		let mut normalized = MultiLocation::Null;
+		let mut iter = self.iter();
+		// We build up the the new normalized path by taking items from the original multi-location.
+		// When the next item we would add is `Parent`, we instead remove the last item assuming
+		// it is non-parent.
+		const EXPECT_MESSAGE: &'static str =
+			"This loop pushes at most N junctions to a new multi-location where N is the number \
```
gavofyork, line 553: please stick to the house style for proofs, as documented in https://github.com/paritytech/substrate/blob/master/docs/STYLE_GUIDE.md.

this means `;` between clauses and finishing with `; qed`.

xcm/src/v0/multi_location.rs
```diff
@@ -541,10 +541,42 @@ impl MultiLocation {
 		}
 	}
 
+	/// This function ensures a multi-junction is in it's canonicalized/normalized form, removing
```
rphmeier, line 544:
```suggestion
	/// This function ensures a multi-junction is in its canonicalized/normalized form, removing
```

rphmeier (approved):
Not sure I completely understand this, but it seems reasonable to minimize/canonicalize paths to ensure that all `parent` junctions are leading. Maybe XCM v1 should prevent this by differentiation the path up the tree with the path down the tree.

How would this work for non-hierarchical (sibling) paths?

```
A <-- B
      ^
      |
      v
C <-- D
```

From B, would something like `(Sibling(D), Parent)` be 'normalized' to point to `A` instead of  `C`?

shawntabrizi:
@rphmeier Gav proposed that we actually remove the parent junction from XCM all together: https://github.com/paritytech/polkadot/issues/3312

Not sure if we want to put this in XCM v1 and have a breaking change in the format for everyone, or actually increment to XCM v2.

As for your scenario, I don't think i quite understand the diagram you made. For canonicalization, a parachain should only ever have one parent, so `Sibling(X), Parent` should always resolve back to where we started.

xcm/src/v0/multi_location.rs
```diff
@@ -541,10 +541,42 @@ impl MultiLocation {
...
+			normalized.push(j.clone()).expect(EXPECT_MESSAGE);
+		}
+
+		core::mem::swap(self, &mut normalized);
+	}
+
+
 	/// Mutate `self` so that it is suffixed with `suffix`. The correct normalized form is returned,
-	/// removing any internal [Non-Parent, `Parent`]  combinations.
+	/// removing any internal `[Non-Parent, Parent`]  combinations.
```
shawntabrizi, line 577:
```suggestion
	/// removing any internal `[Non-Parent, Parent]`  combinations.
```

shawntabrizi:
bot merge

PR #3273, opened 2021-06-16
Add XCM Decode Limit

shawntabrizi:
This PR adds a maximum XCM decode depth of 256. This prevents deeply nested XCMs which can be complex to decode.

https://github.com/paritytech/srlabs_findings/issues/94

### DIFF
--- a/xcm/src/double_encoded.rs
+++ b/xcm/src/double_encoded.rs
@@ -15,7 +15,10 @@
 // along with Polkadot.  If not, see <http://www.gnu.org/licenses/>.
 
 use alloc::vec::Vec;
-use parity_scale_codec::{Encode, Decode};
+use parity_scale_codec::{Encode, Decode, DecodeLimit};
+
+/// Maximum nesting level for XCM decoding.
+pub const MAX_XCM_DECODE_DEPTH: u32 = 256;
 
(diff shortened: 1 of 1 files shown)


gavofyork:
Where did 256 come from? For now I would keep it as small as possible while ensure known needed functionality is covered. I'd suggest 8.

shawntabrizi:
I chose 256 from the extrinsics depth limit, but conservative is good with me 👍

xcm/src/double_encoded.rs
```diff
@@ -15,7 +15,10 @@
 // along with Polkadot.  If not, see <http://www.gnu.org/licenses/>.
 
 use alloc::vec::Vec;
-use parity_scale_codec::{Encode, Decode};
+use parity_scale_codec::{Encode, Decode, DecodeLimit};
+
+/// Maximum nesting level for XCM decoding.
+pub const MAX_XCM_DECODE_DEPTH: u32 = 256;
```
shawntabrizi, line 21:
```suggestion
pub const MAX_XCM_DECODE_DEPTH: u32 = 8;
```

gavofyork (approved):

shawntabrizi:
bot merge

Issue #2842, opened 2021-04-07
Get rid of request multiplexer


eskimor:
By forwarding receivers directly we not only get better performance in theory, but also the possibility of applying back pressure.

In statement distribution we already need this badly, so right now we have a hack to special case statement distribution, which should be removed.

Request input streams should get passed into subsystems at initialization.

PR #3282, opened 2021-06-17
Dispute distribution implementation

### LINKED ISSUE #3163: Implement Dispute Participation Subsystem

### LINKED ISSUE #3399: Recover availability in dispute coordinator for spam detection
See [guide](https://w3f.github.io/parachain-implementers-guide/node/disputes/dispute-distribution.html#reception--spam-considerations) for details.

eskimor:
This PR implements the dispute distribution as described in the [guide](https://w3f.github.io/parachain-implementers-guide/node/disputes/dispute-distribution.html) it also integrates dispute distribution with other subsystems.

# In particular:

- Implement dispute distribution
- Lets the `RuntimeInfo` work with a subsystem sender only, instead of a full `SubsystemContext`.
- Make dispute participation report back on the availability of the candidate
- Make dispute coordinator report back on the validity of an import (spam protection)
- Some general cleanup/code de-duplication
- Make request/response API more expressive

# Todos: 

- [x]  Sending side
- [x]  Handle node restarts
- [x] Receiver
- [x] Tests
- [x] Metrics
- [x] Communication coordinator <-> distribution is bidirectional: Double check for potential dead locks.
- [x] Cleanup imports

Vote recovery will be a second PR, disputes should already be fully functional without it, see https://github.com/paritytech/polkadot/issues/3398

Closes #3163 .
Closes #3399 .


node/network/dispute-distribution/src/sender/mod.rs
```diff
@@ -0,0 +1,361 @@
...
+use std::collections::HashMap;
+use std::collections::HashSet;
+use std::collections::hash_map::Entry;
+
+use futures::Future;
+use futures::channel::mpsc;
+use futures::channel::oneshot;
+use futures::future::RemoteHandle;
+
+use polkadot_node_network_protocol::IfDisconnected;
```
bkchr, line 27: You know that you can combine imports? :see_no_evil:

eskimor: I know that, rust-analyzer used to know that as well, but seems to have forgotten with some update  (Using add missing import on auto complete here). Either way, I will clean that up when I am done.

rphmeier:
> Vote recovery will be a second PR, disputes should already be fully functional without it.

Please file an issue! And yes, agreed. Best to split

node/network/dispute-distribution/src/sender/mod.rs
adoerr: rust-analyzer has an `ImportGranularity` config setting, which you might want to set to `module`.

adoerr: rust-analyzer has an `Import Granularity` config item. For me, `preserve` works best. This will just follow along with the current structure.

eskimor:
Noiice! Thanks for the tip!!

node/network/dispute-distribution/src/receiver/mod.rs
```diff
@@ -0,0 +1,392 @@
...
+
+use crate::LOG_TARGET;
+
+mod error;
+use self::error::{log_error, FatalResult, NonFatalResult, NonFatal, Fatal, Result};
+
+const COST_INVALID_REQUEST: Rep = Rep::CostMajor("Received message could not be decoded");
+const COST_INVALID_SIGNATURE: Rep = Rep::Malicious("Signatures were invalid");
+const COST_INVALID_CANDIDATE: Rep = Rep::Malicious("Reported candidate was not available.");
+const COST_NOT_A_VALIDATOR: Rep = Rep::Malicious("Reporting peer was not a validator.");
```
eskimor, line 56: Maybe too harsh. What if DHT was just not yet populated properly?

node/network/bridge/src/multiplexer.rs
```diff
@@ -68,19 +69,31 @@ impl RequestMultiplexer {
 			})
 			.unzip();
 
+		// Ok this code is ugly as hell, it is also a hack, see #2842. But it works and is executed
```
drahnr, line 72: Full paths to the issues are preferred :)

node/network/bridge/src/multiplexer.rs
```diff
@@ -68,19 +69,31 @@ impl RequestMultiplexer {
...
+		).expect("Statement fetching must be registered.");
 		let statement_fetching = Some(receivers.remove(index).1);
 
+		let index = receivers.iter().enumerate().find_map(|(i, (p, _))|
+			if let Protocol::DisputeSending = p {
+				Some(i)
+			} else {
+				None
+			}
+		).expect("Dispute sending must be registered.");
```
drahnr, line 89: A few too many expects imho that will be transformed into footguns in the future™. I'd prefer early returns and a warning/error message trace.

node/network/bridge/src/multiplexer.rs
```diff
@@ -174,6 +194,9 @@ fn multiplex_single(
 		Protocol::StatementFetching => {
 			panic!("Statement fetching requests are handled directly. qed.");
 		}
+		Protocol::DisputeSending => {
+			panic!("Dispute sending request are handled directly. qed.");
```
drahnr, line 198: `unreachable!` makes the invariant a bit more explicit, not sure why we don't use it much :monocle_face:

node/network/bridge/src/lib.rs
```diff
@@ -295,7 +302,7 @@ impl<N, AD> NetworkBridge<N, AD> {
 impl<Net, AD, Context> Subsystem<Context> for NetworkBridge<Net, AD>
 	where
 		Net: Network + Sync,
-		AD: validator_discovery::AuthorityDiscovery,
+		AD: validator_discovery::AuthorityDiscovery + Clone,
```
drahnr, line 305: Can't we use the first one and then use it with an `Arc`?

node/network/dispute-distribution/src/error.rs
```diff
@@ -0,0 +1,105 @@
...
+
+use polkadot_node_subsystem_util::{Fault, runtime, unwrap_non_fatal};
+use polkadot_subsystem::SubsystemError;
+
+use crate::LOG_TARGET;
+use crate::sender;
+
+#[derive(Debug, Error)]
+#[error(transparent)]
+pub struct Error(pub Fault<NonFatal, Fatal>);
```
drahnr, line 30: :+1: we should consider expanding this notion, or at least brainstorm something based on this for the usage across all subsystems and the interpretation of an _error_

node/network/dispute-distribution/src/lib.rs
```diff
@@ -0,0 +1,224 @@
...
+							self.handle_subsystem_message(&mut ctx, msg).await,
+					};
+					log_error(result, "on FromOverseer")?;
+				}
+				Message::Sender(result) => {
+					self.disputes_sender.on_task_message(
+						result.ok_or(Fatal::SenderExhausted)?
+					).await;
+				}
+			}
```
drahnr, line 133: nit: `fn _inner` to flatten the pyramid a bit

node/network/dispute-distribution/src/lib.rs
```diff
@@ -0,0 +1,224 @@
...
+	Subsystem(FatalResult<FromOverseer<DisputeDistributionMessage>>),
+	/// Messages from spawned sender background tasks.
+	Sender(Option<FromSendingTask>),
+}
+
+impl Message {
+	async fn receive(
+		ctx: &mut impl SubsystemContext<Message = DisputeDistributionMessage>,
+		from_sender: &mut mpsc::Receiver<FromSendingTask>,
+	) -> Message {
```
drahnr, line 208: nit: 
```suggestion
	) -> Self {
```

node/network/dispute-distribution/src/lib.rs
```diff
@@ -0,0 +1,224 @@
...
+					.map_err(Fatal::SpawnTask)?;
+			},
+
+		}
+		Ok(())
+	}
+}
+
+/// Messages to be handled in this subsystem.
+enum Message {
```
drahnr, line 197: `Message` is a bit ambiguous :bike: :house:

node/network/dispute-distribution/src/receiver/mod.rs
```diff
@@ -0,0 +1,399 @@
...
+		).await;
+
+		self.pending_imports.push(peer, confirmation_rx, pending_response);
+		Ok(())
+	}
+
+	/// Await an import and ban any misbehaving peers.
+	fn ban_bad_peer(
+		&mut self,
+		result: Option<NonFatalResult<(PeerId, ImportStatementsResult)>>
```
drahnr, line 280: The optional API seems a bit odd, this should be matched at the caller and skip the entire fn call.

node/network/protocol/src/request_response/request.rs
```diff
@@ -281,6 +278,98 @@ where
...
 
+/// Typed variant of [`netconfig::OutgoingResponse`].
+///
+/// Responses to `IncomingRequest`s.
+pub struct OutgoingResponse<Response> {
+	/// The payload of the response.
+	pub result: Result<Response, ()>,
+
+	/// Reputation changes accrued while handling the request. To be applied to the reputation of
+	/// the peer sending the request.
```
drahnr, line 289: Does this have to be a vec? So far (at least I) went with rep changes based on the first _violation_ detected. Not sure if this simplification can be applied here as well.

node/primitives/src/lib.rs
```diff
@@ -408,3 +415,196 @@ impl SignedDisputeStatement {
...
+	/// So in summary, this smart constructor should be smart enough to prevent from almost all
+	/// programming errors that one could realistically make here.
+	pub fn from_signed_statements(
+		valid_statement: SignedDisputeStatement,
+		valid_index: ValidatorIndex,
+		invalid_statement: SignedDisputeStatement,
+		invalid_index: ValidatorIndex,
+		candidate_receipt: CandidateReceipt,
+		session_info: &SessionInfo,
+	) -> Option<Self> {
```
drahnr, line 472: Should probably be a non-fatal error type, or have a few debug messages what caused the signed statement check to fail.

node/subsystem-util/src/runtime/mod.rs
```diff
@@ -73,32 +86,47 @@ pub struct ValidatorInfo {
 	pub our_group: Option<GroupIndex>,
 }
 
+impl Default for Config {
+	fn default() -> Self {
+		Self {
+			keystore: None,
+			session_cache_lru_size: 2,
```
drahnr, line 93: The old one had some rationale to why `2` is a good idea, I'd prefer to keep that.

node/subsystem-util/src/runtime/mod.rs
```diff
@@ -154,19 +182,19 @@ impl RuntimeInfo {
 	}
 
 	/// Convenience function for checking the signature of something signed.
-	pub async fn check_signature<Context, Payload, RealPayload>(
+	pub async fn check_signature<Sender, Payload, RealPayload>(
 		&mut self,
-		ctx: &mut Context,
+		sender: &mut Sender,
```
drahnr, line 187: sidenote: This is a good idea, but iirc you will need the context to spawn new messages - at least in the long run - the `ToOverseer::Spawn*` should be removed.

drahnr (commented):
Looks pretty well structured, a few small nits - as always - logic needs another pass.

node/network/bridge/src/lib.rs
eskimor: I guess we could, but the Arc is not necessary as the authority discovery service is an mpsc Sender which is perfectly clonable. The problem is, I need it as a trait object and that cannot be `Clone`.

node/network/dispute-distribution/src/receiver/mod.rs
eskimor: Yeah it is odd, but is also only used internally. The problem is, I need to handle the exact same thing twice in the run function (for reasons of flow control) and wanted to make sure those two paths won't diverge. Anyhow, handling the outer `Option` is not a big deal, so I will think about duplicating that handling.

node/network/protocol/src/request_response/request.rs
eskimor: Not sure this is useful at all, I was just mimicking the underlying low level substrate API. Argument against that change would be consistency across APIs - can't think of any other.

node/network/bridge/src/lib.rs
drahnr: In that case, ignore :) there is no point in doing what I suggested.

node/network/dispute-distribution/src/lib.rs
```diff
@@ -0,0 +1,224 @@
...
+			msg = from_overseer => Message::Subsystem(msg.map_err(Fatal::SubsystemReceive)),
+			msg = from_sender.next() => Message::Sender(msg),
+		)
+	}
+}
+
+/// Result of handling signal from overseer.
+enum SignalResult {
+	Conclude,
+	Result(Result<()>),
```
drahnr, line 223: I'd find

```
enum X {
Concluded,
NotSoMuch,
}
//..
Result<X>
```
much more intuitive

node/network/dispute-distribution/src/sender/send_task.rs
```diff
@@ -0,0 +1,310 @@
...
+/// Status of a particular vote/statement delivery to a particular validator.
+enum DeliveryStatus {
+	/// Request is still in flight.
+	Pending(RemoteHandle<()>),
+	/// Succeeded - no need to send request to this peer anymore.
+	Succeeded,
+}
+
+/// Messages from tasks trying to get disputes delievered.
+pub enum FromSendingTask {
```
drahnr, line 75: Make it a struct :)

node/network/dispute-distribution/src/sender/send_task.rs
```diff
@@ -0,0 +1,310 @@
...
+	/// Delivery of statements for given candidate finished for this authority.
+	Finished(CandidateHash, AuthorityDiscoveryId, TaskResult),
+}
```
drahnr, line 78:
```suggestion
/// Delivery of statements for given candidate finished for this 
pub struct TaskFinish {
	candidate_hash: CandidateHash,
	adid: AuthorityDiscoveryId,
	outcome: TaskResult,
}
```

node/network/dispute-distribution/src/receiver/mod.rs
```diff
@@ -0,0 +1,399 @@
...
+
+		let incoming = IncomingRequest::<DisputeRequest>::try_from_raw(
+			raw,
+			vec![COST_INVALID_REQUEST]
+		).map_err(NonFatal::FromRawRequest)?;
+
+		// Immediately drop requests from peers that already have requests in flight or have
+		// been banned recently (flood protection):
+		if self.pending_imports.peer_is_pending(&peer) || self.banned_peers.contains(&peer) {
+			return Ok(())
```
drahnr, line 212: Adding a trace log here is helpful for debugging

node/network/dispute-distribution/src/receiver/mod.rs
```diff
@@ -0,0 +1,399 @@
...
+/// Messages as handled by this receiver internally.
+enum Message {
+	/// An import got confirmed by the coordinator.
+	///
+	/// We need to handle those for two reasons:
+	///
+	/// - We need to make sure responses are actually sent (therefore we need to await futures
+	/// promptly).
+	/// - We need to update banned_peers accordingly to the result.
+	ConfirmedImport(Option<NonFatalResult<(PeerId, ImportStatementsResult)>>),
```
drahnr, line 102: I find the naming of this rather confusing, some :bike: :house: appreciated

drahnr (approved):
The allover flow looks fine, it's a bit on the higher complexity side I'd say, but that might be only due to the remaining :bike: :house: topics, so I'll re-review once those are addressed :+1:

drahnr (dismissed):
See above comment, just miss-clicked :facepalm:

node/network/dispute-distribution/src/lib.rs
```diff
@@ -0,0 +1,224 @@
...
+	async fn handle_subsystem_message<Context: SubsystemContext> (
+		&mut self,
+		ctx: &mut Context,
+		msg: DisputeDistributionMessage
+	) -> Result<()>
+	{
+		match msg {
+			DisputeDistributionMessage::SendDispute(dispute_msg) =>
+				self.disputes_sender.start_sending(ctx, &mut self.runtime, dispute_msg).await?,
+			// This message will only arrive once:
```
drahnr, line 172: :+1:

node/network/dispute-distribution/src/receiver/mod.rs
```diff
@@ -0,0 +1,399 @@
...
+	/// We keep record of the last banned peers.
+	///
+	/// This is needed because once we ban a peer, we will very likely still have pending requests
+	/// in the incoming channel - we should not waste time recovering availability for those, as we
+	/// already know the peer is malicious.
+	banned_peers: LruCache<PeerId, ()>,
+}
+
+/// Messages as handled by this receiver internally.
+enum Message {
```
drahnr, line 94:
```suggestion
enum IncomingMuxMessage {
```

drahnr: After editing the above suggestion for the 3rd time, I am not sure what to call this, I am only certain it's not `Message` :sweat_smile:

eskimor: MuxedMessage at least clarifies its purpose. I will go with it. ;-) Thanks!

node/network/bridge/src/multiplexer.rs
eskimor: I did not bother here, because the code is
a) to be abandoned soon™
b) more importantly - executed at startup and not in some seldomly hit code path.

Anyhow early returns might be fine as well, will check.

node/network/protocol/src/request_response/request.rs
drahnr: Unless we have a good reason, we should cut the `vec![]`s down to one place at the API boundary imho.

node/network/dispute-distribution/src/lib.rs
eskimor: The pyramid looks way worse on github than in the editor. Longest line length is 79, it is also very little code - another indirection would not really improve things, I believe.

node/core/dispute-coordinator/src/lib.rs
```diff
@@ -312,6 +313,10 @@ async fn handle_incoming(
 				session,
 				statements,
 			).await?;
+			// TODO: Recover availability and report back unavailable data:
```
ordian, line 316: is there an issue for this?

node/network/protocol/src/request_response/request.rs
eskimor: I have a "good" reason. If it is not a `Vec` it has to be an `Option`, so we still need to wrap the reputation. To me having to do a `vec![reputation]` is not really any worse than doing `Some(reputation)`, but at the same time the vec is slightly more expressive.

node/network/bridge/src/lib.rs
```diff
@@ -556,6 +563,15 @@ where
 						network_service = ns;
 						authority_discovery_service = ads;
 					}
+					NetworkBridgeMessage::GetAuthorityDiscoveryService(tx) => {
+						if let Err(err) = tx.send(Box::new(authority_discovery_service.clone())) {
```
ordian, line 567: Can we pass it at the configuration instead? Sending `Box<Arc<AuthorityDiscoveryService>>` on a request seems weird.

node/network/dispute-distribution/src/lib.rs
```diff
@@ -0,0 +1,236 @@
...
+
+#[cfg(test)]
+mod tests;
+
+mod metrics;
+//// Prometheus `Metrics` for dispute distribution.
+pub use metrics::Metrics;
+
+// #[cfg(test)]
+// mod tests;
```
ordian, line 63:
```suggestion
```

node/network/bridge/src/lib.rs
eskimor: Yeah, that makes a lot of sense!

eskimor:
CI seems to be having problems.

rphmeier:
A comprehensive PR description would be helpful!

node/core/dispute-coordinator/src/lib.rs
```diff
@@ -312,6 +313,11 @@ async fn handle_incoming(
...
+			// https://github.com/paritytech/polkadot/issues/3399
```
rphmeier, line 317: Dispute participation recovers availability. I assume we should not recover availability on every single imported statement, but only on those that initiate participation.

node/network/bridge/src/multiplexer.rs
```diff
@@ -68,19 +70,31 @@ impl RequestMultiplexer {
...
+		// Ok this code is ugly as hell, it is also a hack, see https://github.com/paritytech/polkadot/issues/2842.
+		// But it works and is executed on startup so, if anything is wrong here it will be noticed immediately.
 		let index = receivers.iter().enumerate().find_map(|(i, (p, _))|
 			if let Protocol::StatementFetching = p {
 				Some(i)
 			} else {
 				None
 			}
-		).expect("Statement fetching must be registered. qed.");
+		).expect("Statement fetching must be registered.");
```
rphmeier, line 81: "prove or remove" - looks like the 'QED' disappeared?

node/network/bridge/src/multiplexer.rs
```diff
@@ -68,19 +70,31 @@ impl RequestMultiplexer {
...
 		let statement_fetching = Some(receivers.remove(index).1);
 
+		let index = receivers.iter().enumerate().find_map(|(i, (p, _))|
+			if let Protocol::DisputeSending = p {
+				Some(i)
+			} else {
+				None
+			}
+		).expect("Dispute sending must be registered.");
```
rphmeier, line 90: "prove or remove"

node/network/protocol/src/authority_discovery.rs
```diff
@@ -0,0 +1,48 @@
...
+
+// Polkadot is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with Polkadot.  If not, see <http://www.gnu.org/licenses/>.
+
+//! All peersets and protocols used for parachains.
```
rphmeier, line 17: Doc comment seems wrong here.

node/network/protocol/src/request_response/request.rs
rphmeier: What's the concern with a `Vec`? Allocation performance? Could be `SmallVec<[UnifiedReputationChange; 1]>`

node/network/protocol/src/request_response/request.rs
```diff
@@ -284,6 +281,98 @@ where
 	}
 }
 
+/// Typed variant of [`netconfig::OutgoingResponse`].
+///
+/// Responses to `IncomingRequest`s.
+pub struct OutgoingResponse<Response> {
+	/// The payload of the response.
+	pub result: Result<Response, ()>,
```
rphmeier, line 289: I am curious about the use of `()` as the `Err` type - should it not be more expressive?

node/network/statement-distribution/src/lib.rs
```diff
@@ -577,7 +577,7 @@ struct FetchingInfo {
 }
 
 /// Messages to be handled in this subsystem.
-enum Message {
+enum MuxedMessage {
```
rphmeier, line 580: Nice

node/network/statement-distribution/src/responder.rs
```diff
@@ -85,35 +83,26 @@ pub async fn respond(
...
-			payload,
-			peer,
-			pending_response,
-		} = raw;
-
-		let payload = match StatementFetchingRequest::decode(&mut payload.as_ref()) {
+		let req =
+			match IncomingRequest::<StatementFetchingRequest>::try_from_raw(
+				raw,
+				vec![COST_INVALID_REQUEST]
```
rphmeier, line 89: nit: trailing comma

node/overseer/src/lib.rs
```diff
@@ -654,6 +664,9 @@ impl ChannelsOut {
 			AllMessages::GossipSupport(msg) => {
 				self.gossip_support.send(make_packet(signals_received, msg)).await
 			},
+			AllMessages::DisputeDistribution(msg) => {
+				self.dispute_distribution.send(make_packet(signals_received, msg)).await
```
rphmeier, line 668: Worth noting that other dispute subsystems aren't yet wired up. Does distribution handle that gracefully? We might want to wire them all up at the same time.

node/primitives/src/lib.rs
```diff
@@ -405,3 +412,227 @@ impl SignedDisputeStatement {
 		self.session_index
 	}
 }
+
+/// A `DisputeMessage` where signatures of statements have not yet been checked.
```
rphmeier, line 416: This is a network message, right? Perhaps the docs or type should clarify that.

node/primitives/src/lib.rs
```diff
@@ -405,3 +412,227 @@ impl SignedDisputeStatement {
...
+		valid_index: ValidatorIndex,
+		invalid_statement: SignedDisputeStatement,
+		invalid_index: ValidatorIndex,
+		candidate_receipt: CandidateReceipt,
+		session_info: &SessionInfo,
+	) -> Option<Self> {
+		let candidate_hash = *valid_statement.candidate_hash();
+		// Check statements concern same candidate:
+		if candidate_hash != *invalid_statement.candidate_hash() {
+			tracing::debug!("Candidate hashes did not match up");
```
rphmeier, line 473: Most of the time, our traces have a `LOG_TARGET` which is useful for observing the behavior of a specific subsystem. While I see the utility in having this exist in a common crate, it does seem that we lose debuggability as a result. I would prefer to have this function return a `Result<Self, DisputeMessageCheckError>` so logging can happen at a higher level.

node/primitives/src/lib.rs
```diff
@@ -405,3 +412,227 @@ impl SignedDisputeStatement {
...
+
+impl From<DisputeMessage> for UncheckedDisputeMessage {
+	fn from(message: DisputeMessage) -> Self {
+		message.0
+	}
+}
+
+/// Any invalid vote (currently only explicit).
+#[derive(Clone, Encode, Decode, Debug)]
+pub struct InvalidDisputeVote {
```
rphmeier, line 614: I think that the `DisputeStatementSet` definition could be improved by the use of this and `ValidDisputeVote` as opposed to the tuples we are currently using. We'd also gain consistency by that change.

node/subsystem-test-helpers/src/lib.rs
```diff
@@ -321,7 +324,7 @@ pub fn subsystem_test_harness<M, OverseerFactory, Overseer, TestFactory, Test>(
 
 	futures::executor::block_on(async move {
 		future::join(overseer, test)
-			.timeout(Duration::from_secs(2))
+			.timeout(Duration::from_secs(8))
```
rphmeier, line 327: Doc comment above needs changing, and also, I'm interested in the motivation of this change.

node/subsystem-test-helpers/src/mock.rs
```diff
@@ -0,0 +1,46 @@
...
+use sc_keystore::LocalKeystore;
+use sp_application_crypto::AppKey;
+use sp_keyring::Sr25519Keyring;
+use sp_keystore::{SyncCryptoStorePtr, SyncCryptoStore};
+
+use polkadot_primitives::v1::{ValidatorId, AuthorityDiscoveryId};
+
+
+/// Get mock keystore with `Ferdie` key.
+pub fn make_ferdie_keystore() -> SyncCryptoStorePtr {
```
rphmeier, line 30: Nice.

node/subsystem-util/src/runtime/mod.rs
```diff
@@ -73,32 +86,48 @@ pub struct ValidatorInfo {
...
+			keystore: cfg.keystore,
 		}
 	}
 
 	/// Retrieve the current session index.
-	pub async fn get_session_index<Context>(
+	pub async fn get_session_index<Sender>(
 		&mut self,
-		ctx: &mut Context,
+		sender: &mut Sender,
```
rphmeier, line 120: Glad to see this change.

node/subsystem/src/messages.rs
```diff
@@ -215,6 +206,12 @@ pub enum DisputeCoordinatorMessage {
 		/// The validator index passed alongside each statement should correspond to the index
 		/// of the validator in the set.
 		statements: Vec<(SignedDisputeStatement, ValidatorIndex)>,
+		/// Inform the requester once we finished importing.
+		///
+		/// This is, we either discarded the votes, just record them because we
+		/// casted our vote already or recovered availability for the candidate
+		/// successfully.
+		pending_confirmation: oneshot::Sender<ImportStatementsResult>
```
rphmeier, line 214: The dispute coordinator accepts all statements on all candidates, even undisputed ones. I implemented this in #3348. So there is some middle ground between "once we've cast our vote or recovered availability".

node/subsystem/src/messages.rs
```diff
@@ -215,6 +206,12 @@ pub enum DisputeCoordinatorMessage {
 		/// The validator index passed alongside each statement should correspond to the index
 		/// of the validator in the set.
 		statements: Vec<(SignedDisputeStatement, ValidatorIndex)>,
+		/// Inform the requester once we finished importing.
+		///
+		/// This is, we either discarded the votes, just record them because we
+		/// casted our vote already or recovered availability for the candidate
```
rphmeier, line 212: typo: casted -> cast

node/network/dispute-distribution/src/lib.rs
```diff
@@ -0,0 +1,230 @@
...
+// Polkadot is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with Polkadot.  If not, see <http://www.gnu.org/licenses/>.
+
+
+/// Sending and receiving of `DisputeRequest`s.
```
rphmeier, line 18: The documentation in this lib.rs should be much more comprehensive. It would be helpful to cover code structure and purpose and highlight key components of the implementation.

node/network/dispute-distribution/src/lib.rs
```diff
@@ -0,0 +1,230 @@
...
+};
+
+/// `DisputeSender` manages the sending side of all initiated disputes.
+///
+/// See the implementers guide for more detail on the general protocol, but in general
+/// `DisputeSender` takes care of getting our vote out to all other relevant validators.
+mod sender;
+use self::sender::{DisputeSender, TaskFinish};
+
+/// Handle receival of dispute requests.
```
rphmeier, line 43: I am not sure that `receival` is a word. `receipt` seems more correct

node/network/dispute-distribution/src/lib.rs
```diff
@@ -0,0 +1,230 @@
...
+				}
+			}
+		}
+	}
+
+	/// Handle overseer signals.
+	async fn handle_signals<Context: SubsystemContext> (
+		&mut self,
+		ctx: &mut Context,
+		signal: OverseerSignal
```
rphmeier, line 151: nit: trailing comma

node/subsystem/src/messages.rs
```diff
@@ -255,6 +261,18 @@ pub enum DisputeParticipationMessage {
...
+/// Messages going to the dispute distribution subsystem.
+#[derive(Debug)]
+pub enum DisputeDistributionMessage {
+
+	/// Tell dispute distribution to distribute an explicit dispute statement to
+	/// validators.
+	SendDispute(DisputeMessage),
+
+	/// Get receiver for receiving incoming network requests for dispute sending.
+	DisputeSendingReceiver(mpsc::Receiver<sc_network::config::IncomingRequest>),
```
rphmeier, line 273: I wonder if this separate `mpsc::Receiver` can introduce any type of deadlock. It might be why we see `statement-distribution`, which is the other subsystem using this pattern, as the one of the only other subsystems which becomes unresponsive.

node/network/dispute-distribution/src/lib.rs
```diff
@@ -0,0 +1,230 @@
...
+
+	/// Handle `DisputeDistributionMessage`s.
+	async fn handle_subsystem_message<Context: SubsystemContext> (
+		&mut self,
+		ctx: &mut Context,
+		msg: DisputeDistributionMessage
+	) -> Result<()>
+	{
+		match msg {
+			DisputeDistributionMessage::SendDispute(dispute_msg) =>
```
rphmeier, line 178: The dispute coordinator doesn't appear to send this message. What does?

rphmeier: If that's beyond the scope of this PR because it depends on the further changes you intend to make, please file a follow-up issue.

eskimor: Yes I have those ready and planned on doing a follow up, but it might actually be better to include those changes right away.

node/network/bridge/src/multiplexer.rs
eskimor: Yeah - it is not really proven, but rather obvious. I think it is fine for the following reasons:
1. It is true.
2. If it were not, this code would panic immediately on startup, so it is impossible to hit production, if we do any testing at all.
3. It is just a hack, until I will finally get rid of this multiplexer all together, see
#2842
and
#3432

ordian: I think @rphmeier refers to the style, specifically `; qed` at the end

node/overseer/src/lib.rs
eskimor: Very likely not, will remove the wiring for now. Thanks for the catch!

node/primitives/src/lib.rs
eskimor: Only indirectly, but yeah I will add a comment about its use.

node/subsystem-test-helpers/src/lib.rs
eskimor: 2 seconds was too low for dispute distribution tests to pass. (Crypto APIs in debug are awfully slow.)

node/network/bridge/src/multiplexer.rs
eskimor: Sure, I can add it back :-)

node/network/protocol/src/request_response/request.rs
eskimor: That's just mimicking the more low level substrate API - copied the comment from there for clarification.

node/primitives/src/lib.rs
eskimor: Not really [`DisputeStatementSet`](https://github.com/paritytech/polkadot/blob/a620156c0cdb46991b8eae89b99d1941aa8d9e18/primitives/src/v1/mod.rs#L1182) does not differentiate between valid and invalid votes.

node/subsystem/src/messages.rs
eskimor: Not one 100% what you mean, but I clarified the comment a bit. Hopefully it makes sense - it is quite late over here already.

node/subsystem/src/messages.rs
eskimor: I would not see how this might happen. They are both queried in their own tasks, not blocking the main subsystem at all.

node/network/dispute-distribution/src/receiver/error.rs
```diff
@@ -0,0 +1,110 @@
...
+/// Fatal errors of this subsystem.
+#[derive(Debug, Error)]
+pub enum Fatal {
+	/// Request channel returned `None`. Likely a system shutdown.
+	#[error("Request channel stream finished.")]
+	RequestChannelFinished,
+
+	/// Errors coming from runtime::Runtime.
+	#[error("Error while accessing runtime information")]
+	Runtime(#[from] #[source] runtime::Fatal),
```
rphmeier, line 59: This appears in both fatal & non-fatal. I don't quite follow why. It seems that runtime info access should not take down the subsystem so I think non-fatal is correct.

ordian: The fatal one is channel being closed, which means we're shutting down.
https://github.com/paritytech/polkadot/blob/519df6ca820539ca06f5c79eda2750dfa88ca5a3/node/subsystem-util/src/runtime/error.rs#L50

eskimor: There are fatal and non fatal errors from querying runtime info. Most are non fatal, there is one that I considered fatal though: https://github.com/paritytech/polkadot/blob/dcd672cb69adb03cc8ebfe764e74bc92348aa811/node/subsystem-util/src/runtime/error.rs#L50

eskimor: Reason, mostly that I would not expect that to happen.

rphmeier (approved):

node/core/dispute-coordinator/src/lib.rs
```diff
@@ -615,12 +658,67 @@ async fn issue_local_statement(
...
 			statements,
+			pending_confirmation,
 		).await?;
 	}
 
 	Ok(())
 }
 
+#[derive(Debug, thiserror::Error)]
+enum MakeDisputeMessageError {
```
drahnr, line 669: nit: consider rename to `DisputeMessageCreationError`

node/core/dispute-coordinator/src/lib.rs
```diff
@@ -615,12 +658,67 @@ async fn issue_local_statement(
...
+				= votes.invalid.get(0).ok_or(MakeDisputeMessageError::NoOppositeVote)?.clone();
+			let other_vote = SignedDisputeStatement::new_checked(
+				DisputeStatement::Invalid(statement_kind),
+				our_vote.candidate_hash().clone(),
+				our_vote.session_index(),
+				validators.get(validator_index.0 as usize).ok_or(MakeDisputeMessageError::InvalidValidatorIndex)?.clone(),
+				validator_signature,
+			).map_err(|()| MakeDisputeMessageError::InvalidStoredStatement)?;
+			(our_vote, our_index, other_vote, validator_index)
+	} else {
```
drahnr, line 701: nit: If else indentation is a bit off

node/network/dispute-distribution/src/metrics.rs
```diff
@@ -0,0 +1,109 @@
...
+
+/// Label for success counters.
+pub const SUCCEEDED: &'static str = "succeeded";
+
+/// Label for fail counters.
+pub const FAILED: &'static str = "failed";
+
+/// Dispute Distribution metrics.
+#[derive(Clone, Default)]
+pub struct Metrics(Option<MetricsInner>);
```
drahnr, line 29: :+1: we should consider extracting metrics to their own files for more subsystems

node/overseer/src/lib.rs
```diff
@@ -395,6 +396,9 @@ pub struct Overseer<SupportsParachains> {
 	#[subsystem(no_dispatch, wip, DisputeParticipationMessage)]
 	dispute_participation: DisputeParticipation,
 
+	#[subsystem(no_dispatch, wip, DisputeDistributionMessage)]
+	dipute_distribution: DisputeDistribution,
```
drahnr, line 400: type: `dispute_distribution`

drahnr (commented):
Post merge review -  LGTM :+1:  a few tiny nits

PR #3299, opened 2021-06-18
Set new staking limits

kianenigma:


runtime/kusama/src/lib.rs
```diff
@@ -1513,10 +1513,31 @@ pub type Executive = frame_executive::Executive<
...
 >;
 /// The payload being signed in the transactions.
 pub type SignedPayload = generic::SignedPayload<Call, SignedExtra>;
 
+pub struct SetStakingLimits;
+impl frame_support::traits::OnRuntimeUpgrade for SetStakingLimits {
+	fn on_runtime_upgrade() -> Weight {
+		// This will be the threshold needed henceforth to become a nominator. All nominators will
+		// less than this amount bonded are at the risk of being chilled by another reporter.
+		let min_nominator_bond = UNITS / 1000;
```
kianenigma, line 1526: While the polkadot limit is more educated, for this one I just pulled `KSM / 1000` out of thin air as a relatively lenient minimum.

kianenigma:
Will set some values for westend as well, as it will allow us to test stuff.

runtime/westend/src/lib.rs
```diff
@@ -1105,10 +1105,23 @@ pub type Executive = frame_executive::Executive<
...
 /// The payload being signed in transactions.
 pub type SignedPayload = generic::SignedPayload<Call, SignedExtra>;
 
+pub struct SetStakingLimits;
+impl frame_support::traits::OnRuntimeUpgrade for SetStakingLimits {
+	fn on_runtime_upgrade() -> Weight {
+		<pallet_staking::MinNominatorBond<Runtime>>::put(1 * UNITS);
+		<pallet_staking::MaxNominatorsCount<Runtime>>::put(1000);
+		<pallet_staking::MinValidatorBond<Runtime>>::put(10 * UNITS);
+		<pallet_staking::MaxValidatorsCount<Runtime>>::put(10);
```
shawntabrizi, line 1119: We want to make westend more complex than polkadot in order to detect issues in the future.

We should have no limits here

kianenigma: I preferred some limits here, and I set them intentionally low, such that we can try chilling others there.

shawntabrizi (approved):

shawntabrizi:
bot merge

runtime/kusama/src/lib.rs
gavofyork:
```suggestion
		let min_nominator_bond = UNITS / 10;
```

runtime/kusama/src/lib.rs
```diff
@@ -1513,10 +1513,31 @@ pub type Executive = frame_executive::Executive<
...
 
+pub struct SetStakingLimits;
+impl frame_support::traits::OnRuntimeUpgrade for SetStakingLimits {
+	fn on_runtime_upgrade() -> Weight {
+		// This will be the threshold needed henceforth to become a nominator. All nominators will
+		// less than this amount bonded are at the risk of being chilled by another reporter.
+		let min_nominator_bond = UNITS / 10;
+		// The absolute maximum number of nominators. This number is set rather conservatively, and
+		// is expected to increase soon after this runtime upgrade via another governance proposal.
+		// The current Polkadot state has more than 30_000 nominators, therefore no other nominator
```
antonkhvorov, line 1529: There are currently around 22418 _active_ nominators on Polkadot, I wonder from where 30 000 number comes from?

shawntabrizi: Its not about active nominators, its about number of nominators in the underlying `Nominators` storage map.

antonkhvorov: > Its not about active nominators, its about number of nominators in the underlying `Nominators` storage map.

I see, thank you! So new nominators would not be able to start staking, or those who have higher stake would be stored, and this will be recalculated for each era?

antonkhvorov: I would assume that only 20 000 with the highest stake will be counted

shawntabrizi: This limit would make it so that only 20_000 accounts could exist in that nominators map. Once we hit that limit, other nominators would not be able to register.

PR #3309, opened 2021-06-18
Follow-up PR: Count no-shows

Lldenaurois:
Quick follow up of #3264. In that PR we modified the `count_no_shows` method to ignore any `ValidatorIndex` that was greater than the length of the `approvals` bitvector. This adds a test showing the same behavior when counting `n_assignments`, where a `ValidatorIndex` that doesn't point to a valid slot in the `assignments` bitvector would be counted towards the number of assignments.

This is remedied by filtering out any indexes that are greater or equal to `n_validators` when counting the number of assignments. It still remains that `n_assignments` and `count_no_shows` inspect the number of validators and approvals, respectively.

Ideally we would add greater safety around ensuring these two values cannot differ.

### DIFF
--- a/node/core/approval-voting/src/approval_checking.rs
+++ b/node/core/approval-voting/src/approval_checking.rs
@@ -385,7 +385,10 @@ pub fn tranches_to_approve(
 				return None;
 			}
 
-			let n_assignments = assignments.len();
+			// Count the number of valid validator assignments.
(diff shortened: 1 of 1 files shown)


rphmeier (approved):

rphmeier:
bot merge

PR #3313, opened 2021-06-19
Improvements to `determine_new_blocks`

rphmeier:
These are extracted from the #3277 PR. Now `determine_new_blocks` only issues ancestry requests if there is a gap. It also avoids fetching the genesis block (0) in a cleaner way.


ordian (approved):
Seems fine.

PR #3334, opened 2021-06-21
Export `fungibles_adapter` Types

apopiak:
This PR exports types that should be usable from parachain runtimes but were private before.


shawntabrizi (approved):

apopiak:
bot merge

PR #3346, opened 2021-06-22
Do not allow any crowdloan contributions during the VRF period

shawntabrizi:
After the ending period of an auction, there is a delay needed to select the random block used to win the auction.

During this time, a crowdloan could receive contributions which would not help win the auction whatsoever.

Because we cannot identify which crowdloans would win a current auction, we must stop all crowdloans from accepting contributions during this VRF delay period.

This PR refactors the way that the Auction pallet and the `Auctioneer` trait reports the status of an auction, making it easy to detect these scenarios, and improving the overall code structure and readability.

This PR also repairs a small bug where the `EndingsCount` was being incremented multiple times per auction. No harm done here, but this is a proper fix for that behavior.

### DIFF
--- a/runtime/common/src/auctions.rs
+++ b/runtime/common/src/auctions.rs
@@ -27,7 +27,7 @@ use frame_support::{
 };
 use primitives::v1::Id as ParaId;
 use crate::slot_range::SlotRange;
-use crate::traits::{Leaser, LeaseError, Auctioneer, Registrar};
+use crate::traits::{Leaser, LeaseError, Auctioneer, Registrar, AuctionStatus};
 use parity_scale_codec::Decode;
 pub use pallet::*;
 
@@ -201,7 +201,7 @@ pub mod pallet {
 			// If the current auction was in its ending period last block, then ensure that the (sub-)range
 			// winner information is duplicated from the previous block in case no bids happened in the
 			// last block.
-			if let Some((offset, _sub_sample)) = Self::is_ending(n) {
+			if let AuctionStatus::EndingPeriod(offset, _sub_sample) = Self::auction_status(n) {
 				weight = weight.saturating_add(T::DbWeight::get().reads(1));
 				if !Winning::<T>::contains_key(&offset) {
 					weight = weight.saturating_add(T::DbWeight::get().writes(1));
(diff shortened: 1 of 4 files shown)


runtime/common/src/traits.rs
```diff
@@ -137,6 +137,47 @@ pub trait Leaser {
...
+	/// An auction has not started yet.
+	NotStarted,
+	/// We are in the opening period of the auction, collecting initial bids.
+	OpeningPeriod,
+	/// We are in the ending period of the auction, where we are taking snapshots of the winning
+	/// bids. This state supports "sampling", where we may only take a snapshot every N blocks.
+	/// In this case, the first number is the current sample number, and the second number
+	/// is the sub-sample. i.e. for sampling every 20 blocks, the 25th block in the ending period
+	/// will be `EndingPeriod(1, 5)`.
+	EndingPeriod(BlockNumber, BlockNumber),
```
emostov, line 152: nit: Just for symetry sake you could stick to opening/closing or starting/ending

shawntabrizi: yup, agreed

runtime/common/src/auctions.rs
```diff
@@ -201,7 +201,7 @@ pub mod pallet {
 			// If the current auction was in its ending period last block, then ensure that the (sub-)range
 			// winner information is duplicated from the previous block in case no bids happened in the
 			// last block.
```
emostov, line 203: Wouldn't it be in case no bids happen in this bock, not the last one?

emostov:
LGTM - the changes are a nice QoL improvement

emostov (approved):

runtime/common/src/auctions.rs
shawntabrizi: This is `on_initialize`, so when this function runs, there would be no transactions

emostov: Yup, that totally makes sense. My hang up then is just on 

> in case no bids happened in the last block.

In the ending period don't we always want to initialize the wining data for the block based on the last block, regardless of if any bids happened? 

(btw, not really concerned with changing the comment, just want to make sure I understand the system :))

shawntabrizi: The ending period has multiple samples and sub-samples (as you might see in some of the changes we made).

When entering a new sample, if there is not any data, we want to backfill it, but within a subsample, if there is already data, we want to update it.

I believe that is what the logic is doing here. We dont want to over-write any data, but we do want to propogate data when it does not exist.

Sample is not always 1:1 with block

gavofyork (approved):

runtime/common/src/crowdloan.rs
```diff
@@ -855,6 +863,12 @@ mod tests {
 	fn bids() -> Vec<BidPlaced> {
 		BIDS_PLACED.with(|p| p.borrow().clone())
 	}
+	fn vrf_delay() -> u64 {
```
kianenigma, line 866: unfortunately no one ever uses them 😁 but I added some pretty convenient ways to create all of these thread local stuff with `parameter_types!` with getter and setters https://crates.parity.io/frame_support/macro.parameter_types.html

runtime/common/src/auctions.rs
```diff
@@ -410,21 +400,24 @@ impl<T: Config> Pallet<T> {
...
+			AuctionStatus::EndingPeriod(o, _) => o,
+			AuctionStatus::VrfDelay(_) => return Err(Error::<T>::AuctionEnded.into()),
+		};
 
 		// Our range.
 		let range = SlotRange::new_bounded(first_lease_period, first_slot, last_slot)?;
 		// Range as an array index.
 		let range_index = range as u8 as usize;
-		let is_ending = Self::is_ending(frame_system::Pallet::<T>::block_number());
 		// The offset into the ending samples of the auction.
```
kianenigma, line 419:
```suggestion
```

runtime/common/src/auctions.rs
```diff
@@ -410,21 +400,24 @@ impl<T: Config> Pallet<T> {
...
-		// We need to check that the auction isn't in the period where it has definitely ended, but yeah we keep the
-		// info around because we haven't yet decided *exactly* when in the `EndingPeriod` that it ended.
-		let now = frame_system::Pallet::<T>::block_number();
-		ensure!(now < late_end, Error::<T>::AuctionEnded);
+		let (first_lease_period, _) = AuctionInfo::<T>::get().ok_or(Error::<T>::NotAuction)?;
+
+		// Get the auction status and the current sample block. For the starting period, the sample
+		// block is zero.
+		let auction_status = Self::auction_status(frame_system::Pallet::<T>::block_number());
+		let offset = match auction_status {
```
kianenigma, line 408:
```suggestion
		// The offset into the ending samples of the auction.
		let offset = match auction_status {
```

kianenigma (approved):
Looks good to me!

andresilva (approved):

runtime/common/src/crowdloan.rs
shawntabrizi: Actually i do normally use it, but copy paste from local code is faster than looking up the syntax lol

runtime/common/src/auctions.rs
```diff
@@ -316,16 +304,28 @@ impl<T: Config> Auctioneer for Pallet<T> {
...
+
+		let after_early_end = match now.checked_sub(&early_end) {
+			Some(after_early_end) => after_early_end,
+			None => return AuctionStatus::StartingPeriod,
+		};
+
+		let ending_period = T::EndingPeriod::get();
+		if after_early_end < ending_period {
+			let sample_length = T::SampleLength::get();
+			let sample = after_early_end / sample_length;
```
shawntabrizi, line 322:
```suggestion
			let sample = after_early_end / sample_length.max(1);
```

shawntabrizi:
bot merge

PR #3354, opened 2021-06-23
Use wasm-builder from git

bkchr:
This brings new features like compressed runtimes out of the box.

### DIFF
--- a/node/core/pvf/src/testing.rs
+++ b/node/core/pvf/src/testing.rs
@@ -31,7 +31,9 @@ pub fn validate_candidate(
 ) -> Result<Vec<u8>, Box<dyn std::error::Error>> {
 	use crate::executor_intf::{prevalidate, prepare, execute, TaskExecutor};
 
-	let blob = prevalidate(code)?;
+	let code = sp_maybe_compressed_blob::decompress(code, 10 * 1024 * 1024).expect("Decompressing code failed");
+
+	let blob = prevalidate(&*code)?;
 	let artifact = prepare(blob)?;
 	let executor = TaskExecutor::new()?;
 	let result = unsafe {
--- a/node/core/pvf/tests/it/main.rs
+++ b/node/core/pvf/tests/it/main.rs
@@ -15,9 +15,7 @@
 // along with Polkadot.  If not, see <http://www.gnu.org/licenses/>.
 
 use polkadot_node_core_pvf::{Pvf, ValidationHost, start, Config, InvalidCandidate, ValidationError};
-use polkadot_parachain::{
-	primitives::{BlockData, ValidationParams, ValidationResult},
-};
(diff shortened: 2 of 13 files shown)


chevdor:
We probably also want to bump up the 2 dependencies under `bridges`. Is that ok for you @tomusdrw ?

HCastano:
> We probably also want to bump up the 2 dependencies under `bridges`. Is that ok for you @tomusdrw ?

Feel free to open a PR in the bridges repo with this change. All our Substrate dependencies are Git based anyways, so I don't think this will be an issue

chevdor:
## Tests 

I tested the build of a few runtimes:
- [x] ✅ Kusama ⇒ Compression : 72.39% 
- [x] ✅ Polkadot ⇒ Compression : 70.64% (10 min)
- [x] ✅ Westend ⇒  Compression : 72.21% (19 min) 
- [x] rococo fails building, see https://github.com/paritytech/polkadot/issues/3357
 
The tests are run with `SRTOOL_TAG=1.53.0 srtool build --package polkadot-runtime` and passing `--runtime-dir` as/when appropriate.

NOTE: I skipped the bridges tests as the code is not current (Millau failed here but the issue is fixed in the bridges repo).

node/core/pvf/src/testing.rs
```diff
@@ -31,7 +31,9 @@ pub fn validate_candidate(
 ) -> Result<Vec<u8>, Box<dyn std::error::Error>> {
 	use crate::executor_intf::{prevalidate, prepare, execute, TaskExecutor};
 
-	let blob = prevalidate(code)?;
+	let code = sp_maybe_compressed_blob::decompress(code, 10 * 1024 * 1024).expect("Decompressing code failed");
+
+	let blob = prevalidate(&*code)?;
```
chevdor, line 36: Do we need this explicit deref ?

chevdor (approved):

node/core/pvf/tests/it/main.rs
```diff
@@ -58,11 +56,14 @@ impl TestHost {
 		params: ValidationParams,
 	) -> Result<ValidationResult, ValidationError> {
 		let (result_tx, result_rx) = futures::channel::oneshot::channel();
+
+		let code = sp_maybe_compressed_blob::decompress(code, 10 * 1024 * 1024).expect("Compression works");
```
pepyakin, line 60: Nit: that's the original value we use

```suggestion
		let code = sp_maybe_compressed_blob::decompress(code, 16 * 1024 * 1024).expect("Compression works");
```

pepyakin (approved):
One nit, not sure if it worth to address

node/core/pvf/tests/it/main.rs
bkchr: This was just randomly picked by me :D

PR #3360, opened 2021-06-24
Only send one collation per relay parent at a time to validators

bkchr:
This changes the way we are sending collations to validators. Before we
answered every collation request immediatley. Now we only answer one
pov request at a time per relay parent. This should bring down the
bandwidth requirements and should help parachains to include bigger
blocks more easily.

### DIFF
--- a/node/network/collator-protocol/src/collator_side/mod.rs
+++ b/node/network/collator-protocol/src/collator_side/mod.rs
@@ -62,7 +62,7 @@ impl Metrics {
 
 	fn on_collation_sent_requested(&self) {
 		if let Some(metrics) = &self.0 {
-			metrics.collations_sent_requested.inc();
+			metrics.collations_send_requested.inc();
 		}
 	}
 
@@ -82,7 +82,7 @@ impl Metrics {
 struct MetricsInner {
 	advertisements_made: prometheus::Counter<prometheus::U64>,
 	collations_sent: prometheus::Counter<prometheus::U64>,
-	collations_sent_requested: prometheus::Counter<prometheus::U64>,
+	collations_send_requested: prometheus::Counter<prometheus::U64>,
 	process_msg: prometheus::Histogram,
 }
 
@@ -98,7 +98,7 @@ impl metrics::Metrics for Metrics {
 				)?,
 				registry,
 			)?,
-			collations_sent_requested: prometheus::register(
+			collations_send_requested: prometheus::register(
 				prometheus::Counter::new(
 					"parachain_collations_sent_requested_total",
 					"A number of collations requested to be sent to validators.",
@@ -203,7 +203,7 @@ struct Collation {
 /// Stores the state for waiting collation fetches.
 #[derive(Default)]
 struct WaitingCollationFetches {
-	/// Is there currently a collation being fetched?
+	/// Is there currently a collation getting fetched?
 	collation_fetch_active: bool,
 	/// The collation fetches waiting to be fulfilled.
 	waiting: VecDeque<IncomingRequest<CollationFetchingRequest>>,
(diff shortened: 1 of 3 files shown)


node/network/collator-protocol/src/collator_side/mod.rs
```diff
@@ -185,6 +200,17 @@ struct Collation {
 	status: CollationStatus,
 }
 
+/// Stores the state for waiting collation fetches.
+#[derive(Default)]
+struct WaitingCollationFetches {
+	/// Is there currently a collation being fetched?
```
eskimor, line 206: getting fetched would make it a bit more clear what's happening, I think.

node/network/collator-protocol/src/collator_side/mod.rs
```diff
@@ -861,17 +930,48 @@ pub(crate) async fn run(
...
-		match msg {
-			Communication { msg } => {
-				log_error(
-					process_msg(&mut ctx, &mut runtime, &mut state, msg).await,
-					"Failed to process message"
-				)?;
+		async fn wait_for_collation_fetch(active: &mut ActiveCollationFetches) -> Hash {
+			loop {
+				if active.is_empty() {
+					futures::pending!()
```
eskimor, line 936: I mean you defined that function locally in the same function that uses it with select, but it still looks like standalone code, but it really is not. A comment referring to the select below would help, I think.

eskimor (approved):
Looks good. What we could do about the DOS by a single bad validator would be to not only move on with the queue on completion of the transfer, but also after some relatively short timeout (short enough so another validator will have enough time to do the fetch). So we won't cancel, but instead if a validator was too slow it will stop getting our full bandwidth at some delay (as apparently it does not utilize it fully anyway).

node/network/collator-protocol/src/collator_side/mod.rs
```diff
@@ -75,6 +82,7 @@ impl Metrics {
 struct MetricsInner {
 	advertisements_made: prometheus::Counter<prometheus::U64>,
 	collations_sent: prometheus::Counter<prometheus::U64>,
+	collations_sent_requested: prometheus::Counter<prometheus::U64>,
```
pepyakin, line 85: Nit: I think grammatically it would be more correct to use "send". In this case "requested" gets the 3rd participle form.

node/network/collator-protocol/src/collator_side/mod.rs
tomaka: `pending!()` doesn't automatically wake up. Is that really what you want here?

eskimor: I think it is - look at the `select!` below. There is not much better to do that returning `Pending` in that case and we will be woken up on new messages, which could lead to filling up the `FuturesUnordered` again. I had a similar use in my code just now and ended up writing the future myself with `poll_fn` as it becomes more clear what is going on that way, also then the code is tightly coupled with the other future, which makes their inter-dependency more obvious. If one sticks to those high level primitives, I would not know a better way of writing this either.

bkchr: Yes I want this ;) This `FuturesUnordered` returns `Ready(None)` if it is empty, so not that helpful and can not be filled anywhere in between.

tomaka: You can use `select_next_some` instead fwiw: https://docs.rs/futures/0.3.15/futures/stream/trait.StreamExt.html#method.select_next_some

bkchr: Nice ty! Much easier.

bkchr:
bot merge

PR #3361, opened 2021-06-24
Don't allow bids for a ParaId where there is an overlapping lease period

shawntabrizi:
This PR introduces a check when calling `handle_bid` which checks that an auction bid is not for a parachain which already has a lease period which overlaps with that bid.

This ensures that parachain teams do not steal potential lease periods from other teams, and that crowdloans will never be able to re-enter other auctions after winning (since the crowdloan has specific lease periods it bids for).


runtime/common/src/integration_tests.rs
```diff
@@ -1092,3 +1092,154 @@ fn gap_bids_work() {
...
 }
+
+// This test verifies that if a parachain already has won some lease periods, that it cannot bid for
+// any of those same lease periods again.
+#[test]
+fn cant_bid_on_existing_lease_periods() {
+	new_test_ext().execute_with(|| {
+		assert!(System::block_number().is_one()); // So events are emitted
+		Balances::make_free_balance_be(&1, 1_000_000_000);
+		// First register 2 parathreads
```
emostov, line 1103: Do you mean register 1 parathread?

runtime/common/src/traits.rs
```diff
@@ -135,6 +135,14 @@ pub trait Leaser {
 
 	/// Returns the current lease period.
 	fn lease_period_index() -> Self::LeasePeriod;
+
+	/// Returns true if the parachain already has a lease in any of lease periods in the inclusive
+	/// range `[first_period, last_period]`.
```
emostov, line 140: I think this may be slightly incorrect: technically our implementation in slots only can check the current lease period and onwards. So if any periods in `first_period..=last_period` are in the past we will not be able to look at those. Additionally we will always return `false` if the entire range is in the past (I don't think this is an issue but just wanted to point it out)

runtime/common/src/integration_tests.rs
```diff
@@ -1092,3 +1092,154 @@ fn gap_bids_work() {
...
+		assert_eq!(Balances::free_balance(&crowdloan_account), total);
+
+		// Finish the auction.
+		run_to_block(starting_block + 110);
+
+		// Appropriate Paras should have won slots
+		assert_eq!(
+			slots::Leases::<Test>::get(ParaId::from(2000)),
+			// -- 1 --- 2 --- 3 --- 4 --- 5 ------------- 6 ------------------------ 7 -------------
+			vec![None, None, None, Some((crowdloan_account, 8855)), Some((crowdloan_account, 8855))],
```
emostov, line 1150:
```suggestion
			// -- 1 --- 2 --- 3 ---- 4 -------------------------------5---------------------------------
			vec![None, None, None, Some((crowdloan_account, 8855)), Some((crowdloan_account, 8855))],
```

emostov: I think the comment is stale: we are only looking at 5 lease periods and the crowdloan is for periods 4 & 5

(Spacing might be a little whacky in my suggestion due to github ui)

runtime/common/src/slots.rs
```diff
@@ -421,6 +421,43 @@ impl<T: Config> Leaser for Module<T> {
...
+
+		// Get the leases, and check each item in the vec which is part of the range we ar checking.
+		let leases = Leases::<T>::get(para_id);
+		for slot in offset ..= offset + period_count {
+			if let Some(maybe_lease) = leases.get(slot) {
+				// If there exists any lease period, we exit early and return true.
+				if maybe_lease.is_some() {
+					return true
+				}
+			}
```
kianenigma, line 455:
```suggestion
			if let Some(Some(_)) = leases.get(slot) {
				// If there exists any lease period, we exit early and return true.
				return true
			}
```

runtime/common/src/slots.rs
```diff
@@ -421,6 +421,43 @@ impl<T: Config> Leaser for Module<T> {
...
+	) -> bool {
+		let current_lease_period = Self::lease_period_index();
+
+		// Can't look in the past, so we pick whichever is the biggest.
+		let start_period = first_period.max(current_lease_period);
+		// Find the offset to look into the lease period list.
+		// Subtraction is safe because of max above.
+		let offset = match (start_period - current_lease_period).checked_into::<usize>() {
+			Some(offset) => offset,
+			None => return false,
```
kianenigma, line 438: to be pedantic you could consider making these an `Err` rather than false, but might be an overkill. If the call site gave a `first_period` so large not to fit into usize then they are clearly wrogn anyhow.

runtime/common/src/traits.rs
kianenigma, line 140:
```suggestion
	/// Returns true if the parachain already has a lease in any of lease periods in the inclusive
	/// range `[first_period, last_period]`, intersected with the unbounded range [`current_lease_period`..] .
```

runtime/common/src/traits.rs
```diff
@@ -135,6 +135,14 @@ pub trait Leaser {
...
+	fn already_leased(
```
emostov, line 141: Thinking out loud a bit here, this could also fail if an auction ever covered any lease periods that are in the past because we simply would not be able to check for intersection. AFAICT this is not an issue because in `do_new_auction` we check

```
ensure!(lease_period_index >= T::Leaser::lease_period_index(), Error::<T>::LeasePeriodInPast);
```

but just wanted to point out in case there is some other method of creating an auction that may be used in the future. (But also why we would create an auction that covers something in the past?).

kianenigma (approved):
LGTM

emostov (commented):
The new logic looks good - planning on adding a test just to confirm my understanding

runtime/common/src/slots.rs
andresilva: I think `return true` would be the safest choice in the unlikely scenario that the conversion fails, worst case you prevent some crowdloan from happening.

runtime/common/src/slots.rs
```diff
@@ -421,6 +421,43 @@ impl<T: Config> Leaser for Module<T> {
...
+		};
+
+		// This calculates how deep we should look in the vec for a potential lease.
+		let period_count = match last_period.saturating_sub(start_period).checked_into::<usize>() {
+			Some(period_count) => period_count,
+			None => return false,
```
andresilva, line 444: Same as above.

runtime/common/src/slots.rs
```diff
@@ -421,6 +421,43 @@ impl<T: Config> Leaser for Module<T> {
...
+		};
+
+		// Get the leases, and check each item in the vec which is part of the range we ar checking.
```
andresilva, line 447:
```suggestion
		// Get the leases, and check each item in the vec which is part of the range we are checking.
```

runtime/common/src/slots.rs
shawntabrizi:
```suggestion
			None => return true,
```

runtime/common/src/slots.rs
shawntabrizi:
```suggestion
			None => return true,
```

runtime/common/src/traits.rs
shawntabrizi: yeah we should not concern ourselves with auctions in the past as they are not relevant for the present or future :)

runtime/common/src/integration_tests.rs
shawntabrizi:
```suggestion
		// First register a parathread
```

runtime/common/src/integration_tests.rs
shawntabrizi:
```suggestion
			// -- 1 --- 2 --- 3 ------------- 4 ------------------------ 5 -------------
			vec![None, None, None, Some((crowdloan_account, 8855)), Some((crowdloan_account, 8855))],
```

runtime/common/src/integration_tests.rs
```diff
@@ -1092,3 +1092,154 @@ fn gap_bids_work() {
...
+		// Start a new auction in the future
+		let starting_block = System::block_number();
+		let duration = 99u32;
+		let lease_period_index_start = 4u32;
+		assert_ok!(Auctions::new_auction(Origin::root(), duration, lease_period_index_start));
+
+		// 2 sessions later they are parathreads
+		run_to_session(2);
+
+		// Open a crowdloan for Para 1 for slots 0-3
```
emostov, line 1121:
```suggestion
		// Open a crowdloan for Para 1 for slots 4-5
```

emostov (approved):
lgtm 

Also some more tests here, but I don't think they are necessary: https://github.com/paritytech/polkadot/pull/3365

s3krit:
bot merge

PR #3365, opened 2021-06-25
Additional test for shawntabrizi-check-bid (#3361)

emostov:
Add a test to assert the following

- A bid that intersects with a won lease slot and a prior, non-won lease slot up for auction is rejected
- A bid that does not intersected with a won lease slot but does intersect with a prior, non-won lease slot is accepted
- A crowdloan that has already won its specified slot range will fail when entering a bid resulting from a poke

### DIFF
--- a/runtime/common/src/integration_tests.rs
+++ b/runtime/common/src/integration_tests.rs
@@ -75,8 +75,7 @@ frame_support::construct_runtime!(
 	}
 );
 
-use crate::crowdloan::Error as CrowdloanError;
-use crate::auctions::Error as AuctionsError;
+use crate::{auctions::Error as AuctionsError, crowdloan::Error as CrowdloanError};
 
 parameter_types! {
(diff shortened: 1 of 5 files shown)


runtime/common/src/integration_tests.rs
```diff
@@ -1243,3 +1243,114 @@ fn cant_bid_on_existing_lease_periods() {
...
+		run_to_block(starting_block + 100);
+
+		// Bid is rejected if intersects and includes a prior, non-intersecting slot
+		assert_noop!(
+			Auctions::bid(
+				Origin::signed(crowdloan_account),
+				ParaId::from(2000),
+				2,
+				lease_period_index_start,
+				lease_period_index_start + 1,
```
emostov, line 1340: this is a valid auction range

emostov:
Closing since the target branch/PR (#3361) is now merged

Issue #3310, opened 2021-06-19
Methods to avoid ever including parachain code in critical-path data


rphmeier:
When performing parachain code upgrades, we currently include the new parachain code.
1. When announcing the code upgrade, the full code appears in the candidate receipt.
2. When applying the code upgrade, the full code appears in the PoV, as it moves from one section of the state to the next.

The code we have empirically for parachains is quite large, typically in the 500K to 800K range.

We would like the new parachain code to never appear in the candidate receipt

This issue will be split into two sections, one for each of these points.

## Solving Code in Candidate Receipts: Hash-based announcements

At the moment, PVFs announce code upgrades by returning the full code when it's allowed, according to the state root of the relay chain. This code then appears, in full, in the candidate commitments. These candidate commitments are, in turn, gossiped among all validators so they can be included into the relay chain by the block author, who is most likely not a backer of the parachain doing the code upgrade.

An improvement to this situation would be for the PVF to only output the hash and size of the code, for inclusion in the candidate commitments.

Upon reaching the relay chain, the future code announcement creates a grace period where any user of the relay chain can upload the code using an `UnsignedTransaction`. These uploads are not on the critical path of parachain execution, and parachain code upgrades need to be delayed anyway for other reasons. Once the code is actually uploaded to the relay chain, the relay chain is ready for the parachain to upgrade its code and after the `code_upgrade_delay`, as specified in the `HostConfiguration`, the code can be upgraded at any time.

## Solving Code in upgrade parablock PoVs: Move code to the PVF parameters and the `AvailableData`.

When a parachain actually triggers its code upgrade, in practice, it involves the PVF moving the new code from one section of the trie to another. Although this is not strictly necessary within the parachain execution model, Cumulus-based parachains store their code in the state trie.

There are two approaches I considered to solve this problem:
  1. Write PVFs in such a way that the code doesn't need to be moved around in its state. What this means for Cumulus is making `:code` less special, or giving a way for `:code` to specify some other trie node which actually holds the real code.
  2. Pass the code, optionally, into the PVF, so it doesn't need to be loaded from the PoV. When it is passed, keep it available in the `AvailableData`.

The problem with approach 1 is that although we avoid including the code in the PoV at the point of the upgrade, we still have include the code in the PoV in some other block, where the code was moved into the storage of the parachain. This makes it a non-solution, so we'll ignore it and look at approach 2.

The idea of approach 2 is to make 3 alterations to parachain primitives:

```rust
struct AvailableData {
[...]

Issue #3211, opened 2021-06-11
Tracking issue: Pre-checking for PVF Compilation time


rphmeier:
We are going to use this issue for tracking progress on the PVF compilation time pre-checking.

- [x] https://github.com/paritytech/substrate/issues/9605
- [x] https://github.com/paritytech/substrate/issues/9606
- [x] https://github.com/paritytech/cumulus/pull/517
- [x] https://github.com/paritytech/polkadot/issues/3720
- [x] https://github.com/paritytech/polkadot/issues/3857
- [x] https://github.com/paritytech/polkadot/issues/4010
- [x] https://github.com/paritytech/polkadot/issues/4009
- [x] #4446
- [x] https://github.com/paritytech/polkadot/pull/4575
- [x] https://github.com/paritytech/polkadot/issues/4610
- [x] https://github.com/paritytech/polkadot/issues/4611
- [x] https://github.com/paritytech/polkadot/pull/4741
- [x] Benchmarks for the paras module logic with PVF pre-checking enabled. (https://github.com/paritytech/polkadot/pull/7138)
- [x] https://github.com/paritytech/polkadot/pull/4683
- [x] #4933
- [x] https://github.com/paritytech/polkadot/pull/4612
- [x] enable PVF pre-checking on Rococo.
- [x] enable PVF pre-checking on Westend.
- [x] enable PVF pre-checking on Kusama.
- [x] enable PVF pre-checking on Polkadot.
- [x] clean-up the code base with the assumption that PVF pre-checking is always on. Adapt the tests that assume PVF pre-checking is true, removing those that only depend on that. (https://github.com/paritytech/polkadot/pull/7138)
- [x] Jump on the next configuration module runtime storage upgrade to remove the configuration field. (https://github.com/paritytech/polkadot/pull/7396)
- [x] https://github.com/paritytech/polkadot/issues/4601 (https://github.com/paritytech/polkadot/pull/7486)
- [ ] finalize the implementor's guide. Remove WIP labels (https://github.com/paritytech/polkadot/pull/7559)

# Original Description

## Problem

If compilation of a Parachain Validation Function (PVF) takes too long or uses too much memory, this can leave a node in limbo as to whether a candidate of that parachain is valid or not. 

The amount of time that a PVF takes to compile is a subjective resource limit and as such PVFs may be maliciously crafted so that there is e.g. a 50/50 split of validators which can and cannot compile and execute the PVF.

This has the following implications:
[...]

PR #3371, opened 2021-06-25
Introduce upgrade go-ahead and upgrade restriction signals

pepyakin:
This patch introduces a new mechanism that will be used in future by parachains to apply validation code upgrade.

Currently, the upgrades are assumed to be activated after a fixed number of relay-chain blocks. Also there are configuration values that the parachain should keep an eye on to not signal an upgrade before an upgrade cooldown (aka `validation_upgrade_frequency`) timer elapses. These mechanics do not work well with our planned changes such as https://github.com/paritytech/polkadot/issues/3211 and https://github.com/paritytech/polkadot/issues/3310.

Thus as a first step to tackle this problem this patch introduces two new signals that should coordinate an upgrade:

- upgrade go ahead signal
- upgrade restriction signal

Instead of counting the number of the relay chain blocks passed since the upgrade was signalled the parachain will be looking at the upgrade go ahead signal through merkle proofs. 

The logic of upgrades themselves is not changed at this time.

This signal is almost always absent. However, at the relay-chain block the upgrade expected to happen, this signal will be turned on. When a parachain block is created on top of that relay-parent or it's descendant, it will observe the go ahead signal and enact the upgrade. The next parablock will be validated using the upgraded validation code.

As soon as the first block with the activated upgrade is included, the signal is reset.

The upgrade restriction signal is set when validation code upgrades are disallowed. We do not give details on why at this moment. To follow the current upgrade logic, this signal is set as soon as the upgrade is announced. The signal is unset only after `validation_upgrade_frequency` is elapsed.

# Rationale

## Why two signals, it seems that it's possible to get away with only one?

It's indeed was initial thinking that one signal will do. 

When I approached this problem I realized that semantically those are different states and cannot be represented by a single state machine. A contrived example: We have a parathread. The parathread doesn't produce blocks too often. The parathread schedules an upgrade. Then the parathread produces a block after `validation_upgrade_frequency`. If the signal had a single state only we would have to signal GoAhead first. We have to delay signalling that the upgrade is allowed  to another relay-block.

Besides being semantically unsound, it was also harder to implement a single signal model.


roadmap/implementers-guide/src/runtime/paras.md
```diff
@@ -137,6 +137,35 @@ PastCodePruning: Vec<(ParaId, BlockNumber)>;
 FutureCodeUpgrades: map ParaId => Option<BlockNumber>;
 /// The actual future code of a para.
 FutureCodeHash: map ParaId => Option<ValidationCodeHash>;
+/// This is used by the relay-chain to communicate to a parachain a go-ahead with in the upgrade procedure.
+///
+/// This value is abscent when there are no upgrades scheduled or during the time the relay chain
```
rphmeier, line 142: typo: abscent -> absent

primitives/src/v1/mod.rs
```diff
@@ -1012,6 +1045,35 @@ pub struct AbridgedHrmpChannel {
 	pub mqc_head: Option<Hash>,
 }
 
+/// A possible upgrade restriction that prevents a parachain from performing an upgrade.
+#[derive(Encode, Decode, PartialEq, RuntimeDebug)]
+pub enum UpgradeRestriction {
```
rphmeier, line 1050: It would be good to have explicit indices on these enums.

runtime/parachains/src/paras.rs
```diff
@@ -621,6 +663,22 @@ impl<T: Config> Pallet<T> {
 			}
 		}
 
+		if !outgoing.is_empty() {
+			// Filter offboarded parachains from the upcoming upgrades list.
+			//
+			// We do it after the offboarding to get away with only a single read/write.
+			<Self as Store>::UpcomingUpgrades::mutate(|upcoming_upgrades| {
```
rphmeier, line 670: Should `UpgradeCooldowns` be pruned as well?

rphmeier: I see that it will be naturally pruned based on block, so it may be redundant.

rphmeier (approved):
Looks like a welcome refactoring and more elegant solution to the upgrade signal process. Is the plan to migrate Cumulus and eventually phase out some of the other storage members?

pepyakin:
Yes, this patch is just a stepping stone. My rough plan is:

- merge this
- deploy this on {rococo, kusama}
- merge https://github.com/paritytech/cumulus/pull/517 and ask parachains to update their collators
- at this point, I imagine, `UpcomingUpgrades` won't be needed and will be substituted by another mechanism to signal the `GoAhead`. E.g. the pre-checking game.

runtime/parachains/src/paras.rs
pepyakin: Yeah, similar to `UpcomingUpgrades`. However, we perform insertions based on `UpcomingUpgrades` but for `UpgradeCooldowns` we only perform removals.

That said, I am ok to add filtering for cooldowns.

primitives/src/v1/mod.rs
pepyakin: https://github.com/paritytech/polkadot/pull/3371/commits/b058f8e6ff600d0a113bf21256f40c7b53b84267

roadmap/implementers-guide/src/runtime/paras.md
pepyakin: https://github.com/paritytech/polkadot/pull/3371/commits/d77aec15e9f35454970d595aa2ce34059c4bf4a7

runtime/parachains/src/paras.rs
pepyakin: https://github.com/paritytech/polkadot/pull/3371/commits/d0c134e794682405a6374e0e293d66671e38803a

rphmeier:
LGTM after merge addressed

PR #3373, opened 2021-06-25
pvf: Enable stack depth metering & NaN canonicalization

pepyakin:
Companion for https://github.com/paritytech/substrate/pull/9069


ordian (approved):
I assume this went through burn-in already.

pepyakin:
Sort of! 

I ran a custom experiment. I built a polkadot node which uses these parameters for the Substrate Runtime execution and synced kusama. The idea was to ensure that the executor can handle a more or less complicated runtime such as kusama's.

athei (approved):

PR #3377, opened 2021-06-28
Enable over-bridge-messaging in Rococo/Wococo runtime

svyatonik:
This PR adds two pallets (every pallet is 'deployed' twice - on Rococo and on Wococo) to Rococo/Wocoo runtime. The messages pallet configuration isn't trivial, hence the separate `bridge_messages.rs` file. Normally there'll be to files like that - one that describes bridge with Chain2 and which is a part of Chain1 runtime and another one that describes bridge with Chain1 and which is a part of Chain2 runtime. Here we have a slightly different scenario when there's one runtime => the single file.

Short pallets description (since it is first reference from the polkadot repo):
- `pallet-bridge-messages`: is the 'core' pallet of this PR. It has the only interesting (for end-user/apps) entrypoint - `send_message`, which enqueues your message (encoded call of the bridged chain) in the internal queue. Eventually, your message will be picked up by the offchain (not to confuse with offchain workers) relayer, which will submit delivery transaction to the bridged chain. Once messages are delivered, relayer will submit confirmation transaction back to the original chain. The side-effect of this tx is the `MessagesDelivered` event, where you could find info on whether message dispatch has succeeded or not;
- `pallet-bridge-dispatch`: a simple pallet that will try to decode the message from the bridged chain as a call of **this** chain (in Rococo<>Wococo case it is the same call). If call is decoded and passes all other checks (like `spec_version` check, weight check, ...), it is dispatched. In any case - for every incoming message, it'll emit event (`MessageWeightMismatch`, `MessageVersionSpecMismatch`, `MessageDispatched`, ...).

Deployment details (mostly for ones, who has some internal knowledge of pallets):
- there are no any parameters (like conversion rate) in messages pallet configuration;
- the owner of pallets will be empty initially (so only root can halt pallet). Could implement runtime upgrade, but it would be easier to transfer ownership with sudo call (or not to do that at all);
- all constants introduced in this PR (which are not a part of `bp_rococo`/`bp_wococo`) are a copy of Millau constants.

### DIFF
--- a/runtime/rococo/src/bridge_messages.rs
+++ b/runtime/rococo/src/bridge_messages.rs
@@ -139,7 +139,7 @@ impl<B, GI> BridgedChainWithMessages for RococoLikeChain<B, GI> {
 		// we don't want to relay too large messages + keep reserve for future upgrades
 		let upper_limit = messages_target::maximal_incoming_message_dispatch_weight(max_extrinsic_weight());
 
-		// we're charging for payload bytes in `With?ococoMessageBridge::transaction_payment` function
+		// we're charging for payload bytes in `With(Wococo | Rococo)MessageBridge::transaction_payment` function
 		//
 		// this bridge may be used to deliver all kind of messages, so we're not making any assumptions about
 		// minimal dispatch weight here
--- a/runtime/rococo/src/bridge_messages.rs
+++ b/runtime/rococo/src/bridge_messages.rs
@@ -1,0 +1,409 @@
+// Copyright 2020 Parity Technologies (UK) Ltd.
+// This file is part of Polkadot.
+
+// Polkadot is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+
+// Polkadot is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with Polkadot.  If not, see <http://www.gnu.org/licenses/>.
+
+//! Over-bridge messaging support for Rococo <> Wococo bridge.
+
+pub use self::{at_rococo::*, at_wococo::*};
+
+use bp_messages::{
+	source_chain::TargetHeaderChain,
+	target_chain::{ProvedMessages, SourceHeaderChain},
+	InboundLaneData, LaneId, Message, MessageNonce,
+};
+use bp_rococo::{EXTRA_STORAGE_PROOF_SIZE, MAXIMAL_ENCODED_ACCOUNT_ID_SIZE, max_extrinsic_size, max_extrinsic_weight};
+use bp_runtime::{ROCOCO_CHAIN_ID, WOCOCO_CHAIN_ID, ChainId};
+use bridge_runtime_common::messages::{
+	BridgedChainWithMessages, ChainWithMessages, MessageBridge, MessageTransaction, ThisChainWithMessages,
+	source as messages_source, target as messages_target,
+};
+use frame_support::{traits::Get, weights::{Weight, WeightToFeePolynomial}, RuntimeDebug};
+use sp_std::{convert::TryFrom, marker::PhantomData, ops::RangeInclusive};
+
+/// Maximal number of pending outbound messages.
+const MAXIMAL_PENDING_MESSAGES_AT_OUTBOUND_LANE: MessageNonce = bp_rococo::MAX_UNCONFIRMED_MESSAGES_AT_INBOUND_LANE;
+/// Maximal weight of single message delivery confirmation transaction on Rococo/Wococo chain.
+///
+/// This value is a result of `pallet_bridge_messages::Pallet::receive_messages_delivery_proof` weight formula
+/// computation for the case when single message is confirmed. The result then must be rounded up to account
+/// possible future runtime upgrades.
+const MAX_SINGLE_MESSAGE_DELIVERY_CONFIRMATION_TX_WEIGHT: Weight = 2_000_000_000;
+/// Increase of delivery transaction weight on Rococo/Wococo chain with every additional message byte.
+///
+/// This value is a result of `pallet_bridge_messages::WeightInfoExt::storage_proof_size_overhead(1)` call. The
+/// result then must be rounded up to account possible future runtime upgrades.
+const ADDITIONAL_MESSAGE_BYTE_DELIVERY_WEIGHT: Weight = 25_000;
+/// Weight of single regular message delivery transaction on Rococo/Wococo chain.
+///
+/// This value is a result of `pallet_bridge_messages::Pallet::receive_messages_proof_weight()` call
+/// for the case when single message of `pallet_bridge_messages::EXPECTED_DEFAULT_MESSAGE_LENGTH` bytes is delivered.
+/// The message must have dispatch weight set to zero. The result then must be rounded up to account
+/// possible future runtime upgrades.
+const DEFAULT_MESSAGE_DELIVERY_TX_WEIGHT: Weight = 1_500_000_000;
+/// Weight of pay-dispatch-fee operation for inbound messages at Rococo/Wococo chain.
+///
+/// This value corresponds to the result of `pallet_bridge_messages::WeightInfoExt::pay_inbound_dispatch_fee_overhead()`
+/// call for your chain. Don't put too much reserve there, because it is used to **decrease**
+/// `DEFAULT_MESSAGE_DELIVERY_TX_WEIGHT` cost. So putting large reserve would make delivery transactions cheaper.
+const PAY_INBOUND_DISPATCH_FEE_WEIGHT: Weight = 600_000_000;
+/// Number of bytes, included in the signed Rococo/Wococo transaction apart from the encoded call itself.
+///
+/// Can be computed by subtracting encoded call size from raw transaction size.
+const TX_EXTRA_BYTES: u32 = 130;
+
+/// Rococo chain as it is seen at Rococo.
+pub type RococoAtRococo = RococoLikeChain<AtRococoWithWococoMessageBridge, crate::RococoGrandpaInstance>;
+
+/// Rococo chain as it is seen at Wococo.
+pub type RococoAtWococo = RococoLikeChain<AtWococoWithRococoMessageBridge, crate::RococoGrandpaInstance>;
+
+/// Wococo chain as it is seen at Wococo.
+pub type WococoAtWococo = RococoLikeChain<AtWococoWithRococoMessageBridge, crate::WococoGrandpaInstance>;
+
+/// Wococo chain as it is seen at Rococo.
+pub type WococoAtRococo = RococoLikeChain<AtRococoWithWococoMessageBridge, crate::WococoGrandpaInstance>;
+
+/// Rococo/Wococo chain from message lane point of view.
+#[derive(RuntimeDebug, Clone, Copy)]
+pub struct RococoLikeChain<B, GI> {
+	__bridge_definition: PhantomData<B>,
+	_at_this_chain_grandpa_pallet_instance: PhantomData<GI>,
+}
+
+impl<B, GI> ChainWithMessages for RococoLikeChain<B, GI> {
+	type Hash = crate::Hash;
+	type AccountId = crate::AccountId;
+	type Signer = primitives::v1::AccountPublic;
+	type Signature = crate::Signature;
+	type Weight = Weight;
+	type Balance = crate::Balance;
+}
+
+impl<B, GI> ThisChainWithMessages for RococoLikeChain<B, GI> {
+	type Call = crate::Call;
+
+	fn is_outbound_lane_enabled(lane: &LaneId) -> bool {
+		*lane == [0, 0, 0, 0]
+	}
+
+	fn maximal_pending_messages_at_outbound_lane() -> MessageNonce {
+		MAXIMAL_PENDING_MESSAGES_AT_OUTBOUND_LANE
+	}
+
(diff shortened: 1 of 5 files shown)


runtime/rococo/src/bridge_messages.rs
```diff
@@ -0,0 +1,409 @@
...
+		MessageTransaction {
+			dispatch_weight: MAX_SINGLE_MESSAGE_DELIVERY_CONFIRMATION_TX_WEIGHT,
+			size: inbound_data_size
+				.saturating_add(EXTRA_STORAGE_PROOF_SIZE)
+				.saturating_add(TX_EXTRA_BYTES),
+		}
+	}
+
+	fn transaction_payment(transaction: MessageTransaction<Weight>) -> crate::Balance {
+		// current fee multiplier is used here
```
svyatonik, line 122: This will also likely be changed (e.g. x2-ed, not sure yet) in the P<>K deployment, because it describes future transactions, where multiplier may increase.

runtime/rococo/src/bridge_messages.rs
```diff
@@ -0,0 +1,409 @@
...
+/// Wococo chain as it is seen at Wococo.
+pub type WococoAtWococo = RococoLikeChain<AtWococoWithRococoMessageBridge, crate::WococoGrandpaInstance>;
+
+/// Wococo chain as it is seen at Rococo.
+pub type WococoAtRococo = RococoLikeChain<AtRococoWithWococoMessageBridge, crate::WococoGrandpaInstance>;
+
+/// Rococo/Wococo chain from message lane point of view.
+#[derive(RuntimeDebug, Clone, Copy)]
+pub struct RococoLikeChain<B, GI> {
+	__bridge_definition: PhantomData<B>,
```
HCastano, line 81:
```suggestion
	_bridge_definition: PhantomData<B>,
```

runtime/rococo/src/bridge_messages.rs
```diff
@@ -0,0 +1,409 @@
...
+impl<B, GI> BridgedChainWithMessages for RococoLikeChain<B, GI> {
+	fn maximal_extrinsic_size() -> u32 {
+		max_extrinsic_size()
+	}
+
+	fn message_weight_limits(_message_payload: &[u8]) -> RangeInclusive<Weight> {
+		// we don't want to relay too large messages + keep reserve for future upgrades
+		let upper_limit = messages_target::maximal_incoming_message_dispatch_weight(max_extrinsic_weight());
+
+		// we're charging for payload bytes in `With?ococoMessageBridge::transaction_payment` function
```
HCastano, line 142: Is this a typo, or a shorthand for `Rococo/Wococo`? If it's a shorthand I would do something like:
```suggestion
		// we're charging for payload bytes in `With(Wococo | Rococo)MessageBridge::transaction_payment` function
```

Instead

runtime/rococo/src/bridge_messages.rs
```diff
@@ -0,0 +1,409 @@
...
+	use super::*;
+
+	/// Message bridge that is 'deployed' at Rococo chain and connecting it to Wococo chain.
+	#[derive(RuntimeDebug, Clone, Copy)]
+	pub struct AtRococoWithWococoMessageBridge;
+
+	impl MessageBridge for AtRococoWithWococoMessageBridge {
+		const THIS_CHAIN_ID: ChainId = ROCOCO_CHAIN_ID;
+		const BRIDGED_CHAIN_ID: ChainId = WOCOCO_CHAIN_ID;
+		const RELAYER_FEE_PERCENT: u32 = 10;
```
HCastano, line 252: I think in the future it would be cool to have relayers dictate this rate themselves in order to create more realistic market dynamics

runtime/rococo/src/bridge_messages.rs
```diff
@@ -0,0 +1,409 @@
...
+	use parity_scale_codec::Encode;
+	use super::*;
+
+	#[test]
+	fn ensure_rococo_messages_weights_are_correct() {
+		// **NOTE**: the main purpose of this test is to be sure that any message that is sumbitted
+		// to (any) inbound lane in Rococo<>Wococo bridge can be delivered to the bridged chain.
+		// Since we deal with testnets here, in case of failure + urgency:
+		//
+		// 1) ping bridges team about this failure;
```
HCastano, line 343:
```suggestion
		// 1) ping bridges team about this failure (see the CODEOWNERS file if you're unsure who to ping);
```

runtime/rococo/src/bridge_messages.rs
```diff
@@ -0,0 +1,409 @@
...
+		// 2) comment/remove the test.
```
HCastano, line 344:
```suggestion
		// 2) comment out or ignore the test.
```

runtime/rococo/src/bridge_messages.rs
```diff
@@ -0,0 +1,409 @@
...
+		);
+	}
+
+	#[test]
+	fn ensure_rococo_tx_extra_bytes_constant_is_correct() {
+		// **NOTE**: this test checks that we're computing transaction fee (for bridged chain, which, in
+		// case of Rococo<>Wococo, means any chain) on-chain properly. If this assert fails:
+		//
+		// 1) just fix the `TX_EXTRA_BYTES` constant to actual (or sightly rounded up) value;
+		// 2) (only if it has changed significantly (> x2 times)) ping the bridges team.
```
HCastano, line 388:
```suggestion
		// 2) (only if it has changed significantly (> x2 times)) ping the bridges team (see the CODEOWNERS file if you're unsure who to ping)
```

This is probably gonna be a long line, so wrap it accordingly :P

runtime/rococo/src/lib.rs
```diff
@@ -245,6 +248,13 @@ construct_runtime! {
 		// Validator Manager pallet.
 		ValidatorManager: validator_manager::{Pallet, Call, Storage, Event<T>},
 
+		// Bridge messages support. The same story as with the bridge grandpa pallet above ^^^ - when we're
+		// running as Rococo we only use `BridgeWococoMessages`/`BridgeWococoMessagesDispatch`, and vice versa.
+		BridgeRococoMessages: pallet_bridge_messages::{Pallet, Call, Storage, Event<T>, Config<T>} = 43,
```
HCastano, line 253: It's a little unfortunate that we're being split by `ValidatorManager`, do you think anything (major) would break if we moved it after the message lane pallets 👀

runtime/rococo/src/lib.rs
```diff
@@ -245,6 +248,13 @@ construct_runtime! {
 		// Validator Manager pallet.
 		ValidatorManager: validator_manager::{Pallet, Call, Storage, Event<T>},
 
+		// Bridge messages support. The same story as with the bridge grandpa pallet above ^^^ - when we're
+		// running as Rococo we only use `BridgeWococoMessages`/`BridgeWococoMessagesDispatch`, and vice versa.
```
HCastano, line 252:
```suggestion
		// Bridge messages support.
		// The same story as with the bridge grandpa pallets: when we're
		// running as Rococo we only use `BridgeWococoMessages`/`BridgeWococoMessagesDispatch`, and vice versa.
```

HCastano (approved):
Looks good!

runtime/rococo/src/bridge_messages.rs
svyatonik: You can't simply ignore failing test, if it blocks CI

runtime/rococo/src/lib.rs
svyatonik: It requires a transaction spec version update. I personally don't care about this split

runtime/rococo/src/bridge_messages.rs
svyatonik: (unless you mean `#[ignore]`, of course)

HCastano: Yeah I meant `#[ignore]`. Should we make that more explicit in the comment?

runtime/rococo/src/lib.rs
HCastano: I think it would nice to have all the bridge stuff side by side, but I won't push too hard on this

tomusdrw: Yeah, I agree Hernando here. @shawntabrizi do you mind if we change the pallet index of `ValidatorManager` in the next restart/runtime upgrade of Rococo? I'd move it above the bridge stuff.

runtime/rococo/src/lib.rs
```diff
@@ -857,6 +867,111 @@ impl pallet_bridge_grandpa::Config<WococoGrandpaInstance> for Runtime {
 	type WeightInfo = pallet_bridge_grandpa::weights::RialtoWeight<Runtime>;
 }
 
+// Instance that is 'deployed' at Wococo chain. Responsible for dispatching Rococo -> Wococo messages.
+pub type AtWococoFromRococoMessagesDispatch = pallet_bridge_dispatch::DefaultInstance;
+impl pallet_bridge_dispatch::Config for Runtime {
```
tomusdrw, line 872:
```suggestion
pub type AtWococoFromRococoMessagesDispatch = pallet_bridge_dispatch::DefaultInstance;
impl pallet_bridge_dispatch::Config<AtWococoFromRococoMessageDispatch> for Runtime {
```

I think it's better to be explicit here, helps in the future in case we move things around.

runtime/rococo/src/lib.rs
```diff
@@ -857,6 +867,111 @@ impl pallet_bridge_grandpa::Config<WococoGrandpaInstance> for Runtime {
...
+		bp_rococo::MAX_UNREWARDED_RELAYER_ENTRIES_AT_INBOUND_LANE;
+	pub const MaxUnconfirmedMessagesAtInboundLane: bp_messages::MessageNonce =
+		bp_rococo::MAX_UNCONFIRMED_MESSAGES_AT_INBOUND_LANE;
+	pub const RootAccountForPayments: Option<AccountId> = None;
+}
+
+// Instance that is 'deployed' at Wococo chain. Responsible for sending Wococo -> Rococo messages
+// and receiving Rococo -> Wococo messages.
+pub type AtWococoWithRococoMessagesInstance = pallet_bridge_messages::DefaultInstance;
+impl pallet_bridge_messages::Config for Runtime {
```
tomusdrw, line 910:
```suggestion
pub type AtWococoWithRococoMessagesInstance = pallet_bridge_messages::DefaultInstance;
impl pallet_bridge_messages::Config<AtWococoWithRococoMessagesInstance> for Runtime {
```

runtime/rococo/src/lib.rs
```diff
@@ -1301,6 +1416,116 @@ sp_api::impl_runtime_apis! {
...
+				).ok()?;
+				Some(bp_messages::MessageDetails {
+					nonce,
+					dispatch_weight: decoded_payload.weight,
+					size: message_data.payload.len() as _,
+					delivery_and_dispatch_fee: message_data.fee,
+					dispatch_fee_payment: decoded_payload.dispatch_fee_payment,
+				})
+			})
+			.collect()
```
tomusdrw, line 1448: That's quite some logic that imho should be abstracted to primitives. Every runtime needs to copy this stuff exactly as-is. Can we add a function to `pallet_bridge_messages::Module`?
Something to address in a follow up PR in `bridges-common` though I think, cause it's not crticial.

runtime/rococo/src/bridge_messages.rs
```diff
@@ -0,0 +1,410 @@
...
+			size: message_payload_len
+				.saturating_add(EXTRA_STORAGE_PROOF_SIZE)
+				.saturating_add(TX_EXTRA_BYTES),
+		}
+	}
+
+	fn transaction_payment(transaction: MessageTransaction<Weight>) -> crate::Balance {
+		// current fee multiplier is used here
+		bridge_runtime_common::messages::transaction_payment(
+			crate::BlockWeights::get().get(frame_support::weights::DispatchClass::Normal).base_extrinsic,
```
tomusdrw, line 178: Should we enable a `CallFilter` for `DispatchClass::Normal` maybe?

tomusdrw (approved):
Awesome! Looking forward to see that deployed 👨‍🚀

runtime/rococo/src/bridge_messages.rs
```diff
@@ -0,0 +1,410 @@
...
+		type BridgedChain = WococoAtRococo;
+		type BridgedMessagesInstance = crate::AtWococoWithRococoMessagesInstance;
+
+		fn bridged_balance_to_this_balance(bridged_balance: bp_wococo::Balance) -> bp_rococo::Balance {
+			bridged_balance
+		}
+	}
+
+	/// Message payload for Rococo -> Wococo messages as it is seen at the Rococo.
+	pub type ToWococoMessagePayload = messages_source::FromThisChainMessagePayload<AtRococoWithWococoMessageBridge>;
```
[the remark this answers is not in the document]
svyatonik: Imo it'll make code more complicated, especially here, where we deal with single runtime=two chains. When you have explicit names, you can interpret configuration pretty easy.

runtime/rococo/src/bridge_messages.rs
svyatonik: Do you mean because here I'm getting base weight of `frame_support::weights::DispatchClass::Normal` tx? But this function is used to 'compute' the cost of delivery/confirmation transactions, which are `normal`. I can add any filter, though if you think we'll need it. I have no clear understanding of what we're going to achieve with this bridge, so I'm open for any suggestions about the set of filters/parameters/... in this PR.

runtime/rococo/src/lib.rs
svyatonik: logged https://github.com/paritytech/parity-bridges-common/issues/1046

runtime/rococo/src/bridge_messages.rs
tomusdrw: Ah, fair enough. I thought this should depend on the type of call we will dispatch across the bridge, but it doesn't matter - thanks for explanations. 
I don't think we need to add any filters currently.

tomusdrw:
bot merge

PR #3380, opened 2021-06-28
Introduce validation_code_and_hash and validation_code_hash in parachain host runtime API

thiolliere:
Those api will later be used by the client in this PR https://github.com/paritytech/polkadot/pull/2907


ordian (approved):

thiolliere:
better to use versioning here, so WIP

thiolliere:
closed, we will increase the runtime API for parachain host API and support both in the client

PR #3393, opened 2021-07-01
Better reward scheme for staking miner

kianenigma:
Still not happy with needing to hardcode "length assumptions". Ideally we should track how much fee is being consumed in the transaction and exactly pay that back later. 

Aside from not taking length into account, the previous model was also wrong to assume that `PerThing` can be larger than `1`.

### DIFF
--- a/runtime/common/src/elections.rs
+++ b/runtime/common/src/elections.rs
@@ -18,10 +18,14 @@
 
 use frame_support::{
 	parameter_types,
-	traits::Get,
-	weights::{DispatchClass, Weight, WeightToFeePolynomial},
+	weights::{DispatchClass, Weight},
 };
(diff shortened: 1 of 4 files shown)


kianenigma:
related: https://github.com/paritytech/substrate/issues/9247

coriolinus (approved):
Yeah, agree that it would be nice to avoid needing to hard-code the current weight, but also agree that this is an improvement on the status quo.

emostov (approved):
lgtm

just a little confused on the intention here vs #3141, which seems to introduce the exact same code.

kianenigma:
Sorry, yeah I added this to the miner PR as well. easier to merge it there.

PR #3403, opened 2021-07-03
paras.rs to FRAME V2

ferrell-code:
relates: #2882 

So users of the `Paras` pallet must be careful about the name they used in `construct_runtime!`. Hence the `runtime-migration label`, which might not be needed depending on the configuration of the `Paras` pallet.

Edit: Ok should work

### DIFF
--- a/runtime/parachains/src/paras.rs
+++ b/runtime/parachains/src/paras.rs
@@ -44,7 +44,7 @@ use sp_core::RuntimeDebug;
 #[cfg(feature = "std")]
 use serde::{Serialize, Deserialize};
 
-pub use crate::Origin;
+pub use crate::Origin as ParachainOrigin;
 
 pub use pallet::*;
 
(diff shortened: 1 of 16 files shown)


runtime/parachains/src/paras.rs
```diff
@@ -279,91 +262,45 @@ pub struct ParaGenesisArgs {
 	pub parachain: bool,
 }
 
-decl_storage! {
-	trait Store for Module<T: Config> as Paras {
```
shawntabrizi, line 283: Pallet must always be imported as `Paras` for this to be backwards compatible

shawntabrizi: Kusama, Westend, and Rococo look okay

KiChjang (approved):

runtime/parachains/src/paras.rs
KiChjang: A re-export should be fine here, i.e. `pub use pallet::Pallet as Paras`

shawntabrizi (approved):

shawntabrizi:
bot merge

PR #3404, opened 2021-07-04
Ensure MultiLocation always has a canonical representation

### LINKED ISSUE #3312: Update XCM MultiLocation
@gavofyork suggested we update the `MultiLocation` syntax to:

```
#[derive(Clone, Eq, PartialEq, Ord, PartialOrd, Encode, Decode, Debug)]
pub struct MultiLocation {
	parents: u8,
	junctions: Junctions,
}

#[derive(Clone, Eq, PartialEq, Ord, PartialOrd, Encode, Decode, Debug)]
pub enum Junctions {
	/// The interpreting consensus system.
	Null,
	/// A relative path comprising 1 junction.
	X1(Junction),
	/// A relative path comprising 2 junctions.
	X2(Junction, Junction),
	/// A relative path comprising 3 junctions.
	X3(Junction, Junction, Junction),
	/// A relative path comprising 4 junctions.
	X4(Junction, Junction, Junction, Junction),
	/// A relative path comprising 5 junctions.
	X5(Junction, Junction, Junction, Junction, Junction),
	/// A relative path comprising 6 junctions.
	X6(Junction, Junction, Junction, Junction, Junction, Junction),
	/// A relative path comprising 7 junctions.
	X7(Junction, Junction, Junction, Junction, Junction, Junction, Junction),
	/// A relative path comprising 8 junctions.
	X8(Junction, Junction, Junction, Junction, Junction, Junction, Junction, Junction),
}
```

And then we eliminate the `Parent` junction. This would force that all `MultiLocations` are always in their canonical/normalized form.

KiChjang:
Fixes #3312.

The API for `MultiLocation` is changed slightly as a result of eliminating the parent junction:

Firstly, all push operations have been split into a parent and non-parent counterpart.

Secondly, take operations that take from the front are denoted with `non_parent` to indicate that we do not modify the parent count.

Thirdly, the `Junctions` enum notably does not have any methods that can cause an overflow. `pushed_with` and `pushed_front_with` are left intact however, because they take ownership and there is no way to obtain ownership of a `Junctions` from a `MultiLocation`.


xcm/src/v1/junction.rs
```diff
@@ -0,0 +1,133 @@
...
+#[derive(Clone, Eq, PartialEq, Ord, PartialOrd, Encode, Decode, Debug)]
+pub enum NetworkId {
+	/// Unidentified/any.
+	Any,
+	/// Some named network.
+	Named(Vec<u8>),
+	/// The Polkadot Relay chain
+	Polkadot,
+	/// Kusama.
+	Kusama,
```
apopiak, line 32: `Any` is a filter, the rest (e.g. `Polkadot`) are identifiers.
See other comments for elaboration.

xcm/src/v1/multi_asset.rs
```diff
@@ -0,0 +1,380 @@
...
+
+	/// Some specific `instance` of the non-fungible asset whose `class` is identified abstractly.
+	AbstractNonFungible { class: Vec<u8>, instance: AssetInstance },
+
+	/// Some specific `amount` of the fungible asset identified by an concrete `id`.
+	ConcreteFungible { id: MultiLocation, #[codec(compact)] amount: u128 },
+
+	/// Some specific `instance` of the non-fungible asset whose `class` is identified concretely.
+	ConcreteNonFungible { class: MultiLocation, instance: AssetInstance },
+}
```
apopiak, line 146: The `All*` variants are filters, the other variants (e.g. `ConcreteFungible`) are (something like) identifiers.
See other comments for elaboration.

apopiak (commented):
Not sure this PR is the right context, but before we're introducing a new XCM version I would like to raise an issue I see with v0:
I see a difference between `NetworkId` and `MultiAsset` on the one hand and other enums/"identifiers" like `MultiLocation` or `BodyPart`. The latter separate the identifying and filtering parts. Identifying is done by the enum instance (e.g. a multilocation might be `X2(Parachain(1), GeneralIndex(42))`) while filtering is done in the XCM implementation (e.g.  the `IsTeleporter` filter). This IMO is a good pattern.
`NetworkId` and `MultiAsset` combine an identifying function with a filter function within the enum which leads to non-filtering code having to manage the case when we get `MultiAsset::AllFungible` even though that is not its purpose. I think we should avoid this in v1 and separate filtering and identifying completely.

Opened https://github.com/paritytech/polkadot/issues/3409 if you don't want to discuss this here.

KiChjang:
@apopiak Yeah I agree, this PR is probably not the best place to discuss things that are not related to the `MultiLocation`. A new XCM version is created simply because it is easier to accommodate changes to the `MultiLocation` types. But I do think that some changes ought to be bundled together if we were to introduce a new version.

shawntabrizi:
not having looked at this PR at all, we should make the evaluation whether maintaining and supporting 2 xcm versions at this point is better than introducing a breaking change. It feels to me, that 2 versions of XCM at this point will be more technical debt than it is worth

xcm/src/v1/junction.rs
```diff
@@ -0,0 +1,133 @@
...
+///
+/// Each item assumes a pre-existing location as its context and is defined in terms of it.
+#[derive(Clone, Eq, PartialEq, Ord, PartialOrd, Encode, Decode, Debug)]
+pub enum Junction {
+	/// An indexed parachain belonging to and operated by the context.
+	///
+	/// Generally used when the context is a Polkadot Relay-chain.
+	Parachain(#[codec(compact)] u32),
+	/// A 32-byte identifier for an account of a specific network that is respected as a sovereign endpoint within
+	/// the context.
```
shawntabrizi, line 94: Could we not just re-export `v0::Junction` from here?

apopiak: But the `Parent` variant is no longer there

xcm/src/v1/mod.rs
```diff
@@ -0,0 +1,317 @@
...
+	/// Withdraw asset(s) (`assets`) from the ownership of `origin` and place them into `holding`. Execute the
+	/// orders (`effects`).
+	///
+	/// - `assets`: The asset(s) to be withdrawn into holding.
+	/// - `effects`: The order(s) to execute on the holding account.
+	///
+	/// Kind: *Instruction*.
+	///
+	/// Errors:
+	#[codec(index = 0)]
```
gavofyork, line 101:
```suggestion
	#[codec(index = 1)]
```

xcm/src/v1/junction.rs
```diff
@@ -0,0 +1,133 @@
...
+
+/// A single item in a path to describe the relative location of a consensus system.
+///
+/// Each item assumes a pre-existing location as its context and is defined in terms of it.
+#[derive(Clone, Eq, PartialEq, Ord, PartialOrd, Encode, Decode, Debug)]
+pub enum Junction {
+	/// An indexed parachain belonging to and operated by the context.
+	///
+	/// Generally used when the context is a Polkadot Relay-chain.
+	Parachain(#[codec(compact)] u32),
```
gavofyork, line 92: `#[codec(index = 1)]`

gavofyork:
Good stuff, but I'd prefer not to need to create XCM v1 on the back of this, when we can very easily just have a custom encoding:

```
/// A single item in a path to describe the relative location of a consensus system.
///
/// Each item assumes a pre-existing location as its context and is defined in terms of it.
///
/// Note for this to encode correctly within v0 XCM, the `parents` value plus the `junctions.len)` must be no greater
/// than 255.
#[derive(Clone, Eq, PartialEq, Ord, PartialOrd, Debug)]
pub struct MultiLocation {
	/// The number of ancestor levels to ascend to find the exterior context. If zero, then it's not an exterior
	/// location.
	pub parents: u8,
	/// The interior components of the exterior context.
	pub interior: Junctions,
}

impl Encode for MultiLocation {
	fn encode_to<T: Output + ?Sized>(&self, output: &mut T) {
		// 255 items max, but we cannot return an error, so we ap silently.
		let parents = self.parents.min(255.saturating_sub(self.junctions.len()) as u8);
		let count = parents.saturating_add(junctions.len());
		output.push_byte(count);
		for _ in 0..parents {
			output.push_byte(0);
		}
		for j in self.interior.iter() {
			j.encode_to(output)
		}
	}
}

impl Decode for MultiLocation {
	fn decode<I: Input>(input: &mut I) -> Result<Self, parity_scale_codec::Error> {
		let count = input.read_byte()?;
		let mut parents = 0u8;
		let mut interior = Junctions::Null;
		for i in 0..count {
			let b = output.read_byte()?;
			if b == 0 {
				if i > s.parents {
					// this would imply that there has been at least one non-parent junction, which further implies
					// that this is non-canonical. we bail.
					return Err(parity_scale_codec::Error::from("Parent after non-Parent forbidden"))
				}
				parents += 1;
			} else {
				interior.push(Junction::decode(&mut &[b])?)
					.map_err(|_| parity_scale_codec::Error::from("Interior too long"))?;
			}
		}
		Ok(MultiLocation { parents, interior })
	}
}
```

This custom `Encode`/`Decode` together with a `#[codec::index = 1]` on the first item in `Junction` should obviate the need to bump the XCM version.

xcm/src/lib.rs
```diff
@@ -38,6 +39,7 @@ pub use double_encoded::DoubleEncoded;
 #[codec(decode_bound())]
 pub enum VersionedXcm<Call> {
 	V0(v0::Xcm<Call>),
+	V1(v1::Xcm<Call>),
```
gavofyork, line 42: No need for this yet if we have a custom encoder for `MultiLocation`.

xcm/src/v1/multi_location.rs
```diff
@@ -0,0 +1,867 @@
...
+/// called a *sub-consensus*, or going up into its parent location. Correct `MultiLocation` values must have all
+/// `Parent` junctions as a prefix to all *sub-consensus* junctions.
+///
+/// This specific `MultiLocation` implementation uses a Rust `enum` in order to make pattern matching easier.
+///
+/// The `MultiLocation` value of `Null` simply refers to the interpreting consensus system.
+#[derive(Clone, Eq, PartialEq, Ord, PartialOrd, Encode, Decode, Debug)]
+pub struct MultiLocation {
+	parents: u8,
+	junctions: Junctions,
```
gavofyork, line 50:
```suggestion
	interior: Junctions,
```

Best to name fields over what they do.

gavofyork (commented):
Minor rework, but overall good.

xcm/src/v1/multi_location.rs
```diff
@@ -0,0 +1,867 @@
...
+		for j in prefix.junctions.into_iter_rev() {
+			self.push_front_non_parent(j).expect(
+				"self junctions len + prefix parent count + prepend len is less than max length; qed"
+			);
+		}
+		Ok(())
+	}
+}
+
+impl From<Junctions> for MultiLocation {
```
gavofyork, line 317: There should alss be an `impl TryFrom<MultiLocation> for Junctions` which fails if `parents > 0`.

KiChjang:
@gavofyork I have a slight issue about making the `MultiLocation` struct fields public -- this allows the user to create an inconsistent state, e.g. there is no way of stopping a user from assigning the `parents` field with `255` and then assigning a `Junctions::X8` to the `interior` field.

I'd like to make it completely impossible for the user to ever create an inconsistent state. Making the fields private would come at the cost of not being able to destructure `MultiLocation` in a match expression, but I think it is worth making the guarantee. If users want to edit the fields, they will have to do so via the APIs that we expose on `MultiLocation`.

The other small issue I have is `MAX_MULTILOCATION_LENGTH`. We currently have it set at 8, yet in the code snippet you provided, it seems to me that you'd like it to be changed to 255? Was that your intention?

xcm/src/v0/multi_location.rs
```diff
@@ -44,155 +44,528 @@ use crate::VersionedMultiLocation;
...
+
+	/// Returns the junction at index `i`, or `None` if the location is a parent or if the location
+	/// does not contain that many elements.
+	pub fn at(&self, i: usize) -> Option<&Junction> {
+		let num_parents = self.parents as usize;
+		if i < num_parents {
+			return None
+		}
+		self.interior.at(i - num_parents)
+	}
```
apopiak, line 237: mmh, not sure I'm a fan of representing the Parent as `None`

xcm/src/v0/multi_location.rs
```diff
@@ -44,155 +44,528 @@ use crate::VersionedMultiLocation;
...
-	X7(Junction, Junction, Junction, Junction, Junction, Junction, Junction),
-	/// A relative path comprising 8 junctions.
-	X8(Junction, Junction, Junction, Junction, Junction, Junction, Junction, Junction),
+#[derive(Clone, Eq, PartialEq, Ord, PartialOrd, Debug)]
+pub struct MultiLocation {
+	parents: u8,
+	interior: Junctions,
+}
+
+impl Encode for MultiLocation {
```
apopiak, line 53: let's add a test asserting that the encoding doesn't change

xcm/src/v0/multi_location.rs
```diff
@@ -44,155 +44,528 @@ use crate::VersionedMultiLocation;
...
+		&self.interior
+	}
+
+	/// Return a mutable reference to the interior field.
+	pub fn junctions_mut(&mut self) -> &mut Junctions {
+		&mut self.interior
+	}
+
+	/// Returns the number of `Parent` junctions at the beginning of `self`.
+	pub fn parent_count(&self) -> usize {
```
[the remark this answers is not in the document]
KiChjang: Yeah, pretty much. I'm imagining the use case for this is similar to `len()`, so I figured that it should return a `usize`.

runtime/rococo/src/lib.rs
```diff
@@ -627,16 +627,16 @@ pub type XcmRouter = (
 	xcm_sender::ChildParachainRouter<Runtime>,
 );
 
-use xcm::v0::{MultiAsset, MultiAsset::AllConcreteFungible, MultiLocation::{Null, X1}, Junction::Parachain};
+use xcm::v0::{MultiAsset, MultiAsset::AllConcreteFungible, Junctions::{Null, X1}, Junction::Parachain};
 parameter_types! {
 	pub const RococoForTick: (MultiAsset, MultiLocation) =
-		(AllConcreteFungible { id: Null }, X1(Parachain(100)));
+		(AllConcreteFungible { id: MultiLocation::empty() }, MultiLocation::with_parachain_interior(100));
```
apopiak, line 633:
```suggestion
		(AllConcreteFungible { id: MultiLocation::empty() }, Parachain(100).into());
```
Should also work, right?

xcm/src/v0/multi_location.rs
```diff
@@ -115,6 +112,40 @@ impl MultiLocation {
...
+		}
+		Ok(MultiLocation {
+			parents,
+			interior: Junctions::Null,
+		})
+	}
+
+	/// Creates a new `MultiLocation` with no parents and a single `Parachain` interior junction
+	/// specified by `para_id`.
+	pub const fn with_parachain_interior(para_id: u32) -> MultiLocation {
```
apopiak, line 137:
```suggestion
	pub const fn from_parachain(para_id: u32) -> MultiLocation {
```
would that also work?

runtime/rococo/src/lib.rs
KiChjang: Yes, however the proper syntax is:

```suggestion
		(AllConcreteFungible { id: MultiLocation::empty() }, X1(Parachain(100)).into());
```

I actually opted to use `with_parachain_interior` because it's clearer on what the code is doing.

xcm/src/v0/multi_location.rs
```diff
@@ -244,7 +244,7 @@ impl MultiLocation {
 
 	/// Consumes `self` and returns a `MultiLocation` suffixed with `new`, or an `Err` with the original value of
 	/// `self` in case of overflow.
-	pub fn pushed_with_non_parent(self, new: Junction) -> result::Result<Self, Self> {
+	pub fn pushed_with_interior(self, new: Junction) -> result::Result<Self, Self> {
```
apopiak, line 257:
```suggestion
	pub fn pushed_interior(self, new: Junction) -> result::Result<Self, Self> {
```

xcm/src/v0/multi_location.rs
```diff
@@ -44,77 +44,424 @@ use parity_scale_codec::{self, Decode, Encode};
...
+	}
+
+	/// Creates a new `MultiLocation` with 0 parents and a `Null` interior.
+	pub const fn empty() -> MultiLocation {
+		MultiLocation { parents: 0, interior: Junctions::Null }
+	}
+
+	/// Creates a new `MultiLocation` with the specified number of parents in the `P` const generic
+	/// parameter and a `Null` interior.
+	pub const fn with_parents_const<const P: u8>() -> MultiLocation {
```
KiChjang, line 130: It is absolutely nuts that this function actually works. I am wondering whether I should call this `with_parents` and remove the existing implementation of that method altogether...

runtime/kusama/src/lib.rs
```diff
@@ -1267,7 +1267,7 @@ pub type XcmRouter = (
 
 parameter_types! {
 	pub const KusamaForStatemint: (MultiAsset, MultiLocation) =
-		(AllConcreteFungible { id: Null }, X1(Parachain(1000)));
+		(AllConcreteFungible { id: MultiLocation::empty() }, MultiLocation::with_parachain_interior(1000));
```
kianenigma, line 1270: I find this to be confusing. `with_parachain_interior` means 1 child, which is a parachain, with this id? what would be the raw equivalent of this?

xcm/pallet-xcm/src/lib.rs
```diff
@@ -344,9 +345,9 @@ where
 pub struct XcmPassthrough<Origin>(PhantomData<Origin>);
 impl<Origin: From<crate::Origin>> ConvertOrigin<Origin> for XcmPassthrough<Origin> {
 	fn convert_origin(origin: MultiLocation, kind: OriginKind) -> Result<Origin, MultiLocation> {
-		match (kind, origin) {
-			(OriginKind::Xcm, l) => Ok(crate::Origin::Xcm(l).into()),
-			(_, origin) => Err(origin),
+		match kind {
```
kianenigma, line 348: matching `origin` here was superfluous, good catch.

xcm/src/v0/multi_location.rs
```diff
@@ -44,77 +44,423 @@ use parity_scale_codec::{self, Decode, Encode};
 /// This specific `MultiLocation` implementation uses a Rust `enum` in order to make pattern matching easier.
```
kianenigma, line 44: not anymore -- doc needs update.

xcm/src/v0/multi_location.rs
```diff
@@ -44,77 +44,423 @@ use parity_scale_codec::{self, Decode, Encode};
...
 ///
 /// The `MultiLocation` value of `Null` simply refers to the interpreting consensus system.
```
kianenigma, line 46: ditto.

runtime/kusama/src/lib.rs
KiChjang: It will just be this:
```rust
MultiLocation {
    parents: 0,
    interior: X1(Parachain(1000)),
}
```

Since I've privatized the fields on `MultiLocation`, it isn't possible to create such a struct outside of the `xcm` crate. Moreover, we're currently in a const eval context, which means we can only call const functions and thereby leading me to create a new `with_parachain_interior` const function on `MultiLocation` to allow construction of this kind of `MultiLocation`.

xcm/src/v0/multi_location.rs
```diff
@@ -44,77 +44,423 @@ use parity_scale_codec::{self, Decode, Encode};
...
-	X5(Junction, Junction, Junction, Junction, Junction),
-	/// A relative path comprising 6 junctions.
-	X6(Junction, Junction, Junction, Junction, Junction, Junction),
-	/// A relative path comprising 7 junctions.
-	X7(Junction, Junction, Junction, Junction, Junction, Junction, Junction),
-	/// A relative path comprising 8 junctions.
-	X8(Junction, Junction, Junction, Junction, Junction, Junction, Junction, Junction),
+#[derive(Clone, Eq, PartialEq, Ord, PartialOrd, Debug)]
+pub struct MultiLocation {
+	parents: u8,
```
kianenigma, line 49: not quite clear to me (yet) how parents should be interpreted, maybe add some docs for it?

xcm/src/v0/multi_location.rs
```diff
@@ -44,77 +44,423 @@ use parity_scale_codec::{self, Decode, Encode};
...
+	/// Creates a new `MultiLocation`, ensuring that the length of it does not exceed the maximum,
+	/// otherwise returns `Err`.
+	pub fn new(parents: u8, junctions: Junctions) -> result::Result<MultiLocation, ()> {
+		if parents as usize + junctions.len() > MAX_MULTILOCATION_LENGTH {
+			return Err(())
+		}
+		Ok(MultiLocation { parents, interior: junctions })
+	}
+
+	/// Creates a new `MultiLocation` with 0 parents and a `Null` interior.
```
kianenigma, line 123:
```suggestion
	/// Creates a new `MultiLocation` with 0 parents and a `Null` interior.
	///
	/// This can be interpreted as the 'current consensus system'. 
	
```

xcm/src/v0/multi_location.rs
kianenigma: will rustc monomorphizes const generics as well? in that case for each use we get a copy of this function lying around, potentially bloating the wasm code. 

And why does this need to be const in the first place?

KiChjang: We can't accept a `Junctions` as a parameter in a const eval context, which is why you'll see no const constructors here that do. The reasoning is because `Junctions` contains a variant with a `Vec<u8>`, which is a type that will run a destructor when it goes out of scope, and running destructors are prohibited in a const eval context.

Not too sure about monomorphization, but I can find that one out.

xcm/src/v0/multi_location.rs
```diff
@@ -44,77 +44,423 @@ use parity_scale_codec::{self, Decode, Encode};
...
+	/// Returns an error if `parents` is greater than `MAX_MULTILOCATION_LENGTH`.
+	pub const fn with_parents(parents: u8) -> result::Result<MultiLocation, ()> {
+		if parents as usize > MAX_MULTILOCATION_LENGTH {
+			return Err(())
+		}
+		Ok(MultiLocation { parents, interior: Junctions::Null })
+	}
+
+	/// Creates a new `MultiLocation` with no parents and a single `Parachain` interior junction
+	/// specified by `para_id`.
```
kianenigma, line 146:
```suggestion
	/// specified by `para_id`.
	///
	/// This can be itnerpreted as the child-parachain of the current consensus system. 
```

xcm/src/v0/multi_location.rs
kianenigma: (reading the docs of `parent_count` clarified it, but maybe write a line here as well)

xcm/src/v0/multi_location.rs
KiChjang: Looks like it does monomorphize const generics: https://github.com/rust-lang/rfcs/blob/master/text/2000-const-generics.md

xcm/src/v0/multi_location.rs
```diff
@@ -44,77 +44,423 @@ use parity_scale_codec::{self, Decode, Encode};
...
+	}
+
+	/// Returns the number of `Parent` junctions at the beginning of `self`.
+	pub const fn parent_count(&self) -> u8 {
+		self.parents
+	}
+
+	/// Returns the number of parents and junctions in `self`.
+	pub const fn len(&self) -> usize {
+		self.parent_count() as usize + self.interior.len()
```
kianenigma, line 173: good place to havea debug_assertion that the length is less than the limit.

xcm/src/v0/multi_location.rs
```diff
@@ -44,77 +44,423 @@ use parity_scale_codec::{self, Decode, Encode};
...
+	}
+
+	/// Returns the first interior junction, or `None` if the location is empty or contains only
+	/// parents.
+	pub fn first_interior(&self) -> Option<&Junction> {
```
kianenigma, line 178: can't you just do `foo.iterior().first()`?

KiChjang: Creating a new iterator just to grab the first element seems a bit inefficient to me; `self.interior.first()` is more direct and creates less stuff on the stack.

xcm/src/v0/multi_location.rs
KiChjang: In practical terms however, I would expect people to only call this with numbers ranging from 1 to 4 to denote their distance from the relay chain. I would not expect to see them iterating from 0 to 255 just to test all possible `u8` digits to fit into this function.

kianenigma: > In practical terms however, I would expect people to only call this with numbers ranging from 1 to 4 to denote their distance from the relay chain. I would not expect to see them iterating from 0 to 255 just to test all possible `u8` digits to fit into this function.

This sounds reasonable, but I didn't receive my answer in the first place that when do we need `with_parents_const` itself?

runtime/kusama/src/lib.rs
kianenigma: This whole const eval context is self-imposed, you don't need to make `parameter_types` const. Just remove the keyword `const` from there and it will not require it to be const anymore.

xcm/src/v0/multi_location.rs
kianenigma: I think you can remove this function overall https://github.com/paritytech/polkadot/pull/3404#discussion_r684411842

xcm/src/v0/multi_asset.rs
```diff
@@ -290,53 +294,64 @@ impl MultiAsset {
 	}
 }
 
-impl From<MultiAsset1> for MultiAsset {
-	fn from(a: MultiAsset1) -> MultiAsset {
+impl TryFrom<MultiAsset1> for MultiAsset {
+	type Error = ();
+
+	fn try_from(a: MultiAsset1) -> result::Result<MultiAsset, ()> {
```
kianenigma, line 300: I think `m` for `multi-asset` or a more verbose name is better than `a`.

xcm/src/v0/multi_asset.rs
```diff
@@ -290,53 +294,64 @@ impl MultiAsset {
 	}
 }
 
-impl From<MultiAsset1> for MultiAsset {
-	fn from(a: MultiAsset1) -> MultiAsset {
+impl TryFrom<MultiAsset1> for MultiAsset {
```
kianenigma, line 297: opinionated suggestion: `v1::Foo` cleaner than needing a manual import and `Foo1`.

xcm/src/v1/multilocation.rs
```diff
@@ -36,83 +39,385 @@ use parity_scale_codec::{self, Decode, Encode};
...
+	/// Creates a new `MultiLocation` with 0 parents and a `Null` interior.
+	///
+	/// The resulting `MultiLocation` can be interpreted as the "current consensus system".
+	pub const fn here() -> MultiLocation {
+		MultiLocation { parents: 0, interior: Junctions::Null }
+	}
+
+	/// Creates a new `MultiLocation` with the specified number of parents in the `P` const generic
+	/// parameter and a `Null` interior.
+	// #FIXME: Use a where clause to evaluate P for well-formedness once const eval of const
```
kianenigma, line 84: needs a linked issue.

kianenigma: also, what do you mean well-formed-ness? given max-length being 255, `P` is always correct. But your comments imply that otherwise this function has no checks. 

While it is an interesting function, I still don't see a need for it.

xcm/src/v0/multi_location.rs
kianenigma: sorry what I mean was `foo.interior().first()`

xcm/src/v1/multilocation.rs
```diff
@@ -36,83 +39,385 @@ use parity_scale_codec::{self, Decode, Encode};
...
+		let (suffix, first) = junctions.split_first();
+		let multilocation = MultiLocation { parents, interior: suffix };
+		(multilocation, first)
+	}
+
+	/// Splits off the last junction, returning the remaining prefix (first item in tuple) and the last element
+	/// (second item in tuple) or `None` if it was empty or if `self` only contains parents.
+	pub fn split_last(self) -> (MultiLocation, Option<Junction>) {
+		let MultiLocation { parents, interior: junctions } = self;
+		let (prefix, last) = junctions.split_last();
```
kianenigma, line 157: If I am `{ parents: 3, interior: vec!{} }`, I'd expect this to chunk my last parent out, but it does nothing. is it intended?

xcm/src/v1/multilocation.rs
```diff
@@ -36,83 +39,385 @@ use parity_scale_codec::{self, Decode, Encode};
...
+		let multilocation = MultiLocation { parents, interior: prefix };
+		(multilocation, last)
+	}
+
+	/// Bumps the parent count up by 1. Returns `Err` in case of overflow.
+	pub fn push_parent(&mut self) -> result::Result<(), ()> {
+		if self.len() >= MAX_MULTILOCATION_LENGTH {
+			return Err(())
+		}
+		self.parents = self.parents.saturating_add(1);
```
kianenigma, line 167:
```suggestion
		self.parents = self.parents.saturating_inc;
```

xcm/src/v1/multilocation.rs
```diff
@@ -36,83 +39,385 @@ use parity_scale_codec::{self, Decode, Encode};
...
+		Ok(())
+	}
+
+	/// Mutates `self`, suffixing its interior junctions with `new`. Returns `Err` in case of overflow.
+	pub fn push_interior(&mut self, new: Junction) -> result::Result<(), ()> {
+		let mut n = Junctions::Null;
+		mem::swap(&mut self.interior, &mut n);
```
kianenigma, line 174: that's an interesting piece of code, why don't you directly do `self.interior.pushed_with`, but rather swap and then assign to `self.interior? `

xcm/src/v1/multilocation.rs
KiChjang: This function is used in a const eval context, and I can't make the existing `with_parents` work in const, because it returns a `Result`, and unwrapping the result in a const eval context is forbidden.

Since `MAX_MULTILOCATION_LENGTH` is configurable, it's not always going to be `255`, which is why we need a where clause to ensure that P is always less than `MAX_MULTILOCATION_LENGTH`.

xcm/src/v1/multilocation.rs
kianenigma: okay I see why, `push_with` oconsumes self.

xcm/src/v1/multilocation.rs
KiChjang: Huh, yeah, I should probably rename this as `split_last_interior` to make the intention clearer.

xcm/src/v1/multilocation.rs
```diff
@@ -36,83 +39,385 @@ use parity_scale_codec::{self, Decode, Encode};
...
+		let num_parents = self.parents as usize;
+		if i < num_parents {
+			return None
+		}
+		self.interior.at_mut(i - num_parents)
+	}
+
+	/// Decrements the parent count by 1.
+	pub fn pop_parent(&mut self) {
+		self.parents = self.parents.saturating_sub(1);
```
kianenigma, line 261:
```suggestion
		self.parents = self.parents.saturating_dec();
```

xcm/src/v1/multilocation.rs
shawntabrizi: add a `const_assert!(MAX_MULTILOCATION_LENGTH == u8::MAX)`

And if that fails, we know this function is potentially unsafe. Some comments around the failing test can hint to that.

xcm/src/v1/multilocation.rs
```diff
@@ -732,90 +1121,45 @@ mod tests {
...
+		let mut m = MultiLocation {
+			parents: 1,
+			interior: X2(Parachain(42), AccountIndex64 { network: Any, index: 23 }),
+		};
+		assert_eq!(m.prepend_with(MultiLocation { parents: 1, interior: X1(OnlyChild) }), Ok(()));
+		assert_eq!(
+			m,
+			MultiLocation {
+				parents: 1,
+				interior: X2(Parachain(42), AccountIndex64 { network: Any, index: 23 })
```
kianenigma, line 1152: is this a good example? `m` basically stays unchanged.

xcm/src/v1/multilocation.rs
KiChjang: It was how `push` was implemented in the old version's `MultiLocation`, and I didn't bother changing its details.

xcm/xcm-builder/src/fungibles_adapter.rs
```diff
@@ -38,18 +38,24 @@ impl<Prefix: Get<MultiLocation>, AssetId: Clone, ConvertAssetId: Convert<u128, A
...
 		let prefix = Prefix::get();
 		let id = id.borrow();
-		if !prefix.iter().enumerate().all(|(index, item)| id.at(index) == Some(item)) {
+		if prefix.parent_count() != id.parent_count() ||
+			prefix
+				.interior()
+				.iter()
+				.enumerate()
+				.any(|(index, junction)| id.interior().at(index) != Some(junction))
+		{
```
kianenigma, line 47: isn't this block basically the definition of `Eq` or `PartialEq` for `MultiLocation`?

xcm/src/v1/multilocation.rs
```diff
@@ -36,83 +39,385 @@ use parity_scale_codec::{self, Decode, Encode};
...
+			Err(old) => {
+				self.interior = old;
+				Err(())
+			},
+		}
+	}
+
+	/// Consumes `self` and returns a `MultiLocation` with its parent count incremented by 1, or
+	/// an `Err` with the original value of `self` in case of overflow.
+	pub fn pushed_with_parent(self) -> result::Result<Self, Self> {
```
shawntabrizi, line 205: Should accept `p: u8` as input, and allow us to push any number of parents

shawntabrizi: with a check

shawntabrizi: with the right checks

runtime/common/src/xcm_sender.rs
```diff
@@ -28,8 +28,8 @@ impl<T: configuration::Config + dmp::Config, W: xcm::WrapVersion> SendXcm
 	for ChildParachainRouter<T, W>
 {
 	fn send_xcm(dest: MultiLocation, msg: Xcm) -> Result {
-		match &dest {
-			MultiLocation::X1(Junction::Parachain(id)) => {
+		match dest.interior() {
+			Junctions::X1(Junction::Parachain(id)) if dest.parent_count() == 0 => {
```
shawntabrizi, line 32: since this is a router for relay chain, we should start the whole function with `if dest.parent_count() > 0 { return Error}`

shawntabrizi: where the check is right now is kind of hidden and may lead to errors in the future

runtime/kusama/src/lib.rs
```diff
@@ -1263,7 +1263,7 @@ pub type XcmRouter = (
 
 parameter_types! {
 	pub const Kusama: MultiAssetFilter = Wild(AllOf { fun: WildFungible, id: Concrete(KsmLocation::get()) });
-	pub const KusamaForStatemint: (MultiAssetFilter, MultiLocation) = (Kusama::get(), X1(Parachain(1000)));
+	pub const KusamaForStatemint: (MultiAssetFilter, MultiLocation) = (Kusama::get(), MultiLocation::with_parachain_interior(1000));
```
shawntabrizi, line 1266: cant you implement `.into()` to make these kinds of transitions easier?

xcm/pallet-xcm/src/mock.rs
```diff
@@ -75,7 +75,7 @@ impl SendXcm for TestSendXcm {
 pub struct TestSendXcmErrX8;
 impl SendXcm for TestSendXcmErrX8 {
 	fn send_xcm(dest: MultiLocation, msg: Xcm) -> XcmResult {
-		if let MultiLocation::X8(..) = dest {
+		if dest.len() >= 8 {
```
shawntabrizi, line 78:
```suggestion
		if dest.len() == 8 {
```

I think this is actually the proper change

xcm/src/v1/multilocation.rs
KiChjang: Can't, `saturating_dec` is only defined in `frame_support`, and we're not going to import that here in `xcm`.

xcm/src/v1/multilocation.rs
KiChjang: This test was directly ported from the old version, and looking at it closely, it is a good example, precisely because we want to ensure that this prepend method didn't change `m` at all.

xcm/xcm-builder/src/fungibles_adapter.rs
KiChjang: Huh, yeah, I wonder why `Eq` wasn't used before in this block...?

KiChjang: Ah ok, I remember why -- `id` and `prefix` may have different lengths. In case `prefix` is shorter, then we want to make sure that `prefix` is a proper prefix of `id`.

xcm/src/v1/multilocation.rs
```diff
@@ -36,83 +39,385 @@ use parity_scale_codec::{self, Decode, Encode};
...
+	///
+	/// Does not modify `self` and returns `Err` with `prefix` in case of overflow.
+	///
+	/// # Example
+	/// ```rust
+	/// # use xcm::v1::{Junctions::*, Junction::*, MultiLocation};
+	/// # fn main() {
+	/// let mut m = MultiLocation::new(2, X1(PalletInstance(3))).unwrap();
+	/// assert_eq!(m.prepend_with(MultiLocation::new(1, X2(Parachain(21), OnlyChild)).unwrap()), Ok(()));
+	/// assert_eq!(m, MultiLocation::new(1, X1(PalletInstance(3))).unwrap());
```
kianenigma, line 329: Can you elaborate this example? This is not what I intuitively would imagine *prefixing* looks like.

xcm/src/v1/multilocation.rs
```diff
@@ -393,79 +740,74 @@ impl MultiLocation {
...
-	pub fn into_iter(self) -> MultiLocationIterator {
-		MultiLocationIterator(self)
+	pub fn iter_rev(&self) -> JunctionsReverseRefIterator {
+		JunctionsReverseRefIterator(&self, 0)
 	}
 
 	/// Consumes `self` and returns an iterator over the junctions in reverse.
-	pub fn into_iter_rev(self) -> MultiLocationReverseIterator {
-		MultiLocationReverseIterator(self)
+	pub fn into_iter_rev(self) -> JunctionsReverseIterator {
```
kianenigma, line 787: again, why not do a normal iterator and just call `.rev()`?

kianenigma: I can imagine that maybe this is more efficient, since `iter().rev()` will need to traverse the normal iterator to obtain the reverse.

xcm/src/v1/multilocation.rs
```diff
@@ -476,254 +818,301 @@ impl MultiLocation {
...
+				j0.try_into()?,
+				j1.try_into()?,
+				j2.try_into()?,
+				j3.try_into()?,
+				j4.try_into()?,
+				j5.try_into()?,
+				j6.try_into()?,
+				j7.try_into()?,
+			)
+			.into()),
```
kianenigma, line 1077: I read all of this impl to double check, I think a macro to generate the match is easier to verify, as a future enhancement.

xcm/xcm-builder/src/location_conversion.rs
```diff
@@ -106,12 +109,14 @@ impl<Network: Get<NetworkId>, AccountId: From<[u8; 32]> + Into<[u8; 32]> + Clone
 	Convert<MultiLocation, AccountId> for AccountId32Aliases<Network, AccountId>
 {
 	fn convert(location: MultiLocation) -> Result<AccountId, MultiLocation> {
-		let id = match location {
-			MultiLocation::X1(Junction::AccountId32 { id, network: NetworkId::Any }) => id,
-			MultiLocation::X1(Junction::AccountId32 { id, network })
-				if &network == &Network::get() =>
-				id,
-			l => return Err(l),
+		let id = match location.interior() {
```
kianenigma, line 112:
```suggestion
		let id = match (localtyion.parent_count(), location.interior()) {
```

All logic is correct to me, but maybe this is a nicer syntax

xcm/xcm-builder/src/location_conversion.rs
```diff
@@ -126,12 +131,14 @@ impl<Network: Get<NetworkId>, AccountId: From<[u8; 20]> + Into<[u8; 20]> + Clone
 	Convert<MultiLocation, AccountId> for AccountKey20Aliases<Network, AccountId>
 {
 	fn convert(location: MultiLocation) -> Result<AccountId, MultiLocation> {
-		let key = match location {
-			MultiLocation::X1(Junction::AccountKey20 { key, network: NetworkId::Any }) => key,
-			MultiLocation::X1(Junction::AccountKey20 { key, network })
-				if &network == &Network::get() =>
-				key,
-			l => return Err(l),
+		let key = match location.interior() {
```
kianenigma, line 134: ditto

xcm/xcm-builder/src/origin_conversion.rs
```diff
@@ -45,9 +45,13 @@ where
 pub struct ParentAsSuperuser<Origin>(PhantomData<Origin>);
 impl<Origin: OriginTrait> ConvertOrigin<Origin> for ParentAsSuperuser<Origin> {
 	fn convert_origin(origin: MultiLocation, kind: OriginKind) -> Result<Origin, MultiLocation> {
-		match (kind, origin) {
-			(OriginKind::Superuser, MultiLocation::X1(Junction::Parent)) => Ok(Origin::root()),
-			(_, origin) => Err(origin),
+		if kind == OriginKind::Superuser &&
+			origin.parent_count() == 1 &&
```
kianenigma, line 49: another suggestion to clean matching on the new multilocation

```
fn is_only_parent(&self, count: usize) -> bool { self.interior.len() == 0 && self.parent_count == count }   
```

xcm/xcm-builder/src/origin_conversion.rs
```diff
@@ -170,7 +178,7 @@ where
 		// We institute a root fallback so root can always represent the context. This
 		// guarantees that `successful_origin` will work.
 		if o.caller() == Origin::root().caller() {
-			Ok(MultiLocation::Here)
+			Ok(MultiLocation::here())
 		} else {
 			Err(o)
 		}
```
kianenigma, line 184: everything in this file is correct, but syntactically I think we can simplify it a bit and remove some of the conditions.

xcm/xcm-executor/src/lib.rs
```diff
@@ -234,7 +234,6 @@ impl<Config: config::Config> XcmExecutor<Config> {
 				None
 			},
 			(origin, Xcm::RelayedFrom { who, message }) => {
-				ensure!(who.is_interior(), XcmError::EscalationOfPrivilege);
```
kianenigma, line 237: explain?

xcm/xcm-simulator/example/src/parachain.rs
```diff
@@ -101,7 +101,7 @@ parameter_types! {
 }
 
 parameter_types! {
-	pub const KsmLocation: MultiLocation = MultiLocation::X1(Parent);
+	pub const KsmLocation: MultiLocation = MultiLocation::with_parents_const::<1>();
```
kianenigma:
```suggestion
	pub KsmLocation: MultiLocation = MultiLocation::with_parents(1);
```

runtime/kusama/src/lib.rs
gavofyork: Agree. `Parachain(1000).into()` is much more readable.

xcm/src/v1/multilocation.rs
KiChjang: Perhaps rewriting the MultiLocation format in a different way would make it more apparent:

`m` originally is in the form `../../PalletInstance(3)`, and now we want to prepend `../Parachain(21)/OnlyChild` to `m`. Or in other words, we start with `../Parachain(21)/OnlyChild` and we append `../../PalletInstance(3)` to it. We see that `m` contains 2 parent junctions, which will cause the `Parachain(21)/OnlyChild` part to be "eliminated" (since we are going back 2 "directories"/junctions up), and now we append `PalletInstance(3)` to the result, which would bring us to `../PalletInstance(3)`.

xcm/src/v1/multilocation.rs
KiChjang: We could, I'm not exactly sure why this exists honestly, since implementing iterators properly involves implementing the `IntoIterator` trait.

Which does give a a good idea -- perhaps instead of making an `into_iter_rev` method, we can implement a custom `rev` method in the `impl Iterator for JunctionsIterator` trait impl and make use of `JunctionsReverseIterator` there...

xcm/xcm-executor/src/lib.rs
KiChjang: Oh wow, I think the original intent here is to recognize that it's absolutely impossible to create a `Junction::Parent` with the new `MultiLocation` changes, but now thinking of it more closely, we really need to check that `who` has a parent count of 0.

runtime/kusama/src/lib.rs
KiChjang: This is unfortunately in a `const` context, and the `into()` function is not `const`. Happy to brainstorm some more ideas to make this more readable however.

xcm/src/v1/multilocation.rs
KiChjang: Nope, impossible, because the `rev()` method returns https://doc.rust-lang.org/stable/std/iter/struct.Rev.html and doesn't return `impl Iterator` :disappointed: .

KiChjang: Oh, but what we can do is implement `DoubleEndedIterator` on `Junctions` and remove `JunctionsReverseIterator` altogether and everything will work!

xcm/src/v0/multi_location.rs
```diff
@@ -696,20 +696,96 @@ impl MultiLocation {
...
+			res.push(Junction::Parent)?;
+		}
+
+		match v1.interior {
+			Null => Ok(res),
+			X1(j0) => res.pushed_with(j0.into()).map_err(|_| ()),
+			X2(j0, j1) => res
+				.pushed_with(j0.into())
+				.map_err(|_| ())?
+				.pushed_with(j1.into())
```
gavofyork, line 715:
```suggestion
				.and_then(|res| res.pushed_with(j1.into()))
```

gavofyork: and the rest...

xcm/src/v1/junction.rs
```diff
@@ -101,26 +118,29 @@ impl BodyPart {
 /// A single item in a path to describe the relative location of a consensus system.
 ///
 /// Each item assumes a pre-existing location as its context and is defined in terms of it.
+///
```
gavofyork, line 121: No need for this...

xcm/src/v1/junction.rs
```diff
@@ -101,26 +118,29 @@ impl BodyPart {
...
 pub enum Junction {
-	/// The consensus system of which the context is a member and state-wise super-set.
-	///
-	/// NOTE: This item is *not* a sub-consensus item: a consensus system may not identify itself trustlessly as
-	/// a location that includes this junction.
-	Parent,
 	/// An indexed parachain belonging to and operated by the context.
 	///
 	/// Generally used when the context is a Polkadot Relay-chain.
+	#[codec(index = 1)]
```
gavofyork, line 131:
```suggestion
```

gavofyork: Remove the others, too.

kianenigma (commented):
All my cooments are addressed and everythign looks good to me. Will abstine from approve for now since I want to give it one last read.

runtime/kusama/src/lib.rs
```diff
@@ -1263,7 +1263,7 @@ pub type XcmRouter = (
 
 parameter_types! {
 	pub const Kusama: MultiAssetFilter = Wild(AllOf { fun: WildFungible, id: Concrete(KsmLocation::get()) });
-	pub const KusamaForStatemint: (MultiAssetFilter, MultiLocation) = (Kusama::get(), X1(Parachain(1000)));
+	pub const KusamaForStatemint: (MultiAssetFilter, MultiLocation) = (Kusama::get(), X1(Parachain(1000)).into());
```
gavofyork, line 1266: Should be possible with `Parachain(1000).into()`

runtime/rococo/src/lib.rs
```diff
@@ -583,9 +583,9 @@ impl parachains_paras::Config for Runtime {
 }
 
 parameter_types! {
-	pub const RocLocation: MultiLocation = MultiLocation::Here;
+	pub const RocLocation: MultiLocation = MultiLocation::here();
```
gavofyork, line 586:
```suggestion
	pub const RocLocation: MultiLocation = Here.into();
```

runtime/common/src/xcm_sender.rs
```diff
@@ -19,7 +19,7 @@
 use parity_scale_codec::Encode;
 use runtime_parachains::{configuration, dmp};
 use sp_std::marker::PhantomData;
-use xcm::opaque::v1::{Error, Junction, MultiLocation, Result, SendXcm, Xcm};
+use xcm::opaque::v1::{Error, Junction, Junctions, MultiLocation, Result, SendXcm, Xcm};
```
shawntabrizi, line 22: probably can be simplified to

```suggestion
use xcm::latest::prelude::*;
```

xcm/src/v0/multi_location.rs
```diff
@@ -16,7 +16,7 @@
 
 //! Cross-Consensus Message format data structures.
 
-use core::{mem, result};
+use core::{convert::TryFrom, mem, result};
 
 use super::{super::v1::MultiLocation as MultiLocation1, Junction};
```
shawntabrizi, line 21: some clarity here would help

xcm/src/v0/mod.rs
```diff
@@ -25,34 +25,36 @@ use core::{
...
-		BodyId, BodyPart,
 		Junction::*,
-		NetworkId, OriginKind,
+		OriginKind,
 		Xcm::{self, *},
 	};
 }
 
-pub use super::v1::{BodyId, BodyPart, Junction, NetworkId, OriginKind, Response};
+pub use super::v1::{BodyPart, NetworkId, OriginKind, Response};
```
shawntabrizi, line 57: This seems dangerous. We should have imported the unchanged v0 parts into v1

xcm/Cargo.toml
```diff
@@ -10,6 +10,7 @@ impl-trait-for-tuples = "0.2.0"
 parity-scale-codec = { version = "2.0.0", default-features = false, features = [ "derive" ] }
 derivative = {version = "2.2.0", default-features = false, features = [ "use_core" ] }
 log = { version = "0.4.14", default-features = false }
+static_assertions = "1.1.0"
```
shawntabrizi, line 13: remove

gavofyork (approved):

shawntabrizi (approved):

KiChjang:
bot merge

KiChjang:
image

PR #3407, opened 2021-07-05
substrate #9202 companion: Multiple vesting schedules

emostov:
substrate companion: https://github.com/paritytech/substrate/pull/9202

Areas for feedback:
- What should `MaxVestingSchedules` be for each runtime? I have initially set it to 24, which is based on if you get 1 schedule a month you can go 2 years without merging any schedules; but this is a very arbitrary goal.

TODO:
- [x] make sure benchmarks work with each runtime's values


emostov:
`VestingInfo` encoded size can be described as `Balance + Balance + BlockNumber` == `u128 + u128 + u32` == `16 + 16 + 4` == 34.

Assuming a `BoundedVec` has no overhead when encoded, we can say the encoded size of the users vesting schedules == `34 * schedule_count`

Below is a map of `schedule count -> encoded size`

```rust
3 -> 102   // closest to 2^7
7 -> 238   // closest to 2^8
14 -> 504  // closest to 2^9
24 -> 816
30 -> 1020 // closest to 2^10
50 -> 1700
60 -> 2040 // closest to 2^11
```

runtime/polkadot/src/lib.rs
```diff
@@ -1592,4 +1594,18 @@ mod test_fees {
...
+		use frame_support::BoundedVec;
+		use parity_scale_codec::MaxEncodedLen;
+
+		let sched_len = VestingInfo::<Balance, BlockNumber>::max_encoded_len();
+		let vec_len = BoundedVec::<VestingInfo<Balance, BlockNumber>, MaxVestingSchedules>::max_encoded_len();
+
+		println!("Vesting schedules max encoded len: {} ({}).", vec_len, sched_len);
+		assert_eq!(sched_len, 36);
+		assert_eq!(vec_len, 1009);
+	}
```
emostov, line 1610: Not sure if I should leave this test in here

runtime/common/src/claims.rs
```diff
@@ -726,6 +726,7 @@ mod tests {
 
 	parameter_types! {
 		pub const MinVestedTransfer: u64 = 0;
+		pub const MinVestedTransfer: u32 = 28;
```
coriolinus, line 729: Looks like you've got a duplicate definition here. You probably meant to name this `MaxVestingSchedules`?

runtime/polkadot/src/lib.rs
coriolinus: This seems like it's served its purpose and isn't really testing anything useful here anymore, so I'd recommend removing it.

coriolinus (approved):

runtime/common/src/claims.rs
emostov: 93554b3

runtime/kusama/src/lib.rs
```diff
@@ -949,6 +949,7 @@ impl pallet_society::Config for Runtime {
 
 parameter_types! {
 	pub const MinVestedTransfer: Balance = 100 * CENTS;
```
emostov, line 951: Do we want to reduce this to align with ED (`1 * CENTS`)

shawntabrizi: no, we want it to be more than ED

shawntabrizi (approved):

runtime/test-runtime/src/lib.rs
```diff
@@ -433,7 +433,7 @@ impl claims::Config for Runtime {
 }
 
 parameter_types! {
-	pub storage MinVestedTransfer: Balance = 100 * DOLLARS;
+	pub const MinVestedTransfer: Balance = 100 * DOLLARS;
```
shawntabrizi, line 436: why is this change being made?

sounds like instead you should be placing something within a externalities environment instead of doing this

PR #3411, opened 2021-07-05
Enable logging in the puppet worker

bkchr:
This is useful for tests where something is failing :D


pepyakin (approved):
Codewise this looks trivially good. I cannot tell if this has the intended effect though.

bkchr:
> Codewise this looks trivially good. I cannot tell if this has the intended effect though.

Yeah it has. I'm currently facing such a situation :(

PR #3412, opened 2021-07-05
Use MaxEncodedLen trait from new parity-scale-codec v2.2

Xanewok:
Companion for paritytech/substrate#9163


Xanewok:
Adding a medium label since it'd be good to have that reviewed and merged soon-ish but I'm having a problem with the wording "elevates a release containing this PR to XYZ". Am I misusing it? By release do we mean a specific version that includes this change and whether it's "important" or not or do we mean the priority of "releasing" (merging and publishing) this particular change?

Xanewok:
> Changes with a priority higher than C1-low *MUST* have a B- label that is not B0-Silent

I guess it's the former, downgrading to the "low" label then

bkchr (approved):

ordian (approved):

Issue #3192, opened 2021-06-07
validation host is too eager to deem a candidate invalid


pepyakin:
Recently we observed that the validation host may fail to communicate with the child process that is responsible for compiling the PVF. This now leads to validation host thinking that the PVF that is broken. Therefore, all following execution requests will fail with "invalid candidate".

The logic there is wrong because failing to communicate with the child process should not lead to claiming that the candidate is invalid. 

Besides that, it is not clear to me if we should deem the candidate invalid if the PVF itself is invalid.

rphmeier:
I think that from the perspective of the high-level code the 'validate candidate' API should make a best effort to validate the candidate and if that fails then to be clear that there has been an internal failure. We must make sure that malicious code cannot cause issues presenting as an internal failure. When internal issues are encountered the candidate validation subsystem might retry a few times to present an API which the high level code can use simply to answer the question of validity

burdges:
Yes exactly, we should report how we fail though when starting a dispute.  We should relevant logs related to disputes too.

pepyakin:
I was thinking about this and I am not sure if I agree.

Imagine a malicious PVF crafted so that it would OOM on 50% of machines during compilation. This would make one side of the chasm to think the candidates for that PVF are valid and the other side would think it is invalid. 

Is that really worse than if the second side thought that this is an internal failure?

rphmeier:
> Imagine a malicious PVF crafted so that it would OOM on 50% of machines during compilation. This would make one side of the chasm to think the candidates for that PVF are valid and the other side would think it is invalid.

Can't this be averted somehow? We must have some kind of threat model for parachains where the compilation can succeed on all machines

Issue #3417, opened 2021-07-06
Validation host is not amenable for thorough testing


pepyakin:
Testing the PVF validation host is quite annoying. This system implements a queue that is handled by workers. Workers are spawned as external processes. Moreover, they are spawned using the `exec` call that points to the same binary with a special command line flags.

The current system tried to anticipate this. Despite that it's still pain to test. Everything that touches effectful IO, especially process creation, is just pain. 

It's possible to separate the system in two parts: a state machine and a dumb IO shell.

The dumb IO shell is responsible for spawning new processes, reporting their death, sending and receiving messages from them. The state machine part handles the queue logic, issues commands for spawning new processes and commands sending messages to specific processes. It's very important that that part doesn't actually perform any logic.

The hope is to isolate the IO shell making the interaction protocol with it very slim. That would enable us to test various behaviors. A puppet worker seems to be enough to cover many scenarios. Things like die before start work, fail to spawn, explode on a message seem all easy enough.

The state machine should be inherently easier to test. Input some commands, catch some requests coming out of it and reply with responses specific to the test scenario.

PR #3418, opened 2021-07-06
PVF: unresponsive worker doesn't mean the candidate is bad

pepyakin:
Closes https://github.com/paritytech/polkadot/issues/3192

This new behavior should be exercised by an existing test: `rip_for_start_work`

https://github.com/paritytech/polkadot/blob/0eb7905aeab120273b795659f34d1d0efcf0ac15/node/core/pvf/src/prepare/queue.rs#L863-L893

However, it is not clear how to test it end-to-end. Here is an entire issue about that https://github.com/paritytech/polkadot/issues/3417

### DIFF
--- a/node/core/pvf/src/prepare/pool.rs
+++ b/node/core/pvf/src/prepare/pool.rs
@@ -179,7 +179,7 @@ async fn purge_dead(
 		}
 	}
 	for w in to_remove {
-		if let Some(_) = spawned.remove(w) {
+		if spawned.remove(w).is_some() {
 			reply(from_pool, FromPool::Rip(w))?;
 		}
 	}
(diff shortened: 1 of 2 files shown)


node/core/pvf/src/prepare/pool.rs
```diff
@@ -308,6 +312,13 @@ fn handle_mux(
 
 					Ok(())
 				}
+				Outcome::Unreachable => {
+					if let Some(_data) = spawned.remove(worker) {
```
bkchr, line 316:
```suggestion
					if spawned.remove(worker).is_some() {
```

pepyakin: Did a bulk change: https://github.com/paritytech/polkadot/pull/3418/commits/35421244269bad933a901e4753abc265c5455e73

bkchr (approved):

PR #3423, opened 2021-07-06
ci: use srtool-actions to build runtimes

chevdor:
- build westend as well
- switch to using srtool-actions@v0.3.0 which defaults to the `paritytech/srtool` docker image
- add compressed runtimes to the artifacts

I have tested at https://github.com/chevdor/polkadot/actions/runs/1007250044

image

The compressed runtime is added to the <chain>-runtime artifacts.

image

### DIFF
--- a/.github/workflows/publish-draft-release.yml
+++ b/.github/workflows/publish-draft-release.yml
@@ -24,25 +24,17 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        runtime: ['polkadot', 'kusama', 'westend']
-    # container:
-    #   image: paritytech/srtool:nightly-2021-03-15
-    #   volumes:
-    #     - ${{ github.workspace }}:/build
-    #   env:
-    #     PACKAGE: ${{ matrix.runtime }}-runtime
-    #     RUSTC_VERSION: nightly-2020-10-27
+        runtime: ["polkadot", "kusama", "westend"]
     steps:
       - uses: actions/checkout@v2
       - name: Cache target dir
         uses: actions/cache@v2
         with:
-          path: '${{ github.workspace }}/runtime/${{ matrix.runtime }}/target'
+          path: "${{ github.workspace }}/runtime/${{ matrix.runtime }}/target"
           key: srtool-target-${{ matrix.runtime }}-${{ github.sha }}
           restore-keys: |
             srtool-target-${{ matrix.runtime }}-
             srtool-target-
-
       - name: Build ${{ matrix.runtime }} runtime
         id: srtool_build
         uses: chevdor/srtool-actions@v0.3.0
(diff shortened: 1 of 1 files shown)


.github/workflows/publish-draft-release.yml
```diff
@@ -65,49 +54,52 @@ jobs:
         uses: actions/upload-artifact@v2
         with:
           name: ${{ matrix.runtime }}-runtime
-          path: "${{ steps.build-runtime.outputs.wasm }}"
+          path: |
+            ${{ steps.srtool_build.outputs.wasm }}
```
s3krit, line 58: Just curious, what happens when you specify 2 paths here?

.github/workflows/publish-draft-release.yml
```diff
@@ -65,49 +54,52 @@ jobs:
...
+            ${{ steps.srtool_build.outputs.wasm_compressed }}
```
s3krit, line 59: Just curious, what happens when you use two lines for the path here?

s3krit (commented):
Lgtm. So long as the runtime srtool output json file hasn't changed format and is still reachable by publish_draft_release.rb, everything should stll Just Work™ (the job failed on your test run due to you currently having no releases). I'll verify that in my polkadot fork I use for testing release stuff then give this an approval. Cheers!

.github/workflows/publish-draft-release.yml
TriplEight: that are the names of two artifacts from `id: srtool_build` job

.github/workflows/publish-draft-release.yml
```diff
@@ -65,49 +54,52 @@ jobs:
...
+      - name: Download srtool json output
+        uses: actions/download-artifact@v2
+      - name: Generate release text
+        env:
+          RUSTC_STABLE: ${{ needs.get-rust-versions.outputs.rustc-stable }}
+          RUSTC_NIGHTLY: ${{ needs.get-rust-versions.outputs.rustc-nightly }}
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          gem install changelogerator git toml
+          ruby $GITHUB_WORKSPACE/polkadot/scripts/github/generate_release_text.rb | tee release_text.md
```
TriplEight, line 85: this looks huge

TriplEight (commented):
every time I see CI installs something I'm tempted to write an image for this.

.github/workflows/publish-draft-release.yml
chevdor: The artifact `${{ matrix.runtime }}-runtime` is an id. Listing several files or a pattern means you end up with several files in this artifact.

.github/workflows/publish-draft-release.yml
chevdor: The PR did not change that part. It looks like the identation changed though.
I did run some formatting.

chevdor:
> the job failed on your test run due to you currently having no releases

I also don't have the Matrix keys/tokens so I did not expect that part to work.

.github/workflows/publish-draft-release.yml
TriplEight: Then PR is good, but in general, it's something that would need some love.

TriplEight (approved):

s3krit (approved):
Cool, verified on my end with a fork that the release generation stuff works as expected

chevdor:
bot merge

PR #3425, opened 2021-07-06
Parachains shared.rs to Frame V2

ferrell-code:
relates: #2882 

So users of the `ParasShared` pallet must be careful about the name they used in `construct_runtime!.` Hence the runtime-migration label, which might not be needed depending on the configuration of the `ParasShared` pallet.

Kusama and westend both already use `ParasShared` in `construct_runtime!` no migration needed. Polkadot does not use the `ParasShared` pallet.

Rococo uses `Shared` in `construct_runtime!`, therefore I altered it to `ParasShared` to keep the pallet name the same as seen in `decl_storage!` of `ParasShared`


runtime/test-runtime/src/lib.rs
```diff
@@ -532,6 +532,7 @@ construct_runtime! {
 		ParasInherent: parachains_paras_inherent::{Pallet, Call, Storage, Inherent},
 		Initializer: parachains_initializer::{Pallet, Call, Storage},
 		Paras: parachains_paras::{Pallet, Call, Storage, Origin, Event},
+		ParasShared: parachains_shared::{Pallet, Call, Storage},
 		Scheduler: parachains_scheduler::{Pallet, Call, Storage},
 		ParasSudoWrapper: paras_sudo_wrapper::{Pallet, Call},
 		SessionInfo: parachains_session_info::{Pallet, Call, Storage},
```
ferrell-code, line 538: pallet must be put in `construct_runtime` for tests... if you make a call into `ParasShared` storage without doing so frame_support will panic as it needs a pallet name (the one that was previously declared in `decl_storage!`. Probably should check if any of cumulus tests do this

thiolliere:
looking at the number of line removed, I guess some master merge got wrong, isn't it ? @ferrell-code

ferrell-code:
Mr. Tabrizi swooping in with those 5 am pushes lol, I force pushed back to the original point haha and resoolved conflicts

KiChjang (approved):

emostov (approved):

KiChjang:
bot merge

shawntabrizi:
@ferrell-code yeah sorry about that.

Was gonna fix it for you, but then got too tired, and gave up lol

ferrell-code:
I got a kick out of it, always appreciate your help :dromedary_camel:

PR #3426, opened 2021-07-07
parachain dmp.rs to Frame V2

ferrell-code:
relates: #2882 

So users of the `Dmp` pallet must be careful about the name they used in `construct_runtime!.` Hence the runtime-migration label, which might not be needed depending on the configuration of the `Dmp` pallet.

Kusama and westend use the name `ParasDmp`. I changed it to `Dmp` so the pallet name matches what was declared in the `decl_storage!`. rococo already used the name `Dmp`.

polkadot does not use `Dmp`


runtime/parachains/src/dmp.rs
```diff
@@ -71,30 +73,43 @@ impl fmt::Debug for ProcessedDownwardMessagesAcceptanceErr {
...
+	#[pallet::pallet]
+	#[pallet::generate_store(pub(super) trait Store)]
+	pub struct Pallet<T>(_);
+
+	#[pallet::config]
+	pub trait Config: frame_system::Config + configuration::Config {}
+
+	/// The downward messages addressed for a certain para.
+	#[pallet::storage]
+	pub type DownwardMessageQueues<T: Config> = StorageMap<
```
KiChjang, line 89:
```suggestion
	pub(crate) type DownwardMessageQueues<T: Config> = StorageMap<
```

runtime/parachains/src/dmp.rs
```diff
@@ -71,30 +73,43 @@ impl fmt::Debug for ProcessedDownwardMessagesAcceptanceErr {
...
+
+	/// A mapping that stores the downward message queue MQC head for each para.
+	///
+	/// Each link in this chain has a form:
+	/// `(prev_head, B, H(M))`, where
+	/// - `prev_head`: is the previous head hash or zero if none.
+	/// - `B`: is the relay-chain block number in which a message was appended.
+	/// - `H(M)`: is the hash of the message being appended.
+	#[pallet::storage]
+	pub type DownwardMessageQueueHeads<T: Config> = StorageMap<_, Twox64Concat, ParaId, Hash, ValueQuery>;
```
KiChjang, line 105:
```suggestion
	pub(crate) type DownwardMessageQueueHeads<T: Config> = StorageMap<_, Twox64Concat, ParaId, Hash, ValueQuery>;
```

KiChjang (approved):

shawntabrizi (approved):
With the rename this looks good

shawntabrizi:
bot merge

PR #3440, opened 2021-07-08
Parachains inclusion.rs to Frame V2

ferrell-code:
relates: #2882 

So users of the `ParaInclusion` pallet must be careful about the name they used in `construct_runtime!.` Hence the runtime-migration label, which might not be needed depending on the configuration of the `ParaInclusion` pallet.

westend and kusama use `ParasInclusion` so I changed them to `ParaInclusion`. rococo uses `Inclusion` also migrated to `ParaInclusion`.

polkadot does not us `ParaInclusion`


runtime/parachains/src/runtime_api_impl/v1.rs
```diff
@@ -300,6 +300,8 @@ where
 				=> CandidateEvent::CandidateIncluded(c, h, core, group),
 			RawEvent::<T>::CandidateTimedOut(c, h, core)
 				=> CandidateEvent::CandidateTimedOut(c, h, core),
+			RawEvent::<T>::__Ignore(_, _)
+				=> unreachable!("__Ignore cannot be used"),
```
ferrell-code, line 304: Maybe change `frame_support` to not expose `__Ignore(_,_)` in Event api?

runtime/parachains/src/inclusion.rs
```diff
@@ -189,33 +193,42 @@ decl_error! {
...
 
-decl_module! {
-	/// The parachain-candidate inclusion module.
-	pub struct Module<T: Config>
-		for enum Call where origin: <T as frame_system::Config>::Origin
-	{
-		type Error = Error<T>;
+	/// The latest bitfield for each validator, referred to by their index in the validator set.
+	#[pallet::storage]
+	pub type AvailabilityBitfields<T: Config> = StorageMap<
```
KiChjang, line 199: Wrong visibility on each of the storage items, this and the other ones below:

```suggestion
	pub(crate) type AvailabilityBitfields<T: Config> = StorageMap<
```

runtime/parachains/src/inclusion.rs
```diff
@@ -189,33 +193,42 @@ decl_error! {
...
+		_,
+		Twox64Concat,
+		ValidatorIndex,
+		AvailabilityBitfieldRecord<T::BlockNumber>
+	>;
+
+	/// Candidates pending availability by `ParaId`.
+	#[pallet::storage]
+	pub type PendingAvailability<T: Config> = StorageMap<
```
KiChjang, line 208:
```suggestion
	pub(crate) type PendingAvailability<T: Config> = StorageMap<
```

runtime/parachains/src/inclusion.rs
```diff
@@ -189,33 +193,42 @@ decl_error! {
...
+		_,
+		Twox64Concat,
+		ParaId,
+		CandidatePendingAvailability<T::Hash, T::BlockNumber>
+	>;
+
+	/// The commitments of candidates pending availability, by `ParaId`.
+	#[pallet::storage]
+	pub type PendingAvailabilityCommitments<T: Config> = StorageMap<
```
KiChjang, line 217:
```suggestion
	pub(crate) type PendingAvailabilityCommitments<T: Config> = StorageMap<
```

KiChjang (approved):

runtime/parachains/src/inclusion.rs
ferrell-code: yeah, I think the other pallets I migrated don't need public facing storage either

runtime/parachains/src/runtime_api_impl/v1.rs
shawntabrizi: i think unavoidable unfortunately, hence the name and stuff

shawntabrizi (commented):
With the import rename, this looks good to me

shawntabrizi (approved):

shawntabrizi:
bot merge

shawntabrizi:
@ferrell-code can you please include your DOT / KSM address again?

ferrell-code:
Polka:
15zKd25HoSDPHCn1HjuuHTioXLT7PA2pZQ8gYjFsHW4r96qS

Or

Kusama:
DSSYqgJqdE6HmFAXpiq3y6eZ5FYZ3ZxmX3joRBkh345mxEN

Thanks @shawntabrizi  :sauropod:, maybe I should just put it in my github bio haha

runtime/parachains/src/runtime_api_impl/v1.rs
thiolliere: It is to capture the generics in case none of the fields capture it.

But it was more to make it handy we could not generate it like decl_event do.
Also right now there is a weird case that construct_runtime doesn't handle non generic or generic on `T` only for event when the pallet is instantiable. so in this case we sometime need an ignore variant.

shawntabrizi:
@ferrell-code the tip has been opened on Polkadot.

Thank you again for your contributions!

PR #3447, opened 2021-07-08
Dispute Coordinator: Move to batch Request - Response model

### LINKED ISSUE #3439: DisputeCoordinator QueryCandidateVotes should be able to answer multiple queries at once
As seen in #3348, the provisioner currently has to make a `QueryCandidateVotes` request for each dispute in `RecentDisputes`. It should be possible to have the provisioner send a single query to the dispute coordinator to be answered all at once, similar to `CandidateBackingMessage::GetBackedCandidates`.

Lldenaurois:
This PR modifies the communication protocol between the provisioner and the dispute-coordinator. In the past, the dispute-coordinator kept track of a FuturesOrdered of oneshot-receivers and constructed a view of the recent disputes in this fashion.

Instead, it is more efficient to allow batch queries and only iterator over the recent disputes once. We leverage a single oneshot channel to reply to the batch `Vec<(SessionIndex, CandidateHash)>` request with the `Vec<(SessionIndex, Candidatehash, CandidateVotes)>` batch response.

Closes #3439 . Depends on #3348 merge.

### DIFF
--- a/node/core/dispute-coordinator/src/lib.rs
+++ b/node/core/dispute-coordinator/src/lib.rs
@@ -437,18 +437,21 @@ async fn handle_incoming(
 			let _ = rx.send(collect_active(recent_disputes, now));
 		}
 		DisputeCoordinatorMessage::QueryCandidateVotes(
-			session,
-			candidate_hash,
+			query,
 			rx
 		) => {
-			let candidate_votes = db::v1::load_candidate_votes(
-				store,
-				&config.column_config(),
-				session,
-				&candidate_hash,
-			)?;
-
-			let _ = rx.send(candidate_votes.map(Into::into));
+			let mut query_output = Vec::new();
+			for (session_index, candidate_hash) in query.into_iter() {
+				if let Some(v) = db::v1::load_candidate_votes(
+					store,
+					&config.column_config(),
+					session_index,
+					&candidate_hash,
+				)? {
+					query_output.push(v.into());
+				}
+			}
+			let _ = rx.send(query_output);
 		}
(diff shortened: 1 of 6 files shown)


node/core/provisioner/src/lib.rs
```diff
@@ -567,38 +568,16 @@ async fn select_disputes(
...
-		}
+	let dispute_candidate_votes: Vec<_> = {
+		let (tx, rx) = oneshot::channel();
+		sender.send_message(DisputeCoordinatorMessage::QueryCandidateVotes(
+			recent_disputes.clone(),
+			tx,
+		).into()).await;
 
-		vote_sets
+		rx.await.unwrap_or_default().into_iter().zip(recent_disputes.into_iter())
```
rphmeier, line 578: I'd like a `debug!` log as done above

node/core/dispute-coordinator/src/lib.rs
```diff
@@ -437,18 +437,21 @@ async fn handle_incoming(
...
-				store,
-				&config.column_config(),
-				session,
-				&candidate_hash,
-			)?;
-
-			let _ = rx.send(candidate_votes.map(Into::into));
+			let mut query_output = Vec::new();
+			for (session_index, candidate_hash) in query.into_iter() {
+				if let Some(v) = db::v1::load_candidate_votes(
```
rphmeier, line 445: It would be better to fail the request if any of the votes aren't present. As such, the `zip` done in the provisioner is inaccurate.

rphmeier (changes requested):
Nits and needs guide update as well.

PR #3449, opened 2021-07-09
Companion: Limit the maximum number of wasm memory pages a runtime can have

### LINKED ISSUE #317: parachain validation function and linear memory
For now it seems that the parachain validation function host environment provides and the parachain wasm imports a linear memory instance.

On the other hand, substrate does it the other way around: the host just imports the memory exported by the wasm runtime. 

In substrate this is not much important since the wasm binary is always trusted. However, this is not the case for the parachain wasm. If we were to make the wasm parachain to export a linear memory then it could declare that that the memory instance requires huge amount of memory and as part of instantiation the wasm engine would happiliy allocate up to 4GB of memory. This is fixable but requires either deserialization of a wasm module and checking its memories or implementing configurable limits in the wasm engine.

I wonder if we should reconsider these design decisions. Also I wonder there are any  implications for cumulus

### LINKED ISSUE #699: Validation context should allow in maximum 1GB of memory
After https://github.com/paritytech/polkadot/pull/665 is merged, we do not check that the validation allocates at most 1GB of memory. We need to bring this check back. I'm not sure if we should stay with the parachain WASM blob providing us the memory or switching back to give it to the WASM blob. If we do the later, we need to add support for this into `sc-executor`.

I think that @pepyakin has an opinion, on what is better :)

pepyakin:
Companion for https://github.com/paritytech/substrate/pull/9308
Closes #699
Closes #317

This is a high priority because this is a potentially breaking change.


rphmeier:
Does this close #699

pepyakin:
Haha, yes, it does. In the end, I forgot to put the tag 🤦 Thanks!

rphmeier (approved):

Issue #3427, opened 2021-07-07
[node] Add additional sanitization checks to overseer


drahnr:
#2962 lays the ground work of adding additional logic to the overseer, such that certain properties can be detected at compile time:

Chores:
* [x] #3773
* [x] #3772
   * [x] #3771
* <s> Remove the `focus()` method and streamline the incoming information flow</s>
* [ ] Handle where clauses for `baggage` generics

Significant enhancements:
* [ ] #3774
    * [x] #3825
    * [x] #3826
    * [x] #5426

rphmeier:
> annotated subsystems which messages are sent by a subsystem

This and everything under it is amazing. Looking forward to it

bkchr:
>     * inject a `unbounded` channel into each cycle, to prevent potential dead-locks until there is a more systematic approach with prioritization

Why do we try to work around the fundamental issue and not solve it properly? The fundamental issue here is that oneshot channels are used inside these messages and that we await them, blocking the entire subsystem. IMO oneshot channels should be forbidden in these messages or only allowed in very rare cases where we can proof that no dead lock ever happens.

Instead of oneshot channels there could just be a closure that constructs a `AllMessages` message and this message would be dispatched again.

Yes that would complicate the design of some subsystems, but it would make it much easier to reason about the implementation of a subsystem and problematic code pieces.

drahnr:
> > ```
> > * inject a `unbounded` channel into each cycle, to prevent potential dead-locks until there is a more systematic approach with prioritization
> > ```
> 
> Why do we try to work around the fundamental issue and not solve it properly? The fundamental issue here is that oneshot channels are used inside these messages and that we await them, blocking the entire subsystem. IMO oneshot channels should be forbidden in these messages or only allowed in very rare cases where we can proof that no dead lock ever happens.
> 
> Instead of oneshot channels there could just be a closure that constructs a `AllMessages` message and this message would be dispatched again.
> 
> Yes that would complicate the design of some subsystems, but it would make it much easier to reason about the implementation of a subsystem and problematic code pieces.

In principle you are correct. 
The above was meant as low-impact change, and that's the precise advantage of it - not to re-architect half of our subsystems logic. 

I also think that spawning additional tasks whenever there is a response channel could get the same effect with less impact. Note that the unbounded channel also solves potential congestion due to loops in our subsystem communication, which is something that is no easier with `AllMessages`-based request-response. I am happy to discuss this further.

[...]

PR #3454, opened 2021-07-09
remove the overseer trait bounds

drahnr:
Follow up addressing the final nits #2962 and moves a few `trait`s into the proc-macro to avoid usage-gunk at the cost of less obvious error messages when implementing `trait Subsystem<Ctx>`.

Ref #3427


drahnr:
Obsoleted by https://github.com/paritytech/polkadot/pull/5314

PR #3459, opened 2021-07-12
Dispute Coordinator: Batch queries

### LINKED ISSUE #3439: DisputeCoordinator QueryCandidateVotes should be able to answer multiple queries at once

Lldenaurois:
This PR modifies the communication protocol between the provisioner and the dispute-coordinator. In the past, the dispute-coordinator kept track of a FuturesOrdered of oneshot-receivers and constructed a view of the recent disputes in this fashion.

Instead, it is more efficient to allow batch queries and only iterator over the recent disputes once. We leverage a single oneshot channel to reply to the batch Vec<(SessionIndex, CandidateHash)> request with the Vec<(SessionIndex, Candidatehash, CandidateVotes)> batch response.

Closes #3439

### DIFF
--- a/node/core/dispute-coordinator/src/lib.rs
+++ b/node/core/dispute-coordinator/src/lib.rs
@@ -444,18 +444,21 @@ async fn handle_incoming(
 			let _ = rx.send(collect_active(recent_disputes, now));
 		}
 		DisputeCoordinatorMessage::QueryCandidateVotes(
-			session,
-			candidate_hash,
+			query,
 			rx
 		) => {
-			let candidate_votes = db::v1::load_candidate_votes(
-				store,
-				&config.column_config(),
-				session,
-				&candidate_hash,
-			)?;
-
-			let _ = rx.send(candidate_votes.map(Into::into));
+			let mut query_output = Vec::new();
+			for (session_index, candidate_hash) in query.into_iter() {
+				if let Some(v) = db::v1::load_candidate_votes(
+					store,
+					&config.column_config(),
+					session_index,
+					&candidate_hash,
+				)? {
+					query_output.push(v.into());
+				}
+			}
+			let _ = rx.send(query_output);
 		}
 		DisputeCoordinatorMessage::IssueLocalStatement(
 			session,
--- a/node/core/dispute-coordinator/src/lib.rs
+++ b/node/core/dispute-coordinator/src/lib.rs
@@ -456,6 +456,12 @@ async fn handle_incoming(
 					&candidate_hash,
 				)? {
 					query_output.push(v.into());
+				} else {
+					return Err(
+						Error::Io(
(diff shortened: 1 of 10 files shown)


node/core/dispute-coordinator/src/lib.rs
```diff
@@ -444,18 +444,27 @@ async fn handle_incoming(
...
+			for (session_index, candidate_hash) in query.into_iter() {
+				if let Some(v) = db::v1::load_candidate_votes(
+					store,
+					&config.column_config(),
+					session_index,
+					&candidate_hash,
+				)? {
+					query_output.push(v.into());
+				} else {
+					return Err(
```
Lldenaurois, line 460: @rphmeier Here is the modification that ensures we only respond if every `(session_index, candidate_hash)` exists in the db or we return an error.

node/core/provisioner/src/lib.rs
```diff
@@ -567,38 +567,25 @@ async fn select_disputes(
...
+		).into()).await;
+
+		match rx.await {
+			Ok(v) => v.into_iter().zip(recent_disputes.into_iter())
+						.map(|(vote, (session_index, candidate_hash))| (session_index, candidate_hash, vote))
+						.collect::<Vec<(SessionIndex, CandidateHash, CandidateVotes)>>(),
+			Err(oneshot::Canceled) => {
+				tracing::debug!(
+					target: LOG_TARGET,
+					"Unable to query candidate votes - subsystem disconnected?",
```
Lldenaurois, line 584: @rphmeier Added the added `debug!` log

Lldenaurois (commented):
Migrated comments from the other PR over.

#3447  was closed when another PR was merged. I just noticed today.

rphmeier:
Is this still missing guide changes?

node/core/dispute-coordinator/src/lib.rs
rphmeier: This is returning an io Error from the entire subsystem. We should probably just log something and continue. There's no need to break the message processing loop.

rphmeier: It's important to make a distinction between errors the subsystem encounters while processing requests and requests the subsystem can't fulfill, but is fine with being able to skip those. As written, any request referencing missing data could interrupt the subsystem and potentially take it down.

Lldenaurois:
Added guide changes and moved away from the all-or-nothing model. Instead, the session_index and candidate_hash are being sent in the response. This removes the need for a clone and seems to handle the overall behavior in a cleaner fashion.

rphmeier (approved):

PR #3460, opened 2021-07-12
Small tweaks to staking miner

kianenigma:
a few small fixes to the staking miner, for the next release. 

More might come, but these are essential to keep it working.

### DIFF
--- a/utils/staking-miner/src/main.rs
+++ b/utils/staking-miner/src/main.rs
@@ -299,7 +299,7 @@ async fn create_election_ext<T: EPM::Config, B: BlockT>(
 	with_staking: bool,
 ) -> Result<Ext, Error> {
 	use frame_support::{storage::generator::StorageMap, traits::PalletInfo};
-	let system_block_hash_key = <frame_system::BlockHash<T>>::prefix_hash();
+	use sp_core::hashing::twox_128;
 
 	Builder::<B>::new()
 		.mode(Mode::Online(OnlineConfig {
@@ -322,7 +322,8 @@ async fn create_election_ext<T: EPM::Config, B: BlockT>(
 			},
 			..Default::default()
 		}))
-		.inject_hashed_prefix(&system_block_hash_key)
+		.inject_hashed_prefix(&<frame_system::BlockHash<T>>::prefix_hash())
+		.inject_hashed_key(&[twox_128(b"System"), twox_128(b"Number")].concat())
 		.build()
 		.await
 		.map_err(|why| Error::RemoteExternalities(why))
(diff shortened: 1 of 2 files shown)


utils/staking-miner/src/main.rs
```diff
@@ -322,7 +322,8 @@ async fn create_election_ext<T: EPM::Config, B: BlockT>(
 			},
 			..Default::default()
 		}))
-		.inject_hashed_prefix(&system_block_hash_key)
+		.inject_hashed_prefix(&<frame_system::BlockHash<T>>::prefix_hash())
+		.inject_hashed_key(&[twox_128(b"System"), twox_128(b"Number")].concat())
```
kianenigma, line 326: This is needed, otherwise we can't create the extrinsic correctly -- the `fn additional_signed` of `CheckMortality` would fail.

utils/staking-miner/src/monitor.rs
```diff
@@ -122,10 +122,16 @@ macro_rules! monitor_cmd_for { ($runtime:tt) => { paste::paste! {
...
 			let mut tx_subscription: Subscription<
 				TransactionStatus<<Block as BlockT>::Hash, <Block as BlockT>::Hash>
-			> = client
+			> = match client
 				.subscribe(&"author_submitAndWatchExtrinsic", params! { bytes }, "author_unwatchExtrinsic")
 				.await
-				.unwrap();
+			{
+				Ok(sub) => sub,
+				Err(why) => {
```
kianenigma, line 130:
```suggestion
				Err(why) => {
				// This usually happens when we've been busy with mining for a few blocks, and now we're receiving the subscriptions of blocks in which we were busy. 
				// In these blocks, we still don't have a solution, so we re-compite a new solution
				// and submit it with an outdated `Nonce`, which yields most often `Stale` error.
				// NOTE: to improve this overall, and to be able to intoroduce an array of other fancy features, 
				//  we should make this multi-threaded and do the computation outside of this callback.
```

utils/staking-miner/src/main.rs
emostov: Does the module hash or module + storage item hash exist somewhere on the module itself? I think I recall Shawn mentioning something to that affect, but maybe you would need to import a runtime

emostov (approved):
LGTM

Just want to highlight you note, since it seems like the next major thing for the staking miner once we are confident the mortality stuff works:

NOTE: to improve this overall, and to be able to introduce an array of other fancy features, we should make this multi-threaded and do the computation outside of this callback.

utils/staking-miner/src/main.rs
shawntabrizi: unfortunately `number` is not pub, so cannot be read this way... maybe with `StoargeInfo` it is possible though

shawntabrizi (approved):

shawntabrizi:
bot merge

PR #3462, opened 2021-07-12
Dispute coordinator overlay

Lldenaurois:
This diff moves the dispute coordinator to the common Backend, OverlayedBackend pattern used in both the Approval Voting and Chain Selection subsystems.


node/core/dispute-coordinator/src/backend.rs
```diff
@@ -0,0 +1,160 @@
+use polkadot_primitives::v1::{CandidateHash, SessionIndex};
```
rphmeier, line 1: License header & module docs, please

rphmeier (approved):

rphmeier:
Looks good! Needs another merge; last one must have missed something.

PR #3465, opened 2021-07-13
Reduce staking miner reward (companion `substrate/pull/9395`)

emostov:
EDIT 

We now use the call fee plus a fixed reward base to determine the staking miner reward. 

substrate companion: https://github.com/paritytech/substrate/pull/9395

---
Below is the outdated approach 

The plan is to reduce the staking miner reward until all the components of the signed phase solution submission apparatus have been battle tested

Currently, the polkadot reward is around 24 DOTs per era. Given that the signed solution submission process is still in the early stages of rollout we rather keep the reward low while things are honed in, and then look into raising it from there. Long term we expect teams to spend near-full-time effort on building staking miners, and thus a decent reward.  

> How can this be the right solution?

This PR uses the approach of hardcoding a decrease for `SignedRewardBase`. In order to lower the reward, we decrease the expected signed solution length in all the calculations. The lengths used here are based on current solution lengths for each chain.

Another approach would involve keeping track of the exact fee paid for the submission (https://github.com/paritytech/substrate/issues/9247) and then using that to create the `SignedRewardBase`.

> And how was this decrease amount selected?

The decrease here stems from decreasing the expected length of the submission. The justification lies in the fact that the new lengths closely follow what we currently see on chain.

cc: @kianenigma 
rel: https://github.com/paritytech/polkadot/pull/3393


runtime/kusama/src/lib.rs
```diff
@@ -365,8 +365,8 @@ parameter_types! {
 		sp_runtime::FixedU128::saturating_from_rational(12, 10),
 		// maximum weight possible.
 		weights::pallet_election_provider_multi_phase::WeightInfo::<Runtime>::submit(SignedMaxSubmissions::get()),
-		// assume a solution of 100kb length.
-		100 * 1024
+		// assume a solution of 20k length.
```
shawntabrizi, line 368:
```suggestion
		// assume a solution of 20kb length.
```

shawntabrizi:
I am very confused. You want to fix the fee here, but you are decreasing some assumption about how long the solution message is?

How can this ever be the right solution?

kianenigma:
@shawntabrizi https://github.com/paritytech/substrate/pull/9395 is the right solution for this.

emostov:
@kianenigma ✅ (just can't officially do it since I opened the PR)

shawntabrizi (approved):

shawntabrizi:
seems to be missing the estimatefee config

kianenigma (approved):

PR #3475, opened 2021-07-15
Parachains hrmp.rs to Frame V2

ferrell-code:
relates: #2882 

So users of the `Hrmp` pallet must be careful about the name they used in `construct_runtime!.` Hence the runtime-migration label, which might not be needed depending on the configuration of the `Hrmp` pallet.

rococo uses `Hrmp` in `construct_runtime!`,  so no mirgration needed

kusama and westend use `ParasHrmp` in `construct_runtime!` so I migrated them to `Hrmp`

polkadot does not us `Hrmp`.


rphmeier:
Sorry for the delay on reviewing these.

ferrell-code:
No worries, I'm chillin

shawntabrizi (approved):

runtime/parachains/src/hrmp.rs
```diff
@@ -215,139 +215,46 @@ impl fmt::Debug for OutboundHrmpAcceptanceErr {
...
-#[cfg(feature = "std")]
-fn initialize_storage<T: Config>(preopen_hrmp_channels: &[(ParaId, ParaId, u32, u32)]) {
-	let host_config = configuration::Module::<T>::config();
-	for &(sender, recipient, max_capacity, max_message_size) in preopen_hrmp_channels {
-		if let Err(err) = preopen_hrmp_channel::<T>(sender, recipient, max_capacity, max_message_size) {
-			panic!("failed to initialize the genesis storage: {:?}", err);
-		}
+	#[pallet::event]
+	#[pallet::generate_deposit(pub(super) fn deposit_event)]
+	pub enum Event<T: Config> {
```
thiolliere, line 246: I think event doesn't needs to be generic, however I'm ok with it anyway.

thiolliere (approved):

thiolliere:
bot merge

PR #3479, opened 2021-07-15
overseer gen minor

drahnr:
Minor adjustments, mostly missing `#support_crate ::` prefixes which became apparent when refactoring again in a larger scope


ordian (approved):
if it compiles it works :shrug:

drahnr:
> if it compiles it works shrug

Now, but with #3454 this caused some PITA

PR #3491, opened 2021-07-19
add rustfmt toml

drahnr:
Copies the substrate `rustfmt.toml` to polkadot, to be used with `cargo +nightly fmt` or `rustfmt +nightly file.rs`.

This PR does _NOT_ apply the format, it only provides it to eventually be applied subsystem by subsystem.

Ref https://github.com/paritytech/substrate/pull/8982

### DIFF
--- a/rustfmt.toml
+++ b/rustfmt.toml
@@ -1,0 +1,22 @@
+# Basic
+hard_tabs = true
+max_width = 100
+use_small_heuristics = "Max"
+# Imports
+imports_granularity = "Crate"
+reorder_imports = true
+# Consistency
+newline_style = "Unix"
+normalize_comments = true
+normalize_doc_attributes = true
+# Misc
+chain_width = 80
+spaces_around_ranges = false
+binop_separator = "Back"
+reorder_impl_items = false
+match_arm_leading_pipes = "Preserve"
+match_arm_blocks = false
+match_block_trailing_comma = true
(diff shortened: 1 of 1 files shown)


ordian (approved):
I wonder whether we really need those nightly features, but since the PR has been merged already, let's keep it consistent.

One thing we might add is a couple more options as suggested in https://github.com/paritytech/substrate/pull/8982#issuecomment-882535182, also maybe `edition = "2018"`?

drahnr:
The second revert of issues is https://github.com/rust-lang/rustfmt/issues/4909

drahnr:
> I wonder whether we really need those nightly features, but since the PR has been merged already, let's keep it consistent.
> 
> One thing we might add is a couple more options as suggested in [paritytech/substrate#8982 (comment)](https://github.com/paritytech/substrate/pull/8982#issuecomment-882535182), also maybe `edition = "2018"`?

Edition is auto detected _iff_ used via `cargo +nightly fmt`, which is what we should use imho instead of hardcoding, especially with `edition = "2021"` being around the corner.

rphmeier (approved):

drahnr:
bot merge

andresilva:
Let's make sure the changes here are propagated back to substrate.

drahnr:
> Let's make sure the changes here are propagated back to substrate.

Left a comment on the merged PR

Issue #3500, opened 2021-07-20
how to transfer native assets between two relaychain account from parachain side?


yrong:
the requirement here is just to transfer some ksm from relaychain accountA to accountB, but instead of straightforwardly using polkadot js I would like to invoke from parachain, currently I'm using  [cumlulus ParentAsUmp route] (https://github.com/yrong/bifrost/blob/167d8eadd0c5de6cd9d5b2e38f3483a987b6557a/runtime/asgard/src/lib.rs#L767) to send message but the problem is that origin contains only parachain account and the source accountId part missing. 

Then I notice that there is [RelayedFrom xcm format](https://github.com/paritytech/xcm-format#relayedfrom) which will keep the original accountId so I wrap a [xcmp call with `TransferAsset` embed in `RelayedFrom` message](https://github.com/yrong/bifrost/blob/167d8eadd0c5de6cd9d5b2e38f3483a987b6557a/xcm-support/src/lib.rs#L167-L176)

And then another issue comes is that current polkadot xcm executor can not extract accountId in RelayedFrom message with the default accountId convertor so do some hack in [extension of SovereignAccountOf types](https://github.com/yrong/polkadot/blob/166239e326df7091879c089ca6be8139e97d170b/runtime/rococo/src/lib.rs#L610) to achieve that.

The whole hack is in this [commit](https://github.com/yrong/polkadot/commit/1b5c9b6f924afc46e6d2021373a26af4e4d5d7c6#diff-913ae0b4a6608e2af7b0141218c195418841c4fd46afd33f4d83a862e3777d25) and after all above steps I can transfer assets from parachain side but I would like to know is this the proper way or is there any way without hacking the current polkadot implementation?

bkchr:
CC @apopiak

apopiak:
I don't expect the relaychain to trust a community chain enough to give it control over its accounts directly like this.
If I understand what you want to do correctly, a parachain could take control of e.g. the treasury account which we would want to avoid ;-)
The scenario that will be supported will likely be the following:
There could be a deterministic account id generated (e.g. sth like `hash(parachain id, parachain account)`) for origins coming from parachains. You could then send funds to that generated id and remote control it from the parachain.

yrong:
[...]

PR #3505, opened 2021-07-20
XCM Account Derivation

apopiak:
This PR adds a struct to derive a local account id from a parachain + account combination and adds it to Kusama, Rococo and Westend.

### DIFF
--- a/xcm/xcm-builder/src/location_conversion.rs
+++ b/xcm/xcm-builder/src/location_conversion.rs
@@ -123,6 +123,7 @@ impl<
 	}
 }
 
+/// Extracts the `AccountKey20` from the passed `location` if the network matches.
 pub struct AccountKey20Aliases<Network, AccountId>(PhantomData<(Network, AccountId)>);
 impl<
 	Network: Get<NetworkId>,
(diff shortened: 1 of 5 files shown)


apopiak:
related to https://github.com/paritytech/polkadot/issues/3500

xcm/xcm-builder/src/location_conversion.rs
```diff
@@ -143,6 +144,41 @@ impl<
...
+	AccountId: From<[u8; 32]> + Into<[u8; 32]> + Clone,
+> Convert<MultiLocation, AccountId> for DerivedParachainAccountId32<Network, AccountId> {
+	fn convert(location: MultiLocation) -> Result<AccountId, MultiLocation> {
+		use Junction::*;
+		use MultiLocation::*;
+		let id = match location {
+			X2(Parachain(para_id), AccountId32 { id, network })
+				if network == NetworkId::Any || &network == &Network::get() =>
+			{
+				(para_id, id).using_encoded(blake2_256)
```
gavofyork, line 164: Should have a disabusing prefix.

gavofyork (changes requested):

xcm/xcm-builder/src/location_conversion.rs
apopiak: Like generate a 28byte array and prepend a u32 with the para id?

PR #3507, opened 2021-07-21
Improve test in bridge

thiolliere:
This test is testing using the current weights for pallet system. But the weight for pallet system is changing often, I think this rewrite of the test is correct and is not breaking.

### DIFF
--- a/bridges/modules/dispatch/src/lib.rs
+++ b/bridges/modules/dispatch/src/lib.rs
@@ -397,7 +397,7 @@ mod tests {
 	#![allow(clippy::from_over_into)]
 
 	use super::*;
-	use frame_support::{parameter_types, weights::Weight};
+	use frame_support::{parameter_types, weights::Weight, dispatch::GetDispatchInfo};
 	use frame_system::{EventRecord, Phase};
 	use sp_core::H256;
 	use sp_runtime::{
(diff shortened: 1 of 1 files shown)


bridges/modules/dispatch/src/lib.rs
```diff
@@ -599,9 +599,11 @@ mod tests {
 	fn should_fail_on_weight_mismatch() {
 		new_test_ext().execute_with(|| {
 			let id = [0; 4];
-			let mut message =
-				prepare_root_message(Call::System(<frame_system::Call<TestRuntime>>::remark(vec![1, 2, 3])));
+			let call = Call::System(<frame_system::Call<TestRuntime>>::remark(vec![1, 2, 3]));
+			let call_weight = call.get_dispatch_info().weight;
+			let mut message = prepare_root_message(call);
 			message.weight = 7;
+			assert!(call_weight != 7, "needed for test to be correct");
```
HCastano, line 606: Can you explain the comment here a little more?

HCastano (approved):
Can you also open a PR in the `parity-bridges-common` repo with this change?

bridges/modules/dispatch/src/lib.rs
thiolliere: the test is testing for a weight mismatch, so we must not match the correct weight

thiolliere:
> `parity-bridges-common`

yes done https://github.com/paritytech/parity-bridges-common/pull/1067

tomusdrw:
bot merge

PR #3511, opened 2021-07-22
technical committee is using the weight of council, but should have its own generated weight instead

thiolliere:
currently technical committee is using the weight of council, but should have its own generated weight instead

when I read https://substrate.dev/rustdocs/latest/frame_benchmarking/macro.add_benchmark.html I see that in case of instances the path is the concatenation of both pallet path and pallet name

cc @shawntabrizi

* [x] requires https://github.com/paritytech/substrate/pull/9626

### TODO

Rerun benchmarks:

* [x] benchmark runtime polkadot pallet_collective
* [x] benchmark runtime kusama pallet_collective


shawntabrizi (approved):

thiolliere:
the bot generates an implementation of the trait which is incomplete, I need to look into it

thiolliere:
actually I think we should simply use unique names, the doc given on add_benchmark is confusing to me as I don't understand in which context does the output file name get the suffix. And I think having multiple benchmark with the same name is making this weird result.

Instead I know use unique names for each pallets.

thiolliere:
/benchmark runtime polkadot pallet_collective_technical_committee

thiolliere:
/benchmark runtime polkadot pallet_collective_technical_committee

thiolliere:
ah but this still doesn't work, because it writes: 
```
impl<T: frame_system::Config> pallet_collective_council::WeightInfo for WeightInfo<T> {
```
where I want
```
impl<T: frame_system::Config> pallet_collective::WeightInfo for WeightInfo<T> {
```

Maybe we should give an optional name to add_benchmark/list_benchmark for giving a unique name for the benchmark which is different from the pallet name. :-/

thiolliere:
this should be now good to go

shawntabrizi (approved):

shawntabrizi:
thanks for figuring out the issues with the pipeline

thiolliere:
bot merge

apopiak (approved):

thiolliere:
bot merge

PR #3515, opened 2021-07-23
[CI] Add manual Gitlab job for running all benchmarks

s3krit:
This change adds a Gitlab CI job for running *all* benchmarks for westend, kusama and polkadot. 
Three new benchmark machines got provisioned [as requested](https://github.com/paritytech/devops/issues/989) and I've [added](https://gitlab.parity.io/parity/devops/-/merge_requests/176) them to Gitlab as runners, restricted to *only* running the jobs added in this PR.

I also added instructions for how to use these new jobs at the point of initiating a release - Currently, it will actually just generate a git patch file that we can manually apply and create a commit out of. This was done to avoid getting into the mess of having CI create commits and pull requests, though this could definitely be worked on and expanded.
This is more like an MVP to just get *something* automated for running all our benchmarks to update weights prior to a release.

WIP while I wait for the job to complete and can actually verify the patch files work and the weights aren't ridiculous due to something unforeseen.

### DIFF
--- a/.github/ISSUE_TEMPLATE/release.md
+++ b/.github/ISSUE_TEMPLATE/release.md
@@ -134,8 +134,22 @@ date to include them.
 
 ### Benchmarks
 
-Run the benchmarking suite with the new runtime and update any function weights
-if necessary.
+There are three benchmarking machines reserved for updating the weights at
+release-time. To initialise a benchmark run for each production runtime
+(westend, kusama, polkadot):
+* Go to https://gitlab.parity.io/parity/polkadot/-/pipelines?page=1&scope=branches&ref=master
+* Click the link to the last pipeline run for master
+* Start each of the manual jobs:
+  * 'update_westend_weights'
+  * 'update_polkadot_weights'
+  * 'update_kusama_weights'
+* When these jobs have completed (it takes a few hours), a git PATCH file will
+    be available to download as an artifact. 
+* On your local machine, branch off master
+* Download the patch file and apply it to your branch with `git patch patchfile.patch`
+* Commit the changes to your branch and submit a PR against master
+* The weights should be (Currently manually) checked to make sure there are no
+    big outliers (i.e., twice or half the weight).
 
 ### Polkadot JS
 
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -335,6 +335,16 @@ publish-adder-collator-image:
 update_polkadot_weights:
   <<:                              *update-weights
 
+update_kusama_weights:
+  <<:                              *update-weights
+  variables:
+    RUNTIME:                       kusama
+
+update_westend_weights:
+  <<:                              *update-weights
+  variables:
+    RUNTIME:                       westend
+
 #### stage:                        publish
 
 publish-s3-release:                &publish-s3
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -125,7 +125,6 @@ test-build-linux-stable:
     - if: $CI_COMMIT_REF_NAME =~ /^v[0-9]+\.[0-9]+.*$/              # i.e. v1.0, v2.1rc1
     - if: $CI_COMMIT_REF_NAME =~ /^[0-9]+$/                         # PRs
     - if: $CI_COMMIT_REF_NAME == "rococo-v1"
-    - if: $CI_COMMIT_REF_NAME == "mp-weights-CI"
   script:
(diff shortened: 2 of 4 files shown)


.gitlab-ci.yml
```diff
@@ -318,6 +319,33 @@ publish-adder-collator-image:
...
+.update_weights:                   &update-weights
+  stage:                           test
+  when:                            manual
+  tags:
+    - weights
+  variables:
+    RUNTIME:                       polkadot
+  artifacts:
+    paths:
+      - ${RUNTIMES}_weights_${CI_COMMIT_SHORT_SHA}.patch
```
chevdor, line 331: Is it really `RUNTIMES` and not `RUNTIME` ?

chevdor (changes requested):
LGTM beside the minor line breaking and the question about `RUNTIMES` vs `RUNTIME`.

scripts/run_benches_for_runtime.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash
+
+# Runs all benchmarks for all pallets, for a given runtime, provided by $1
+# Should be run on a reference machine to gain accurate benchmarks
+# current reference machine: https://github.com/paritytech/substrate/pull/5848
+
+runtime="$1"
+standard_args="--release --locked --features=runtime-benchmarks"
```
chevdor, line 8: we can add `--execution=wasm --wasm-execution=compiled` here

chevdor: `--chain "${runtime}-dev"` as well

chevdor: My bad, the 2 comments above would be a second variable.

chevdor: So the first variable is `cargo_args` and the second variable could be `benchmark_args`

chevdor (approved):
LGTM

.github/ISSUE_TEMPLATE/release.md
```diff
@@ -134,8 +134,22 @@ date to include them.
...
+release-time. To initialise a benchmark run for each production runtime
+(westend, kusama, polkadot):
+* Go to https://gitlab.parity.io/parity/polkadot/-/pipelines?page=1&scope=branches&ref=master
+* Click the link to the last pipeline run for master
+* Start each of the manual jobs:
+  * 'update_westend_weights'
+  * 'update_polkadot_weights'
+  * 'update_kusama_weights'
+* When these jobs have completed (it takes a few hours), a git PATCH file will
+    be available to download as an artifact.
```
TriplEight, line 147: we could safely use a separate master-tracking branch which would only be committed to by CI system.

.gitlab-ci.yml
```diff
@@ -125,6 +125,7 @@ test-build-linux-stable:
     - if: $CI_COMMIT_REF_NAME =~ /^v[0-9]+\.[0-9]+.*$/              # i.e. v1.0, v2.1rc1
     - if: $CI_COMMIT_REF_NAME =~ /^[0-9]+$/                         # PRs
     - if: $CI_COMMIT_REF_NAME == "rococo-v1"
+    - if: $CI_COMMIT_REF_NAME == "mp-weights-CI"
```
TriplEight, line 128: this one is a debug line, right?

.gitlab-ci.yml
```diff
@@ -318,6 +319,35 @@ publish-adder-collator-image:
...
       dotenv: ./artifacts/collator.env
 
+.update_weights:                   &update-weights
+  stage:                           test
+  when:                            manual
+  tags:
+    - weights
+  variables:
+    RUNTIME:                       polkadot
+    # sccache causing issues...
```
TriplEight, line 329: what's wrong with it? Is it (and redis) installed to those hosts?

TriplEight (commented):
How bad is the situation with caching here?

.github/ISSUE_TEMPLATE/release.md
s3krit: I thought about that and sure, going forwards this is an improvement I will add. But for now this just gets *a* process in place which we can iterate on.

.gitlab-ci.yml
s3krit: yeah I think it was due to redis. Should be resolved so I can remove this line now

TriplEight (approved):
LGTM, maybe a couple empty lines would help in `scripts/run_benches_for_runtime.sh`

PR #3516, opened 2021-07-23
Parachains configuration.rs FrameV2

ferrell-code:
relates: #2882 

So users of the `Configuration` pallet must be careful about the name they used in `construct_runtime!.` Hence the runtime-migration label, which might not be needed depending on the configuration of the `Configuration` pallet.

kusama, westend and rococo uses `ParachainsConfiguration` in `construct_runtime!`,  so I migrated them to `Configuration`

polkadot does not use `Configuration`.


runtime/parachains/src/configuration.rs
```diff
@@ -258,37 +255,68 @@ impl<BlockNumber: Zero> HostConfiguration<BlockNumber> {
...
+		}
+	}
+
+	#[pallet::genesis_build]
+	impl<T: Config> GenesisBuild<T> for GenesisConfig<T> {
+		fn build(&self) {
+			ActiveConfig::<T>::put(&self.config);
+			self.config.check_consistency();
+		}
+	}
```
ferrell-code, line 313: I Believe that is equivalent to the old macro... but thought it would be could to highlight

shawntabrizi: Yes, it looks right, but you could add a unit test to confirm if you like :)

shawntabrizi: Also the order of these lines should be swapped. First we check consistency, then we put the value into storage.

shouldn't practically matter since ultimately it results in a panic, but reading the code doesnt make sense

shawntabrizi (approved):

runtime/parachains/src/configuration.rs
ferrell-code: ah good point doesn't do much checking before we put the config storage lol

runtime/parachains/src/configuration.rs
```diff
@@ -258,37 +255,68 @@ impl<BlockNumber: Zero> HostConfiguration<BlockNumber> {
...
-}
 
-decl_module! {
-	/// The parachains configuration module.
-	pub struct Module<T: Config> for enum Call where origin: <T as frame_system::Config>::Origin {
-		type Error = Error<T>;
+	/// The active configuration for the current session.
+	#[pallet::storage]
+	#[pallet::getter(fn config)]
+	pub type ActiveConfig<T: Config> = StorageValue<
```
KiChjang, line 278:
```suggestion
	pub(crate) type ActiveConfig<T: Config> = StorageValue<
```

runtime/parachains/src/configuration.rs
```diff
@@ -258,37 +255,68 @@ impl<BlockNumber: Zero> HostConfiguration<BlockNumber> {
...
+		_,
+		HostConfiguration<T::BlockNumber>,
+		ValueQuery
+	>;
+
+	/// Pending configuration (if any) for the next session.
+	#[pallet::storage]
+	pub type PendingConfig<T: Config> = StorageMap<
```
KiChjang, line 286:
```suggestion
	pub(crate) type PendingConfig<T: Config> = StorageMap<
```

KiChjang (approved):
Great work as always!

KiChjang:
bot merge

PR #3517, opened 2021-07-24
Companion for Generate storage info for pallet authority_discovery #9428

georgesdib:
companion to parity substrate #9428


runtime/kusama/src/lib.rs
```diff
@@ -735,7 +736,9 @@ impl pallet_offences::Config for Runtime {
 	type OnOffenceHandler = Staking;
 }
 
-impl pallet_authority_discovery::Config for Runtime {}
+impl pallet_authority_discovery::Config for Runtime {
+	type MaxAuthorities = MaxAuthorities;
```
shawntabrizi, line 740: not sure if this actually works this simple. Kusama currently has just under 1,000 authorities, and probably we want to pick a number which allows kusama to scale significantly.

KiChjang (approved):

PR #3526, opened 2021-07-27
Companion for Substrate#9165

bkchr:
https://github.com/paritytech/substrate/pull/9165


runtime/kusama/src/lib.rs
```diff
@@ -1525,11 +1525,22 @@ pub type Executive = frame_executive::Executive<
...
 >;
 /// The payload being signed in the transactions.
 pub type SignedPayload = generic::SignedPayload<Call, SignedExtra>;
 
+/// Migrate from `PalletVersion` to the new `StorageVersion`
+pub struct MigratePalletVersionToStorageVersion;
+
+impl OnRuntimeUpgrade for MigratePalletVersionToStorageVersion {
+	fn on_runtime_upgrade() -> frame_support::weights::Weight {
+		frame_support::migrations::migrate_from_pallet_version_to_storage_version::<AllPalletsWithSystem>(
```
thiolliere, line 1538: this will write the new version to the storage before executing the pallet specific hooks,
If some hooks rely on the stored version, this upgrade can break them.
I'll double check now, but I guess we should double check before the next releases as well.

thiolliere (approved):
looks good to me, about https://github.com/paritytech/polkadot/pull/3526/files#r677620613 the only runtime upgrade that looks into the stored version is pallet-contract but we don't use it here.

AurevoirXavier:
Report label `E1-runtimemigration` missing.

We miss this migration...

PR #3528, opened 2021-07-27
Parachains ump.rs to FrameV2

ferrell-code:
relates: #2882 

So users of the `Ump` pallet must be careful about the name they used in `construct_runtime!.` Hence the runtime-migration label, which might not be needed depending on the configuration of the `Ump` pallet.

kusama and westend use `ParasUmp` in `construct_runtime!`,  so I migrated them to `Ump`

rococo already uses `Ump`

polkadot does not use `Ump`.


emostov (approved):
Thanks for the contribution!

shawntabrizi (approved):

shawntabrizi:
bot merge

PR #3531, opened 2021-07-29
Parachains para_inherent.rs to FrameV2

ferrell-code:
relates: #2882 

So users of the `ParaInherent` pallet must be careful about the name they used in `construct_runtime!.` Hence the runtime-migration label, which might not be needed depending on the configuration of the `ParaInherent` pallet.

kusama, rococo and westend use `ParasInherent` in `construct_runtime!`,  so I migrated them to `ParaInherent` 

polkadot does not use `ParaInherent`.


runtime/parachains/src/disputes.rs
```diff
@@ -1152,6 +1152,9 @@ mod tests {
 		while System::block_number() < to {
 			let b = System::block_number();
 			if b != 0 {
+				// circumvent requirement to have bitfields and headers in block for testing purposes
+				crate::paras_inherent::Included::<Test>::set(Some(()));
+
 				AllPallets::on_finalize(b);
```
ferrell-code, line 1158: The addition of ParaInherent in the mock `construct_runtime!` causes the ParaInherent `on_finalize` hook to now fire and will panic as no paras inherent have been entered. I believe this to be reasonable for the disputes tests but always good to double check

shawntabrizi (approved):

runtime/parachains/src/disputes.rs
shawntabrizi: looks fine to me

runtime/parachains/src/paras_inherent.rs
```diff
@@ -73,30 +69,89 @@ decl_error! {
...
+							disputes: Vec::new(),
+							parent_header: inherent_data.parent_header,
+						}
+					},
+				};
+
+			Some(Call::enter(inherent_data))
+		}
+
+		fn is_inherent(call: &Self::Call) -> bool {
```
KiChjang, line 141: This function seems to be unused, I think we can remove it?

KiChjang (approved):

KiChjang:
bot merge

PR #3534, opened 2021-07-30
XCM benchmark pallet

kianenigma:


xcm/src/v0/mod.rs
```diff
@@ -277,32 +291,40 @@ impl<Call> TryFrom<VersionedXcm<Call>> for Xcm<Call> {
 }
 
 impl<Call> Xcm<Call> {
-	pub fn into<C>(self) -> Xcm<C> { Xcm::from(self) }
+	pub fn into<C>(self) -> Xcm<C> {
+		Xcm::from(self)
+	}
 	pub fn from<C>(xcm: Xcm<C>) -> Self {
```
KiChjang, line 297: We should make use of the `From` trait so that we can take advantage of its reflexive property and obviate the need to have an `into` method.

KiChjang: Nevermind, this is impossible since the type parameters of the `Xcm` argument is not the same as `self`.

KiChjang: jk, requires specialization

PR #3538, opened 2021-08-01
XCM simulator

shaunxw:
Compared to the original [xcm-simulator](https://github.com/shaunxw/xcm-simulator), all cumulus dependencies are removed. The new impl uses `XcmpMessageHandler` and `DmpMessageHandler` abstractions, instead of directly calling cumulus pallets to handle xcmp and dmp. Users need to pass these implementations when declaring test parachain.

In order to make the new impl work, the following definitions are moved to polkadot from cumulus(will need a cumulus companion PR):
- `XcmpMessageHandler` trait
- `DmpMessageHandler` trait
- `XcmpMessageFormat` enum


xcm/xcm-simulator/src/lib.rs
```diff
@@ -0,0 +1,244 @@
...
+
+			let encoded = VersionedXcm::<()>::from(message).encode();
+			outbound.extend_from_slice(&encoded[..]);
+			outbound
+		},
+	}
+}
+
+#[macro_export]
+macro_rules! decl_test_relay_chain {
```
apopiak, line 65: It would be great if you could add some doc comments for these macros.

apopiak (approved):
Awesome contribution to allow for integration testing :+1:

shawntabrizi (approved):

xcm/xcm-simulator/src/lib.rs
shawntabrizi: lets follow up on that :)

https://github.com/paritytech/polkadot/issues/3542

shawntabrizi:
bot merge

jhonalino:
such a beautiful piece of code

PR #3539, opened 2021-08-01
Add unit tests to pallet-xcm

4meta5:
Adds unit tests for all dispatchables in `pallet-xcm`. It also tests a custom `XcmRouter` which blocks sending messages of a specific format.


4meta5:
@apopiak please review when you have a free moment

apopiak (approved):

xcm/pallet-xcm/src/mock.rs
```diff
@@ -0,0 +1,218 @@
...
+pub struct TestSendXcm;
+impl SendXcm for TestSendXcm {
+	fn send_xcm(dest: MultiLocation, msg: Xcm) -> XcmResult {
+		SENT_XCM.with(|q| q.borrow_mut().push((dest, msg)));
+		Ok(())
+	}
+}
+/// Sender that returns error if `X3` junction and stops routing
+pub struct TestSendXcmErr;
+impl SendXcm for TestSendXcmErr {
```
shawntabrizi, line 76:
```suggestion
pub struct TestSendXcmErrX3;
impl SendXcm for TestSendXcmErrX3 {
```

suggested clarity

xcm/pallet-xcm/src/mock.rs
```diff
@@ -0,0 +1,218 @@
...
+	type XcmTeleportFilter = All<(MultiLocation, Vec<MultiAsset>)>;
+	type XcmReserveTransferFilter = All<(MultiLocation, Vec<MultiAsset>)>;
+	type Weigher = FixedWeightBounds<BaseXcmWeight, Call>;
+}
+
+impl origin::Config for Test {}
+
+pub const ALICE: AccountId = AccountId::new([0u8; 32]);
+pub const PARA_ID: u32 = 2000;
+pub const INITIAL_BALANCE: u128 = 100_000_000_000;
```
shawntabrizi, line 195: probably weight, unit weight cost, and initial balance should probably all be close to one another and dependent on one another?

xcm/pallet-xcm/src/tests.rs
```diff
@@ -0,0 +1,180 @@
...
+use crate::mock::*;
+use frame_support::{assert_noop, assert_ok, traits::Currency};
+use polkadot_parachain::primitives::{AccountIdConversion, Id as ParaId};
+use xcm::{
+	opaque::v0::prelude::*,
+	v0::{Junction, Xcm},
+};
+
+#[test]
+fn send_reserve_asset_deposit_works() {
```
shawntabrizi, line 26: should just be called `send_works`. I dont see anything specific about reserve asset deposit happening here

xcm/pallet-xcm/src/tests.rs
```diff
@@ -0,0 +1,180 @@
...
+	let balances =
+		vec![(ALICE, INITIAL_BALANCE), (ParaId::from(PARA_ID).into_account(), INITIAL_BALANCE)];
+	new_test_ext_with_balances(balances).execute_with(|| {
+		let amount = 10 * ExistentialDeposit::get();
+		let weight = 2 * BaseXcmWeight::get();
+		assert_eq!(Balances::total_balance(&ALICE), INITIAL_BALANCE);
+		assert_ok!(XcmPallet::teleport_assets(
+			Origin::signed(ALICE),
+			RelayLocation::get(),
+			RelayLocation::get(),
```
shawntabrizi, line 94: Why wouldnt this at least send to another user on the chain and you can confirm the balance is sent without fees?

xcm/pallet-xcm/src/tests.rs
```diff
@@ -0,0 +1,180 @@
...
+			Order::DepositAsset { assets: vec![All], dest: Parachain(PARA_ID).into() },
+		];
+		assert_eq!(Balances::total_balance(&ALICE), INITIAL_BALANCE);
+		assert_ok!(XcmPallet::execute(
+			Origin::signed(ALICE),
+			Box::new(Xcm::WithdrawAsset {
+				assets: vec![ConcreteFungible { id: MultiLocation::Null, amount }],
+				effects: vec![
+					buy_execution(weight),
+					Order::InitiateTeleport {
```
shawntabrizi, line 165: If `Teleport` here is arbitrary, might as well test `Withdraw` -> `Deposit` and check that another user got the funds or something?

shawntabrizi (approved):
would like some of the improvements mentioned below

xcm/pallet-xcm/src/tests.rs
4meta5: We can verify the amount is sent, but we cannot verify the recipient balance increases (effects on destination chain) because the context of these tests is limited to the sending chain.

shawntabrizi:
bot merge

PR #3543, opened 2021-08-01
Do not expire HRMP open channel requests

### LINKED ISSUE #3451: Get rid of timeouts for open channel requests
We had a discussion with @shawntabrizi where he suggested that timeouts for pending open channel requests are superfluous.

First off, timing introduces unneeded churn. It makes the recipient to act quickly to accept the request. Alternatively, the sender and recipient may have to agree between each other to coordinate opening the channel so that it happens at the same time.

Then we also are doing some work on each session to check what pending requests are expired. We could avoid that.

While we don't have parathreads there is a very limited venue for spamming the requests given that they are required to point to existing para ids and no duplicates are permitted. Moreover, the spamming can be limited economically by the already existing deposit mechanism.

We may hold off with changing this for two reasons:

- we can increase the open timeout to be very long, so that it won't be bothersome anymore
- XCMP module can take this into account when implemented and start with non expiring requests from the start.

pepyakin:
Closes #3451


roadmap/implementers-guide/src/types/runtime.md
```diff
@@ -85,8 +85,10 @@ struct HostConfiguration {
 	/// decide to do with its PoV so this value in practice will be picked as a fraction of the PoV
 	/// size.
 	pub max_downward_message_size: u32,
-	/// Number of sessions after which an HRMP open channel request expires.
-	pub hrmp_open_request_ttl: u32,
+	/// NOTE: this field is deprecated. Channel open requests became non-expiring. Changing this value
+	/// doesn't have any effect. This field doesn't have an `deprecated` attribute because that would
```
shawntabrizi, line 89:
```suggestion
	/// doesn't have any effect. This field doesn't have a `deprecated` attribute because that would
```

runtime/parachains/src/configuration.rs
```diff
@@ -92,8 +91,10 @@ pub struct HostConfiguration<BlockNumber> {
 	pub hrmp_max_parachain_outbound_channels: u32,
 	/// The maximum number of outbound HRMP channels a parathread is allowed to open.
 	pub hrmp_max_parathread_outbound_channels: u32,
-	/// Number of sessions after which an HRMP open channel request expires.
-	pub hrmp_open_request_ttl: u32,
+	/// NOTE: this field is deprecated. Channel open requests became non-expiring. Changing this value
+	/// doesn't have any effect. This field doesn't have an `deprecated` attribute because that would
```
shawntabrizi, line 95:
```suggestion
	/// doesn't have any effect. This field doesn't have a `deprecated` attribute because that would
```

runtime/parachains/src/hrmp.rs
```diff
@@ -430,6 +440,22 @@ pub mod pallet {
 			Ok(())
 		}
 
+		/// This cancels a pending open channel request. It can be canceled be either of the sender
+		/// or the recipient for that request. The origin must be either of those.
+		///
+		/// The cancelling happens immediatelly. It is not possible to cancel the request if it is
+		/// already accepted.
```
4meta5, line 447: If request is already accepted, why not interpret the `cancel_request` as `close_channel`?

Otherwise most callers of cancel open request will wait until this Errs and then call `close_channel`

pepyakin: Correction: we can't issue `close_channel` for a channel that is not yet opened. Nonetheless, I got your idea: why not just permit the cancelling a confirmed open request?

Well, this is just what @shawntabrizi and me decided when we spoke about it. I thought it is a tad bit simpler to  implement this way. Also is not that important in the end, since this is a rather an edge case.

WDYT, @shawntabrizi ?

roadmap/implementers-guide/src/runtime/hrmp.md
```diff
@@ -233,15 +238,12 @@ the parachain executed the message.
     1. Remove all outbound channels of `P`, i.e. `(P, _)`,
     1. Remove `HrmpOpenChannelRequestCount` for `P`
     1. Remove `HrmpAcceptedChannelRequestCount` for `P`.
+    1. Remove `HrmpOpenChannelRequests` and `HrmpOpenChannelRequestsList` for `(P, _)` and `(_, P)`.
+        1. For each removed channel request `C`:
+            1. Unreserve the sender's deposit if the sender is not present in `outgoing_paras`
+            1. Unreserve the recipient`s deposit if `C` is confirmed and the recipient is not present in `outgoing_paras`
```
4meta5, line 244:
```suggestion
            1. Unreserve the recipient's deposit if `C` is confirmed and the recipient is not present in `outgoing_paras`
```

runtime/parachains/src/hrmp.rs
```diff
@@ -1167,6 +1237,35 @@ impl<T: Config> Pallet<T> {
 	}
 }
 
+impl<T: Config> Pallet<T> {
+	/// Decreases the open channel request count for the given sender. If the value reaches zero
+	/// if is removed completely.
```
4meta5, line 1242:
```suggestion
	/// it is removed completely.
```

runtime/parachains/src/hrmp.rs
```diff
@@ -1167,6 +1237,35 @@ impl<T: Config> Pallet<T> {
...
+	/// Decreases the open channel request count for the given sender. If the value reaches zero
+	/// it is removed completely.
+	fn decrease_open_channel_request_count(sender: ParaId) {
+		let new_open_channel_req_cnt =
+			<Self as Store>::HrmpOpenChannelRequestCount::get(&sender).saturating_sub(1);
+		if new_open_channel_req_cnt != 0 {
+			<Self as Store>::HrmpOpenChannelRequestCount::insert(&sender, new_open_channel_req_cnt);
+		} else {
+			<Self as Store>::HrmpOpenChannelRequestCount::remove(&sender);
+		}
```
shawntabrizi, line 1250: this could be done a bit more cleanly with a mutate_exists function i think.

shawntabrizi (approved):
lgtm

shawntabrizi:
how can we eventually remove these deprecated fields?

runtime/parachains/src/hrmp.rs
pepyakin: Thanks for the suggestion!

pepyakin:
> how can we eventually remove these deprecated fields?

Unfortunately, I don't have a good answer. I think once we get enough of those so that they start to bother us we can run a migration. Alternatively, we may wait up to the next time we have to do migration and then we can remove all the deprecated fields along the way.

PR #3537, opened 2021-07-31
add integration tests to xcm-builder

apopiak:
This PR adds some integration tests aimed to simulate the relay-chain part for some common scenarios:
+ `withdraw` and `deposit`
+ using `QueryHolding` to get a notification of success
+ teleport to common good parachain
+ reserve based transfer to community parachain

PR #3546, opened 2021-08-01
Improve xcm-simulator

4meta5:
Adds tests from #3537 to xcm-simulator (except teleport_assets)

Also makes message processing asynchronous and pulls it out from the test externalities environment to allow for the processing of messages that go back to the origin chain. The previous implementation ran into borrowing issues in the underlying `RefCell`s when sending messages back to the origin.
Messages are processed after the closure passed to `test_ext.execute_with()`.


4meta5:
@apopiak ready for your review

@shaunxw thanks for making the simulator, let me know if I can use it better

xcm/xcm-simulator/example/src/relay_chain.rs
```diff
@@ -140,7 +158,7 @@ pub type LocalOriginToLocation = SignedToAccountId32<Origin, AccountId, KusamaNe
 impl pallet_xcm::Config for Runtime {
 	type Event = Event;
 	type SendXcmOrigin = xcm_builder::EnsureXcmOrigin<Origin, LocalOriginToLocation>;
-	type XcmRouter = XcmRouter;
+	type XcmRouter = (XcmRouter, TestSendXcm);
```
shaunxw, line 161: `TestSendXcm` might not be needed anymore, as `XcmRouter` always sends messages out.

4meta5: We need something in the test to track messages sent from the relay chain. I thought adding the `TestSendXcm` impl would work, but I see now that it does not. It short circuits on the first successful message sent.

This means I must change `XcmRouter` impl for the relay chain to store all messages sent.

4meta5: Instead of adding state to the macro, I think I will just change the `MockMsgQueue` pallet to store all received messages.

xcm/xcm-simulator/example/src/lib.rs
```diff
@@ -202,14 +216,165 @@ mod tests {
...
+			let query_id = 1234;
+			let message = WithdrawAsset {
+				assets: vec![ConcreteFungible { id: Null, amount }],
+				effects: vec![
+					buy_execution(weight),
+					Order::DepositAsset {
+						assets: vec![All],
+						dest: Parachain(2).into(), // valid destination
+					},
+					Order::QueryHolding { query_id, dest: Parachain(2).into(), assets: vec![All] },
```
4meta5, line 350: The purpose of this test is to check that if all the previous messages execute successfully, a `Order::QueryResponse` will be sent to this `dest`.

The `dest` should be `Parachain(1)` instead of `Parachain(2)`, but when I make the change shown below, there is an `BorrowMut` error because it triggers `execute_with` inside of another `execute_with`
```suggestion
					Order::QueryHolding { query_id, dest: Parachain(1).into(), assets: vec![All] },
```
The idea of a solution is to make message execution less automatic so that SEND, RECEIVE, and PROCESS are separate steps.

apopiak: @shaunxw This is the issue I mentioned in chat.

shaunxw: As discussed in the call, the less automatic approach sounds useful in some scenarios.

I'm just thinking that, as in most unit tests here and in orml-xtokens, we're expecting message passing in an automatic way. Probably we can make it automatic by default, while providing a flag in the mock network decl macro to allow users to make it manual. This could keep APIs still simple to use.

4meta5: #3552

xcm/xcm-simulator/src/lib.rs
```diff
@@ -196,44 +210,47 @@ macro_rules! decl_test_network {
 			}
 		}
 
+		fn process_messages() -> $crate::XcmResult {
```
apopiak, line 213:
```suggestion

		/// Process any XCMs in the message buses.
		///
		/// Called automatically by `TestExt::execute_with`.
		fn process_messages() -> $crate::XcmResult {
```

xcm/xcm-simulator/src/lib.rs
```diff
@@ -196,44 +210,47 @@ macro_rules! decl_test_network {
...
+			process_para_messages()
+		}
+
 		/// XCM router for parachain.
 		pub struct ParachainXcmRouter<T>($crate::PhantomData<T>);
 
-		impl<T: $crate::Get<$crate::ParaId>> ParachainXcmRouter<T> {
-			fn process_messages() -> $crate::XcmResult {
-				use $crate::{UmpSink, XcmpMessageHandlerT};
+		fn process_para_messages() -> $crate::XcmResult {
```
apopiak, line 221:
```suggestion
		/// Process all messages originating from parachains.
		fn process_para_messages() -> $crate::XcmResult {
```

xcm/xcm-simulator/src/lib.rs
```diff
@@ -261,29 +278,27 @@ macro_rules! decl_test_network {
 		/// XCM router for relay chain.
 		pub struct RelayChainXcmRouter;
 
-		impl RelayChainXcmRouter {
-			fn process_messages() -> $crate::XcmResult {
-				use $crate::DmpMessageHandlerT;
+		fn process_relay_messages() -> $crate::XcmResult {
```
apopiak, line 281:
```suggestion
		/// Process all messages originating from the relay chain.
		fn process_relay_messages() -> $crate::XcmResult {
```

xcm/xcm-simulator/example/src/lib.rs
```diff
@@ -190,25 +200,191 @@ mod tests {
...
+
+		let send_amount = 10;
+		let mut amount_spent = 0;
+		let mut amount_received = 0;
+		let weight_for_execution = 3 * relay_chain::BaseXcmWeight::get();
+		let query_id_set = 1234;
+
+		// First send a message which fails on the relay chain
+		ParaA::execute_with(|| {
+			//let para_account_b: relay_chain::AccountId = ParaId::from(2).into_account();
```
4meta5, line 310:
```suggestion
```

xcm/xcm-simulator/src/lib.rs
```diff
@@ -161,13 +169,28 @@ macro_rules! __impl_ext {
...
+			fn execute_with<R>(execute: impl FnOnce() -> R) -> R {
+				let result = $ext_name.with(|v| v.borrow_mut().execute_with(execute));
+				process_messages().expect("message processing failure");
+				result
+			}
 		}
 	};
 }
 
+use sp_std::collections::vec_deque::VecDeque;
```
4meta5, line 185: move to top of file

xcm/xcm-simulator/src/lib.rs
```diff
@@ -161,13 +168,26 @@ macro_rules! __impl_ext {
...
 			}
 
-			fn execute_with<R>(execute: impl FnOnce() -> R) -> R {
+			fn execute_without_auto_dispatch<R>(execute: impl FnOnce() -> R) -> R {
 				$ext_name.with(|v| v.borrow_mut().execute_with(execute))
 			}
+
+			fn execute_with<R>(execute: impl FnOnce() -> R) -> R {
+				let result = $ext_name.with(|v| v.borrow_mut().execute_with(execute));
+				process_messages().expect("message processing failure");
```
shaunxw, line 177: Including returned `XcmResult` in `expect` str would be helpful to give more details on panic.

xcm/xcm-simulator/src/lib.rs
```diff
@@ -161,13 +168,26 @@ macro_rules! __impl_ext {
...
+			}
 		}
 	};
 }
 
+thread_local! {
+	pub static PARA_MESSAGE_BUS: RefCell<VecDeque<(ParaId, MultiLocation, Xcm<()>)>>
+		= RefCell::new(VecDeque::new());
+	pub static RELAY_MESSAGE_BUS: RefCell<VecDeque<(MultiLocation, Xcm<()>)>>
+		= RefCell::new(VecDeque::new());
```
shaunxw, line 188: We should clean message buses in test network `reset` fn.

shaunxw (commented):
LGTM 👍

xcm/xcm-simulator/src/lib.rs
```diff
@@ -187,30 +207,67 @@ macro_rules! decl_test_network {
 			}
 		}
 
+		/// Process any XCMs in the message buses.
+		///
+		/// Called automatically by `TestExt::execute_with`.
+		fn process_messages() -> $crate::XcmResult {
+			process_relay_messages()?;
+			process_para_messages()
```
shaunxw, line 215: In the scenario where `ParaA` sends relay chain tokens to `ParaB`, there would be a para message firstly (ump), and then a relay message (dmp). This impl won't process the relay message in this case.

IMO the messages process should loop process until there is no more messages in *any* buses, like:

```rust
while has_messages_in_any_bus {
    process_relay_messages();
    process_para_messages();
}
```

xcm/xcm-simulator/example/src/lib.rs
```diff
@@ -182,33 +192,201 @@ mod tests {
...
 		});
 	}
 
+	/// Scenario:
+	/// A user Alice sends funds from the relaychain to a parachain.
+	///
+	/// Asserts that the correct XCM is sent and the balances are set as expected.
 	#[test]
-	fn reserve_transfer() {
+	fn reserve_transfer_assets() {
```
kianenigma, line 208: reserve_asset_transfer is correct, since `ReserveAsset` is basically one word, has a specific meaning and should not be separated.

xcm/xcm-simulator/example/src/lib.rs
```diff
@@ -182,33 +192,201 @@ mod tests {
...
+		ParaA::execute_and_dispatch_xcm(|| {
+			let message = WithdrawAsset {
+				assets: vec![ConcreteFungible { id: Null, amount: send_amount }],
+				effects: vec![
+					buy_execution(weight_for_execution),
+					Order::DepositAsset { assets: vec![All], dest: Parachain(2).into() },
+				],
+			};
+			// Send withdraw and deposit
+			assert_ok!(ParachainPalletXcm::send_xcm(Null, X1(Parent), message.clone(),));
```
kianenigma, line 277:
```suggestion
			assert_ok!(ParachainPalletXcm::send_xcm(Null, X1(Parent), message.clone()));
```

xcm/xcm-simulator/example/src/lib.rs
```diff
@@ -182,33 +192,201 @@ mod tests {
...
+		});
+
+		amount_received += send_amount;
```
kianenigma, line 280: this is superfluous in the case of 1 transfer. You initialize it, increment it by `send_amount` and then check.  This should go away and the last line would be: 

```
assert_eq!(relay_chain::Balances::free_balance(para_account_b), send_amount);
```

```suggestion
```

xcm/xcm-simulator/example/src/lib.rs
```diff
@@ -182,33 +192,201 @@ mod tests {
...
+	/// It sends a `QueryHolding` after the deposit to get notified on success.
+	///
+	/// Asserts that the balances are updated correctly and the expected XCM is sent.
+	#[test]
+	fn query_holding() {
+		MockNet::reset();
+
+		let send_amount = 10;
+		let mut amount_spent = 0;
+		let mut amount_received = 0;
```
kianenigma, line 304: same.

xcm/xcm-simulator/example/src/lib.rs
```diff
@@ -182,33 +192,201 @@ mod tests {
...
+		let weight_for_execution = 3 * relay_chain::BaseXcmWeight::get();
+		let query_id_set = 1234;
+
+		// First send a message which fails on the relay chain
+		ParaA::execute_and_dispatch_xcm(|| {
```
kianenigma, line 309: this failure is not part of the test dsecription.

xcm/xcm-simulator/example/src/lib.rs
```diff
@@ -182,33 +192,201 @@ mod tests {
...
+			};
+			// Send withdraw and deposit with query holding
+			assert_ok!(ParachainPalletXcm::send_xcm(Null, X1(Parent), message.clone(),));
+		});
+
+		amount_spent += send_amount;
+
+		// Check that no transfer was executed and no response message was sent
+		Relay::execute_and_dispatch_xcm(|| {
+			let para_account_a: relay_chain::AccountId = ParaId::from(1).into_account();
```
kianenigma, line 334: seems like `fn para_account_id(para_id: u32) -> AccountId` is used a lot.

kianenigma (approved):
LGTM except https://github.com/paritytech/polkadot/pull/3546/files#r681906407

xcm/xcm-simulator/example/src/lib.rs
4meta5: removed it, I agree

xcm/xcm-simulator/example/src/lib.rs
4meta5: The function tested from `pallet_xcm` is named `reserve_transfer_assets` https://github.com/paritytech/polkadot/blob/2882304560ef5b8f2811ddd6b055b78f5de44448/xcm/pallet-xcm/src/lib.rs#L195 

Should this test still be renamed?

xcm/xcm-simulator/src/lib.rs
```diff
@@ -32,13 +29,30 @@ pub use polkadot_runtime_parachains::{
...
+	/// Execute code in the context of the test externalities, without automatic
+	/// message processing. All messages in the message buses can be processed
+	/// by calling `Self::dispatch_xcm_queue()`.
+	fn execute_and_queue_xcm<R>(execute: impl FnOnce() -> R) -> R;
+	/// Process all messages in the message buses
+	fn dispatch_xcm_queue();
+	/// Execute some code in the context of the test externalities, with
+	/// automatic message processing.
+	/// Messages are dispatched once the passed closure completes.
+	fn execute_and_dispatch_xcm<R>(execute: impl FnOnce() -> R) -> R {
```
shaunxw, line 51: I would prefer the previous fn name `execute_with`(and `execute_with_no_dispatch`). In real life networks, messages are dispatched automatically after sending extrinsics, which is the same behavior with here, so there is no need to explicitly say it would dispatch xcm. Also, `execute_with` is same to `TestExternalities`'s API and it's intuitive to use.

4meta5: I agree. In general, the common path should have short names and the uncommon path can have longer names.

There was some discussion over the names so I changed them, but I prefer your suggestion (changing them back now).

4meta5:
Just merged #2815 into this branch, it'll be easier to review this PR again once that PR is merged into master

4meta5:
Ah I keep getting wrecked by auto-merge 😬 I'll open a new one

apopiak:
superseded by https://github.com/paritytech/polkadot/pull/3618

PR #3551, opened 2021-08-02
Change pallet naming in test-runtime

bkchr:
This is required to make the `well_know_keys::ACTIVE_CONFIG` match the
`ActiveConfig` key.

### DIFF
--- a/node/test/service/src/chain_spec.rs
+++ b/node/test/service/src/chain_spec.rs
@@ -174,7 +174,7 @@ fn polkadot_testnet_genesis(
 		},
 		vesting: runtime::VestingConfig { vesting: vec![] },
 		sudo: runtime::SudoConfig { key: root_key },
-		configuration: runtime::ParachainsConfigurationConfig {
+		configuration: runtime::ConfigurationConfig {
 			config: polkadot_runtime_parachains::configuration::HostConfiguration {
 				validation_upgrade_frequency: 10u32,
 				validation_upgrade_delay: 5,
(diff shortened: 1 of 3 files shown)


pepyakin:
Oh, I am not familiar with this `integrity_test`, I will look into the docs. But you are aware that there is a test for that in this module?

bkchr:
Yeah I know and that didn't catch this :P 

We changed this to `FRAME V2` and this means the prefix of the storage item is derived from the name of the pallet in the runtime. So, the prefix isn't static anymore.

The `integrity_test` basically boils down to some `test` that is generated in the runtime to check for certain stuff.

pepyakin:
Meaning that this replaces the test? If so, could you please remove it? Or is it still valuable?

bkchr:
It still tests if the stuff decodes and contains the expected values. I would still see this somewhat useful.

bkchr:
But I also can remove it.

bkchr:
(It still ensures that the type of `ActiveConfig` doesn't change)

shawntabrizi:
I talked to @pepyakin yesterday about this, and mentioned it would make sense to me for this to be a custom key from end to end, rather than a pallet storage item (which can be changed from frame version to frame version)

shawntabrizi (approved):

bkchr:
> I talked to @pepyakin yesterday about this, and mentioned it would make sense to me for this to be a custom key from end to end, rather than a pallet storage item (which can be changed from frame version to frame version)

I'm clearly not happy with the status, but it is the easiest solution for now. In the future we really need to have some better utility for parachains to make use of this. As outlined here https://github.com/paritytech/cumulus/issues/320#issuecomment-775188888 by @pepyakin

bkchr:
bot merge

PR #3553, opened 2021-08-02
Remove unused type

thiolliere:
the type QueueSize probably comes from an old configuration requirement, it is now not used and without doc, I don't see any reason not to remove it.

thanks @ascjones to find it.


kianenigma (approved):

ascjones (approved):

thiolliere:
bot merge

Issue #3463, opened 2021-07-13
Incentive-Alignment: Approval Voting


Lldenaurois:
**Problem Statement**

In the current protocol, validators have no incentive to complete the Approval Voting step and advance parachain block finality. This needn't necessarily present an attack surface since we assume that `2f` among `3f+1` validators are honest and do not deviate from the consensus protocol, but it could present a mis-alignment of incentives that causes some validators who run on less performant hardware to fall-behind without any incentives to upgrade their hardware. If multiple validators fall subject to this equilibrium, the entire network could incur significant finality lag.

**Overview**

Therefore, we present a simplified lottery scheme that enables validators to incur a benefit for their approval assignment and subsequent approval vote. One particular aspect of the scheme we present below is that validators will only incur an incentive if they are able to issue their approval vote in the `0-th` tranche. The reason for this is two-fold:

1.  By issuing incentives only for Approval Votes in the `0-th` tranche, we ensure that the incentive structure is heavily skewed towards preferring to complete Approval Voting as fast as possible, thereby incentivizing validators to run Approval Voting in the most efficient fashion.
2. The scheme we present does not apply to later tranches because in any tranche after the `0-th` tranche there exists an incentive to collude with other validators to present Approval Votes without the work having ever been computed honestly.

**Background**

`PendingMessage::Assignment` includes an `IndirectAssignmentCert`, which includes a relay_block hash, validator index and `AssignmentCert`.

In addition, the `AssignmentCert` includes both a `VRFOutput` and `VRFProof`.

When a validator receives a `PendingMessage::Approval`, this includes a signature on the `ApprovalVote(CandidateHash)` and the `SessionIndex` under their ValidatorKey.

Both `Assignments` and `Approvals` are gossip'd among all validators in the validator network via the `ApprovalDistribution` subsystem.

**Construction**

TL;DR: Validators keep track of the `VRFOutput` for all `PendingMessage::Assignment`s. When `PendingMessage::Approval`s are received over-the-wire for approvals in the `0-th` tranche, the receiving validator can sign the approving validator's `VRFOutput` with their own `VRFKey` to issue an `ApprovalTicket`. `ApprovalTickets` are valid extrinsics that will issue rewards on-chain for both the Voter and the Ticket issuer.

[...]

PR #3558, opened 2021-08-02
runtime/parachains: Keep track of included candidates and historical babe vrfs in `shared` module

### LINKED ISSUE #3469: Track included candidates and cores in the shared module
Part of #3463 .

We'll need the runtime to keep track of all included candidates as well as the core index they occupied when they became available as part of the approval lottery ticket checking procedure on-chain. The `shared` module is a natural candidate for this.

In fact, the disputes runtime PR (#2947) already included a mapping of included blocks based on candidate hash and session index, so that will become obsolete after the implementation of this issue is complete.

This issue will have 2 parts:

1. Add a mapping of all included candidates to the `shared` module, to be filled out as part of the `paras_inherent` processing pipeline. This should keep all candidates for recent sessions, and should be designed in such a way that pruning of ancient sessions' data is performed gradually over the course of many blocks. Attempting to delete all included candidates in one block will become a scaling bottleneck and will prove to be an issue.
2. Remove the `Included` mapping from the `Disputes` runtime module and change it to rely on the `shared` module's mapping instead.

### LINKED ISSUE #3470: Track all historical BABE VRFs in the `shared` module
Part of #3463 

The runtime needs to keep track of recent BABE VRFs in order to verify the `RelayVRFModulo` assignments that are used within the approval ticket system. All BABE VRFs, corresponding to block number, should be kept for recent sessions. Pruning of this mapping on session change should be spread out over the course of many blocks.

Lldenaurois:
This PR attempts to take the first step in adding incentive structures to the Approval Voting system, c.f. #3463

Closes #3469  

The goal is three-fold:

1. Keep track of all historical `CandidateHash`es
2. Maintain the `CandidateHash` -> `CoreIndex` mapping for some historical `CandidateHash`es
3. Prune `CandidateHash`es for a historical session over the course of an ongoing session

Specifically, we add a StorageDoubleMap to the shared module. This StorageDoubleMap maintains the (SessionIndex, CandidateHash) -> Coreindex mapping (Note that we also include the BlockNumber in the value of the Map).

The additional indirection of maintaining the Session allows to implement more coherent pruning logic.

Pruning is currently simplified in that we prune a fixed number of Candidates at a time. This is achieved by prefix-query for a particular SessionIndex in the DoubleMap, thereby pruning CandidateHashes over time until a SessionIndex is exhausted.

Closes #3470 

In addition, this PR also adds a StorageDoubleMap to the shared module which keeps track of historical Babe VRFs for the purpose of comparing these VRFs against the Approval Assignments that are issued throughout the Approval Voting process.

The pruning algorithm for Historical VRFs is modified from the above pruning logic for candidates. Within every invocation of the pruning function, we remove one block's worth of historical VRFs for a single pruneable session. However, if all Candidates for a specfic Session have been exhausted, then all Historical VRFs across all blocks in that session are pruned.

The parameter for how many candidates to prune per invocation of the pruning function should be more adaptive in order to ensure we prune all blocks (and historical VRFs) in a session approximately as quickly as we prune all candidates in a session.

### DIFF
--- a/runtime/parachains/src/disputes.rs
+++ b/runtime/parachains/src/disputes.rs
@@ -619,10 +619,12 @@ impl<T: Config> Pallet<T> {
 			for to_prune in to_prune {
 				// This should be small, as disputes are rare, so `None` is fine.
 				<Disputes<T>>::remove_prefix(to_prune, None);
-
-				// This is larger, and will be extracted to the `shared` module for more proper pruning.
-				// TODO: https://github.com/paritytech/polkadot/issues/3469
-				shared::Pallet::<T>::prune_included_candidates(to_prune);
+				// Mark the session as pruneable so that its candidates can be incrementally
+				// removed over the course of many block inclusions.
+				shared::Pallet::<T>::mark_session_pruneable(to_prune);
+				// TODO(ladi): remove this call, currently allows unit tests to pass. Need to
+				// figure out how to invoke paras_inherent::enter in run_to_block.
+				shared::Pallet::<T>::prune_ancient_sessions(shared::MAX_CANDIDATES_TO_PRUNE);
 				SpamSlots::<T>::remove(to_prune);
 			}
 
--- a/runtime/parachains/src/disputes.rs
+++ b/runtime/parachains/src/disputes.rs
@@ -17,10 +17,9 @@
 //! Runtime component for handling disputes of parachain candidates.
 
 use crate::{
-	shared,
 	configuration::{self, HostConfiguration},
 	initializer::SessionChangeNotification,
-	session_info,
+	session_info, shared,
 };
 use bitvec::{bitvec, order::Lsb0 as BitOrderLsb0};
 use frame_support::{ensure, traits::Get, weights::Weight};
@@ -554,10 +553,8 @@ impl<T: Config> Pallet<T> {
 				dispute.concluded_at = Some(now);
 				<Disputes<T>>::insert(session_index, candidate_hash, &dispute);
 
-				let is_local = <shared::Pallet<T>>::is_included_candidate(
-					&session_index,
-					&candidate_hash,
-				);
+				let is_local =
+					<shared::Pallet<T>>::is_included_candidate(&session_index, &candidate_hash);
 				if is_local {
 					// Local disputes don't count towards spam.
 
@@ -780,7 +777,8 @@ impl<T: Config> Pallet<T> {
 		};
 
 		// Apply spam slot changes. Bail early if too many occupied.
-		let is_local = <shared::Pallet<T>>::is_included_candidate(&set.session, &set.candidate_hash);
+		let is_local =
+			<shared::Pallet<T>>::is_included_candidate(&set.session, &set.candidate_hash);
 		if !is_local {
 			let mut spam_slots: Vec<u32> =
 				SpamSlots::<T>::get(&set.session).unwrap_or_else(|| vec![0; n_validators]);
@@ -912,10 +910,8 @@ impl<T: Config> Pallet<T> {
 		};
 
 		// Apply spam slot changes. Bail early if too many occupied.
-		let is_local = <shared::Pallet<T>>::is_included_candidate(
-			&set.session,
-			&set.candidate_hash,
-		);
+		let is_local =
+			<shared::Pallet<T>>::is_included_candidate(&set.session, &set.candidate_hash);
 		if !is_local {
 			let mut spam_slots: Vec<u32> =
 				SpamSlots::<T>::get(&set.session).unwrap_or_else(|| vec![0; n_validators]);
@@ -989,10 +985,9 @@ impl<T: Config> Pallet<T> {
 
 		// Freeze if just concluded against some local candidate
 		if summary.new_flags.contains(DisputeStateFlags::AGAINST_SUPERMAJORITY) {
-			if let Some((revert_to, _)) = <shared::Pallet<T>>::get_included_candidate(
-				&set.session,
-				&set.candidate_hash,
-			) {
+			if let Some((revert_to, _)) =
+				<shared::Pallet<T>>::get_included_candidate(&set.session, &set.candidate_hash)
+			{
 				Self::revert_and_freeze(revert_to);
 			}
 		}
@@ -1111,7 +1106,6 @@ fn check_signature(
 #[cfg(test)]
 mod tests {
 	use super::*;
-	use primitives::v1::CoreIndex;
 	use crate::mock::{
 		new_test_ext, AccountId, AllPallets, Initializer, MockGenesisConfig, System, Test,
 		PUNISH_VALIDATORS_AGAINST, PUNISH_VALIDATORS_FOR, PUNISH_VALIDATORS_INCONCLUSIVE,
@@ -1122,7 +1116,7 @@ mod tests {
 		traits::{OnFinalize, OnInitialize},
 	};
 	use frame_system::InitKind;
-	use primitives::v1::BlockNumber;
+	use primitives::v1::{BlockNumber, CoreIndex};
 	use sp_core::{crypto::CryptoType, Pair};
 
 	// All arguments for `initializer::on_new_session`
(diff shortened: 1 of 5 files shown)


runtime/parachains/src/disputes.rs
```diff
@@ -186,12 +186,12 @@ impl<T: Config> DisputesHandler<T::BlockNumber> for pallet::Pallet<T> {
 		pallet::Pallet::<T>::provide_multi_dispute_data(statement_sets)
 	}
 
-	fn note_included(
+	fn process_included(
 		session: SessionIndex,
 		candidate_hash: CandidateHash,
-		included_in: T::BlockNumber,
+		revert_to: T::BlockNumber,
```
rphmeier, line 192: it's not exactly revert_to, though, because I think we revert to `included_in - 1`

runtime/parachains/src/disputes.rs
```diff
@@ -632,7 +622,7 @@ impl<T: Config> Pallet<T> {
 
 				// This is larger, and will be extracted to the `shared` module for more proper pruning.
 				// TODO: https://github.com/paritytech/polkadot/issues/3469
-				<Included<T>>::remove_prefix(to_prune, None);
+				shared::Pallet::<T>::prune_included_candidates(to_prune);
```
rphmeier, line 625: The `last_pruned` update logic needs to be changed because pruning here may only be partial.

I suggest having the `shared` module own all the pruning logic anyway. It's weird to have `shared` pruning logic triggered by the disputes module.

runtime/parachains/src/shared.rs
```diff
@@ -114,6 +129,53 @@ impl<T: Config> Pallet<T> {
...
+	pub(crate) fn get_included_candidate(
+		session: &SessionIndex,
+		candidate_hash: &CandidateHash,
+	) -> Option<(T::BlockNumber, CoreIndex)> {
+		<IncludedCandidates<T>>::get(session, candidate_hash)
+	}
+
+	/// Prunes all candidates that were included in the `to_prune` session.
+	pub(crate) fn prune_included_candidates(to_prune: SessionIndex) {
+		<IncludedCandidates<T>>::remove_prefix(to_prune, None);
```
rphmeier, line 169: Specifically we're meant to avoid this `None`

runtime/parachains/src/disputes.rs
Lldenaurois: How about we rename it to `revert_tail`, since it's the tail of the chain that gets removed in reversion

kianenigma: not an expert thus can't give good terminology advice, but seems like `included_in` was a beteter name, when there's a clear document saying `we revert to included_in - 1 when ...`.

runtime/parachains/src/disputes.rs
```diff
@@ -565,7 +553,9 @@ impl<T: Config> Pallet<T> {
 				dispute.concluded_at = Some(now);
 				<Disputes<T>>::insert(session_index, candidate_hash, &dispute);
 
-				if <Included<T>>::contains_key(&session_index, &candidate_hash) {
+				let is_local =
```
kianenigma, line 556: nit: not clear why you want this variable.

runtime/parachains/src/disputes.rs
```diff
@@ -565,7 +553,9 @@ impl<T: Config> Pallet<T> {
...
+					<shared::Pallet<T>>::is_included_candidate(&session_index, &candidate_hash);
+				if is_local {
 					// Local disputes don't count towards spam.
 
 					weight += T::DbWeight::get().reads_writes(1, 1);
```
kianenigma, line 561: please use saturating math here.

runtime/parachains/src/disputes.rs
```diff
@@ -629,10 +619,12 @@ impl<T: Config> Pallet<T> {
...
 				<Disputes<T>>::remove_prefix(to_prune, None);
-
-				// This is larger, and will be extracted to the `shared` module for more proper pruning.
-				// TODO: https://github.com/paritytech/polkadot/issues/3469
-				<Included<T>>::remove_prefix(to_prune, None);
+				// Mark the session as pruneable so that its candidates can be incrementally
+				// removed over the course of many block inclusions.
+				shared::Pallet::<T>::mark_session_pruneable(to_prune);
+				// TODO(ladi): remove this call, currently allows unit tests to pass. Need to
+				// figure out how to invoke paras_inherent::enter in run_to_block.
```
kianenigma, line 626: why can't you call this in your `run_to_block`? `enter` is a dispatchable and is public

runtime/parachains/src/disputes.rs
```diff
@@ -1171,6 +1159,21 @@ mod tests {
...
+		block_number: BlockNumber,
+	) {
+		if let Some(revert_to) = shared::Pallet::<Test>::note_included_candidate(
+			session,
+			candidate_hash.clone(),
+			block_number,
+			CoreIndex(0),
+		) {
+			Pallet::<Test>::process_included(session, candidate_hash.clone(), revert_to);
+		}
```
kianenigma, line 1174: what should happen in `else`?

runtime/parachains/src/shared.rs
```diff
@@ -64,6 +71,40 @@ pub mod pallet {
...
+	>;
+
+	/// The set of all past sessions that have yet to fully pruned. Sessions are added to this set
+	/// upon moving to new current session, and removed after all of their included candidates
+	/// have been removed. This set is tracked independently so that the cost of removing included
+	/// candidates can be amortized over multiple blocks rather than removing all candidates upon
+	/// transitioning to a new session.
+	#[pallet::storage]
+	#[pallet::getter(fn pruneable_sessions)]
+	pub(super) type PruneableSessions<T: Config> = StorageMap<_, Twox64Concat, SessionIndex, ()>;
```
kianenigma, line 95: @thiolliere if this pattern is legit, it would be easy to abstract it behind a `StorageSet`.

runtime/parachains/src/shared.rs
```diff
@@ -114,6 +155,148 @@ impl<T: Config> Pallet<T> {
...
+	/// Prunes up to `max_candidates_to_prune` candidates from `IncludedCandidates` that belong to
+	/// non-active sessions.
+	pub(crate) fn prune_ancient_sessions(max_candidates_to_prune: usize) {
+		let mut to_prune_candidates = Vec::new();
+		let mut to_prune_blocks = Vec::new();
+
+		let mut n_candidates = 0;
+
+		let mut incomplete_session = None;
+		for session in PruneableSessions::<T>::iter_keys() {
```
kianenigma, line 174:
```suggestion
		for session in PruneableSessions::<T>::iter_keys().take(max_candidates_to_prune) {
```

kianenigma: I haven't read yet the rest of the code but I really think this is what you wanted

runtime/parachains/src/shared.rs
```diff
@@ -114,6 +155,148 @@ impl<T: Config> Pallet<T> {
...
+	/// Records an included candidate, returning the block height that should be reverted to if the
+	/// block is found to be invalid. This method will return `None` if and only if `included_in`
+	/// is zero.
+	pub(crate) fn note_included_candidate(
+		session: SessionIndex,
+		candidate_hash: CandidateHash,
+		included_in: T::BlockNumber,
+		core_index: CoreIndex,
+	) -> Option<T::BlockNumber> {
+		if included_in.is_zero() {
```
kianenigma, line 228: when does this happen?

runtime/parachains/src/shared.rs
```diff
@@ -114,6 +155,148 @@ impl<T: Config> Pallet<T> {
...
+
+	/// Returns the revert-to height and core index for a given candidate, if one exists.
+	pub(crate) fn get_included_candidate(
+		session: &SessionIndex,
+		candidate_hash: &CandidateHash,
+	) -> Option<(T::BlockNumber, CoreIndex)> {
+		<IncludedCandidates<T>>::get(session, candidate_hash)
+	}
+
+	#[cfg(test)]
```
kianenigma, line 254: why is this code not in the `mod tests`?

PR #3570, opened 2021-08-04
[CI] Fix listing pallets for benchmarks jobs

s3krit:
Due to https://github.com/paritytech/substrate/pull/9373 we now have a `--list` flag for the benchmarks command, and the previous method no longer works, yielding an empty list.

### DIFF
--- a/scripts/run_benches_for_runtime.sh
+++ b/scripts/run_benches_for_runtime.sh
@@ -11,7 +11,8 @@ echo "[+] Running all benchmarks for $runtime"
 
 # shellcheck disable=SC2086
 cargo +nightly run $standard_args benchmark \
-  --chain "${runtime}-dev" --list |\
+    --chain "${runtime}-dev" \
+    --list |\
   tail -n+2 |\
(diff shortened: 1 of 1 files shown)


scripts/run_benches_for_runtime.sh
```diff
@@ -11,13 +11,9 @@ echo "[+] Running all benchmarks for $runtime"
...
 # shellcheck disable=SC2086
 cargo +nightly run $standard_args benchmark \
-    --chain "${runtime}-dev" \
-    --execution=wasm \
-    --wasm-execution=compiled \
-    --pallet "*" \
-    --extrinsic "*" \
-    --repeat 0 | \
-  sed -r -e 's/Pallet: "([a-z_:]+)".*/\1/' | \
+  --chain "${runtime}-dev" --list |\
```
chevdor, line 14: You may want to keep the previous indentation to easily distinguish the `cargo` from the following piped calls.

chevdor (approved):

PR #3575, opened 2021-08-05
UMP: Support Overweight messages

pepyakin:
The purpose of this PR is to handle the UMP messages that exceeded the weight budget for an individual message. This is done by putting such messages aside. At a later time, an entity interested in servicing the message, will execute the `service_outweight` dispatchable.

The biggest part of this PR is actually not related to UMP but configuration. We introduce a new field called `ump_max_individual_weight`, which sets the weight threshold for being counted as overweight. Adding another field is not backward compatible, thus this requires a migration.

I've taken a look at several strategies here. The alternatives were like using a separate storage field for this particular value, a private configuration for the ump module, or a custom Decode implementation that injects the defaults for the missing values. 

I figured that this value (or any other value for that matter) that cannot just be added to `AbridgedHostConfiguration` may be required by the parachains someday. Also it is plausible we add more or change the existing configuration values. Therefore the infrastructure for that can be useful in future.

Please look extra carefully at the migration code. It is my first time doing it.

⚠️ This PR performs an migration for the host configuration pallet. It adds a new field required by this PR and it also removes a field that was deprecated in #3543.


runtime/parachains/src/ump.rs
```diff
@@ -197,6 +209,26 @@ pub mod pallet {
...
+		/// used.
+		///
+		/// \[ overweight_index, used \]
+		OverweightServiced(OverweightIndex, Weight),
+	}
+
+	#[pallet::error]
+	pub enum Error<T> {
+		/// The message index given is unknown.
+		Unknown,
```
4meta5, line 229:
```suggestion
		UnknownMessageIndex,
```

runtime/parachains/src/ump.rs
```diff
@@ -197,6 +209,26 @@ pub mod pallet {
...
+		/// The amount of weight given is possibly not enough for executing the message.
+		OverLimit,
```
4meta5, line 231:
```suggestion
		WeightOverLimit,
```

runtime/parachains/src/ump.rs
```diff
@@ -599,51 +693,64 @@ pub(crate) mod mock_sink {
...
+		// the result returned from the `process_upward_message` for this expected message.
+		mock_result: Result<Weight, (MessageId, Weight)>,
 	}
 
 	std::thread_local! {
 		// `Some` here indicates that there is an active probe.
 		static HOOK: RefCell<Option<VecDeque<UmpExpectation>>> = RefCell::new(None);
 	}
 
+	/// Yeet an expectation into the active probe instance. Panics if there is no active probe.
```
KiChjang, line 713: I actually don't know what yeet means here?

pepyakin: to toss )

runtime/parachains/src/configuration.rs
```diff
@@ -213,6 +220,11 @@ impl<BlockNumber: Default + From<u32>> Default for HostConfiguration<BlockNumber
 			hrmp_max_parachain_outbound_channels: Default::default(),
 			hrmp_max_parathread_outbound_channels: Default::default(),
 			hrmp_max_message_num_per_candidate: Default::default(),
+
+			//
+			// FOR REVIEW: what is a sensible value here?
+			//
+			ump_max_individual_weight: 10 * WEIGHT_PER_MILLIS,
```
shawntabrizi, line 227: should probably be some percentage of the block weight. Something like 1% of the block weight does not sound unreasonable. That would be .02 seconds.

runtime/parachains/src/configuration/migration.rs
```diff
@@ -0,0 +1,288 @@
...
+#[cfg(test)]
+mod tests {
+	use super::*;
+	use crate::mock::{new_test_ext, Test};
+
+	#[test]
+	fn test_migrate_to_v1() {
+		// Host configuration has lots of fields. However, in this migration we only add a single
+		// field. The most important part to check are a couple of the last fields. We also pick
+		// extra fields to check arbitrary, e.g. depending on their position (i.e. the middle) and
```
shawntabrizi, line 214:
```suggestion
		// extra fields to check arbitrarily, e.g. depending on their position (i.e. the middle) and
```

runtime/parachains/src/ump.rs
shawntabrizi: i agree with these name changes

runtime/parachains/src/ump.rs
```diff
@@ -434,6 +515,19 @@ impl<T: Config> Pallet<T> {
...
+	/// Puts a given upward message into the list of overweight messages allowing it to be executed
+	/// later.
+	fn stash_overweight(sender: ParaId, upward_message: Vec<u8>) -> OverweightIndex {
+		let index = <Self as Store>::OverweightCount::mutate(|count| {
+			let index = *count;
+			*count += 1;
+			index
+		});
+
+		<Self as Store>::Overweight::insert(index, (sender, upward_message));
```
shawntabrizi, line 528: not that it matters, but index 0 is never used?

runtime/parachains/src/ump.rs
```diff
@@ -972,4 +1096,85 @@ mod tests {
...
+			}
+			.build(),
+		)
+		.execute_with(|| {
+			// HACK: Start with the block number 1. This is needed because should an event be
+			// emitted during the genesis block they will be implicitly wiped.
+			System::set_block_number(1);
+
+			{
+				// This one is overweight. However, the weight is plenty and we can afford to execute
```
shawntabrizi, line 1131: is this one overweight? what does that mean in this context?

runtime/parachains/src/ump.rs
```diff
@@ -972,4 +1096,85 @@ mod tests {
...
+					Ump::service_overweight(Origin::root(), 0, 1000),
+					Error::<Test>::OverLimit
+				);
+				drop(probe);
+			}
+
+			{
+				let mut probe = Probe::new();
+				probe.assert_msg(para_a, a_msg_3.clone(), 1000);
+				assert_ok!(Ump::service_overweight(Origin::root(), 0, 1000));
```
shawntabrizi, line 1174: the ump total weight is 900, why does this succeed?

shawntabrizi (approved):
looks okay to me

runtime/parachains/src/ump.rs
pepyakin: How so? 

When there are no overweight messages, then when we enter here `count` will be 0. On L523 we save 0 into `index` and return it on L525, which will in turn assign 0 to `index` defined on L522.

runtime/parachains/src/configuration.rs
pepyakin: Put `20 * WEIGHT_PER_MILLIS`. FWIW, this is 1/5 of the `ump_total_service_weight`.

runtime/parachains/src/ump.rs
shawntabrizi: my bad :P

pepyakin:
@shawntabrizi in https://github.com/paritytech/polkadot/pull/3543#issuecomment-908858622 you asked if we could remove the deprecated field and I said we can wait until the next migration. This next migration is actually happening in this PR so I went ahead and removed the deprecated field here.

pepyakin:
@apopiak could you please take a look at this one? I would be glad to get your review on the runtime parts of this, especially the runtime migration. Thanks!

runtime/parachains/src/configuration/migration.rs
```diff
@@ -0,0 +1,284 @@
...
+		pub max_code_size: u32,
+		pub max_head_data_size: u32,
+		pub max_upward_queue_count: u32,
+		pub max_upward_queue_size: u32,
+		pub max_upward_message_size: u32,
+		pub max_upward_message_num_per_candidate: u32,
+		pub hrmp_max_message_num_per_candidate: u32,
+		pub validation_upgrade_frequency: BlockNumber,
+		pub validation_upgrade_delay: BlockNumber,
+		pub max_pov_size: u32,
```
shawntabrizi, line 53: shouldnt this struct have `_hrmp_open_request_ttl`?

runtime/parachains/src/configuration/migration.rs
```diff
@@ -0,0 +1,284 @@
...
+				hrmp_channel_max_message_size: Default::default(),
+				hrmp_max_parachain_outbound_channels: Default::default(),
+				hrmp_max_parathread_outbound_channels: Default::default(),
+				hrmp_max_message_num_per_candidate: Default::default(),
+			}
+		}
+	}
+}
+
+fn migrate_to_v1<T: Config>() -> Weight {
```
apopiak, line 132:
```suggestion

/// Migrates the `HostConfiguration` from v0 (with deprecated `hrmp_open_request_ttl` and without 
/// `ump_max_individual_weight`) to v1 (without hrmp ttl and with max individual weight).
/// Uses the `Default` implementation of `HostConfiguration` to choose a value for `ump_max_individual_weight`.
///
/// NOTE: Only use this function if you know what you are doing. Default to using `migrate_to_latest`.
pub fn migrate_to_v1<T: Config>() -> Weight {
```
Let's expose the function so users can decide which migration to run?

runtime/parachains/src/configuration/migration.rs
```diff
@@ -0,0 +1,284 @@
...
+
+use crate::configuration::{self, Config, Pallet, Store};
+use frame_support::{pallet_prelude::*, traits::StorageVersion, weights::Weight};
+use frame_system::pallet_prelude::BlockNumberFor;
+
+/// The current storage version.
+pub const STORAGE_VERSION: StorageVersion = StorageVersion::new(1);
+
+/// Call this during an next runtime upgrade for this module.
+pub fn on_runtime_upgrade<T: Config>() -> Weight {
```
apopiak, line 27:
```suggestion
/// Migrates the pallet storage to the most recent version, checking and setting the `StorageVersion`.
pub fn migrate_to_latest<T: Config>() -> Weight {
```

apopiak (commented):
Looks sane except for the issue spotted by Shawn.
Left some suggestions (most of them fairly nitpicky).

runtime/parachains/src/configuration.rs
```diff
@@ -28,6 +28,10 @@ use sp_std::prelude::*;
 
 pub use pallet::*;
 
+mod migration;
```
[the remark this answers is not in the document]
pepyakin: Thanks for the pointer, did not know this may be useful.

runtime/parachains/src/configuration/migration.rs
```diff
@@ -0,0 +1,284 @@
...
+	let mut weight = 0;
+
+	if StorageVersion::get::<Pallet<T>>() == 0 {
+		weight += migrate_to_v1::<T>();
+		StorageVersion::new(1).put::<Pallet<T>>();
```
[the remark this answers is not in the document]
pepyakin: I think it would be better to leave it like this. The reason is: here we want to migrate from version 0 to version 1, not to the current storage version. The difference is subtle, but it boils down that this is more robust to changes.

If the constant was bumped, there is a chance that it won't be noticed to be updated here. The fix would be to replace `STORAGE_VERSION` with `StorageVersion::new(1)` which is what we see here already. 

And if you look from the opposite side, we do not want this number to be updated because it will always be a migration from version 0 to version 1.

runtime/parachains/src/configuration/migration.rs
```diff
@@ -0,0 +1,284 @@
...
+	// - make it easier to verify that fields assign what they supposed to assign.
+	// - this code is transient and will be removed after all migrations are done.
+	// - this code is important enough to optimize for legibility sacrificing consistency.
+	#[rustfmt::skip]
+	let translate =
+		|pre: v0::HostConfiguration<BlockNumberFor<T>>| -> configuration::HostConfiguration<BlockNumberFor<T>>
+	{
+		super::HostConfiguration {
+
+max_code_size                            : pre.max_code_size,
```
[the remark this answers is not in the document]
pepyakin: see a justification from a couple of lines above. Lmk if that is not a good one

runtime/parachains/src/ump.rs
```diff
@@ -70,10 +71,18 @@ impl UmpSink for () {
...
+pub type OverweightIndex = u64;
+
 /// A specific implementation of a `UmpSink` where messages are in the XCM format
 /// and will be forwarded to the XCM Executor.
 pub struct XcmSink<XcmExecutor, Config>(PhantomData<(XcmExecutor, Config)>);
 
+/// Returns a [`MessageId`] for the given upward message payload.
+fn upward_message_id(data: &[u8]) -> MessageId {
+	sp_io::hashing::blake2_256(data)
+}
```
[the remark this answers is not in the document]
pepyakin: Not sure yet. I think it would be good to leave as is. In case somebody needs it that somebody can make it public

runtime/parachains/src/ump.rs
```diff
@@ -242,8 +274,49 @@ pub mod pallet {
 	#[pallet::storage]
 	pub type NextDispatchRoundStartWith<T: Config> = StorageValue<_, ParaId>;
 
+	/// The messages that exceeded max individual message weight budget.
+	///
+	/// These messages stay there until manually dispatched.
+	#[pallet::storage]
+	pub type Overweight<T: Config> =
+		StorageMap<_, Twox64Concat, OverweightIndex, (ParaId, Vec<u8>), OptionQuery>;
```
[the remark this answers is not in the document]
pepyakin: This is actually not an original work, but rather a port of the similar mechanism from Gav's code found on the parachain side, which you can find [here](https://github.com/paritytech/cumulus/blob/276e5bdd38c49976e3e1598722bd11c056077317/pallets/dmp-queue/src/lib.rs). It uses the same mechanism. I am not sure what's the reason behind choosing the one that stores messages, in contrast to what you are proposing here, but it seems to make sense.

runtime/parachains/src/configuration/migration.rs
pepyakin: Can you elaborate? I was under impression that the pallet owns it's data and performs it's migrations whenever needed and that can only be triggered on runtime upgrade.

Exposing this function will allow users to run the storage upgrade whenever, which... I do not know when that can be useful apart from a runtime upgrade.

runtime/parachains/src/ump.rs
```diff
@@ -490,16 +586,27 @@ impl QueueCache {
...
 		}
+	}
 
+	/// Returns if the queue for the given para is empty.
+	///
+	/// That is, if returned `true` then next call to [`peek_front`] will return `None`.
+	///
+	/// Does not mutate the queue.
+	fn is_empty<T: Config>(&mut self, para: ParaId) -> bool {
+		let cache_entry = self.ensure_cached::<T>(para);
```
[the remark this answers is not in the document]
pepyakin: Yep, that's right.

runtime/parachains/src/configuration/migration.rs
pepyakin: I did not accept this suggestion and went ahead pushing this myself cause applying this will break the code since it doesn't account for the call site change.

runtime/parachains/src/configuration/migration.rs
pepyakin: (I applied the rustdoc part in https://github.com/paritytech/polkadot/pull/3575/commits/7236e5e27a81cfaa8c1dcabdab661487f095f9aa)

runtime/kusama/src/lib.rs
```diff
@@ -1117,6 +1117,7 @@ impl parachains_ump::Config for Runtime {
 	type Event = Event;
 	type UmpSink = crate::parachains_ump::XcmSink<XcmExecutor<XcmConfig>, Runtime>;
 	type FirstMessageFactorPercent = FirstMessageFactorPercent;
+	type ExecuteOverweightOrigin = EnsureRoot<AccountId>;
```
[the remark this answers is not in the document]
pepyakin: I do not feel confident answering this.

I am actually wondering why don't we allow for anyone to dispatch an overweight item given that the transaction supplies enough weight.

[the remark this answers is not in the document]
apopiak: True, you could probably relax it even more

runtime/parachains/src/configuration/migration.rs
apopiak: What you mention is the common case. Unfortunately it can happen that you cannot follow the common upgrade path and need to run a custom migration.

Maybe I'm thinking too far ahead/too much of the edge case :shrug: 
I just remember having to fork Substrate and change visibility of migrations to be able to trigger them in a custom upgrade scenario.

runtime/parachains/src/configuration.rs
```diff
@@ -215,6 +217,7 @@ impl<BlockNumber: Default + From<u32>> Default for HostConfiguration<BlockNumber
 			hrmp_max_parachain_outbound_channels: Default::default(),
 			hrmp_max_parathread_outbound_channels: Default::default(),
 			hrmp_max_message_num_per_candidate: Default::default(),
+			ump_max_individual_weight: 20 * WEIGHT_PER_MILLIS,
```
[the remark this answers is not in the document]
pepyakin: I see what you mean however that won't make it better: all the other values are zero and this is basically a non-workable config. While it may be worth addressing, I think this should be out of scope for this PR

runtime/parachains/src/configuration/migration.rs
pepyakin: That is enough of justification and I exposed the fields

scripts/gitlab/lingua.dic
```diff
@@ -292,6 +292,7 @@ WND/S
 Wococo
 XCM/S
 XCMP/M
+yeet
```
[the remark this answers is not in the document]
pepyakin: let me know if you think we should rephrase

runtime/parachains/src/ump.rs
```diff
@@ -242,8 +274,49 @@ pub mod pallet {
...
+			origin: OriginFor<T>,
+			index: OverweightIndex,
+			weight_limit: Weight,
+		) -> DispatchResultWithPostInfo {
+			T::ExecuteOverweightOrigin::ensure_origin(origin)?;
+
+			let (sender, data) =
+				Overweight::<T>::get(index).ok_or(Error::<T>::UnknownMessageIndex)?;
+			let used = T::UmpSink::process_upward_message(sender, &data[..], weight_limit)
+				.map_err(|_| Error::<T>::WeightOverLimit)?;
```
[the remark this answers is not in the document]
apopiak: :facepalm: sorry for the brain fart :sweat_smile:

apopiak (approved):

PR #3578, opened 2021-08-05
Avoid bad pattern of wildcard fees

gavofyork:
Alters the two XCM dispatchables so that the `BuyExecution` uses the first Asset in the list to purchase fees, rather than `MultiAsset::All`, which is not a pattern that will be supported in the upcoming XCM v1.

## Migration

`pallet_xcm::Config` has an extra item `type LocationInverter`. This should be the same definition as declared in the `xcm_executor::Config`, typically:

```
type LocationInverter = LocationInverter<Ancestry>;
```


KiChjang (commented):
Code-wise looks ok, I'm just wondering how we can enforce that the first element in the `assets` vector is going to be used for fee payment (or whether it is at all desirable in this point in time to have any such kind of enforcement, programmatically or otherwise).

KiChjang (approved):

PR #3580, opened 2021-08-05
Chore: Mass replace `,);` pattern

pepyakin:
Mass replace `,);` pattern to `);`. 

This is an artifact left by rustfmt being conservative. Specifically, it is not dare to remove the comma from macro invocations because it cannot know if it changes the semantics.

The exception is with type declarations `(Type,);` which denotes type which is a tuple of one element. If we remove the comma it will become the type itself.

Obviously this PR doesn't intend to change semantics.

### DIFF
--- a/bridges/modules/currency-exchange/src/lib.rs
+++ b/bridges/modules/currency-exchange/src/lib.rs
@@ -470,7 +470,7 @@ mod tests {
 			assert_ok!(Exchange::import_peer_transaction(
 				Origin::signed(SUBMITTER),
 				(true, transaction),
-			),);
+			));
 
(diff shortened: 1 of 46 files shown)


HCastano (approved):
If you want your OCD forever satisfied you'll also need to open a PR in `parity-bridges-common` with this 😛

pepyakin:
One day, one day... But is this fine that I modify the bridges code here?

HCastano:
> One day, one day... But is this fine that I modify the bridges code here?

Yeah it's fine, but next time the subtree gets updated these changes will be blown out and you'll have to re-fix them

bkchr (approved):

PR #3555, opened 2021-08-02
Guard against XCM recursive bombs by setting a recursion limit

KiChjang:
Part of #2841.

PR #3563, opened 2021-08-03
Add Benchmarking Pipeline to XCM

shawntabrizi:
supersedes https://github.com/paritytech/polkadot/pull/3534

PR #3586, opened 2021-08-06
Harden XCM v1 for Recursions

shawntabrizi:
Taking: https://github.com/paritytech/polkadot/pull/3555

And making it work for XCM v1


runtime/test-runtime/src/lib.rs
```diff
@@ -488,6 +489,31 @@ impl parachains_ump::Config for Runtime {
...
 
+parameter_types! {
+	pub const BaseXcmWeight: frame_support::weights::Weight = 1_000;
+	pub const AnyNetwork: xcm::latest::NetworkId = xcm::latest::NetworkId::Any;
+}
+
+pub type LocalOriginToLocation = xcm_builder::SignedToAccountId32<Origin, AccountId, AnyNetwork>;
+
+impl pallet_xcm::Config for Runtime {
+	// The config types here are entirely configurable, since the only one that is sorely needed
```
kianenigma, line 500:
```suggestion
	// The config types here are entirely configurable, since the only one that is surely needed
```

?

runtime/test-runtime/src/xcm_config.rs
```diff
@@ -0,0 +1,91 @@
...
+);
+
+pub struct DoNothingRouter;
+impl SendXcm for DoNothingRouter {
+	fn send_xcm(_dest: MultiLocation, _msg: Xcm<()>) -> XcmResult {
+		Ok(())
+	}
+}
+
+pub type Barrier = AllowUnpaidExecutionFrom<All<MultiLocation>>;
```
kianenigma, line 46: This is better than `struct YesItShould` that I put for benchmarking mock, perhaps use this there as well: https://github.com/paritytech/polkadot/pull/3563/files#diff-4956fec0e6b5a0cf3ad1a1980b05285ac79c5a91d5edb15863d19b1519e8ae40R21

xcm/xcm-executor/integration-tests/src/lib.rs
```diff
@@ -0,0 +1,141 @@
...
+		polkadot_test_runtime::Call::Xcm(pallet_xcm::Call::execute(
+			Box::new(msg.clone()),
+			1_000_000_000,
+		)),
+		sp_keyring::Sr25519Keyring::Alice,
+	);
+
+	block_builder.push_polkadot_extrinsic(execute).expect("pushes extrinsic");
+
+	let block = block_builder.build().expect("Finalizes the block").block;
```
kianenigma, line 69:
```suggestion
	let block = block_builder.build().expect("finalizes the block").block;
```

xcm/xcm-executor/integration-tests/src/lib.rs
```diff
@@ -0,0 +1,141 @@
...
+	let block_hash = block.hash();
+
+	futures::executor::block_on(client.import(sp_consensus::BlockOrigin::Own, block))
+		.expect("imports the block");
+
+	client
```
kianenigma, line 75: why so complicated? why not just use the runtime primitives to test this?

kianenigma: You can test the decode limit (which is checked at client-runtime boundary) with manual encode/decode. Also the execution limit can be checked here as well via the xcm-executor. 

I recall that it had something to do with a strange rustc issue. Now's is a good time to rethink why we're doing it this way.

kianenigma (approved):
Looks correct to me.

xcm/xcm-executor/integration-tests/src/lib.rs
KiChjang: We had to run the test in WASM in order to simulate a real-ish environment. Local testing with a native client would blow the stack with a depth of just 6, while on WASM, it began to OOM at around 84.

xcm/xcm-simulator/example/src/parachain.rs
```diff
@@ -238,7 +240,10 @@ pub mod mock_msg_queue {
 
 				let mut remaining_fragments = &data_ref[..];
 				while !remaining_fragments.is_empty() {
-					if let Ok(xcm) = VersionedXcm::<T::Call>::decode(&mut remaining_fragments) {
+					if let Ok(xcm) = VersionedXcm::<T::Call>::decode_all_with_depth_limit(
+						xcm::MAX_XCM_DECODE_DEPTH,
+						&mut remaining_fragments,
+					) {
```
KiChjang, line 246: Needs a revert:

```suggestion
					if let Ok(xcm) = VersionedXcm::<T::Call>::decode(&mut remaining_fragments) {
```

xcm/xcm-simulator/example/src/parachain.rs
```diff
@@ -256,8 +261,11 @@ pub mod mock_msg_queue {
 		) -> Weight {
 			for (_i, (_sent_at, data)) in iter.enumerate() {
 				let id = sp_io::hashing::blake2_256(&data[..]);
-				let maybe_msg =
-					VersionedXcm::<T::Call>::decode(&mut &data[..]).map(Xcm::<T::Call>::try_from);
+				let maybe_msg = VersionedXcm::<T::Call>::decode_all_with_depth_limit(
+					xcm::MAX_XCM_DECODE_DEPTH,
+					&mut &data[..],
+				)
+				.map(Xcm::<T::Call>::try_from);
```
KiChjang, line 268:
```suggestion
				let maybe_msg =
					VersionedXcm::<T::Call>::decode(&mut &data[..]).map(Xcm::<T::Call>::try_from);
```

xcm/xcm-simulator/example/src/parachain.rs
```diff
@@ -147,7 +147,9 @@ impl Config for XcmConfig {
 #[frame_support::pallet]
 pub mod mock_msg_queue {
 	use super::*;
+	use codec::DecodeLimit;
 	use frame_support::pallet_prelude::*;
+	use xcm_executor::MAX_RECURSION_LIMIT;
```
KiChjang, line 152:
```suggestion
	use frame_support::pallet_prelude::*;
```

KiChjang (approved):

KiChjang:
bot merge

PR #3597, opened 2021-08-08
Fix Backwards Compatability with v0 Response

shawntabrizi:
In https://github.com/paritytech/polkadot/pull/2815, items that were unchanged between XCM v0 and v1 were moved to XCM v1, and then a dependency was placed from v0 to v1.

This PR switches the behavior, keeping v0 XCM stuff unchanged and unmoved from v0, and instead having v1 rely on the v0 syntax.

This caught a problem where the `Response` type from v1 was actually different than v0, and this also fixes that.


KiChjang (approved):
@gavofyork I know you mentioned that `BodyId::Index` and `Junction::GeneralIndex` were originally struct enum variants and should be changed into tuple enum variants, but I confirmed with @shawntabrizi that the encoding format is completely the same between the two, so we can simply make the change without worry.

xcm/src/v1/mod.rs
```diff
@@ -45,7 +43,7 @@ pub use traits::{Error, ExecuteXcm, Outcome, Result, SendXcm};
 /// A prelude for importing all types typically used when interacting with XCM messages.
 pub mod prelude {
 	pub use super::{
-		junction::{
+		super::v0::{
```
shawntabrizi, line 46:
```suggestion
		// These parts of XCM v0 have been unchanged in XCM v1
		super::v0::{
```

gavofyork (approved):

shawntabrizi:
bot merge

PR #3601, opened 2021-08-09
Impl WeightTrader for tuple

shaunxw:
This will allow to configure multiple traders for XCM execution fee.

### DIFF
--- a/xcm/xcm-executor/src/traits/weight.rs
+++ b/xcm/xcm-executor/src/traits/weight.rs
@@ -58,6 +58,11 @@ pub trait UniversalWeigher {
 }
 
 /// Charge for weight in order to execute XCM.
+///
+/// A `WeightTrader` may also be put into a tuple, in which case the default behavior of
+/// `buy_weight` and `refund_weight` would be to attempt to call each tuple element's own
+/// implementation of these two functions, in the order of which they appear in the tuple,
+/// returning early when a successful result is returned.
 pub trait WeightTrader: Sized {
 	/// Create a new trader instance.
 	fn new() -> Self;
--- a/xcm/xcm-executor/src/traits/weight.rs
+++ b/xcm/xcm-executor/src/traits/weight.rs
@@ -83,12 +83,16 @@ impl WeightTrader for Tuple {
 	}
 
 	fn buy_weight(&mut self, weight: Weight, payment: Assets) -> Result<Assets, Error> {
+		let mut last_error = None;
 		for_tuples!( #(
-			if let Ok(assets) = Tuple.buy_weight(weight, payment.clone()) {
-				return Ok(assets);
+			match Tuple.buy_weight(weight, payment.clone()) {
+				Ok(assets) => return Ok(assets),
+				Err(e) => { last_error = Some(e) }
 			}
 		)* );
-		Err(Error::TooExpensive)
+		let last_error = last_error.unwrap_or(Error::TooExpensive);
+		log::trace!(target: "xcm::buy_weight", "last_error: {:?}", last_error);
+		Err(last_error)
 	}
 
 	fn refund_weight(&mut self, weight: Weight) -> Option<MultiAsset> {
--- a/xcm/xcm-executor/src/traits/weight.rs
+++ b/xcm/xcm-executor/src/traits/weight.rs
@@ -76,11 +76,27 @@ pub trait WeightTrader: Sized {
 	}
 }
 
-impl WeightTrader for () {
+#[impl_trait_for_tuples::impl_for_tuples(30)]
+impl WeightTrader for Tuple {
 	fn new() -> Self {
-		()
+		for_tuples!( ( #( Tuple::new() ),* ) )
 	}
-	fn buy_weight(&mut self, _: Weight, _: Assets) -> Result<Assets, Error> {
-		Err(Error::Unimplemented)
+
+	fn buy_weight(&mut self, weight: Weight, payment: Assets) -> Result<Assets, Error> {
+		for_tuples!( #(
+			if let Ok(assets) = Tuple.buy_weight(weight, payment.clone()) {
+				return Ok(assets);
+			}
+		)* );
+		Err(Error::TooExpensive)
+	}
+
+	fn refund_weight(&mut self, weight: Weight) -> Option<MultiAsset> {
+		for_tuples!( #(
+			if let Some(asset) = Tuple.refund_weight(weight) {
+				return Some(asset);
+			}
+		)* );
+		None
 	}
 }
--- a/xcm/xcm-builder/src/tests.rs
+++ b/xcm/xcm-builder/src/tests.rs
@@ -15,9 +15,9 @@
 // along with Polkadot.  If not, see <http://www.gnu.org/licenses/>.
 
 use super::{mock::*, *};
+use frame_support::{assert_err, weights::constants::WEIGHT_PER_SECOND};
 use xcm::latest::prelude::*;
 use xcm_executor::{traits::*, Config, XcmExecutor};
-use frame_support::{weights::constants::WEIGHT_PER_SECOND, assert_err};
 
 #[test]
 fn basic_setup_works() {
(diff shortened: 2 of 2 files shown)


xcm/xcm-executor/src/traits/weight.rs
```diff
@@ -76,11 +76,27 @@ pub trait WeightTrader: Sized {
...
+		)* );
+		Err(Error::TooExpensive)
+	}
+
+	fn refund_weight(&mut self, weight: Weight) -> Option<MultiAsset> {
+		for_tuples!( #(
+			if let Some(asset) = Tuple.refund_weight(weight) {
+				return Some(asset);
+			}
+		)* );
```
shawntabrizi, line 99: does this make sense?

We could buy weight with one asset and return weight with another?

shaunxw: From my understanding, the sensible `WeightTrader` impl would be that if weights are not bought here, then it won't refund. This is the case for all current traders like `FixedRateOfFungible` and `UsingComponents`.

This case is covered in the added unit tests `weight_trader_tuple_should_work`.

shawntabrizi: yeah that makes sense

shawntabrizi (approved):

xcm/xcm-executor/src/traits/weight.rs
```diff
@@ -76,11 +76,27 @@ pub trait WeightTrader: Sized {
...
+		for_tuples!( ( #( Tuple::new() ),* ) )
 	}
-	fn buy_weight(&mut self, _: Weight, _: Assets) -> Result<Assets, Error> {
-		Err(Error::Unimplemented)
+
+	fn buy_weight(&mut self, weight: Weight, payment: Assets) -> Result<Assets, Error> {
+		for_tuples!( #(
+			if let Ok(assets) = Tuple.buy_weight(weight, payment.clone()) {
+				return Ok(assets);
+			}
```
apopiak, line 89: This swallows any errors that might happen along the way. Not sure we want to do that.

shawntabrizi: I think the only thing we care about is if all of them fail, which is the final error below. Otherwise, we simply want to check with each trader if it is possible to buy weight.

shaunxw:
Is this PR good to go, or we have to wait for more approvals?

xcm/xcm-executor/src/traits/weight.rs
KiChjang: That seems to be the case if we assume that weight can only be bought by the native token of a chain. I'm trying to understand if this is still valid for assets that are non-native, e.g. USDC on Statemint/Statemine.

Would it be possible to buy weights using non-native tokens? If so, then a `WeightTrader` tuple may have elements that both accept the non-native token, and when `refund_weight` is called, the first `WeightTrader` in the tuple would refund first, which may not be the correct behaviour.

shaunxw: > If so, then a `WeightTrader` tuple may have elements that both accept the non-native token, and when `refund_weight` is called, the first `WeightTrader` in the tuple would refund first, which may not be the correct behaviour.

The tuple impl tries buy/refund weights in the same order. If I understand your example correctly, in this case, the first trader buys weight, so it's fine that it refunds.

For instance, we have a tuple with two non-native traders `(UsdcTrader, UsdtTrader)`, if the assets are:
1. Sufficient USDC. `UsdcTrader` do the buying and refund. `UsdtTrader` won't have the chance to try.
2. Sufficient USDT. `UsdtTrader` buys and refunds. `UsdcTrader` tried both first but failed.
3. Insufficient USDC/USDT. Both trader failed.

KiChjang: That seems to carry an assumption that each token has a corresponding `WeightTrader` associated with it, e.g. USDC has a `UsdcTrader` abd USDT has a `UsdtTrader`. Is this relationship guaranteed to always be 1:1? If so, then this logic looks to be sufficient enough.

What I'm more concerned about is when you have multiple `WeightTrader`s that references the same token to trade. I'm imagining that some DeFi protocols may be permissive enough to allow for multiple tokens to be used to pay for fees, and each `WeightTrader` in their use case corresponds to a liquidity pool. Let's say both traders reference USDC as a token that they accept for paying for the weights. You'll then have `Lp1Trader` and `Lp2Trader` in a tuple, and the default behaviour here would then be to always buy and refund with `Lp1Trader` first, before `Lp2Trader` gets queried, which may result in an incorrect behaviour.

Though practically speaking, such a use case is complex enough to justify for a custom implementation of the `WeightTrader` trait on a self-defined type, and I wouldn't expect the developers of such a DeFi protocol to simply use a tuple to wrap the 2 `WeightTrader`s. However, if that's the case, then I think it would be good to document what the default behaviour is when multiple `WeightTrader`s are put inside of a tuple.

shawntabrizi:
@shaunxw needs a second approve

xcm/xcm-executor/src/traits/weight.rs
apopiak: I think that retry behavior is generally fine, but I think it would be good to at least surface the last error (returned by the last impl of buy_weight).
We should also add some logging (similar to the other XCM tracing I added) and maybe that can be enough.

apopiak: I just want to make sure it's possible to debug this at all (which I fear the current version is not).

apopiak (changes requested):
I would like to see *some* way to get information about the inner errors if this fails so it is more debuggable.

xcm/xcm-executor/src/traits/weight.rs
shawntabrizi: is a log trace message enough?

shawntabrizi: is a trace log enough?

apopiak: What do you think of:
```Rust
fn buy_weight(&mut self, weight: Weight, payment: Assets) -> Result<Assets, Error> {
		let mut last_error = None;
		for_tuples!( #(
			match Tuple.buy_weight(weight, payment.clone()) {
				Ok(assets) => return Ok(assets),
				Err(e) => { last_error = Some(e) }
			}
		)* );
		log::trace!(target: "xcm::buy_weight", "last_error: {:?}", last_error);
		Err(last_error)
	}
```

apopiak (approved):

KiChjang (approved):

xcm/xcm-executor/src/traits/weight.rs
shaunxw: The tuple impl doesn't assume 1:1 between tokens and traders. It's similar to [`SendXcm` tuple](https://github.com/paritytech/polkadot/blob/83d18b4e2858338f9847397b6e227cedceb9a036/xcm/src/v1/traits.rs#L259) that making no assumption of who sends what kind of messages, but simply try one by one until the msg is successfully sent. I think it's reasonable to make only one of the traders buy/refund weights, as the required weights shouldn't be bought/refunded twice, just like an XCM message should be sent twice. And yeah it's helpful to document this behavior.

I agree that for more complex requirements, ppl should really impl a trader themselves instead of expecting this tuple impl to help. Actually we're planning to impl a DEX trader in Karura, like in your example, and it will be used with other backup traders together.

xcm/xcm-executor/src/traits/weight.rs
shaunxw: It's helpful to add a trace log for debugging. Not sure about returning the last error, as it doesn't represent the reason why all traders failed.

xcm/xcm-executor/src/traits/weight.rs
shaunxw: @KiChjang Thanks for helping to add documentations. 👍

KiChjang:
bot merge

PR #3606, opened 2021-08-09
Track historical vrfs

Lldenaurois:
Part of #3463 

This PR intends to add a `StorageDoubleMap` to the `shared` module which keeps track of historical Babe VRFs for the purpose of comparing these VRFs against the Approval Assignments that are issued throughout the Approval Voting process.

### DIFF
--- a/runtime/parachains/src/disputes.rs
+++ b/runtime/parachains/src/disputes.rs
@@ -619,10 +619,12 @@ impl<T: Config> Pallet<T> {
 			for to_prune in to_prune {
 				// This should be small, as disputes are rare, so `None` is fine.
 				<Disputes<T>>::remove_prefix(to_prune, None);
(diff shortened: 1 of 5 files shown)


runtime/parachains/src/disputes.rs
```diff
@@ -629,10 +619,12 @@ impl<T: Config> Pallet<T> {
 			for to_prune in to_prune {
 				// This should be small, as disputes are rare, so `None` is fine.
 				<Disputes<T>>::remove_prefix(to_prune, None);
-
-				// This is larger, and will be extracted to the `shared` module for more proper pruning.
-				// TODO: https://github.com/paritytech/polkadot/issues/3469
```
rphmeier, line 634: Does this close #3469?

PR #3610, opened 2021-08-10
Use Holding Register term consistently

apopiak:
...and give a short description of it.

Also fixes some variable refences and adjusts some formatting.


gavofyork:
It's a thing and its name is the Holding Register. This rewording does little to help newbies understand that it's Something New which they should be aware of.

gavofyork:
Happy to consider actual improvements, but I don't think this is one so closing.

KiChjang (approved):

kianenigma:
seems stale

PR #3611, opened 2021-08-10
use WEIGHT_PER_SECOND in FixedRateOfFungible

apopiak:
Presumably it should be the same as the usage of `WEIGHT_PER_SECOND` above?


xcm/xcm-builder/src/weight.rs
```diff
@@ -191,7 +191,8 @@ impl<T: Get<(AssetId, u128)>, R: TakeRevenue> WeightTrader for FixedRateOfFungib
 	fn refund_weight(&mut self, weight: Weight) -> Option<MultiAsset> {
 		let (id, units_per_second) = T::get();
 		let weight = weight.min(self.0);
-		let amount = units_per_second * (weight as u128) / 1_000_000_000_000u128;
+		use frame_support::weights::constants::WEIGHT_PER_SECOND;
```
shawntabrizi, line 194: can we just import this at the top?

shawntabrizi (approved):

KiChjang (approved):

KiChjang:
bot merge

PR #3612, opened 2021-08-10
collect better memory stats

drahnr:
Introduces new metrics to collect total allocation and resident memory size in bytes and publish to prometheus. Little aid in tracking down OOM conditions by providing better data on allocations.


bkchr:
Isn't that already provided by prometheus? We already have memory statistics about the nodes.

bkchr:
https://github.com/paritytech/substrate/pull/6822

drahnr:
The idea is to provide a less coarse view on the allocations. The system view of jemalloc is too coarse for good correlations between other event occurences and the total consumption, with this we can obtain the inner stats of jemalloc synced to the time of i.e. channel fill snapshots.

node/metrics/src/metronome.rs
```diff
@@ -0,0 +1,51 @@
+use futures::prelude::*;
```
drahnr, line 1: TODO: missing header

bkchr:
> The idea is to provide a less coarse view on the allocations. The system view of jemalloc is too coarse for good correlations between other event occurences and the total consumption, with this we can obtain the inner stats of jemalloc synced to the time of i.e. channel fill snapshots.
> 
> TL;DR yes, can be done on OS level, but this is more useful for root cause analysis

How do you want to do a root cause analysis with this? You will just see that memory usage goes up, the same as the system will tell you. You want to say that the system update interval is not good enough for the memory usage?

drahnr:
For the linux client we use jemalloc, as such you will get somewhat jagged / fuzzy correlations given that multiple arenas are used with jemalloc.

Now this exhibits a rough representation of the true allocations to the system which then can be reported by a system level metrics collector. But this gives little insight on how many allocations, what granularity and only a rough estimate on time to event correlations.


Using the internal stats avoids this. Now you can remove the _correct_ amount of memory consumed by i.e. the channels (channel based memory consumption reporting TBD) and work with a less noisy residue memory allocation and correlate these in grafana. And this is helpful when digging into OOM case root causes.

eskimor:
Hi @koute ! How did you track down the issue in [statement distribution](https://github.com/paritytech/polkadot/pull/3519)? Would this PR have made it easier?

eskimor:
Nice, thanks @koute !

eskimor (approved):
Looks good to me. Also not sure, it is going to help much in practice though.

node/metrics/src/memory_stats.rs
```diff
@@ -0,0 +1,66 @@
...
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with Polkadot.  If not, see <http://www.gnu.org/licenses/>.
+
+//! Memory tracking statistics.
+//!
+//! Many subsystems have common interests such as canceling a bunch of spawned jobs,
+//! or determining what their validator ID is. These common interests are factored into
+//! this module.
```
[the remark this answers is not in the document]
drahnr: Ouch. Fixing now.

drahnr:
@koute the features you desire will be added in a separate PR, jemalloc does the heavy lifting there too (statistical profiling), but this is not in scope for this PR.

node/metrics/src/memory_stats.rs
```diff
@@ -0,0 +1,66 @@
...
+//!
+//! This crate also reexports Prometheus metric types which are expected to be implemented by subsystems.
```
drahnr, line 23:
```suggestion
//! Exposes the internal statistics of `jemalloc`.
```

node/overseer/src/metrics.rs
```diff
@@ -182,6 +201,24 @@ impl metrics::Metrics for Metrics {
 				)?,
 				registry,
 			)?,
+
+			#[cfg(feature = "memory-stats")]
+			memory_stats_allocated: prometheus::register(
+				prometheus::Gauge::<prometheus::U64>::new(
+					"parachain_overseer_memory_stats_allocated",
```
bkchr, line 208: This is neither "overseer" nor "parachain" specific. Naming doesn't make sense.

PR #3614, opened 2021-08-10
Add an empty `.cargo/config` file

expenses:
This makes it slightly easier to do substrate companion PRs, as one only has to add `paths = ["path/to/substrate"]` to this file in order to override the sources.

https://github.com/paritytech/substrate/blob/master/docs/CONTRIBUTING.adoc#updating-polkadot-as-well should be updated as well.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,4 +10,3 @@ polkadot.*
 !polkadot.service
 !.rpm/*
 .DS_Store
-.cargo


shawntabrizi:
Are you saying for a companion PR we should temporarily commit a `.cargo` file with the appropriate branch so that the polkadot side compiles successfully?

Or what specifically is this solving?

(I do this already locally, and just dont commit the `.cargo` file)

expenses:
@shawntabrizi well the main advantage is that the file just exists, instead of everyone having to create the directory and then the file themselves.

shawntabrizi:
eh, with that in mind, i think i would not want this PR :)

`touch ./.cargo/config` is not that hard

expenses:
😭  fine

PR #3617, opened 2021-08-10
Make RelayedFrom typesafe

gavofyork:
Some additional alterations to API of `Junctions`/`MultiLocation` to ensure data doesn't accidentally get swallowed.

### DIFF
--- a/xcm/src/v1/mod.rs
+++ b/xcm/src/v1/mod.rs
@@ -372,7 +372,7 @@ impl<Call> TryFrom<Xcm0<Call>> for Xcm<Call> {
 			Xcm0::Transact { origin_type, require_weight_at_most, call } =>
 				Transact { origin_type, require_weight_at_most, call: call.into() },
 			Xcm0::RelayedFrom { who, message } => RelayedFrom {
-				who: who.try_into()?,
+				who: MultiLocation::try_from(who)?.try_into()?,
 				message: alloc::boxed::Box::new((*message).try_into()?),
 			},
 		})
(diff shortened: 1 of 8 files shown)


xcm/src/v1/mod.rs
```diff
@@ -267,14 +267,11 @@ pub enum Xcm<Call> {
...
 	///
-	/// Safety: `who` must be an interior location of the context. This basically means that no `Parent`
-	/// junctions are allowed in it. This should be verified at the time of XCM execution.
-	///
 	/// Kind: *Instruction*
 	///
 	/// Errors:
 	#[codec(index = 10)]
-	RelayedFrom { who: MultiLocation, message: alloc::boxed::Box<Xcm<Call>> },
+	RelayedFrom { who: Junctions, message: alloc::boxed::Box<Xcm<Call>> },
```
gavofyork, line 274: This is the API change - `MultiLocation` can be wrong and must otherwise be subject to an dynamic check as `RelayedFrom`'s `who` must strictly be interior.

xcm/src/v1/mod.rs
```diff
@@ -267,14 +267,11 @@ pub enum Xcm<Call> {
 	/// A message to indicate that the embedded XCM is actually arriving on behalf of some consensus
 	/// location within the origin.
 	///
-	/// Safety: `who` must be an interior location of the context. This basically means that no `Parent`
```
gavofyork, line 270: No need for this with the new type in place.

apopiak (approved):

PR #3620, opened 2021-08-11
New Github Workflow to check extrinsic ordering

chevdor:
This PR makes checking for extrinsic ordering easier by introducing a new Github Workflow that runs the steps required to produce a small report comparing the new runtime metadata and the current one.

Since it follows the steps previously mentioned in the release todos, the explanation there has been replaced by a simple call of the workflow.

The new workflow is called manually for now, mainly due to the need for inputs such a link to a built binary.

### DIFF
--- a/.github/ISSUE_TEMPLATE/release.md
+++ b/.github/ISSUE_TEMPLATE/release.md
@@ -113,7 +113,7 @@ The things to look for in the output are lines like:
   - If no indices have changed, every modules line should look something like `[Identity] idx 25 (calls 15)`
 
 Note: Adding new functions to the runtime does not constitute a breaking change
-as long as teh indexes did not change.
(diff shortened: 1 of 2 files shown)


chevdor:
A sample run and output can be see [here](https://github.com/chevdor/polkadot/actions/runs/1119672822).

s3krit (approved):
Awesome stuff! Will definitely make releasing an easier process (which is the direction we should be going - collect steps in a checklist, then automate what we can).

chevdor:
Here is [the link](https://github.com/paritytech/polkadot/actions/workflows/extrinsic-ordering-check-from-bin.yml) for the sake of documentation, since it only shows up otherwise in the release todo when a release is being baked.

PR #3621, opened 2021-08-11
More standard staking miner deposits

kianenigma:
the current staking miner deposit for polkadot is a wee bit too high (400 DOTs), this lowers it to around 50 DOTs for current solutions + make it standard across all runtimes. 

The deposits are stored on-chain, so this change of config is 100% safe.

### DIFF
--- a/runtime/kusama/src/lib.rs
+++ b/runtime/kusama/src/lib.rs
@@ -345,11 +345,8 @@ parameter_types! {
 
 	// signed config
 	pub const SignedMaxSubmissions: u32 = 16;
-	pub const SignedDepositBase: Balance = deposit(1, 0);
-	// A typical solution occupies within an order of magnitude of 50kb.
-	// This formula is currently adjusted such that a typical solution will spend an amount equal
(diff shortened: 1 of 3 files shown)


runtime/kusama/src/lib.rs
```diff
@@ -1919,3 +1918,17 @@ sp_api::impl_runtime_apis! {
...
+mod tests_fess {
+	use super::*;
+	use sp_runtime::assert_eq_error_rate;
+
+	#[test]
+	fn signed_deposit_is_sensible() {
+		// ensure this number does not change, or that it is checked after each change.
+		// a 1 MB solution should need around 0.16 KSM deposit
+		let deposit = SignedDepositBase::get() + (SignedDepositByte::get() * 1024 * 1024);
+		assert_eq_error_rate!(deposit, UNITS * 16 / 100, UNITS / 100);
```
emostov, line 1932: Why use `assert_eq_error_rate!` and not just `assert_eq!` ? Seems like the values should be constant relative to each other

emostov (approved):

PR #3624, opened 2021-08-11
Remove request multiplexer

eskimor:
Fixes issue #2842  and paves the way for proper back pressuring in request/response protocols.

There is now a new typed receiver type `IncomingRequestReceiver<ReqType>`, which gets passed into subsystems at initialization time. So instead of the request multiplexer sending incoming requests to concerned subsystems, each subsystem gets their own `IncomingRequestReceiver` which does the decoding for them, so they only have to deal with typed data.

By having access to the receiver directly, they also have the possibility of applying back pressure to that channel, which was not possible before. (This PR does not introduce any new back pressure, but we now have the means to do it in a proper way.)

Also this PR gets rid of `Fault` type, as I realized that what was left with all type system limitations was not worth to keep.


node/network/availability-distribution/src/responder.rs
```diff
@@ -171,15 +239,16 @@ where
...
-async fn query_available_data<Context>(
-	ctx: &mut Context,
+async fn query_available_data<Sender>(
+	sender: &mut Sender,
 	candidate_hash: CandidateHash,
-) -> Result<Option<AvailableData>>
+) -> NonFatalResult<Option<AvailableData>>
 where
-	Context: SubsystemContext,
+	Sender: SubsystemSender,
```
drahnr, line 247: :+1:

node/network/dispute-distribution/src/sender/error.rs
```diff
@@ -20,33 +20,30 @@
...
-	fn from(e: NonFatal) -> Self {
-		Self(Fault::from_non_fatal(e))
-	}
-}
-
-impl From<Fatal> for Error {
-	fn from(f: Fatal) -> Self {
-		Self(Fault::from_fatal(f))
-	}
+pub enum Error {
```
drahnr, line 28: This seems to be a pattern, so making this a generic and only having to define type aliases in the subsystems might make sense.

node/network/protocol/src/request_response/incoming/mod.rs
```diff
@@ -0,0 +1,232 @@
...
+	}
+}
+
+/// Typed variant of [`netconfig::OutgoingResponse`].
+///
+/// Responses to `IncomingRequest`s.
+pub struct OutgoingResponse<Response> {
+	/// The payload of the response.
+	///
+	/// `Err(())` if none is available e.g. due an error while handling the request.
```
drahnr, line 193:
```suggestion
	/// `Err(())` if none is available e.g. due to an error while handling the request.
```

node/network/statement-distribution/src/lib.rs
```diff
@@ -1542,6 +1549,16 @@ impl StatementDistribution {
 		// Sender/Receiver for getting news from our responder task.
 		let (res_sender, mut res_receiver) = mpsc::channel(1);
 
+		ctx.spawn(
```
drahnr, line 1552: Iiuc, this is done to avoid introducing backpressure for now, which we will eventually want. Correct me if I am wrong.

node/overseer/src/lib.rs
```diff
@@ -920,32 +920,3 @@ where
...
-impl From<IncomingRequest<req_res_v1::CollationFetchingRequest>> for AllMessages {
-	fn from(req: IncomingRequest<req_res_v1::CollationFetchingRequest>) -> Self {
-		From::<CollatorProtocolMessage>::from(From::from(req))
-	}
-}
-impl From<IncomingRequest<req_res_v1::AvailableDataFetchingRequest>> for AllMessages {
-	fn from(req: IncomingRequest<req_res_v1::AvailableDataFetchingRequest>) -> Self {
-		From::<AvailabilityRecoveryMessage>::from(From::from(req))
-	}
-}
```
drahnr, line 951: :heart: hell yeah!

drahnr (approved):
A few nits, but overall, a great improvement!

node/network/dispute-distribution/src/sender/error.rs
eskimor: It is, but due to limitations in the type system, I was not able to abstract this in a meaningful way. I haven't tried macros, there might also be ways, I was not thinking of ... so we could discuss this over some beverage at some point ;-)

node/network/statement-distribution/src/lib.rs
eskimor: No, we spawn a task handling incoming requests and it will back pressure on those. We spawn a task, so we can back pressure incoming requests without blocking the whole subsystem.

PR #3626, opened 2021-08-11
backing-availability-audit: Move ErasureChunk Proof to BoundedVec

Lldenaurois:
Addresses: https://github.com/paritytech/srlabs_findings/issues/100

This PR moved the `proof` field in `ErasureChunk` to a nested set of `BoundedVec`s.


node/core/av-store/src/lib.rs
```diff
@@ -1177,7 +1178,7 @@ fn store_available_data(
 	let erasure_chunks = chunks.iter().zip(branches.map(|(proof, _)| proof)).enumerate().map(
 		|(index, (chunk, proof))| ErasureChunk {
 			chunk: chunk.clone(),
-			proof,
+			proof: Proof::try_from(proof).unwrap(),
```
Lldenaurois, line 1181: Note: I am not sure what the best approach is here.

Since the `obtain_chunks` function (c.f. [here](https://github.com/paritytech/polkadot/blob/master/erasure-coding/src/lib.rs#L136)) and `branches` function (c.f. [here](https://github.com/paritytech/polkadot/blob/686e3180af8fee86b2b541ffb9b09d7bc66eba9c/erasure-coding/src/lib.rs#L276).

One option would be to have the erasure-coding return the Proof as oppose to the `AsRef<[u8]>` and the `Vec<Vec<u8>>`. But before I go ahead and do that it maybe be best to sync up regarding the proper approach.

node/core/av-store/src/tests.rs
```diff
@@ -16,11 +16,14 @@
 
 use super::*;
 
+use std::convert::TryFrom;
```
Lldenaurois, line 19: Since there's no use of this trait anywhere (it seems), I believe it may be incorrect to implement the `TryFrom` trait as oppose to just making a Method on `Proof`.

node/primitives/src/lib.rs
```diff
@@ -287,6 +291,96 @@ pub struct AvailableData {
...
+impl Encode for Proof {
+	fn size_hint(&self) -> usize {
+		MERKLE_NODE_MAX_SIZE * MERKLE_PROOF_MAX_DEPTH
+	}
+
+	fn using_encoded<R, F: FnOnce(&[u8]) -> R>(&self, f: F) -> R {
+		let mut out = Vec::new();
+		for element in self.0.iter() {
+			let temp = element.as_vec();
+			let mut element_vec = [0u8; MERKLE_NODE_MAX_SIZE];
```
Lldenaurois, line 355: This is not the tersest way to store Merkle proofs, but it works. I wanted to have something working quickly in order to get some of the higher-level questions out of the way.

Strangely, the Merkle proofs generated in the polkadot tests seem to always an outer length 2, which I currently don't fully understand. The inner length varies from 3 to 347 and I also don't fully understand why.

Based on how we store these proofs as a Vec<Vec<u8>>, we can choose the best way to bound the inner and out vectors.

node/primitives/src/lib.rs
```diff
@@ -51,6 +52,9 @@ pub use disputes::{
 	SignedDisputeStatement, UncheckedDisputeMessage, ValidDisputeVote,
 };
 
+const MERKLE_NODE_MAX_SIZE: usize = 347;
+const MERKLE_PROOF_MAX_DEPTH: usize = 3;
```
Lldenaurois, line 56: Note that these constants would not work in practice. I've just set them to what I observed being crated in the tests. It is my belief that the depth is no more than 3 (in fact it's not more than 2) throughout all tests because we don't ever create a Merkle Tree with that many elements, i.e. intermediate nodes.

The correct values to use here are:

MERKLE_NODE_MAX_SIZE = 512;
MERKLE_PROOF_MAX_DEPTH = 8;

I just found it interesting that the node max size was uneven. I may be mistaken, but I'm quite certain the max proof node created is 347 bytes.

eskimor:
With the limits in place, how big can a proof become now?

Lldenaurois:
4096 (= 512 * 8) bytes is the maximum size of the Proof with this change, as oppose to the 500KB protocol-level limit.

node/core/av-store/src/tests.rs
drahnr: How would it be incorrect?

node/primitives/src/lib.rs
```diff
@@ -51,6 +52,9 @@ pub use disputes::{
 	SignedDisputeStatement, UncheckedDisputeMessage, ValidDisputeVote,
 };
 
+const MERKLE_NODE_MAX_SIZE: usize = 512;
+const MERKLE_PROOF_MAX_DEPTH: usize = 8;
```
drahnr, line 56: Adding comments as to why these were chosen as they are would be appreciated.

node/primitives/src/lib.rs
```diff
@@ -287,6 +291,82 @@ pub struct AvailableData {
...
+
+impl Proof {
+	/// This function allows to convert back to the standard nested Vec format
+	pub fn as_vec(&self) -> Vec<Vec<u8>> {
+		self.0.as_vec().iter().map(|v| v.as_vec().clone()).collect()
+	}
+}
+
+impl TryFrom<Vec<Vec<u8>>> for Proof {
+	type Error = &'static str;
```
drahnr, line 306: I would argue for a

```
#[derive(thiserror::Error)]
enum Error {
    #[error("erkle max proof depth exceeded {0} > {} .", MERKLE_PROOF_MAX_DEPTH)]
    MerkleProofDepthExceeded(depth),
    // ..
}
```

node/primitives/src/lib.rs
```diff
@@ -287,6 +291,82 @@ pub struct AvailableData {
...
+
+	fn try_from(input: Vec<Vec<u8>>) -> Result<Self, Self::Error> {
+		if input.len() > MERKLE_PROOF_MAX_DEPTH {
+			return Err("Merkle max proof depth exceeded.")
+		}
+		let mut out = Vec::new();
+		for element in input.into_iter() {
+			let data: BoundedVec<u8, 1, MERKLE_NODE_MAX_SIZE> =
+				BoundedVec::from_vec(element).map_err(|_| "Merkle node max size exceeded.")?;
```
drahnr, line 315: CC

node/primitives/src/lib.rs
```diff
@@ -287,6 +291,82 @@ pub struct AvailableData {
...
+			out.push(data);
+		}
+		Ok(Proof(BoundedVec::from_vec(out).expect("Buffer size is deterined above. QED")))
```
drahnr, line 318:
```suggestion
		Ok(Proof(BoundedVec::from_vec(out).expect("Buffer size is already checked. QED")))
```

node/primitives/src/lib.rs
```diff
@@ -287,6 +291,82 @@ pub struct AvailableData {
 	pub validation_data: PersistedValidationData,
 }
 
+/// This is a convenience type to allow the Erasure chunk proof to Decode into a nested BoundedVec
+#[derive(PartialEq, Eq, Clone, Debug, Hash)]
+pub struct Proof(BoundedVec<BoundedVec<u8, 1, MERKLE_NODE_MAX_SIZE>, 1, MERKLE_PROOF_MAX_DEPTH>);
+
+impl Proof {
+	/// This function allows to convert back to the standard nested Vec format
+	pub fn as_vec(&self) -> Vec<Vec<u8>> {
```
drahnr, line 300: nit/future: Do we truely require a `Vec<Vec<_>>`? Or could we avoid all those allocations by employing a single `BoundedVec<_>` and providing a `Vec<&[]>` or even `&[&[]]`? Again, not for this PR, a potential future optimization.

drahnr (changes requested):
I would prefer the error handling to _not_ be static strings, with the exception of `CodecError` unless there is a particular requirement to do so.

Besides that, :+1:

node/primitives/src/lib.rs
```diff
@@ -302,20 +304,33 @@ impl Proof {
...
+pub enum MerkleProofError {
+	#[error("Merkle max proof depth exceeded {0} > {} .", MERKLE_PROOF_MAX_DEPTH)]
+	/// This error signifies that the Proof length exceeds the trie's max depth
+	MerkleProofDepthExceeded(usize),
+
+	#[error("Merkle node max size exceeded {0} > {} .", MERKLE_NODE_MAX_SIZE)]
+	/// This error signifies that a Proof node exceeds the 16-ary max node size
+	MerkleProofNodeSizeExceeded(usize),
+}
+
```
drahnr, line 318: :heart:

node/primitives/src/lib.rs
```diff
@@ -52,7 +52,9 @@ pub use disputes::{
 	SignedDisputeStatement, UncheckedDisputeMessage, ValidDisputeVote,
 };
 
+// For a 16-ary Merkle Prefix Trie, we can expect at most 16 32-byte hashes per node.
```
drahnr, line 55: :+1:

node/primitives/src/lib.rs
```diff
@@ -302,20 +304,33 @@ impl Proof {
 	}
 }
 
+#[derive(thiserror::Error, Debug)]
+///
+pub enum MerkleProofError {
```
drahnr, line 309: nit: if `#[allow(missing_docs)]` works, that'd be preferable

node/primitives/src/lib.rs
```diff
@@ -302,20 +304,33 @@ impl Proof {
...
 		for element in input.into_iter() {
-			let data: BoundedVec<u8, 1, MERKLE_NODE_MAX_SIZE> =
-				BoundedVec::from_vec(element).map_err(|_| "Merkle node max size exceeded.")?;
+			let length = element.len();
+			let data: BoundedVec<u8, 1, MERKLE_NODE_MAX_SIZE> = BoundedVec::from_vec(element)
+				.map_err(|_| Self::Error::MerkleProofNodeSizeExceeded(length))?;
 			out.push(data);
 		}
-		Ok(Proof(BoundedVec::from_vec(out).expect("Buffer size is deterined above. QED")))
+		Ok(Proof(BoundedVec::from_vec(out).expect("Buffer size is deterined above. qed")))
```
drahnr, line 333: nit: isn't it `determined`?

drahnr (approved):

erasure-coding/src/lib.rs
```diff
@@ -258,13 +260,12 @@ impl<'a, I: AsRef<[u8]>> Iterator for Branches<'a, I> {
...
 			Some(_) => {
-				let nodes = recorder.drain().into_iter().map(|r| r.data).collect();
+				let nodes: Vec<Vec<u8>> = recorder.drain().into_iter().map(|r| r.data).collect();
 				let chunk = self.chunks.get(self.current_pos).expect(
 					"there is a one-to-one mapping of chunks to valid merkle branches; qed",
 				);
-
 				self.current_pos += 1;
-				Some((nodes, chunk.as_ref()))
+				Proof::try_from(nodes).ok().map(|proof| (proof, chunk.as_ref()))
```
eskimor, line 268: Not sure it is the right behavior to just let the iterator end if the `Proof` could not be constructed. Can this even happen, without the code having a logic error?

node/network/availability-distribution/src/requester/fetch_task/mod.rs
```diff
@@ -363,7 +363,7 @@ impl RunningTask {
 
 	fn validate_chunk(&self, validator: &AuthorityDiscoveryId, chunk: &ErasureChunk) -> bool {
 		let anticipated_hash =
-			match branch_hash(&self.erasure_root, &chunk.proof, chunk.index.0 as usize) {
+			match branch_hash(&self.erasure_root, &chunk.proof_as_vec(), chunk.index.0 as usize) {
```
eskimor, line 366: Can't we fix `branch_hash` in order to get rid of that conversion?

node/primitives/src/lib.rs
```diff
@@ -287,6 +293,95 @@ pub struct AvailableData {
...
+			let length = element.len();
+			let data: BoundedVec<u8, 1, MERKLE_NODE_MAX_SIZE> = BoundedVec::from_vec(element)
+				.map_err(|_| Self::Error::MerkleProofNodeSizeExceeded(length))?;
+			out.push(data);
+		}
+		Ok(Proof(BoundedVec::from_vec(out).expect("Buffer size is deterined above. qed")))
+	}
+}
+
+impl Decode for Proof {
```
eskimor, line 337: I am missing tests, to make sure Encode/Decode works and is compatible to what we had before.

PR #3627, opened 2021-08-11
Make most XCM APIs accept an Into<MultiLocation> where MultiLocation is accepted

KiChjang:
This PR makes it so that users do not need to sweat the details of constructing a `MultiLocation` object when using APIs like `send_xcm` or `execute_xcm` by allowing for syntax such as the following:

```rust
send_xcm(Parent, ...);
execute_xcm((1, Parachain(id.into())), ...);
```

A slight issue that this may introduce is that trait functions that now accept `impl Into<MultiLocation>` would now make its containing trait not be object-safe, since these functions are now generic. In practical terms, this means a type that implements e.g. `SendXcm` cannot be typecasted into a `Box<dyn SendXcm>` or a `&dyn SendXcm` anymore.


xcm/xcm-simulator/example/src/lib.rs
```diff
@@ -166,7 +166,7 @@ mod tests {
 		ParaA::execute_with(|| {
 			assert_ok!(ParachainPalletXcm::send_xcm(
 				Here,
-				MultiLocation::new(1, X1(Parachain(2))),
+				(1, Parachain(2)),
```
KiChjang, line 169: This makes me feel like I should implement `From<(Parent, Junction)>` for `MultiLocation`. Can do in this PR or in a future one.

KiChjang: Done in #3635.

gavofyork (approved):

shawntabrizi (approved):

KiChjang:
bot merge

PR #3628, opened 2021-08-12
check runtime version in staking miner

kianenigma:
- will check the runtime version at startup + upon monitor and fail-safe if a runtime upgrade has happened. 
- print balances in a nicer way. 
- do more checks on dry-run
- print weight and deposit info


utils/staking-miner/src/dry_run.rs
```diff
@@ -35,18 +36,42 @@ fn force_create_snapshot<T: EPM::Config>(ext: &mut Ext) -> Result<(), Error> {
...
+		let deposit = EPM::Pallet::<T>::deposit_for(&raw_solution, snapshot_size);
+		log::info!(
+			target: LOG_TARGET,
+			"solution score {:?} / deposit {:?}",
+			&raw_solution.score.iter().map(|x| Token::from(*x)).collect::<Vec<_>>(),
+			Token::from(deposit),
+		);
+	});
+
+	let info = rpc::<Vec<u8>>(client, "payment_queryInfo", params! { extrinsic }).await;
```
kianenigma, line 73: we're still working on this with @niklasad1, actual return type should be fixed.

niklasad1: this is the API https://github.com/paritytech/substrate/blob/master/frame/transaction-payment/rpc/src/lib.rs#L90?

ok, this should return https://docs.rs/pallet-transaction-payment-rpc-runtime-api/3.0.0/pallet_transaction_payment_rpc_runtime_api/struct.RuntimeDispatchInfo.html instead?

basically, if you put `Vec<u8>` here and it returns a `Object/Map` decoding it will fail because of the generics.....

kianenigma: Have tried that as well, it also fails. The `Vec<u8>` was an experiment.

kianenigma: Okay I've no idea what's happening, now it worked.

niklasad1: Otherwise put `JsonValue` here and convert it manually, but based on the [serialize impl in RuntimeDispatchInfo](https://docs.rs/pallet-transaction-payment/3.0.0/src/pallet_transaction_payment/types.rs.html#147) it should be serialized to a `Map/Object`.

niklasad1 (approved):
LGTM

utils/staking-miner/src/dry_run.rs
```diff
@@ -64,9 +64,10 @@ async fn print_info<T: EPM::Config>(
 		let deposit = EPM::Pallet::<T>::deposit_for(&raw_solution, snapshot_size);
 		log::info!(
 			target: LOG_TARGET,
-			"solution score {:?} / deposit {:?}",
+			"solution score {:?} / deposit {:?} / length {:?}",
 			&raw_solution.score.iter().map(|x| Token::from(*x)).collect::<Vec<_>>(),
 			Token::from(deposit),
+			raw_solution.encode().len(),
```
emostov, line 70: This seems quite expensive just to do a log message - is there somewhere else this is already encoded and we can just pul the length from that? Or maybe we can use `size_hint(&self) -> usize`?

utils/staking-miner/src/dry_run.rs
```diff
@@ -76,24 +112,40 @@ macro_rules! dry_run_cmd_for { ($runtime:ident) => { paste::paste! {
...
+			transfers funds out of it, or get slashed. If it does not exist at this point, \
+			it is likely due to a bug, or the signer got slashed. Terminating."
+		);
 		let tip = 0 as Balance;
 		let era = sp_runtime::generic::Era::Immortal;
-		let extrinsic = ext.execute_with(|| create_uxt(raw_solution, witness, signer.clone(), nonce, tip, era));
+		let extrinsic = ext.execute_with(|| create_uxt(raw_solution.clone(), witness, signer.clone(), nonce, tip, era));
 
 		let bytes = sp_core::Bytes(extrinsic.encode().to_vec());
+		print_info::<Runtime>(client, &mut ext, &raw_solution, bytes.clone()).await;
```
emostov, line 135: what would happen if we didn't block on this call? Would it still print correctly? If so it could be  a performance improvement

utils/staking-miner/src/monitor.rs
```diff
@@ -86,6 +86,9 @@ macro_rules! monitor_cmd_for { ($runtime:tt) => { paste::paste! {
 			let hash = now.hash();
 			log::debug!(target: LOG_TARGET, "new event at #{:?} ({:?})", now.number, hash);
 
+			// if the runtime version has changed, terminate
+			crate::check_versions::<Runtime>(client, false).await?;
```
emostov, line 90: Not neccessary for this PR, but is there a way we could parrell-ize the calls to `check_versions`, `ensure_signed_phase` & `create_election_ext`? Basicaly a Promise.all and then we can analyze the results of each independtly and figure out how to proceed

utils/staking-miner/src/signer.rs
```diff
@@ -69,6 +73,12 @@ pub(crate) async fn read_signer_uri<
 	let _info = get_account_info::<T>(&client, &account, None)
```
emostov, line 73: why underscore?

utils/staking-miner/src/main.rs
```diff
@@ -296,29 +302,20 @@ struct Opt {
 async fn create_election_ext<T: EPM::Config, B: BlockT>(
```
emostov, line 302: Maybe update this doc comment to `Build the `Ext` at `hash` with all the data of `ElectionProviderMultiPhase` and any `additional` pallets.

emostov (approved):

utils/staking-miner/src/dry_run.rs
kianenigma: yeah it could, but it would make the code look less ergonomic and since this is dry-run and not at all time-sensitive, I intentionally overlooked it.

utils/staking-miner/src/monitor.rs
kianenigma: yeah this is a good improvement. Also a critical improvement. is that currently we need to scrape the data of the whole pallet to check if we have a solution in this round or not (since we pass the entire `ext`), this can be done much better by just querying the appropriate storage item (see `ensure_no_previous_solution`). 

we should be able to early-exit from this check, without the need to scrape the entire `ext`.

utils/staking-miner/src/dry_run.rs
kianenigma: not sure what you mean, elaborate?

utils/staking-miner/src/monitor.rs
kianenigma: https://github.com/paritytech/polkadot/issues/3646

utils/staking-miner/src/signer.rs
kianenigma: because it is only used in logging and if that log target is not enabled, this variable would not be used at all.

kianenigma:
bot merge

PR #3629, opened 2021-08-12
XCM v2: Scripting, Query responses & Error reporting

gavofyork:
Moves XCM away from a hybrid script/structured language and into a pure script model. This means no recusion and an equivalence between `Order` and instruction (`Xcm`).

Advantages are:
- Simpler to learn & understand - it's just another scripting language.
- No issues with stack depth. 
- No need for shallow/deep weight distinction.
- No need for distinguishing between order & instruction, particularly problematic for juggling `BuyExecution` and `ReportOutcome`.

Disadvantages are:
- Less overall structure, which means a greater possibility of nonsensical or buggy fragments and of making mistakes in construction.
- More global state in the XCM interpreter: `holding` is now available to instructions as well as orders, outcome reporting must be done at the global level.

To alleviate the latter issue and especially if global machine state increases further, a later version of XCM may introduce instructions which utilise hierarchy to reduce global state.

Also, introduces query response handling and XCM instruction outcome reporting and `pallet_xcm` can (and probably should) now be used as an `OnResponse` handler.

This introduces three new `pub` functions to `pallet_xcm::Pallet`:
- `on_report`: A "do it all" function which is able to alter a given XCM into one whose outcome (success/failure) will be passed into a provided handler (which must be a dispatchable function).
- `new_query`: A helper function to return a unique and registered query ID.
- `new_notify_query`: A helper function to return a unqiue query ID registered to dispatch a given handler once the outcome is returned.

Also:
- New XCM instruction `ClearOrigin` which removes the origin (thereby preventing further instructions from executing with its associated privileges, useful for `InitiateTeleport` and `InitiateReserveWithdraw`).
- New XCM instruction `ReportOutcome`.
Alters `Response` type, `Order::QueryBalance` and `Xcm::QueryResponse`.

All `Xcm` items and `Order` items are merged into `Instruction` enum. `Xcm` becomes a unit struct of inner type `Vec<Instruction>`.

All `Xcm` items which included a field `Vec<Order>` _*to be executed in the local context*_ now have those fields removed (they're no longer needed since the `Order`s can be placed as `Instruction`s immediately afterwards in the `Xcm` vector). For the remote context, `xcm: Xcm` is used as the field name/type.

## Compatibility

`v1` messages can be auto-converted to `v2` provided:

- `BuyExecution::instructions` is empty.
- `RelayedFrom` is unused.

`v2` messages can be auto-converted to `v1` provided:

- Only a single item corresponding to a `v1::Xcm` item is used for local execution in a single `v2::Xcm` and it must be at the begining.
- `BuyExecution::weight_limit` is `Limited`.
- `ClearOrigin` if used, may only follow immediately after `ReceiveTeleportedAsset` or `ReserveAssetDeposited`. The latter two may not appear with it.
- `DescendOrigin`, `ReportOutcome`, `RefundSurplus` are not used.

### XCM fragments

- `Xcm::` and `Order::` become `Instruction::`.
- `orders`/`effects` fields describing effects on the remote side are renamed `xcm`. 
- For `WithdrawAsset`, `ReceiveTeleportedAsset`, `DepositReserveAsset` have no `effects` field; contents should be moved to follow the instruction in the containing `Vec`.
- `BuyExecution`:
  - New field `weight_limit`, which replaces `debt` and `weight`; it should be set to `Limited` with the inner value being the sum of `weight` and `debt`, or, alternatively just `Unlimited`.
  - If `instructions` is not empty, then contents should be moved to follow `BuyExecution` in the containing `Vec`.
  - Only fields remaining are `fees` and `weight`.
- `RelayedFrom` becomes `DescendOrigin`; `RelayedFrom::instruction` removed and contents should be moved to follow `RelayedFrom` in the containing `Vec`.

### XCM Pallet

`pallet_xcm::Config` now contains two more items `Origin` and `Call`; they should just be defined to the corresponding outer types:

```
type Origin = Origin;
type Call = Call;
```

`pallet_xcm` has a special `Origin` which must be specified in the `construct_runtime!` macro. 

## TODO
- [ ] Test `on_report`
- [ ] Test `new_query`
- [ ] Test `new_query_notify`
- [ ] Test conversions `v1` -> `v2`
- [ ] Test conversions `v2` -> `v1`


xcm/src/v0/traits.rs
```diff
@@ -186,7 +186,7 @@ impl<C> ExecuteXcm<C> for () {
 ///
 /// # Example
 /// ```rust
-/// # use xcm::v0::{MultiLocation, Xcm, Junction, Error, OriginKind, SendXcm, Result};
+/// # use xcm::v2::{MultiLocation, Xcm, Junction, Junctions, Error, OriginKind, SendXcm, Result, Parent};
```
KiChjang, line 189: Doesn't quite make sense to me that this uses v2 types when we're under the v0 directory...

apopiak:
Also: `Noop` was removed, right? Is it (sufficiently) replaced by `Xcm(vec![])`?

xcm/pallet-xcm/src/lib.rs
```diff
@@ -319,19 +353,130 @@ pub mod pallet {
...
+			message: &mut Xcm<()>,
+			responder: MultiLocation,
+			notify: impl Into<<T as Config>::Call>,
+			timeout: T::BlockNumber,
+		) {
+			let dest = T::LocationInverter::invert_location(&responder);
+			let notify: <T as Config>::Call = notify.into();
+			let max_response_weight = notify.get_dispatch_info().weight;
+			let query_id = Self::new_notify_query(responder, notify, timeout);
+			message.0.insert(0, ReportOutcome { dest, query_id, max_response_weight });
```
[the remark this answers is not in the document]
gavofyork: Are you suggesting that we alter the `Xcm` datastructure, which is used by numerous crates, codebases, runtimes and pallets, and which is chosen for high performance under most of those circumstances, and change it to something else based on one single use in one single utility function?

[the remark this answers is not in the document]
apopiak: No. That would indeed not make sense at all. 
Sorry for the noise, I was just sharing an observation.

xcm/src/lib.rs
```diff
@@ -52,6 +53,94 @@ impl Decode for Unsupported {
...
 }
 
+/// A single `MultiLocation` value, together with its version code.
+#[derive(Derivative, Encode, Decode)]
+#[derivative(Clone(bound = ""), Eq(bound = ""), PartialEq(bound = ""), Debug(bound = ""))]
+#[codec(encode_bound())]
+#[codec(decode_bound())]
+pub enum VersionedMultiLocation {
+	V0(v0::MultiLocation),
+	V1(v1::MultiLocation),
```
KiChjang, line 63: Is the intention here to indicate that types like `MultiLocation` can evolve independently from the XCM versioning? If so, then I worry that this may introduce complexity with versioning mismatches, an example being v1 XCM messages working on v2 MultiLocations but not on v0 MultiLocations.

I'm inclined towards a model where all XCM types follow the versioning scheme of XCM in general. That is, if we bump XCM to v2, then all its types would automatically be v2. I realize this is the existing behavior, so perhaps I'm missing some more of the context behind this change?

xcm/src/v2/mod.rs
```diff
@@ -0,0 +1,675 @@
...
+	///
+	/// Errors:
+	ClearOrigin,
+
+	/// Mutate the origin to some interior location.
+	///
+	/// Kind: *Instruction*
+	///
+	/// Errors:
+	DescendOrigin(InteriorMultiLocation),
```
KiChjang, line 310: There's a break in convention with this enum variant -- all other variants use named fields even when they only have 1 field, whereas `DescendOrigin` in particular does not. Is this intended?

gavofyork: yes - others will be switched over in due course of this PR.

xcm/src/lib.rs
gavofyork: `VersionedMultiLocation` is needed for when we store a `MultiLcation` value between potential runtime upgrades.

KiChjang: Ah okay, I see what's going on now. Because these types are exposed in a public API in XCM, there may be situations where an incoming XCM is using an older representation of these types, and we'll need to know how to parse them accordingly.

As for my own concerns, I think I can answer them -- the complexity of maintaining an up-to-date `MultiLocation` or other types is inherent within the design of the system, and the versioning scheme is simply a way to make it explicit and manageable. It's also not going to be as complex as I think it would be, because all incoming XCM messages are always versioned, and each XCM version locks onto a specific version of `MultiLocation` and so is the case for the other types as well.

A follow-up question here is whether we should be seeing more versioned types. I think I only see it being used in `QueryStatus`, but does it make sense for functions that take in a `MultiLocation` to instead take in its versioned counterpart?

gavofyork: For dispatchables (i.e. those in transactions) yes it does. Ideally it should be introduced prior to the release of 0.9.10 (XCM v1) since that will change those transaction formats anyway.

As for `QueryStatus`, yes it probably makes sense to introduce versioning to that also.

gavofyork: Now introduced in https://github.com/paritytech/polkadot/pull/3693.

xcm/src/v2/mod.rs
```diff
@@ -0,0 +1,705 @@
...
+	/// This may be used by the XCM author to ensure that later instructions cannot command the
+	/// authority of the origin (e.g. if they are being relayed from an untrusted source, as often
+	/// the case with `ReserveAssetDeposited`).
+	///
+	/// Safety: No concerns.
+	///
+	/// Kind: *Instruction*.
+	///
+	/// Errors:
+	ClearOrigin,
```
KiChjang, line 305: It may be useful to formally describe the kind of virtual machine that the default XCM executor looks like. Right now, we most definitely have a register machine due to the existence of a holding register, but it also has this notion of an origin which I think is also another register?

xcm/src/v2/mod.rs
```diff
@@ -0,0 +1,705 @@
...
+	Ancestor, AncestorThen, AssetId, AssetInstance, BodyId, BodyPart, Fungibility,
+	InteriorMultiLocation, Junction, Junctions, MultiAsset, MultiAssetFilter, MultiAssets,
+	MultiLocation, NetworkId, OriginKind, Parent, ParentThen, WildFungibility, WildMultiAsset,
+};
+
+#[derive(Derivative, Encode, Decode)]
+#[derivative(Clone(bound = ""), Eq(bound = ""), PartialEq(bound = ""), Debug(bound = ""))]
+#[codec(encode_bound())]
+#[codec(decode_bound())]
+pub struct Xcm<Call>(pub Vec<Instruction<Call>>);
```
KiChjang, line 44: I'm wondering if there's a way to enforce some kind of well-formedness with the XCMs, because some instructions don't make sense by their own without another happening prior, e.g. `WithdrawAsset` should have been executed before calling `BuyExecution`. We currently have a sort of idempotence when instructions such as `BuyExecution` is called before `WithdrawAsset` by simply failing to execute, but I wonder if we can leverage Rust's static analysis to aid in XCM authors in writing well-formed XCMs.

It's definitely going to be tricky to think about, since XCMs are not necessarily self-contained and state transitions across XCMs should be allowed (one processed XCM could've already called `WithdrawAsset`, and the next one calls `BuyExecution`).

xcm/pallet-xcm/src/tests.rs
```diff
@@ -1,4 +1,4 @@
-// Copyright 2020 Parity Technologies (UK) Ltd.
+// Copyright 2020 Parity Technologies (UK) Lt dest: (), max_response_weight: () d.
```
KiChjang, line 1: That's a weird kind of company...

xcm/src/v2/mod.rs
xlc: I think the only way is build a state machine and each instruction is an action, and for some state, some actions are not allowed.
Not sure if there is already existing tools to define this, but I image it will be cumbersome to use.
Alternately we make this a proper scripting language and have some linter to check it.

xlc:
What will happen to https://github.com/paritytech/xcm-format ? do we want to keep it up to date?

xcm/src/v2/traits.rs
```diff
@@ -0,0 +1,308 @@
...
+			SendError::DestinationUnsupported => Error::DestinationUnsupported,
+			SendError::ExceedsMaxMessageSize => Error::ExceedsMaxMessageSize,
+		}
+	}
+}
+
+pub type Result = result::Result<(), Error>;
+
+/// Local weight type; execution time in picoseconds.
+pub type Weight = u64;
```
KiChjang, line 119: Really dumb question: did we decide that we can't reuse the weights system in FRAME, and thus we have to define the `Weight` type here?

xcm/xcm-builder/src/barriers.rs
```diff
@@ -19,53 +19,64 @@
...
-				) =>
-				Ok(()),
+		let mut iter = message.0.iter_mut();
+		let i = iter.next().ok_or(())?;
+		match i {
+			ReceiveTeleportedAsset(..) | WithdrawAsset(..) | ReserveAssetDeposited(..) => (),
+			_ => return Err(()),
+		}
+		let mut i = iter.next().ok_or(())?;
+		if let ClearOrigin = i {
```
KiChjang, line 68: Should this be a `while let` here just to eat up all potential `ClearOrigin` instructions that follow? While it's definitely unusual, I don't think we should error out either since `ClearOrigin` is idempotent, if I'm understanding it correctly.

xcm/xcm-builder/src/tests.rs
```diff
@@ -336,32 +344,22 @@ fn paid_transacting_should_refund_payment_for_unused_weight() {
...
-			},
-		],
-	};
+	let message = Xcm::<TestCall>(vec![
+		WithdrawAsset((Parent, 100).into()), // enough for 100 units of weight.
+		BuyExecution { fees, weight_limit: Limited(100) },
+		Transact {
+			origin_type: OriginKind::Native,
+			require_weight_at_most: 50,
+			// call estimated at 70 but only takes 10.
```
KiChjang, line 353: Is it still correct that the weight is estimated at 70? It seemed to look like it's 50 to me according to the code, but I could be mistaken.

xcm/src/v2/mod.rs
gavofyork: A linter could probably identify *definitely* bad programs, but I don't think a general validator is possible.

xcm/src/v2/mod.rs
gavofyork: Yeah, it's a very specialised register machine now.

gavofyork:
> What will happen to https://github.com/paritytech/xcm-format ? Do we want to keep it up to date?

Yes.

> Definitely need some RPC or something to help estimate weights on dest chain.

Exporting the `WeightBounds` impl in a crate would allow runtimes to estimate each other's weights.

> This assumes transaction_version remains the same. Maybe it should be included in the payload somewhere to assert that's the case?

Not sure what you mean here. We already have https://github.com/paritytech/xcm-format/pull/22 as a later change to manage transction versioning for `Transact`. Outcide of `Transact`, I don't think transaction versioning is an issue.

xcm/src/v2/traits.rs
gavofyork: XCM is independent of Frame.

xcm/xcm-builder/src/barriers.rs
gavofyork: Yeah I guess it can be more forgiving.

xcm/xcm-executor/src/lib.rs
```diff
@@ -227,142 +286,130 @@ impl<Config: config::Config> XcmExecutor<Config> {
...
+				Config::XcmSender::send_xcm(reserve, Xcm(message)).map_err(Into::into)
 			},
-			Order::InitiateTeleport { assets, dest, effects } => {
+			InitiateTeleport { assets, dest, xcm } => {
 				// We must do this first in order to resolve wildcards.
-				let assets = holding.saturating_take(assets);
+				let assets = self.holding.saturating_take(assets);
 				for asset in assets.assets_iter() {
-					Config::AssetTransactor::check_out(&origin, &asset);
+					Config::AssetTransactor::check_out(&dest, &asset);
```
KiChjang, line 353: This checks out assets from the destination instead of the origin?

PR #3631, opened 2021-08-12
GHA: extrinsic ordering, new filter to generate a summary

chevdor:
In #3620, a new Github Worflow was added to compare metadata between releases and catch changes that need to trigger a transaction version bump.

This PR appends a new `Summary` section where the *interesting* changes are highlighted (= the changes that needs special attention such as removal, index changes and indexes decreasing).

A sample output can be seen [here](https://github.com/chevdor/polkadot/runs/3311714422?check_suite_focus=true). Note that the CI run in this example was run to compare Westend and Kusama for the solely purpose to show case while working on more differences.

The summary for this (fantasy) run looks like:

------------------------------ SUMMARY ------------------------------- 
⚠️ This filter is here to help spotting changes that should be reviewed carefully. 
⚠️ It catches only index changes, deletions and index decreases. 
## Deletions
 14: [-] modules: Sudo, ParachainsConfiguration, ParasInclusion, ParasInherent, ParasScheduler

## Index changes
 27: [Identity] idx: 17 -> 25 (calls: 15, storage: 4)
28: [Recovery] idx: 18 -> 27 (calls: 9, storage: 3)
29: [Vesting] idx: 19 -> 28 (calls: 4, storage: 1)
30: [Scheduler] idx: 20 -> 29 (calls: 6, storage: 3)
31: [Proxy] idx: 22 -> 30 (calls: 10, storage: 2)
32: [Multisig] idx: 23 -> 31 (calls: 4, storage: 2)
33: [ElectionProviderMultiPhase] idx: 24 -> 37 (calls: 4, storage: 10)
36: [Paras] idx: 47 -> 56 (calls: 5, storage: 13 -> 17)
38: [Registrar] idx: 60 -> 70 (calls: 6, storage: 3)
39: [Slots] idx: 61 -> 71 (calls: 3, storage: 1)
40: [Auctions] idx: 63 -> 72 (calls: 3, storage: 4)
41: [Crowdloan] idx: 64 -> 73 (calls: 8, storage: 4)

## Decreases
 n/a 
----------------------------------------------------------------------

### DIFF
--- a/.github/workflows/extrinsic-ordering-check-from-bin.yml
+++ b/.github/workflows/extrinsic-ordering-check-from-bin.yml
@@ -6,7 +6,7 @@ on:
     inputs:
       reference_url:
         description: The WebSocket url of the reference node
-        default: wss://rpc.polkadot.io
+        default: wss://kusama-rpc.polkadot.io
         required: true
       binary_url:
         description: A url to a Linux binary for the node containing the runtime to test
@@ -14,7 +14,7 @@ on:
         required: true
       chain:
         description: The name of the chain under test. Usually, you would pass a local chain
-        default: polkadot-local
+        default: kusama-local
         required: true
 
 jobs:
@@ -27,6 +27,8 @@ jobs:
       REF_URL: ${{github.event.inputs.reference_url}}
 
     steps:
+      - uses: actions/checkout@v2
+
       - name: Fetch binary
         run: |
           echo Fetching $BIN_URL
@@ -46,17 +48,26 @@ jobs:
           echo "Date: $(date)" >> output.txt
           echo "Reference: $REF_URL" >> output.txt
           echo "Target version: $VERSION" >> output.txt
-          echo "-------------------------------------------" >> output.txt
+          echo "Chain: $CHAIN" >> output.txt
+          echo "----------------------------------------------------------------------" >> output.txt
+
+      - name: Pull polkadot-js-tools image
+        run: docker pull jacogr/polkadot-js-tools
 
       - name: Compare the metadata
         run: |
-          CMD="docker run --network host jacogr/polkadot-js-tools metadata $REF_URL ws://localhost:9944"
+          CMD="docker run --pull always --network host jacogr/polkadot-js-tools metadata $REF_URL ws://localhost:9944"
           echo -e "Running:\n$CMD"
           $CMD >> output.txt
           sed -z -i 's/\n\n/\n/g' output.txt
+          cat output.txt | egrep -n -i ''
+          SUMMARY=$(./scripts/github/extrinsic-ordering-filter.sh output.txt)
+          echo -e $SUMMARY
+          echo -e $SUMMARY >> output.txt
 
       - name: Show result
-        run: cat output.txt
+        run: |
+          cat output.txt
 
       - name: Stop our local node
         run: pkill polkadot
--- a/scripts/github/extrinsic-ordering-filter.sh
+++ b/scripts/github/extrinsic-ordering-filter.sh
@@ -1,0 +1,55 @@
+#!/usr/bin/env bash
+# This script is used in a Github Workflow. It helps filtering out what is interesting
+# when comparing metadata and spot what would require a tx version bump.
+
+# shellcheck disable=SC2002,SC2086
+
+FILE=$1
+
+# Higlight indexes that were deleted
+function find_deletions() {
+    echo "\n## Deletions\n"
+    RES=$(cat "$FILE" | grep -n '\[\-\]' | tr -s " ")
+    if [ "$RES" ]; then
+        echo "$RES" | awk '{ printf "%s\\n", $0 }'
+    else
+        echo "n/a"
+    fi
+}
+
+# Highlight indexes that have been deleted
+function find_index_changes() {
+    echo "\n## Index changes\n"
+    RES=$(cat "$FILE" | grep -E -n -i 'idx:\s*([0-9]+)\s*(->)\s*([0-9]+)' | tr -s " ")
+    if [ "$RES" ]; then
+        echo "$RES" | awk '{ printf "%s\\n", $0 }'
+    else
+        echo "n/a"
+    fi
+}
+
+# Highlight values that decreased
+function find_decreases() {
+    echo "\n## Decreases\n"
+    OUT=$(cat "$FILE" | grep -E -i -o '([0-9]+)\s*(->)\s*([0-9]+)' | awk '$1 > $3 { printf "%s;", $0 }')
+    IFS=$';' LIST=("$OUT")
+    unset RES
+    for line in "${LIST[@]}"; do
+        RES="$RES\n$(cat "$FILE" | grep -E -i -n \"$line\" | tr -s " ")"
+    done
+
+    if [ "$RES" ]; then
+        echo "$RES" | awk '{ printf "%s\\n", $0 }' | sort -u -g | uniq
+    else
+        echo "n/a"
+    fi
+}
+
+echo "\n------------------------------ SUMMARY -------------------------------"
+echo "\n⚠️ This filter is here to help spotting changes that should be reviewed carefully."
+echo "\n⚠️ It catches only index changes, deletions and value decreases".
+
+find_deletions "$FILE"
+find_index_changes "$FILE"
+find_decreases "$FILE"
+echo "\n----------------------------------------------------------------------\n"


.github/workflows/extrinsic-ordering-check-from-bin.yml
```diff
@@ -46,17 +48,26 @@ jobs:
           echo "Date: $(date)" >> output.txt
           echo "Reference: $REF_URL" >> output.txt
           echo "Target version: $VERSION" >> output.txt
-          echo "-------------------------------------------" >> output.txt
+          echo "Chain: $CHAIN" >> output.txt
+          echo "----------------------------------------------------------------------" >> output.txt
+
+      - name: Pull polkadot-js-tools image
+        run: docker pull jacogr/polkadot-js-tools
```
TriplEight, line 55: could add `--pull=always` to `docker run` below instead. But it doesn't make much sense since GHA are executed on the preemptible VMs, unless you'd bring docker cache with you.

TriplEight (changes requested):

.github/workflows/extrinsic-ordering-check-from-bin.yml
chevdor: I actually tried using caching of the docker images with GHA which ended up a hassle for a very low gain in terms of time. So since I am not using caching `--pull=always` is implied. Do you see a benefit to add the option nonetheless? That would even get on the way to supporting caching later on if someone wants to give it a shot.

TriplEight: > Do you see a benefit to add the option nonetheless?

Just to be more explicit.

> That would even get on the way to supporting caching

Then it's better to move it to GitLab CI anyway.

chevdor: done

shawntabrizi:
What is an index change vs an index decrease?

chevdor:
@shawntabrizi you can see a sample input [here](https://github.com/chevdor/polkadot/runs/3299630664?check_suite_focus=true) under `result`.

I guess I could be more explicit indeed.

Obviously the deletion of an index is something noticeable we want to spot.

An index change is significant because that means a call/module will not longer be reachable at the previous index.

** Decreases** are no **index decreases** they spot decreasing values (and we would hopefully never see one).
One case I can think of is the runtime version going from a value to a lower one, which is something we would want to highlight (and block !).

chevdor:
So **index changes** are patterns like:
```
idx: X -> Y
```

whereas **decreases** are patterns like:
```
N -> N - X 
```
with X positive.

chevdor:
@shawntabrizi I see where you saw *index decrease*, I can fix that if you think this is confusing.
To be honnest, this PR should be short lived and hopefully soon replaced by something better.

scripts/github/extrinsic-ordering-filter.sh
```diff
@@ -0,0 +1,52 @@
...
+    if [ "$RES" ]; then
+        echo "$RES" | awk '{ printf "%s\\n", $0 }' | sort -u -g
+    else
+        echo "n/a"
+    fi
+}
+
+echo "\n------------------------------ SUMMARY -------------------------------"
+echo "\n⚠️ This filter is here to help spotting changes that should be reviewed carefully."
+echo "\n⚠️ It catches only index changes, deletions and index decreases".
```
apopiak, line 47:
```suggestion
echo "\n⚠️ It catches only index changes, deletions and value decreases".
```
based on the discussion above, like this?

s3krit (changes requested):
Looks good to me. One thing I would request is that we fix the [shellcheck](https://github.com/koalaman/shellcheck) nits in `extrinsic-ordering-filter.sh` - I'm happy to do that if you say it's ok for me to push to your branch. I'll also make a CI/CD ticket for us to add this as a check (probs a Github Action), since shellcheck really is indispensable for maintaining shell scripts. Full shellcheck output below:

```
 % shellcheck scripts/github/extrinsic-ordering-filter.sh 

In scripts/github/extrinsic-ordering-filter.sh line 4:
#!/usr/bin/env bash
^-- SC1128: The shebang must be on the first line. Delete blanks and move comments.


In scripts/github/extrinsic-ordering-filter.sh line 9:
    echo "\n## Index changes\n"
         ^--------------------^ SC2028: echo may not expand escape sequences. Use printf.


In scripts/github/extrinsic-ordering-filter.sh line 10:
    RES=$(cat $FILE | egrep -n -i 'idx:\s*([0-9]+)\s*(->)\s*([0-9]+)' | tr -s " ")
              ^---^ SC2086: Double quote to prevent globbing and word splitting.
              ^---^ SC2002: Useless cat. Consider 'cmd < file | ..' or 'cmd file | ..' instead.
                      ^---^ SC2196: egrep is non-standard and deprecated. Use grep -E instead.

Did you mean: 
    RES=$(cat "$FILE" | egrep -n -i 'idx:\s*([0-9]+)\s*(->)\s*([0-9]+)' | tr -s " ")


In scripts/github/extrinsic-ordering-filter.sh line 19:
    echo "\n## Deletions\n"
         ^----------------^ SC2028: echo may not expand escape sequences. Use printf.


In scripts/github/extrinsic-ordering-filter.sh line 20:
    RES=$(cat $FILE | grep -n '\[\-\]' | tr -s " ")
              ^---^ SC2086: Double quote to prevent globbing and word splitting.
              ^---^ SC2002: Useless cat. Consider 'cmd < file | ..' or 'cmd file | ..' instead.

Did you mean: 
    RES=$(cat "$FILE" | grep -n '\[\-\]' | tr -s " ")


In scripts/github/extrinsic-ordering-filter.sh line 29:
    echo "\n## Decreases\n"
         ^----------------^ SC2028: echo may not expand escape sequences. Use printf.


In scripts/github/extrinsic-ordering-filter.sh line 30:
    OUT=$(cat $FILE | egrep -i -o '([0-9]+)\s*(->)\s*([0-9]+)' | awk '$1 > $3 { printf "%s;", $0 }')
              ^---^ SC2086: Double quote to prevent globbing and word splitting.
              ^---^ SC2002: Useless cat. Consider 'cmd < file | ..' or 'cmd file | ..' instead.
                      ^---^ SC2196: egrep is non-standard and deprecated. Use grep -E instead.

Did you mean: 
    OUT=$(cat "$FILE" | egrep -i -o '([0-9]+)\s*(->)\s*([0-9]+)' | awk '$1 > $3 { printf "%s;", $0 }')


In scripts/github/extrinsic-ordering-filter.sh line 31:
    IFS=$';' LIST=($OUT)
                   ^--^ SC2206: Quote to prevent word splitting/globbing, or split robustly with mapfile or read -a.


In scripts/github/extrinsic-ordering-filter.sh line 34:
        echo $line
             ^---^ SC2086: Double quote to prevent globbing and word splitting.

Did you mean: 
        echo "$line"


In scripts/github/extrinsic-ordering-filter.sh line 35:
        RES="$RES\n$(cat $FILE | egrep -i -n "$line" | tr -s " ")"
                         ^---^ SC2086: Double quote to prevent globbing and word splitting.
                         ^---^ SC2002: Useless cat. Consider 'cmd < file | ..' or 'cmd file | ..' instead.
                                 ^---^ SC2196: egrep is non-standard and deprecated. Use grep -E instead.

Did you mean: 
        RES="$RES\n$(cat "$FILE" | egrep -i -n "$line" | tr -s " ")"


In scripts/github/extrinsic-ordering-filter.sh line 45:
echo "\n------------------------------ SUMMARY -------------------------------"
     ^-- SC2028: echo may not expand escape sequences. Use printf.


In scripts/github/extrinsic-ordering-filter.sh line 46:
echo "\n⚠️ This filter is here to help spotting changes that should be reviewed carefully."
     ^-- SC2028: echo may not expand escape sequences. Use printf.


In scripts/github/extrinsic-ordering-filter.sh line 47:
echo "\n⚠️ It catches only index changes, deletions and value decreases".
     ^-- SC2028: echo may not expand escape sequences. Use printf.


In scripts/github/extrinsic-ordering-filter.sh line 49:
find_deletions $FILE
               ^---^ SC2086: Double quote to prevent globbing and word splitting.

Did you mean: 
find_deletions "$FILE"


In scripts/github/extrinsic-ordering-filter.sh line 50:
find_index_changes $FILE
                   ^---^ SC2086: Double quote to prevent globbing and word splitting.

Did you mean: 
find_index_changes "$FILE"


In scripts/github/extrinsic-ordering-filter.sh line 51:
find_decreases $FILE
               ^---^ SC2086: Double quote to prevent globbing and word splitting.

Did you mean: 
find_decreases "$FILE"


In scripts/github/extrinsic-ordering-filter.sh line 52:
echo "\n----------------------------------------------------------------------\n"
     ^-- SC2028: echo may not expand escape sequences. Use printf.

For more information:
  https://www.shellcheck.net/wiki/SC1128 -- The shebang must be on the first ...
  https://www.shellcheck.net/wiki/SC2206 -- Quote to prevent word splitting/g...
  https://www.shellcheck.net/wiki/SC2028 -- echo may not expand escape sequen...

```

chevdor:
I will run `shellcheck` locally and fix those nits.

chevdor:
I fixed according to the linter's suggestion but did disable one type that would make the commands way less readable and harder to test and debug.

chevdor:
I will bring another fix in, as the linter's recommandations are not without consequences...

chevdor:
So this linter is not amazing... it enforces double quotes but fails to recognize an escaped one, effectively inviting you to screw up your script 😞
I had to silence one more rule although all the hints of this rules have been fixed except the escaped-double-quote issue.

.github/workflows/extrinsic-ordering-check-from-bin.yml
TriplEight: now, this step can be removed.

TriplEight: reusing docker cache in GHA is hardly possible: it runs on preemptible VMs, hence the only way is to upload it in the previous job and download it here. Sum of these times sometimes is bigger than rebuilding the payload.

TriplEight (changes requested):

.github/workflows/extrinsic-ordering-check-from-bin.yml
chevdor: I am considering reverting your suggestion actually as I would like to keep the 2 steps pull and run separate. It makes the logs more readable as the pull is quite verbose. So I aware we **could** have those 2 steps in one and I used to have it this way but it is better split.

TriplEight (approved):

chevdor:
Rebased on master

.github/workflows/extrinsic-ordering-check-from-bin.yml
```diff
@@ -6,15 +6,15 @@ on:
     inputs:
       reference_url:
         description: The WebSocket url of the reference node
-        default: wss://rpc.polkadot.io
+        default: wss://kusama-rpc.polkadot.io
```
s3krit, line 9: Any reason we choose to default to kusama? Just curious :)

s3krit (approved):
LGTM :D

.github/workflows/extrinsic-ordering-check-from-bin.yml
chevdor: There are 2 reasons :)
- to go from Kusama to Polkadot you just need to remove the Kusama part
- Kusama is usually what we check first for the releases

PR #3640, opened 2021-08-14
Remove BuyExecution::orders

gavofyork:
Current code is invalid since the weight for orders never gets checked.


kianenigma (approved):
will it be re-introduced in a different way, potentially with proper weighing?

KiChjang (approved):

gavofyork:
No; XCM v2 reworks the `Xcm`/`Order` model into a single-level `Instruction`. Recursion is removed, which simplifies things quite a bit at the cost of introducing more nonsense expressions and the loss of some structure.

apopiak:
@gavofyork Could you add that rework to https://github.com/paritytech/polkadot/issues/3644 and/or https://github.com/paritytech/polkadot/issues/2841 ?

PR #3655, opened 2021-08-17
Accept PVF code hashes in validation host

### LINKED ISSUE #2866: PVF validation host: pull not push PVFs
Currently candidate validation pushes PVF preimages to the candidate validation host. This is non-ideal, because PVF validation host caches compiled PVFs and does not need the original code blobs.

To fix this we should update PVF validation host so that it accepts not only raw PVFs but also PVF code hashes. In case of a cache miss, the PVF validation host should request the validation code preimage from candidate-validation.

It seems we cannot completely get rid of the mode where PVFs are passed directly as preimages. For example, for disputes, the subsystem that initiates candidate validation will have the validation code at hand, however, there is no guarantee the code can be found on the chain. In this case, it is just more convenient to pass the validation code directly.

slumber:
Resolves #2866


node/core/candidate-validation/src/lib.rs
```diff
@@ -196,73 +236,58 @@ where
...
-enum AssumptionCheckOutcome {
-	Matches(PersistedValidationData, ValidationCode),
+enum AssumptionCheckOutcome<T> {
+	Matches(T, OccupiedCoreAssumption),
 	DoesNotMatch,
 	BadRequest,
 }
 
-async fn check_assumption_validation_data<Context>(
+async fn check_assumption_validation_data<Context, T>(
```
pepyakin, line 245: TBH, I am not really a fan of this approach - i.e. using the trait and trying to reuse this code. Sure, it perhaps minimizes the diff but it does so at the cost of legibility. If possible I would prefer straight line code.

Luckily, I think this should be possible to avoid.

Specifically, look at the `CandidateDescriptor`. It contains the hash of validation code against which the candidate was produced. This enables us to avoid querying the validation code before and allows us to query the validation code by the hash when it is needed. No need to try to match it against the assumption or anything alike.

pepyakin (changes requested):
My main concern is the candidate-validation changes. I think the code can be improved there by leveraging the fact that the candidate descriptor carries the code hash which we can query.

node/core/candidate-validation/src/lib.rs
```diff
@@ -500,7 +558,7 @@ fn perform_basic_checks(
 		return Err(InvalidCandidate::PoVHashMismatch)
 	}
 
-	if *validation_code_hash != candidate.validation_code_hash {
+	if let Some(false) = validation_code_hash.map(|hash| *hash == candidate.validation_code_hash) {
```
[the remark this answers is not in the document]
slumber: In fact it is necessary to preserve it, candidate validation subsystem still can receive `ValidateFromExhaustive` request.

node/core/pvf/src/host.rs
```diff
@@ -72,7 +74,7 @@ impl ValidationHost {
 	/// situations this function should return immediately.
 	///
 	/// Returns an error if the request cannot be sent to the validation host, i.e. if it shut down.
-	pub async fn heads_up(&mut self, active_pvfs: Vec<Pvf>) -> Result<(), String> {
+	pub async fn heads_up(&mut self, active_pvfs: Vec<PvfPreimage>) -> Result<(), String> {
```
pepyakin, line 77: This is the public interface of this crate. As far as I can see, `PvfPreImage` is not exported from this crate and this is undesired (if not making the API unusable)

node/core/candidate-validation/src/lib.rs
```diff
@@ -421,6 +466,8 @@ async fn validate_candidate_exhaustive(
 			Ok(ValidationResult::Invalid(InvalidCandidate::ExecutionError(
 				"ambigious worker death".to_string(),
 			))),
+		Err(ValidationError::ArtifactNotFound) =>
+			Err(ValidationFailed("Code not found".to_string())),
```
pepyakin, line 470: Nit: Let's try to make the error message a bit more clear. Imagine if you were going through the logs and just saw this error message. I imagine you wouldn't say thanks for the past you who left this message. If you were more descriptive, if it didn't light a light bulb, then it would perhaps at least help to grep the origin of the string.

```suggestion
			Err(ValidationFailed("validation host can't find artifact even though just received it".to_string())),
```

(I am not sure if you can apply this suggestion directly because of rustfmt and the line width constaints, and I am also not insisting in this particular wording)

node/core/candidate-validation/src/lib.rs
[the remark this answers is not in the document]
pepyakin: Yeah, what I am saying above is basically: it seems fine this way. The rationale is:
- if the validation code was provided via the exhaustive validation request, then we do have to perform this check. The `ValidationCodeHash` will be Some and we do the check.
- if we are on the validation from onchain state path, then the code hash won't be provided but that's fine because the only way to get it via querying code by hash which implicitly preserves this invariant.

node/core/candidate-validation/src/lib.rs
```diff
@@ -296,30 +302,53 @@ where
...
+	};
 
-	let validation_result = validate_candidate_exhaustive(
-		validation_host,
-		validation_data,
-		validation_code,
+	let mut validation_result = validate_candidate_exhaustive(
+		&mut validation_host.clone(),
+		validation_data.clone(),
+		None,
```
pepyakin, line 320: Nit: Reading this code I can't help but notice that this function is a misnomer now. It is not exhaustive now since we do not necessarily provide validation code.

node/core/candidate-validation/src/lib.rs
```diff
@@ -296,30 +302,53 @@ where
...
-		pov,
+		pov.clone(),
 		metrics,
 	)
 	.await;
 
+	// In case preimage for the supplied code hash was not found by the
+	// validation host, request the code from Runtime API and try again.
+	if let Ok(Err(ref err)) = validation_result {
+		if err.0.as_str() == "Code not found" {
```
pepyakin, line 330: No, you cannot do that. I perceive this unreliable. At the very least you can do an enum or otherwise make sure the compiler helps you to avoid a disaster.

I think we can move the code query into the `validate_candidate_exhaustive`. Specifically, if the caller didn't pass preimage, then if `validation_backend.validate_candidate` returned `ArtifactNotFound` it would ask the runtime for the validation code by hash and retry. This can be abstracted as a function.

pepyakin (changes requested):
candidate-validation looks way better now! However, there is another point which made me sad: "Code not found" thingy.

node/core/candidate-validation/src/lib.rs
slumber: I agree here, it's just that the error holds a string in the first place. That's a good hint to handle an error inside of the exhaustive function

pepyakin: > I agree here, it's just that the error holds a string in the first place. 

Yeah but I think it's fine. The error has multiple purposes: reporting errors to humans and signalling a condition to trigger some recovery logic. This type was used only for reporting errors and a string could do, but IMO a string is an inappropriate tool for matching as a condition. It sometimes happens that new requirements can lead to the need on matching and in that case we should do a refactoring. Only in very rare and extreme cases we can reach to string matching counting on eventual refactoring. This particular case is far from that though.

node/core/candidate-validation/src/lib.rs
```diff
@@ -398,16 +417,51 @@ async fn validate_candidate_exhaustive(
...
+						tracing::debug!(target: LOG_TARGET, err=?e, "Invalid validation code");
+
+						// If the validation code is invalid, the candidate certainly is.
+						return Ok(Ok(ValidationResult::Invalid(
+							InvalidCandidate::CodeDecompressionFailure,
+						)))
+					},
+				}
+				.to_vec(),
+			);
```
pepyakin, line 462: Nit: it's getting visually hard to get a mental structure of this code when reading. First there is a call to `decompress`, then either break the control flow or yield `code`. If we returned code we `to_vec` it and then do `from_code`.

I think if we extracted the actual argument to `from_code` into a separate variable and moved `to_vec` closer. The end result would be a more straightforward code like, ideally just top-to-bottom.

node/core/pvf/src/host.rs
```diff
@@ -416,12 +418,17 @@ async fn handle_execute_pvf(
...
+			//
+			artifacts.insert_preparing(artifact_id.clone());
+			send_prepare(prepare_queue, prepare::ToQueue::Enqueue { priority, pvf: inner }).await?;
 
-		awaiting_prepare.add(artifact_id, params, result_tx);
+			awaiting_prepare.add(artifact_id, params, result_tx);
+		} else {
+			// Expect another request with PVF provided.
+			let _ = result_tx.send(Err(ValidationError::ArtifactNotFound));
+		};
```
pepyakin, line 431: Nit: `;` seems to be superfluous. I wonder why rustfmt allows that

```suggestion
		}
```

node/core/candidate-validation/src/lib.rs
```diff
@@ -421,7 +475,17 @@ async fn validate_candidate_exhaustive(
 			Ok(ValidationResult::Invalid(InvalidCandidate::ExecutionError(
 				"ambigious worker death".to_string(),
 			))),
-
+		Err(ValidationError::ArtifactNotFound) => {
+			// The code was supplied on the second attempt, this
+			// error should be unreachable.
+			tracing::error!(
+				target: LOG_TARGET,
+				"Unexpected error received from the validation host"
```
pepyakin, line 483: I think this is too broad. Why can't we just copy the message below?

node/core/candidate-validation/src/lib.rs
```diff
@@ -362,22 +373,30 @@ async fn validate_candidate_exhaustive(
...
-		Err(e) => {
-			tracing::debug!(target: LOG_TARGET, err=?e, "Invalid validation code");
+	let validation_code = if let Some(validation_code) = validation_code {
+		let raw_code = match sp_maybe_compressed_blob::decompress(
+			&validation_code.0,
+			VALIDATION_CODE_BOMB_LIMIT,
+		) {
+			Ok(code) => code,
+			Err(e) => {
+				tracing::debug!(target: LOG_TARGET, err=?e, "Invalid validation code");
```
pepyakin, line 388: This message is a bit too broad. Maybe we can reuse the error here, like "code decompression failure"?

node/core/candidate-validation/src/lib.rs
```diff
@@ -362,22 +373,30 @@ async fn validate_candidate_exhaustive(
...
+				// If the validation code is invalid, the candidate certainly is.
+				return Ok(Ok(ValidationResult::Invalid(InvalidCandidate::CodeDecompressionFailure)))
+			},
+		}
+		.to_vec();
+		Pvf::from_code(raw_code)
+	} else {
+		// In case validation code is not provided, ask the backend to obtain
+		// it from the cache using the hash.
+		Pvf::Hash(ValidationCodeHash::from(descriptor.persisted_validation_data_hash))
```
pepyakin, line 399: We need `validation_code_hash` here.

pepyakin (changes requested):

node/core/candidate-validation/src/lib.rs
```diff
@@ -491,14 +509,14 @@ trait ValidationBackend {
 impl ValidationBackend for ValidationHost {
 	async fn validate_candidate(
 		&mut self,
-		raw_validation_code: Vec<u8>,
+		validation_code: Pvf,
```
drahnr, line 515: :+1:

node/core/candidate-validation/src/lib.rs
```diff
@@ -539,7 +557,7 @@ fn perform_basic_checks(
 		return Err(InvalidCandidate::PoVHashMismatch)
 	}
 
-	if *validation_code_hash != candidate.validation_code_hash {
+	if let Some(false) = validation_code_hash.map(|hash| *hash == candidate.validation_code_hash) {
```
drahnr, line 563: This should be expressable as, right? That'd be a little more concise.

```suggestion
	if Some(&candidate.validation_code_hash) == validation_code_hash {
```

node/core/candidate-validation/src/lib.rs
```diff
@@ -459,7 +472,15 @@ async fn validate_candidate_exhaustive(
...
 			))),
-
+		Err(ValidationError::ArtifactNotFound) => {
+			// The code was supplied on the second attempt, this
+			// error should be unreachable.
+			let err = ValidationFailed(
+				"Validation host failed to find artifact even though it was supplied".to_string(),
+			);
+			tracing::error!(target: LOG_TARGET, error = ?err, "Unexpected error reported by the validation backend");
+			Err(err)
```
drahnr, line 482: This rubs me the wrong way. There are too many strings invovled here and `ValidationFailed` should become a proper error enum, but that's out of scope for this PR. I also don't see a point to print the error here, the caller should take care of printing the `Err(err)`.

drahnr (approved):
LGTM, minus the TODO and the nits annotated. Please talk to @eskimor about the remaining open question(s).

node/core/candidate-validation/src/lib.rs
slumber: In fact, this could be replaced with `unreachable`, but since we try to avoid panicking the error simply gets logged. Caller (as I understand the component that sent the validation request?) should never receive this error, so we handle it here. Though I totally agree on the part of turning strings into enum. 

The PR is on ice due to dependency on the recent runtime upgrade, so probably I could fix `ValidationFailed` too, but I wonder how would anybody handle variants? Isn't it a kind of internal error that means something went wrong and we should skip the candidate?

drahnr: What my comment boils down to, is that 

```

/// Blanket error for validation failing for internal reasons.
#[derive(Debug, Error)]
#[error("Validation failed with {0:?}")]
pub struct ValidationFailed(pub String);
```
exists which is an error antipattern, and makes the code less legible than it could be. It should be an error enum with explicit variants and annotations using `thiserror`.

node/core/candidate-validation/src/lib.rs
slumber: Not quite, we want to check whether it's *not* equal to the expected one.
But that's a good catch, after the upgrade the validation code hash is always present so this place can be fixed

node/core/candidate-validation/src/lib.rs
```diff
@@ -446,6 +423,40 @@ async fn validate_candidate_exhaustive(
...
+		if let &ValidationError::ArtifactNotFound = e {
+			tracing::debug!(
+				target: LOG_TARGET,
+				"Validation host failed to find artifact by provided hash",
+			);
+
+			let validation_code = match request_validation_code_by_hash(sender, &descriptor).await {
+				Ok(Some(validation_code)) => validation_code,
+				Ok(None) =>
+				// TODO: code not found by hash by the runtime, is this candidate invalid?
```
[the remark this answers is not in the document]
eskimor: I would say, that's a fair assumption -yes.

node/core/pvf/src/host.rs
```diff
@@ -433,6 +435,10 @@ async fn handle_precheck_pvf(
 	pvf: Pvf,
 	result_sender: PrepareResultSender,
 ) -> Result<(), Fatal> {
+	let pvf = match pvf {
+		Pvf::Code(code) => code,
+		Pvf::Hash(_) => todo!(),
+	};
```
slumber, line 441: @pepyakin we could either expect `PvfCode` here or allow requesting by hash.

drahnr: Why not handle this outside of this fn and just expected `PvfCode` as input?

slumber: We could possibly accept the hash here and look for it in artifacts cache, return `ArtifactNotFound` otherwise.
It's left as a question and thus `todo`

slumber: I decided to keep it as is: request the code from the runtime right away:
https://github.com/paritytech/polkadot/blob/e4edd445f472a2610b25f1aaee525f9e16ed95fe/node/core/candidate-validation/src/lib.rs#L318-L321

node/core/candidate-validation/src/lib.rs
```diff
@@ -551,9 +554,17 @@ async fn validate_candidate_exhaustive(
 			Ok(ValidationResult::Invalid(InvalidCandidate::ExecutionError(
 				"ambiguous worker death".to_string(),
 			))),
+		Err(ValidationError::ArtifactNotFound) => {
+			// The code was supplied on the second attempt, this
+			// error should be unreachable.
+			let err = ValidationFailed(
+				"Validation host failed to find artifact even though it was supplied".to_string(),
+			);
+			tracing::error!(target: LOG_TARGET, error = ?err, "Unexpected error reported by the validation backend");
```
pepyakin, line 563: Nit: IMO unnecessary exceeds the soft limit of 100 characters per line

node/primitives/src/lib.rs
```diff
@@ -231,6 +231,8 @@ pub enum InvalidCandidate {
 	ParaHeadHashMismatch,
 	/// Validation code hash does not match.
 	CodeHashMismatch,
+	/// The corresponding validation code is not present in the Runtime storage.
+	MissingValidationCode,
```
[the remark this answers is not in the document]
slumber: Uggh, there were no discussion about this, I just failed to translate the behavior from assumptions workflow to the new one.
Good catch, thank you

roadmap/implementers-guide/src/node/utility/candidate-validation.md
```diff
@@ -32,11 +32,11 @@ For a [`CandidateValidationMessage`][CVM]`::ValidateFromExhaustive`, these param
...
+For this reason this subsystem uses a convenient Runtime API endpoint — [`AssumedValidationData`](../../types/overseer-protocol.md#runtime-api-message). It accepts two parameters: parachain ID and the expected hash of the validation data, — the one we obtain from the `CandidateDescriptor`. Then the runtime tries to construct the validation data for the given parachain first without force enacting the core and, if its hash doesn't match the expected one, tries again with it. In case of a match the API returns the constructed validation data along with the corresponding validation code hash, or `None` otherwise. The latter means that the validation data hash in the descriptor is not based on the relay parent and thus given candidate is invalid.
```
burdges, line 35: It's likely worth explaining force enact somewhat better somewhere, because the hash differing here winds up confusing.

slumber:
@pepyakin note that given async backing changes (in particular https://github.com/paritytech/polkadot/pull/5557) `ValidateFromChainState` is not used anymore, backing requires explicitly providing persisted validation data (which makes sense).

IMO this PR should be closed as it introduces rather unnecessary changes for now. The caching can be implemented later but would require some additional work.

m-cat:
I guess I should have asked around more before tackling this!

If this PR is closed then should [this issue](https://github.com/paritytech/polkadot/issues/3697) be closed as well? I believe it depends on this PR.

slumber:
> I guess I should have asked around more before tackling this!
> 
> If this PR is closed then should [this issue](https://github.com/paritytech/polkadot/issues/3697) be closed as well? I believe it depends on this PR.

No worries, there was no decision on closing it (yet), that's just my comment to give more context on whether this issue is still relevant. I started a discussion on how to proceed with it.
Another obvious option would be to rebase it straight to the async backing feature branch.

pepyakin:
> @pepyakin note that given async backing changes (in particular https://github.com/paritytech/polkadot/pull/5557) ValidateFromChainState is not used anymore, backing requires explicitly providing persisted validation data (which makes sense).

Explicitly providing PVD? But this one is about passing PVF hashes instead of providing the preimages, right?

But yeah in general I am not surprised of this outcome, candidate-validation was asking for rework anyway.

slumber:
> Explicitly providing PVD? But this one is about passing PVF hashes instead of providing the preimages, right?

Yes, for this reason a PR is still relevant but requires additional work.

For example, ValidateFrom... message types could be unified into one and we would always attempt to look up pvf in the cache first.

PR #3656, opened 2021-08-18
Back port of connectivity warnings to 0.9.8

eskimor:
* Warn on low connectivity.

* Better timeout and docs.

* Review remarks.


eskimor:
I had that deployed on a validator before that release. Still using it for debugging - I will make it a draft to avoid confusion.

PR #3658, opened 2021-08-18
Removed unneeded deps

gilescope:
Tentatively these deps seem not to be needed.


bridges/primitives/runtime/Cargo.toml
```diff
@@ -13,16 +13,16 @@ num-traits = { version = "0.2", default-features = false }
...
 sp-state-machine = { git = "https://github.com/paritytech/substrate", branch = "master", default-features = false }
-sp-std = { git = "https://github.com/paritytech/substrate", branch = "master" , default-features = false }
-sp-trie = { git = "https://github.com/paritytech/substrate", branch = "master" , default-features = false }
+sp-std = { git = "https://github.com/paritytech/substrate", branch = "master", default-features = false }
+sp-trie = { git = "https://github.com/paritytech/substrate", branch = "master", default-features = false }
 
-[dev-dependencies]
-sp-core = { git = "https://github.com/paritytech/substrate", branch = "master" }
+# [dev-dependencies]
+# sp-core = { git = "https://github.com/paritytech/substrate", branch = "master" }
```
HCastano, line 25: I suspect the reason why this is fine is because the bridge tests don't get run in the Polkadot repo. 

I would avoid touching the bridges code as part of this PR, and instead do a follow up directly as part of the `parity-bridges-common` repo to see what unneeded deps exist there

gilescope:
Great idea - will do.

node/malus/Cargo.toml
```diff
@@ -24,7 +24,6 @@ polkadot-cli = { path = "../../cli", default-features = false, features = [
 polkadot-node-subsystem = { path = "../subsystem" }
 polkadot-node-subsystem-util = { path = "../subsystem-util" }
 polkadot-node-core-candidate-validation = { path = "../core/candidate-validation" }
-# parity-util-mem = { version = "0.10.0", default-features = false, features = ["jemalloc-global"] }
```
ordian: I believe it was added here on purpose to enable jemalloc and have similar execution environment to the polkadot binary. cc @drahnr 
If that's the case, we could add a comment here stating the purpose.

drahnr: I think the change is sane, but we should include a note where `parity-util-mem` is included, that `jemalloc-ctl` crate is used on `*nix` platforms, so changes have to be applied with care.

node/malus/Cargo.toml
```diff
@@ -17,11 +17,13 @@ readme = "README.md"
...
 [dependencies]
-polkadot-cli = { path = "../../cli", default-features = false, features = [ "cli", "malus" ]  }
+polkadot-cli = { path = "../../cli", default-features = false, features = [
+    "cli",
+    "malus",
+] }
 polkadot-node-subsystem = { path = "../subsystem" }
 polkadot-node-subsystem-util = { path = "../subsystem-util" }
 polkadot-node-core-candidate-validation = { path = "../core/candidate-validation" }
-parity-util-mem = { version = "0.10.0", default-features = false, features = ["jemalloc-global"] }
```
drahnr, line 24: Please avoid changing this file, https://github.com/paritytech/polkadot/pull/3639 addresses part of it.

drahnr (commented):
Cleanup of the dependencies is much appreciated!

I am not sure we want the additional line breaks for feature lists.

runtime/westend/src/lib.rs
```diff
@@ -113,7 +113,7 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
 	impl_name: create_runtime_str!("parity-westend"),
 	authoring_version: 2,
 	spec_version: 9090,
-	impl_version: 0,
+	impl_version: 1,
```
ordian, line 116: as well

runtime/rococo/src/lib.rs
```diff
@@ -104,7 +104,7 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
 	impl_name: create_runtime_str!("parity-rococo-v1.6"),
 	authoring_version: 0,
 	spec_version: 9004,
-	impl_version: 0,
+	impl_version: 1,
```
ordian, line 107: ditto

runtime/kusama/src/lib.rs
```diff
@@ -115,7 +115,7 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
 	impl_name: create_runtime_str!("parity-kusama"),
 	authoring_version: 2,
 	spec_version: 9090,
-	impl_version: 0,
+	impl_version: 1,
```
ordian, line 118: this is resolved in master, no need for bump

runtime/parachains/Cargo.toml
```diff
@@ -46,14 +47,14 @@ rand_chacha = { version = "0.3.1", default-features = false }
...
-sp-trie = { git = "https://github.com/paritytech/substrate", branch = "master" }
-pallet-babe = { git = "https://github.com/paritytech/substrate", branch = "master" }
-sp-application-crypto = { git = "https://github.com/paritytech/substrate", branch = "master" }
-pallet-staking-reward-curve = { git = "https://github.com/paritytech/substrate", branch = "master" }
-pallet-treasury = { git = "https://github.com/paritytech/substrate", branch = "master" }
+#sp-trie = { git = "https://github.com/paritytech/substrate", branch = "master" }
+#pallet-babe = { git = "https://github.com/paritytech/substrate", branch = "master" }
+#sp-application-crypto = { git = "https://github.com/paritytech/substrate", branch = "master" }
+#pallet-staking-reward-curve = { git = "https://github.com/paritytech/substrate", branch = "master" }
+#pallet-treasury = { git = "https://github.com/paritytech/substrate", branch = "master" }
```
ordian, line 54:
```suggestion
```

runtime/parachains/Cargo.toml
```diff
@@ -46,14 +47,14 @@ rand_chacha = { version = "0.3.1", default-features = false }
...
 frame-support-test = { git = "https://github.com/paritytech/substrate", branch = "master" }
-serde_json = "1.0.61"
-sp-version = { git = "https://github.com/paritytech/substrate", branch = "master", default-features = false }
+#serde_json = "1.0.61"
+#sp-version = { git = "https://github.com/paritytech/substrate", branch = "master", default-features = false }
```
ordian, line 57:
```suggestion
```

runtime/polkadot/src/lib.rs
```diff
@@ -97,7 +97,7 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
 	impl_name: create_runtime_str!("parity-polkadot"),
 	authoring_version: 0,
 	spec_version: 9090,
-	impl_version: 0,
+	impl_version: 1,
```
ordian, line 100: same

runtime/kusama/src/lib.rs
gilescope: thx

gilescope:
@drahnr the line brakes come from vscode's "even better toml" extension that formats toml files. I'm keen on _something_ auto-formatting Cargo.toml files and so far that seems to be about the only thing that does.

scripts/gitlab/lingua.dic
```diff
@@ -28,6 +29,8 @@ broadcast/UDSMG
 BTC/S
 canonicalize/D
 CentOS
+Chain1
+Chain2
```
ordian, line 33: is this still needed? you only touch toml files here, I don't see how this might trigger spellcheck cc @drahnr

ordian (approved):
Thanks for tracking down unused deps!

scripts/gitlab/lingua.dic
drahnr: There are a few things to note: The spellcheck run is done on files that changed in respect to the changes git knows. This builds on-top of changes files between the base commit of a branch and the branch's head commit. Now it seems occasionally (for whatever reason) gitlab picks up more files than anticipated `${CI_COMMIT_SHA} $(git merge-base ${CI_COMMIT_SHA} ${CI_DEFAULT_BRANCH})` so this might be necessary to make that check green.

Besides that, I don't  think we should add those on-off items.  It should be marked as code I guess, like you do on github `ThisChain`, `Chain1`, and `Chain2`.

node/network/availability-distribution/Cargo.toml
```diff
@@ -7,16 +7,16 @@ edition = "2018"
...
 polkadot-subsystem = { package = "polkadot-node-subsystem", path = "../../subsystem" }
 polkadot-node-network-protocol = { path = "../../network/protocol" }
 polkadot-node-subsystem-util = { path = "../../subsystem-util" }
 polkadot-node-primitives = { path = "../../primitives" }
-polkadot-node-core-runtime-api = { path = "../../core/runtime-api" }
-sp-application-crypto = { git = "https://github.com/paritytech/substrate", branch = "master" }
-sp-core = { git = "https://github.com/paritytech/substrate", branch = "master", features = ["std"]  }
+sp-core = { git = "https://github.com/paritytech/substrate", branch = "master", features = [
+    "std",
+] }
```
drahnr, line 19: nit: single item feature lists don't necessarily need linebreaks, is there a way to teach the toml formatter that?

drahnr (approved):
A few nits, other than that LGTM & thank you for doing this!

scripts/gitlab/lingua.dic
gilescope: Isn't it just going to nag someone else later if I don't include them? Is there a special way to mark a word as not a word but to be ignored if found?

ordian: I think @drahnr suggests to put such words in backticks (`) instead of adding to the dictionary

gilescope: ok - as these are in the bridge subtree I will revert the dict changes here and will add backticks when I remove the unneeded dependencies from the bridges repo.

node/network/availability-distribution/Cargo.toml
gilescope: There is but that's a whole new PR. For now I've manually put them back to what they were.

node/core/approval-voting/Cargo.toml
```diff
@@ -23,23 +26,21 @@ polkadot-primitives = { path = "../../../primitives" }
...
-sp-application-crypto = { git = "https://github.com/paritytech/substrate", branch = "master", default-features = false, features = ["full_crypto"] }
+sp-application-crypto = { git = "https://github.com/paritytech/substrate", branch = "master", default-features = false, features = [
+    "full_crypto",
+] }
 sp-runtime = { git = "https://github.com/paritytech/substrate", branch = "master", default-features = false }
 
 [dev-dependencies]
 parking_lot = "0.11.1"
-rand_core = "0.5.1" # should match schnorrkel
+rand_core = "0.5.1"                                                                        # should match schnorrkel
```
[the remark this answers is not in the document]
gilescope: The `#` needs to be near the crate but we can achieve the same effect and keep the formatter happy.

PR #3661, opened 2021-08-18
Provide dummy dispute coordinator by default.

eskimor:
As all dispute coordination is handled by well the the dispute coordinator, a dummy implementation doing nothing, should prevent anything dispute related.


node/core/dispute-coordinator/src/dummy.rs
```diff
@@ -0,0 +1,252 @@
...
+		DisputeCoordinatorMessage::DetermineUndisputedChain {
+			base_number,
+			block_descriptions,
+			tx,
+		} => {
+			let undisputed_chain = block_descriptions
+				.last()
+				.map(|e| (base_number + block_descriptions.len() as BlockNumber, e.block_hash));
+
+			let _ = tx.send(undisputed_chain);
```
eskimor, line 244: Please double check, that this is correct.

eskimor:
I picked C3-medium ... don't really know, if that is the right level.

node/core/dispute-coordinator/src/dummy.rs
rphmeier: Arithmetic looks correct - chain of 1 with base B would return B+1 with hash(B+1)

rphmeier (approved):
@drahnr - How would you use these changes to deploy a dispute testnet?

drahnr:
The easiest approach would be to add another malus impl, which replaces the subsystem with this implementation, and that's it.

node/core/dispute-coordinator/src/dummy.rs
```diff
@@ -0,0 +1,252 @@
...
+}
+
+impl Error {
+	fn trace(&self) {
+		match self {
+			// don't spam the log with spurious errors
+			Self::RuntimeApi(_) | Self::Oneshot(_) =>
+				tracing::debug!(target: LOG_TARGET, err = ?self),
+			// it's worth reporting otherwise
+			_ => tracing::warn!(target: LOG_TARGET, err = ?self),
```
drahnr, line 104: Nit: this obfuscates the origin of the message, since `tracing::*!` does some "magic" to determine the line where it is logged from.

node/core/dispute-coordinator/src/dummy.rs
drahnr: Is correct iff the first element in the `block_descriptions` has block number `base_number`, which holds by impl of the current `relay_chain_selection`.

drahnr (approved):

Issue #3674, opened 2021-08-20
Re-enable dispute coordinator


eskimor:
- [x] Test
- [x] Fix bugs
- [x] Revert https://github.com/paritytech/polkadot/pull/3672
- [x] Remove "allow dead code" again: https://github.com/paritytech/polkadot/pull/3676 https://github.com/paritytech/polkadot/pull/3685
- [ ] Remove dummy implementation (once we are sure we won't need it anymore)

eskimor:
Closed via Bernhard's PR. Not necessary to keep track of removing dummy implementation.

PR #3676, opened 2021-08-20
Bypass chain-selection subsystem until disputes are enabled.

drahnr:
<del>The `cfg!(feature = "fx")` does not exist yet, but it does not matter for compilation.</del> Uses the `feature = "disputes"` as gate for the relay chain selection rule.

### DIFF
--- a/node/service/src/relay_chain_selection.rs
+++ b/node/service/src/relay_chain_selection.rs
@@ -185,7 +185,7 @@ where
 			return longest_chain_best
 		}
 		self.selection
-			.finality_target(target_hash, longest_chain_best, maybe_max_number)
+			.finality_target_with_fallback(target_hash, longest_chain_best, maybe_max_number)
 			.await
 	}
 }
@@ -309,6 +309,20 @@ where
 		self.block_header(best_leaf)
 	}
 
+	async fn finality_target(
+		&self,
+		_target_hash: Hash,
+		_maybe_max_number: Option<BlockNumber>,
+	) -> Result<Option<Hash>, ConsensusError> {
+		unreachable!("Must call finality_target_with_fallback instead. qed");
+	}
+}
+
+impl<B, OH> SelectRelayChain<B, OH>
+where
+	B: HeaderProviderProvider<PolkadotBlock>,
+	OH: OverseerHandleT,
+{
 	/// Get the best descendant of `target_hash` that we should attempt to
 	/// finalize next, if any. It is valid to return the `target_hash` if
 	/// no better block exists.
@@ -318,7 +332,7 @@ where
 	///
 	/// It will also constrain the chain to only chains which are fully
 	/// approved, and chains which contain no disputes.
-	async fn finality_target(
+	pub async fn finality_target_with_fallback(
 		&self,
 		target_hash: Hash,
 		best_leaf: Hash,
--- a/node/service/src/relay_chain_selection.rs
+++ b/node/service/src/relay_chain_selection.rs
@@ -455,7 +455,7 @@ where
 		};
 
 		// 4. Apply the maximum safeguard to the finality lag.
-		if lag < MAX_FINALITY_LAG {
+		if lag > MAX_FINALITY_LAG {
 			// We need to constrain our vote as a safety net to
 			// ensure the network continues to finalize.
 			let safe_target = initial_leaf_number - MAX_FINALITY_LAG;
--- a/node/service/src/relay_chain_selection.rs
+++ b/node/service/src/relay_chain_selection.rs
@@ -332,7 +332,7 @@ where
 	///
 	/// It will also constrain the chain to only chains which are fully
 	/// approved, and chains which contain no disputes.
-	pub async fn finality_target_with_fallback(
+	async fn finality_target_with_fallback(
 		&self,
 		target_hash: Hash,
 		best_leaf: Option<Hash>,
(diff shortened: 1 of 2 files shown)


eskimor (approved):

rphmeier:
(as discussed, needs to feature-gate disputes lag checking as well)

drahnr:
The check labels bot is wrong.

s3krit:
check_labels should be happy now

node/service/src/relay_chain_selection.rs
```diff
@@ -419,30 +427,35 @@ where
...
+			let lag_disputes = initial_leaf_number.saturating_sub(subchain_number);
+			self.metrics.note_disputes_finality_lag(lag_disputes);
+			lag_disputes
+		} else {
+			lag
+		};
 
 		// 4. Apply the maximum safeguard to the finality lag.
-		if lag > MAX_FINALITY_LAG {
+		if lag < MAX_FINALITY_LAG {
```
eskimor: This does not look right.

node/service/src/relay_chain_selection.rs
```diff
@@ -313,14 +332,15 @@ where
 	///
 	/// It will also constrain the chain to only chains which are fully
 	/// approved, and chains which contain no disputes.
-	async fn finality_target(
+	pub async fn finality_target_with_fallback(
```
rphmeier, line 335: Why  is this now pub?

node/service/src/relay_chain_selection.rs
```diff
@@ -304,6 +309,20 @@ where
 		self.block_header(best_leaf)
 	}
 
+	async fn finality_target(
+		&self,
+		_target_hash: Hash,
+		_maybe_max_number: Option<BlockNumber>,
+	) -> Result<Option<Hash>, ConsensusError> {
+		unreachable!("Must call finality_target_with_fallback instead. qed");
```
rphmeier, line 317: This exists in a trait implementation which is called by the GRANDPA code. Wouldn't that unconditionally hit the panic in this case?

node/service/src/relay_chain_selection.rs
```diff
@@ -419,27 +444,32 @@ where
...
-				},
-				std::any::type_name::<Self>(),
-			)
-			.await;
-		let (subchain_number, subchain_head) = rx
-			.await
-			.map_err(Error::OverseerDisconnected)
-			.map_err(|e| ConsensusError::Other(Box::new(e)))?
-			.unwrap_or_else(|| (subchain_number, subchain_head));
+		let lag = if cfg!(feature = "disputes") {
```
rphmeier, line 447: 👍

node/service/src/relay_chain_selection.rs
drahnr: no, we only pass a wrapping type which contains both `selection` and `fallback`. This wrapper will be removed once the overseer proc macro gains the ability to use closures for initialization and avoid the `connected` /`disconnected` states.

Hence the unreachable.

node/service/src/relay_chain_selection.rs
drahnr:
```suggestion
	async fn finality_target_with_fallback(
```

node/service/src/relay_chain_selection.rs
rphmeier: I'd prefer to just remove the unused trait implementation and move the other methods to inherent impl

drahnr: Yeah, you are correct. The initial idea of just removing `struct RelayChainSelWithFallback` is obsolete now that we need fallback unconditionally. 👍

node/service/src/relay_chain_selection.rs
```diff
@@ -339,6 +344,11 @@ where
 				None => return Ok(Some(target_hash)),
 				Some(best) => best,
 			}
+		} else {
+			match best_leaf {
+				None => return Ok(Some(target_hash)),
+				Some(best_leaf) => best_leaf,
+			}
```
drahnr, line 351: @rphmeier could double check this one? The case where we did not get a best/longest chain in the first place. I think this is the best we can do in that case.

rphmeier: Yeah, I also agree. We'll need to rethink when we re-enable chain selection

node/service/src/tests.rs
```diff
@@ -1,3 +1,4 @@
+#![allow(dead_code)]
```
shawntabrizi, line 1: maybe create an issue / todo so this doesnt live forever in this file by accident

node/service/src/relay_chain_selection.rs
```diff
@@ -419,27 +429,32 @@ where
...
+			// The the total lag accounting for disputes.
+			let lag_disputes = initial_leaf_number.saturating_sub(subchain_number);
+			self.metrics.note_disputes_finality_lag(lag_disputes);
+			lag_disputes
+		} else {
+			lag
+		};
 
 		// 4. Apply the maximum safeguard to the finality lag.
 		if lag > MAX_FINALITY_LAG {
```
shawntabrizi, line 460: im reviewing out of context of all this code, but the behavior here before and after is different.

This check before was always checking against `lag`, and with the new code, could check against `lag_disputes` with the feature flag.

Was this a bug before? or a bug now?

rphmeier: Bug before, addressed here. Approval and disputes both apply constraints to the finalizable chain and previously the disputes lag wasn't accounted for in the 'max' check.

node/service/src/tests.rs
rphmeier: @drahnr

eskimor: Done: https://github.com/paritytech/polkadot/issues/3674

PR #3680, opened 2021-08-20
staking-miner: remove need of a file to pass the seed

chevdor:
Currently the seed needs to be  passed as a `PathBuf` pointing to a file containing the seed.

This PR allows to:
- renames `--account-seed` into `--seed`
- pass the seed directly as a string (`--seed 0x1234...`)
- pass the seed as string via ENV (``)
- pass the node URI as ENV

Those changes make it easier to package the staking miner and use it as a container.


utils/staking-miner/src/main.rs
```diff
@@ -276,15 +276,15 @@ struct DryRunConfig {
 #[derive(Debug, Clone, StructOpt)]
 struct SharedConfig {
 	/// The `ws` node to connect to.
-	#[structopt(long, short, default_value = DEFAULT_URI)]
+	#[structopt(long, short, default_value = DEFAULT_URI, env = "URI")]
```
bkchr, line 279:
```suggestion
	#[structopt(long, default_value = DEFAULT_URI, env = "URI")]
```

utils/staking-miner/src/main.rs
```diff
@@ -276,15 +276,15 @@ struct DryRunConfig {
...
-	/// The file from which we read the account seed.
+	/// The seed of a funded account in hex.
 	///
-	/// WARNING: don't use an account with a large stash for this. Based on how the bot is
-	/// configured, it might re-try lose funds through transaction fees/deposits.
-	#[structopt(long, short)]
-	account_seed: std::path::PathBuf,
+	/// WARNING: Don't use an account with a large stash for this. Based on how the bot is
+	/// configured, it might re-try and lose funds through transaction fees/deposits.
+	#[structopt(long, short, env = "SEED")]
```
bkchr, line 286:
```suggestion
	#[structopt(long, env = "SEED")]
```

utils/staking-miner/src/signer.rs
```diff
@@ -51,26 +52,23 @@ pub(crate) async fn get_account_info<T: frame_system::Config>(
...
 >(
-	path: P,
+	seed: &str,
 	client: &WsClient,
 ) -> Result<Signer, Error> {
-	let uri = std::fs::read_to_string(path)?;
-
 	// trim any trailing garbage.
-	let uri = uri.trim_end();
+	let uri = seed.trim_end();
```
bkchr, line 67:
```suggestion
	let seed = seed.trim_end();
```

?

kianenigma:
isn't it a bad practice to pass the seed like that to the command, since it will be saved in all sorts of history files etc.? This is why I initially pref red passing a file in.

Otherwise looks good to me.

utils/staking-miner/src/main.rs
chevdor: I double checked and this works fine.

It allows users to use `-u <uri>` (short) as well as `--uri <uri>` (long). The benefit is however admitedly very limited and saves only a little typing.

If there are strong opinions against it, I don't mind removing those. There are not many flags and no conflicts/ambiguities so I think that leaving `short` is fine.

chevdor:
> isn't it a bad practice to pass the seed like that to the command, since it will be saved in all sorts of history files etc.? This is why I initially pref red passing a file in.

Totally!
The best recommandation is to pass an ENV, thus this PR. It may be useful however for testers and devs to have a simple way to pass the seed for now. If we want to really harden this, we could totally remove both `short` and `long`, effectively forcing the user to pass the seed as a `SEED` ENV var.

Note that the security still remains in the hand of the user, who can still call the staking miner with (⚠️ do NOT do the following at home, this is bad :)):
```
SEED=0x1234... staking-miner ...
```
Effectively leaving the `SEED` in the history exactly like it happens calling `staking-miner --seed 0x1234...`

NOTE: There a little trick if you do not want to leave *traces* in your history though. The history can be totally disabled but that removes the convenience... or use:
```
 staking-miner --seed 0x1234... ...
```
instead of:
```
staking-miner --seed 0x1234... ...
```
(Notice the barely visible leading space).

I also added `.env` to the list of ignored files in the repo in this PR. That helps also regarding security as one can then create for dev purposes a `.env` file with the content:
```
SEED=0x1234
```
This file can be loaded manually or automatically. Making a call as `staking-miner --uri ws://... dry-run` possible, without having to pass the ENV at all.

utils/staking-miner/src/signer.rs
chevdor: Totally wrong ! Thanks @bkchr.

kianenigma (approved):
Okay, sounds good to me. Only note is that this will need to be redeployed after the first runtime release that contains this, since the flag has changed. But not a blocker, since a proper upgrade mechanism for staking miner does not exist and is purely manual for now.

PR #3681, opened 2021-08-20
Fill up request slots via `launch_parallel_requests`

eskimor:
in case waiting for the next response takes too long. I intend to improve this further, but this should already help a lot with the finality lag issues.

### DIFF
--- a/node/network/availability-recovery/src/lib.rs
+++ b/node/network/availability-recovery/src/lib.rs
@@ -21,6 +21,7 @@
 use std::{
 	collections::{HashMap, VecDeque},
 	pin::Pin,
+	time::Duration,
 };
 
 use futures::{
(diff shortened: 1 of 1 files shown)


drahnr (approved):
LGTM, but definitely needs a burnin. As @rphmeier stated, maybe defer to next release +1

node/network/availability-recovery/src/lib.rs
```diff
@@ -285,7 +290,11 @@ impl RequestChunksPhase {
 
 	async fn wait_for_chunks(&mut self, params: &InteractionParams) {
 		// Wait for all current requests to conclude or time-out, or until we reach enough chunks.
-		while let Some(request_result) = self.requesting_chunks.next().await {
+		// We will also stop, if there has not been a response for `MAX_CHUNK_WAIT`, so
+		// `launch_parallel_requests` cann fill up slots again.
+		while let Some(request_result) =
+			self.requesting_chunks.next().timeout(MAX_CHUNK_WAIT).await.flatten()
```
rphmeier, line 296: Previously, a timeout would lead to an invocation of `self.shuffling.push_front(validator_index)`. Now, it no longer does. This could also lead to mistaken "unavailable" returns from the process.

eskimor: It still does. I did not change that code. The above timeout is a soft timeout, we still wait for the futures to return and they will still return with a timeout, triggering the `push_front` call. The only difference is that in the meantime we filled up free slots with new futures.

ordian (approved):

node/network/availability-recovery/src/lib.rs
rphmeier: Ah, I see. The timeout causes the loop to break so we do another loop iteration where we populate chunks.

PR #3690, opened 2021-08-23
Improve MultiLocation conversion functions in xcm-procedural

KiChjang:
Make use of fully qualified paths when generating code in proc macros, and ensures a canonicalized v0 MultiLocation before converting to v1 MultiLocation.

### DIFF
--- a/xcm/procedural/src/v1/multilocation.rs
+++ b/xcm/procedural/src/v1/multilocation.rs
@@ -172,7 +172,7 @@ fn generate_conversion_from_v0() -> TokenStream {
 							if #( #parent_idents.is_parent() )&&* =>
 							Ok(MultiLocation {
 								parents: #parent_count,
-								interior: #junction_variant( #( #junction_idents.try_into()? ),* ),
+								interior: #junction_variant( #( core::convert::TryInto::try_into(#junction_idents)? ),* ),
 							}),
 					}
 				})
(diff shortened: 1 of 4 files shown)


xcm/procedural/src/v1/multilocation.rs
```diff
@@ -184,16 +184,18 @@ fn generate_conversion_from_v0() -> TokenStream {
...
 
 	quote! {
-		impl TryFrom<crate::v0::MultiLocation> for MultiLocation {
+		impl core::convert::TryFrom<crate::v0::MultiLocation> for MultiLocation {
 			type Error = ();
-			fn try_from(v0: crate::v0::MultiLocation) -> core::result::Result<Self, ()> {
+			fn try_from(mut v0: crate::v0::MultiLocation) -> core::result::Result<Self, ()> {
 				use Junctions::*;
+
+				v0.canonicalize();
```
bkchr, line 198: This wasn't called before?

bkchr (approved):

xcm/procedural/src/v1/multilocation.rs
KiChjang: Yeah it wasn't, even before the migration to the proc macro.

bkchr: What I should have asked, do you added this by intention? :D

KiChjang: Yes, I thought that the conversion should always try and canonicalize the `MultiLocation` first, since it's an infallible operation. However, it does change the internal representation of the `MultiLocation` irreversibly, but I think that's ok because the non-canonical representation would've been "erroneous" anyway.

xcm/src/v1/multilocation.rs
```diff
@@ -898,7 +895,7 @@ mod tests {
 		);
 		assert_eq!(
 			v0::MultiLocation::X2(v0::Junction::Parachain(88), v0::Junction::Parent).try_into(),
-			Err::<MultiLocation, ()>(()),
+			Ok(MultiLocation::here()),
```
kianenigma, line 898: why is this not an error anymore? because of `v0.canonicalize();`?

kianenigma (approved):

xcm/src/v1/multilocation.rs
KiChjang: Correct, what results after canonicalization is exactly `Here`.

KiChjang:
bot merge

PR #3693, opened 2021-08-23
XCM: Introduce versioning to dispatchables' params

gavofyork:
Ensure that dispatchables (which can be used for transactions) have versioned XCM params, since bare XCM datatypes are expected to change between runtimes which can cause problems with queued transactions which persist between runtime versions.

This should go in prior to the release of a version with XCM v1 to minimise alterations to transactions that infrastructure providers need to manage.

CC @s3krit @jacogr


KiChjang (approved):
LGTM

PR #3716, opened 2021-08-25
Better overlength block handling

Lldenaurois:
This PR should address the overlength block vulnerability raised by srlabs. In particular, this PR modifies the `select_candidates` function to keep track of two things:

1. The `cumulative_size` of all candidates in the selection
2. The single validation code containing candidate (if any)

In the issue, jakoblell mentions that it may be prudent to randomize the selection set so that the validation code containing candidate can not be influenced externally. This would be a suitable use of reservoir sampling.

### DIFF
--- a/node/core/provisioner/src/lib.rs
+++ b/node/core/provisioner/src/lib.rs
@@ -25,6 +25,7 @@ use futures::{
 	prelude::*,
 };
 use futures_timer::Delay;
+use parity_scale_codec::Encode;
 use polkadot_node_subsystem::{
 	errors::{ChainApiError, RuntimeApiError},
 	jaeger,
@@ -53,6 +54,8 @@ mod tests;
 
 /// How long to wait before proposing.
 const PRE_PROPOSE_TIMEOUT: std::time::Duration = core::time::Duration::from_millis(2000);
+const MAX_CANDIDATES_SIZE: usize = 53_000;
+const MAX_SINGLE_CANDIDATE_SIZE: usize = 1_500;
 
 const LOG_TARGET: &str = "parachain::provisioner";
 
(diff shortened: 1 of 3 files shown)


node/core/provisioner/src/lib.rs
```diff
@@ -466,18 +469,42 @@ async fn select_candidates(
...
-	});
+		// If candidate contains new validation code, we update cumulative_size
+		if c.candidate
+			.commitments
+			.new_validation_code
+			.as_ref()
+			.map(|_| {
+				cumulative_size.saturating_sub(
+					with_validation_code
+						.replace(c.clone())
```
Lldenaurois, line 498: If our candidate has `new_validation_code`, we want to replace the old `with_validation_code` via `fn replace`, but we must also remove the old candidate's size from `cumulative_size` in order to ensure we don't overcount the block length.

If `with_validation_code` is `None`, then we can `unwrap_or_default()` to 0.

node/core/provisioner/src/lib.rs
```diff
@@ -466,18 +469,42 @@ async fn select_candidates(
...
-		if c.candidate.commitments.new_validation_code.is_some() {
-			if with_validation_code {
-				return false
-			}
+	let mut with_validation_code = None;
+	let mut cumulative_size = 0usize;
+	let mut out = Vec::new();
+
+	// Drain the candidates list to extract a suitable list to propose
+	while let Some(c) = candidates.pop() {
```
Lldenaurois, line 477: In theory, we could achieve this via `.into_iter().map(...)`, but I feel like the while let loop is easier to read.

bkchr:
We already have the following logic in the runtime: https://github.com/paritytech/polkadot/blob/ddd1a90beca9a680c04ecb8f4a29ca67a8d60e88/runtime/parachains/src/paras_inherent.rs#L288-L322

Not sure why we need something, that can estimate the weight even less optimal here in the node.

PR #3727, opened 2021-08-26
XCM: Allow reclaim of assets dropped from holding

### LINKED ISSUE #3694: XCM: Asset trap to avoid accidental burning of Holding Register contents

gavofyork:
Introduces a mechanism for systematically allowing the rescue of lost contents of the Holding Register.

Fixes #3694 

## Trapping assets

XCM Executor's config now includes `AssetTrap` type which is a hook to which the contents of the Holding Register is passed at the close of execution. Like `OnResponse`, a (recommended) implementation is provided by the XCM pallet.

This will record a hash of the assets left in the Holding Register at the finish of execution together with the origin. A later claim may be made from the origin in order to resurrect the assets. This is done through the single-function trait `ClaimAssets` which verifies that the assets to be claimed match a hash placed by the same origin.

If the version with which the `MultiAssets` have been stored is not the most recent version, then the version index must be included within the claim `ticket` parameter, a `MultiLocation` as a `GeneralIndex`.

## `ClaimAsset`

A new instruction is provided, `ClaimAsset`, which gains access to any `ClaimAssets` implementations exposed to the XCM Executor. If the `pallet_xcm::Pallet` implementation is exposed and the origin has unclaimed assets trapped, then those assets may be provided (exactly, as a hash of them is used to validate the claim) together with a ticket reflecting the version of XCM at the time of the trap (or `Here` if unchanged) and they will be resurrected into the Holding Register.

## Additional

Also introduces the `Trap` instruction and error type to force an error to be thrown, and be able to recognise it. It comes with a `u64` parameter to allow some degree of information to be conveyed with it (without bloating the stack footprint of `XcmError`).

### `xcm_executor::Config`

Two new type definitions are required, and they'll generally want to point at the XCM Pallet concrete type if there is one, e.g.:

```rust
impl xcm_executor::Config for ExecuteXcm {
	/* snip */
	type AssetTrap = XcmPallet;
	type AssetClaims = XcmPallet;
}
```

Note that destinations which use v1 or v0 for message transport will not be able to use these features (even if the destination supports v2).

## TODO
- [x] `ShouldTrap` filter type to allow runtimes to dictate which assets/origins should be trapped.
- [x] Unit test.

### DIFF
--- a/xcm/pallet-xcm/src/lib.rs
+++ b/xcm/pallet-xcm/src/lib.rs
@@ -249,7 +249,7 @@ pub mod pallet {
 
 	/// The existing asset traps.
 	///
-	/// Key is the blake2 256 hash of (origin, versioned multiassets) pair. Value is the number of
+	/// Key is the blake2 256 hash of (origin, versioned `MultiAssets`) pair. Value is the number of
 	/// times this pair has been trapped (usually just 1 if it exists at all).
 	#[pallet::storage]
 	#[pallet::getter(fn asset_trap)]
--- a/xcm/pallet-xcm/src/lib.rs
+++ b/xcm/pallet-xcm/src/lib.rs
@@ -50,7 +50,8 @@ pub mod pallet {
 		pallet_prelude::*,
 	};
 	use frame_system::{pallet_prelude::*, Config as SysConfig};
-	use sp_runtime::traits::{AccountIdConversion, BlockNumberProvider};
+	use sp_core::H256;
+	use sp_runtime::traits::{AccountIdConversion, BlakeTwo256, BlockNumberProvider, Hash};
 	use xcm_executor::{Assets, traits::{ClaimAssets, DropAssets, InvertLocation, OnResponse, WeightBounds}};
 
 	#[pallet::pallet]
@@ -170,6 +171,10 @@ pub mod pallet {
 		///
 		/// \[ id \]
 		ResponseTaken(QueryId),
+		/// Some assets have been placed in an asset trap.
+		/// 
+		/// \[ hash, origin, assets \]
+		AssetsTrapped(H256, MultiLocation, VersionedMultiAssets),
 	}
 
 	#[pallet::origin]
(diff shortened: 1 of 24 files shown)


xcm/src/v2/mod.rs
```diff
@@ -505,21 +524,54 @@ pub enum Instruction<Call> {
...
 
-	/// Clear the error register.
+	/// Clear the Error Register.
+	///
+	/// Kind: *Instruction*
+	///
+	/// Errors: None.
 	ClearError,
+
+	/// Create some assets which are being held on behalf of the origin.
```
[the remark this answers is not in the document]
gavofyork: In this context they're interchangeable - the Assets will be "created" in the sense that the `MultiAsset` instance(s) will be created.

xcm/xcm-executor/src/config.rs
```diff
@@ -58,4 +58,12 @@ pub trait Config {
...
 	/// What to do when a response of a query is found.
 	type ResponseHandler: OnResponse;
+
+	/// The general asset trap - handler for when assets are left in the Holding Register at the
+	/// end of execution.
+	type AssetTrap: DropAssets;
+
+	/// The general asset trap - handler for when assets are left in the Holding Register at the
+	/// end of execution.
+	type AssetClaims: ClaimAssets;
```
KiChjang, line 68: The docs for `AssetClaims` is exactly the same as `AssetTrap`, it doesn't seem to be intended?

xcm/pallet-xcm/src/lib.rs
```diff
@@ -553,16 +569,54 @@ pub mod pallet {
...
 
+	impl<T: Config> DropAssets for Pallet<T> {
+		fn drop_assets(origin: &MultiLocation, assets: Assets) -> Weight {
+			if assets.is_empty() {
+				return 0
+			}
+			let versioned = VersionedMultiAssets::from(MultiAssets::from(assets));
+			let hash = BlakeTwo256::hash_of(&(&origin, &versioned));
+			AssetTraps::<T>::mutate(hash, |n| *n += 1);
+			Self::deposit_event(Event::AssetsTrapped(hash, origin.clone(), versioned));
```
KiChjang, line 580: This event becomes really, really important if people want to look for their own assets that were trapped inside of the pallet. Wonder if it's a good idea to allow some amount of querying by the sender into the map to see how much assets were trapped?

gavofyork: The map only stores hashes, by design. If they care, they'll need to check the event.

KiChjang (approved):
Nothing really stands out as concerning, aside from some minor documentation stuff.

PR #3728, opened 2021-08-26
Introduce new Runtime API endpoint for fetching the validation data

### LINKED ISSUE #3700: New Parachains Runtime API tailored for backing.
Right now during candidate validation for backing we query persisted validation data. We do it once with the assumption that the core is free and if the hash of persisted validation data does not match the hash found in the candidate we try once more with assumption that the current candidate occupied the core is enacted. Once we figured out what persisted validation data matches we query runtime API for the current validation code. This amounts to 3 runtime API requests which seems to be inefficient.

Furthermore, during working on https://github.com/paritytech/polkadot/pull/3655 we've realized that we actually need to get the validation code that parachain currently has. This functionality exists but is not exposed as a runtime API.

In order to solve two problems at once I propose a new one-stop shop API which will be tailored specifically for backing. 

Next follows a rough description of the API:

The API accepts the `para_id`, `expected_persisted_validation_data_hash` and returns the `validation_code_hash` and the preimage of `persisted_validation_data`. The API is supposed to be called against the `descriptor.relay_parent`.

The implementation will do a check:

Obtain the `persisted_validation_data` for the given `para_id` and hash it. If the hash matches the passed `expected_persisted_validation_data_hash` then query the current validation code hash and return the persisted validation data preimage.

This check is performed twice actually: first without force enacting the core and then with it. If the core is not occupied we can skip the second check since it won't do any difference. If the match was found on the first attempt the implementation should return immediately.

slumber:
The corresponding implementers guide section will be updated along with the candidate validation subsystem as a part of #3655 
Resolves #3700

cumulus companion: https://github.com/paritytech/cumulus/pull/693

### DIFF
--- a/runtime/kusama/src/lib.rs
+++ b/runtime/kusama/src/lib.rs
@@ -49,9 +49,7 @@ use authority_discovery_primitives::AuthorityId as AuthorityDiscoveryId;
 use beefy_primitives::crypto::AuthorityId as BeefyId;
 use frame_support::{
 	construct_runtime, parameter_types,
-	traits::{
-		Contains, Everything, InstanceFilter, KeyOwnerProofSystem, LockIdentifier, Nothing,
-	},
+	traits::{Contains, Everything, InstanceFilter, KeyOwnerProofSystem, LockIdentifier, Nothing},
 	weights::Weight,
 	PalletId, RuntimeDebug,
 };


bkchr:
We should probably also bump the version of the runtime api? And add some fallback for when it is not available?

pepyakin:
The code that depends on this PR, https://github.com/paritytech/polkadot/pull/3655, is actually a low-priority one. We can as well delay the dependent PR until after this gets on the relay-chain. This way we will save us some effort by not introducing a fallback.

Do you think it is sensible?

bkchr:
Okay, sounds fine.

PR #3730, opened 2021-08-26
Return a Result in invert_location

KiChjang:
Fixes https://github.com/paritytech/srlabs_findings/issues/105.

Removes an unproven assertion in `invert_location` and instead make it fallible and return a Result.


gavofyork (approved):

gavofyork:
spelling needs to be fixed.

PR #3736, opened 2021-08-28
XCM: Automatic Version Negotiation

gavofyork:
This introduces autonomous version negotiation in XCM, removing the need for manual intervention in order to ensure that chains use the most recent common version of XCM which they support.

It introduces a stateful `WrapVersion` implementation through the XCM Pallet, together with a new `SubscribeVersion` instruction (as well as `UnsubscribeVersion`) to provide a `QueryResponse` with the latest XCM a runtime supports together with further similar notifications when the version changes. When a message is sent to a destination whose version is unknown, the pallet queues up this destination for version querying and each block drains some number from this queue and posts a query. As they come in, it checks for responses and (assuming it knowingly subscribed for it) records it.

The Pallet maintains a list of locations which have asked for version notification subscriptions and whenever there is an upgrade sends them notifications of the new upgrade. This is done through a new trait `VersionChangeNotifier`.

### XCM Pallet

`pallet_xcm::Config` trait get one new const `VERSION_DISCOVERY_QUEUE_SIZE` and one new type `AdvertizeXcmVersion`. A sensible value for the former is 100. For the latter, it's best to just reference `pallet_xcm::CurrentXcmVersion`. e.g.:

```rust
	/* snip */
	const VERSION_DISCOVERY_QUEUE_SIZE: u32 = 100;
	type AdvertizeXcmVersion = pallet_xcm::CurrentXcmVersion;
}
```

### XCM Config

Ensure that any `xcm_executor::Config` implementations includes a definition for the new type `SubscriptionService`. Generally this should be set to `PalletXcm`, the `pallet_xcm` concrete type. i.e.:

```rust
impl xcm_executor::Config for XcmConfig {
	/* snip */
	type SubscriptionService = XcmPallet;
}
```

### Routers & `WrapVersion`

Any routers (XCM transport methods) which have a `WrapVersion` item (including `cumulus_pallet_xcmp_queue` and `ParentAsUmp`) should have that item set to be implemented by `PalletXcm`, the `pallet_xcm` concrete type. e.g.:

```rust
impl cumulus_pallet_xcmp_queue::Config for Runtime {
	/* snip */
	type VersionWrapper = ();
}
/* snip */
pub type XcmRouter = (
	cumulus_primitives_utility::ParentAsUmp<ParachainSystem, ()>,
	XcmpQueue,
);
```

would become:

```rust
impl cumulus_pallet_xcmp_queue::Config for Runtime {
	/* snip */
	type VersionWrapper = XcmPallet;
}
/* snip */
pub type XcmRouter = (
	cumulus_primitives_utility::ParentAsUmp<ParachainSystem, XcmPallet>,
	XcmpQueue,
);
```
**NOTE** In addition to this change, you should do at least one of two things to initialize your chain and ensure that XCM messages can be sent:

- Whitelist destinations to which you wish to send messages with the `force_xcm_version` (root-callable) extrinsic.
- Provide a general fallback version with the `force_xcm_default_version` (root-callable) extrinsic.

If you do the latter, then the automatic XCM version tracking code with work as expected and destination XCM versions will be kept up to date.

### Barrier

An additional barrier `AlllowSubscriptionsFrom` should be included in your `Barriers` tuple. You can probably set the origin filter to `Everything`, unless you want to filter which locations can ask to be notified of your chain's XCM version changes:

```rust
pub type Barrier = (
	TakeWeightCredit,
	/* snip */
	AllowSubscriptionsFrom<Everything>,
);
```

## TODO

- [x] Automatic subscription
- [x] Automatic unsubscription
- [x] Unsubscription extrinsic
- [x] Barrier
- [x] Expose remote origin in VM and ensure it is unchanged from actual origin in subscription instructions.
- [x] Cleanup for orphaned subscription requests
- [x] Unit tests for XCM executor
- [x] Unit tests for XCM pallet

Maybe here or another PR:
- [x] Remove `top_level`
- [x] Allow `on_runtime_upgrade` notifications to happen over multiple blocks
- [ ] Break apart `fn on_response`

### DIFF
--- a/roadmap/implementers-guide/src/types/overseer-protocol.md
+++ b/roadmap/implementers-guide/src/types/overseer-protocol.md
@@ -441,7 +441,7 @@ enum DisputeCoordinatorMessage {
         pending_confirmation: oneshot::Sender<ImportStatementsResult>
     },
     /// Fetch a list of all recent disputes that the co-ordinator is aware of.
-    /// These are disputes which have occured any time in recent sessions, which may have already concluded.
+    /// These are disputes which have occurerd any time in recent sessions, which may have already concluded.
     RecentDisputes(ResponseChannel<Vec<(SessionIndex, CandidateHash)>>),
     /// Fetch a list of all active disputes that the co-ordinator is aware of.
     /// These disputes are either unconcluded or recently concluded.
--- a/roadmap/implementers-guide/src/types/overseer-protocol.md
+++ b/roadmap/implementers-guide/src/types/overseer-protocol.md
@@ -699,7 +699,7 @@ The Runtime API subsystem is responsible for providing an interface to the state
 
 This is fueled by an auxiliary type encapsulating all request types defined in the Runtime API section of the guide.
 
-> TODO: link to the Runtime API section. Not possible currently because of https://github.com/Michael-F-Bryan/mdbook-linkcheck/issues/25. Once v0.7.1 is released it will work.
+> To do: link to the Runtime API section. Not possible currently because of https://github.com/Michael-F-Bryan/mdbook-linkcheck/issues/25. Once v0.7.1 is released it will work.
 
 ```rust
 enum RuntimeApiRequest {
--- a/xcm/pallet-xcm/src/lib.rs
+++ b/xcm/pallet-xcm/src/lib.rs
@@ -260,9 +260,6 @@ pub mod pallet {
 		}
 	}
 
-	/// Value of a query, must be unique for each query.
-	pub type QueryId = u64;
-
 	/// The latest available query index.
 	#[pallet::storage]
 	pub(super) type QueryCount<T: Config> = StorageValue<_, QueryId, ValueQuery>;
--- a/xcm/pallet-xcm/src/lib.rs
+++ b/xcm/pallet-xcm/src/lib.rs
@@ -25,7 +25,7 @@ mod tests;
 
 use codec::{Decode, Encode, EncodeLike};
 use frame_support::traits::{Contains, EnsureOrigin, Get, OriginTrait};
-use sp_runtime::{traits::BadOrigin, RuntimeDebug};
+use sp_runtime::{traits::{BadOrigin, Saturating}, RuntimeDebug};
 use sp_std::{
 	boxed::Box,
 	convert::{TryFrom, TryInto},
@@ -106,6 +106,9 @@ pub mod pallet {
 			+ GetDispatchInfo
 			+ IsType<<Self as frame_system::Config>::Call>
 			+ Dispatchable<Origin = <Self as Config>::Origin, PostInfo = PostDispatchInfo>;
+
+		/// The maximum number of items that we can store in the version discovery queue.
+		type VersionDiscoveryQueueSize: Get<u32>;
 	}
 
 	/// The maximum number of distinct assets allowed to be transferred in a single helper extrinsic.
@@ -325,8 +328,37 @@ pub mod pallet {
 		OptionQuery,
 	>;
 
+	/// Destinations whose latest XCM version we would like to know. Duplicates allowed. The block
+	/// number is the most recent block which pushed to it.
+	#[pallet::storage]
+	pub(super) type VersionDiscoveryQueue<T: Config> = StorageValue<
+		_,
+		BoundedVec<(VersionedMultiLocation, u32), T::VersionDiscoveryQueueSize>,
+		ValueQuery,
+	>;
+
 	#[pallet::hooks]
 	impl<T: Config> Hooks<BlockNumberFor<T>> for Pallet<T> {
+		fn on_initialize(_n: BlockNumberFor<T>) -> Weight {
+			// Here we aim to get one successful version negotiation request sent per block, ordered
+			// by the destinations being most sent to.
+			let mut q = VersionDiscoveryQueue::<T>::get().into_inner();
+			q.sort_by_key(|i| i.1);
+			while let Some((versioned_dest, _)) = q.pop() {
+				if let Ok(dest) = versioned_dest.try_into() {
+					if Self::request_version_notify(dest).is_ok() {
+						break;
+					}
+				}
+			}
+			// Should never fail since we only removed items. But better safe than panicking as it's
+			// way better to drop the queue than panic on initialize.
+			if let Ok(q) = BoundedVec::try_from(q) {
+				VersionDiscoveryQueue::<T>::put(q);
+			}
+			// TODO: correct weight.
+			0
+		}
 		fn on_runtime_upgrade() -> Weight {
 			for v in 0..XCM_VERSION {
 				for (old_key, value) in SupportedVersion::<T>::drain_prefix(v) {
@@ -798,6 +830,20 @@ pub mod pallet {
 				None
 			}
 		}
+
+		/// Note that a particular destination to whom we would like to send a message is unknown
+		/// and queue it for version discovery.
+		fn note_unknown_version(dest: &MultiLocation) {
+			let versioned_dest = VersionedMultiLocation::from(dest.clone());
+			VersionDiscoveryQueue::<T>::mutate(|q| {
+				if let Some(index) = q.iter().position(|i| &i.0 == &versioned_dest) {
+					// exists - just bump the count.
+					q[index].1.saturating_inc();
+				} else {
+					let _ = q.try_push((versioned_dest, 0));
+				}
+			});
+		}
 	}
 
 	impl<T: Config> WrapVersion for Pallet<T> {
@@ -806,7 +852,10 @@ pub mod pallet {
 			xcm: impl Into<VersionedXcm<Call>>,
 		) -> Result<VersionedXcm<Call>, ()> {
 			SupportedVersion::<T>::get(XCM_VERSION, LatestVersionedMultiLocation(dest))
-				.or_else(SafeXcmVersion::<T>::get)
+				.or_else(|| {
+					Self::note_unknown_version(dest);
+					SafeXcmVersion::<T>::get()
+				})
 				.ok_or(())
 				.and_then(|v| xcm.into().into_version(v))
 		}
--- a/xcm/pallet-xcm/src/lib.rs
+++ b/xcm/pallet-xcm/src/lib.rs
@@ -236,6 +236,8 @@ pub mod pallet {
 		/// The given location could not be used (e.g. because it cannot be expressed in the
 		/// desired version of XCM).
 		BadLocation,
+		/// The referenced subscription could not be found.
+		NoSubscription,
 	}
 
 	/// The status of a query.
@@ -610,6 +612,25 @@ pub mod pallet {
 			ensure_root(origin)?;
 			SafeXcmVersion::<T>::set(maybe_xcm_version);
 			Ok(())
+		}
+
+		/// Require that a particular destination should no longer notify us regarding any XCM
+		/// version changes.
+		///
+		/// - `origin`: Must be Root.
+		/// - `location`: The location to which we are currently subscribed for XCM version
+		///   notifications which we no longer desire.
+		#[pallet::weight(100_000_000u64)]
+		pub fn force_unsubscribe_version_notify(
+			origin: OriginFor<T>,
+			location: Box<MultiLocation>,
+		) -> DispatchResult {
+			ensure_root(origin)?;
+			let location = *location;
+			Self::unrequest_version_notify(location).map_err(|e| match e {
+				XcmError::InvalidLocation => Error::<T>::NoSubscription,
+				_ => Error::<T>::InvalidOrigin,
+			}.into())
 		}
 	}
 
--- a/xcm/pallet-xcm/src/lib.rs
+++ b/xcm/pallet-xcm/src/lib.rs
@@ -780,16 +780,15 @@ pub mod pallet {
 			if stage == NotifyCurrentTargets {
 				for (key, value) in VersionNotifyTargets::<T>::iter_prefix(XCM_VERSION) {
 					let (query_id, max_weight, mut target_xcm_version) = value;
-					let new_key: MultiLocation =
-						match key.clone().try_into() {
-							Ok(k) if target_xcm_version != xcm_version => k,
-							_ => {
-								// We don't early return here since we need to be certain that we
-								// make some progress.
-								weight_used.saturating_accrue(todo_vnt_already_notified_weight);
-								continue
-							},
-						};
+					let new_key: MultiLocation = match key.clone().try_into() {
+						Ok(k) if target_xcm_version != xcm_version => k,
+						_ => {
+							// We don't early return here since we need to be certain that we
+							// make some progress.
+							weight_used.saturating_accrue(todo_vnt_already_notified_weight);
+							continue
+						},
+					};
 					let response = Response::Version(xcm_version);
 					let message = Xcm(vec![QueryResponse { query_id, response, max_weight }]);
 					let event = match T::XcmRouter::send_xcm(new_key.clone(), message) {
@@ -836,7 +835,8 @@ pub mod pallet {
 						} else {
 							// Need to notify target.
 							let response = Response::Version(xcm_version);
-							let message = Xcm(vec![QueryResponse { query_id, response, max_weight }]);
+							let message =
+								Xcm(vec![QueryResponse { query_id, response, max_weight }]);
 							let event = match T::XcmRouter::send_xcm(new_key.clone(), message) {
 								Ok(()) => {
 									VersionNotifyTargets::<T>::insert(
--- a/xcm/pallet-xcm/src/lib.rs
+++ b/xcm/pallet-xcm/src/lib.rs
@@ -342,9 +342,13 @@ pub mod pallet {
 					let new_key = match MultiLocation::try_from(old_key.clone()) {
 						Ok(k) => k,
 						Err(()) => {
-        					Self::deposit_event(Event::NotifyTargetDropped(old_key, value.0, XcmError::InvalidLocation));
-							return 0;
-    					},
+							Self::deposit_event(Event::NotifyTargetDropped(
+								old_key,
+								value.0,
+								XcmError::InvalidLocation,
+							));
+							return 0
+						},
 					};
 					let instruction = QueryResponse {
 						query_id: value.0,
@@ -362,8 +366,12 @@ pub mod pallet {
 						},
 						Err(e) => {
 							let new_key = new_key.into();
-							Self::deposit_event(Event::NotifyTargetDropped(new_key, value.0, e.into()));
-						}
+							Self::deposit_event(Event::NotifyTargetDropped(
+								new_key,
+								value.0,
+								e.into(),
+							));
+						},
 					}
 				}
 			}
--- a/xcm/pallet-xcm/src/lib.rs
+++ b/xcm/pallet-xcm/src/lib.rs
@@ -810,7 +810,7 @@ pub mod pallet {
 					weight_used.saturating_accrue(todo_vnt_notify_weight);
 					if weight_used >= weight_cutoff {
 						let last = Some(iter.last_raw_key().into());
-						return (weight_used, Some(NotifyCurrentTargets(last)));
+						return (weight_used, Some(NotifyCurrentTargets(last)))
 					}
 				}
 				stage = MigrateAndNotifyOldTargets;
--- a/xcm/pallet-xcm/src/lib.rs
+++ b/xcm/pallet-xcm/src/lib.rs
@@ -23,7 +23,7 @@ mod mock;
 #[cfg(test)]
 mod tests;
 
-use codec::{Decode, Encode};
+use codec::{Decode, Encode, EncodeLike};
 use frame_support::traits::{Contains, EnsureOrigin, Get, OriginTrait};
 use sp_runtime::{traits::BadOrigin, RuntimeDebug};
 use sp_std::{
@@ -34,10 +34,7 @@ use sp_std::{
 	result::Result,
 	vec,
 };
-use xcm::{
-	latest::prelude::*, VersionedMultiAssets, VersionedMultiLocation, VersionedResponse,
-	VersionedXcm,
-};
+use xcm::prelude::*;
 use xcm_executor::traits::ConvertOrigin;
 
 use frame_support::PalletId;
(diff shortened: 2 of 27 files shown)


roadmap/implementers-guide/src/types/overseer-protocol.md
```diff
@@ -441,7 +441,7 @@ enum DisputeCoordinatorMessage {
         pending_confirmation: oneshot::Sender<ImportStatementsResult>
     },
     /// Fetch a list of all recent disputes that the co-ordinator is aware of.
-    /// These are disputes which have occured any time in recent sessions, which may have already concluded.
+    /// These are disputes which have occurerd any time in recent sessions, which may have already concluded.
```
shawntabrizi, line 444:
```suggestion
    /// These are disputes which have occured any time in recent sessions, which may have already concluded.
```

xcm/pallet-xcm/src/mock.rs
```diff
@@ -306,3 +322,14 @@ pub(crate) fn new_test_ext_with_balances(
 	ext.execute_with(|| System::set_block_number(1));
 	ext
 }
+/*
+pub(crate) fn run_to_block(n: BlockNumber) {
+	while System::block_number() < n {
+		XcmPallet::on_finalize(System::block_number());
```
shawntabrizi, line 328: comment code

xcm/pallet-xcm/src/lib.rs
```diff
@@ -106,6 +116,13 @@ pub mod pallet {
 			+ GetDispatchInfo
 			+ IsType<<Self as frame_system::Config>::Call>
 			+ Dispatchable<Origin = <Self as Config>::Origin, PostInfo = PostDispatchInfo>;
+
+		/// The maximum number of items that we can store in the version discovery queue.
+		type VersionDiscoveryQueueSize: Get<u32>;
+
+		/// The latest supported version that we advertise. Generally just set it to
+		/// `pallet_xcm::CurrentXcmVersion`.
+		type AdvertizeXcmVersion: Get<XcmVersion>;
```
shawntabrizi, line 125:
```suggestion
		type AdvertisedXcmVersion: Get<XcmVersion>;
```

maybe doesnt need to be past tense, but should be US english for a public api

xcm/pallet-xcm/src/lib.rs
```diff
@@ -106,6 +116,13 @@ pub mod pallet {
 			+ GetDispatchInfo
 			+ IsType<<Self as frame_system::Config>::Call>
 			+ Dispatchable<Origin = <Self as Config>::Origin, PostInfo = PostDispatchInfo>;
+
+		/// The maximum number of items that we can store in the version discovery queue.
+		type VersionDiscoveryQueueSize: Get<u32>;
```
shawntabrizi, line 121: We should avoid using `Get` where we expect a const, else someone may implement this in a bad way :|

```suggestion
		const VersionDiscoveryQueueSize: u32;
```

shawntabrizi: there are places in substrate we need to do the same

xcm/pallet-xcm/src/lib.rs
```diff
@@ -255,8 +309,103 @@ pub mod pallet {
...
+			q.sort_by_key(|i| i.1);
+			while let Some((versioned_dest, _)) = q.pop() {
+				if let Ok(dest) = versioned_dest.try_into() {
+					if Self::request_version_notify(dest).is_ok() {
+						break
+					}
+				}
+			}
+			// Should never fail since we only removed items. But better safe than panicking as it's
+			// way better to drop the queue than panic on initialize.
```
shawntabrizi, line 398: I think this comment implies you actually want to do a `take` instead of a `get` above, which makes sense to me, else we will repeat this error condition over and over.

xcm/pallet-xcm/src/lib.rs
```diff
@@ -448,9 +597,203 @@ pub mod pallet {
...
+			})
+		}
 	}
 
 	impl<T: Config> Pallet<T> {
+		fn check_xcm_version_change() -> Weight {
+			// We assume that supported XCM version only ever increases, so just cycle through lower
+			// XCM versioned from the current.
+			for v in 0..XCM_VERSION {
+				for (old_key, value) in SupportedVersion::<T>::drain_prefix(v) {
```
shawntabrizi, line 691: it is important that this will be bounded, which i believe it is to the number of parachains in the immediate relay chain system, but may not scale to the nested relay chain future with unbounded number of locations.

shawntabrizi: probably same below

xcm/pallet-xcm/src/lib.rs
```diff
@@ -448,9 +597,203 @@ pub mod pallet {
...
+				}
+				for (old_key, value) in VersionNotifiers::<T>::drain_prefix(v) {
+					if let Ok(new_key) = old_key.into_latest() {
+						VersionNotifiers::<T>::insert(XCM_VERSION, new_key, value);
+					}
+				}
+			}
+			let xcm_version = T::AdvertizeXcmVersion::get();
+			let response = Response::Version(xcm_version);
+			for (key, mut value) in VersionNotifyTargets::<T>::iter_prefix(XCM_VERSION) {
```
shawntabrizi, line 704: also potentially unbounded

xcm/pallet-xcm/src/lib.rs
```diff
@@ -448,9 +597,203 @@ pub mod pallet {
...
+		}
+
+		/// Request that `dest` informs us of its version.
+		pub fn request_version_notify(dest: MultiLocation) -> XcmResult {
+			let versioned_dest = VersionedMultiLocation::from(dest.clone());
+			let already = VersionNotifiers::<T>::contains_key(XCM_VERSION, &versioned_dest);
+			ensure!(!already, XcmError::InvalidLocation);
+			let query_id = QueryCount::<T>::mutate(|q| {
+				let r = *q;
+				*q += 1;
```
shawntabrizi, line 774: practically speaking, this should be a checked add, and can fail to mutate if overflow

xcm/pallet-xcm/src/lib.rs
```diff
@@ -448,9 +597,203 @@ pub mod pallet {
...
+			Ok(())
+		}
+
+		/// Request that `dest` ceases informing us of its version.
+		pub fn unrequest_version_notify(dest: MultiLocation) -> XcmResult {
+			let versioned_dest = LatestVersionedMultiLocation(&dest);
+			let query_id = VersionNotifiers::<T>::take(XCM_VERSION, versioned_dest)
+				.ok_or(XcmError::InvalidLocation)?;
+			T::XcmRouter::send_xcm(dest.clone(), Xcm(vec![UnsubscribeVersion]))?;
+			Queries::<T>::remove(query_id);
```
shawntabrizi, line 793: need to decrease query count

xcm/pallet-xcm/src/lib.rs
```diff
@@ -579,8 +922,71 @@ pub mod pallet {
...
 		}
+
+		/// Note that a particular destination to whom we would like to send a message is unknown
+		/// and queue it for version discovery.
+		fn note_unknown_version(dest: &MultiLocation) {
+			let versioned_dest = VersionedMultiLocation::from(dest.clone());
+			VersionDiscoveryQueue::<T>::mutate(|q| {
+				if let Some(index) = q.iter().position(|i| &i.0 == &versioned_dest) {
+					// exists - just bump the count.
+					q[index].1.saturating_inc();
```
shawntabrizi, line 933: this second value in the queue is just a priority hint?

xcm/pallet-xcm/src/lib.rs
gavofyork: Why should that be a *practical* issue?

xcm/pallet-xcm/src/lib.rs
```diff
@@ -579,8 +922,71 @@ pub mod pallet {
...
+			Ok(())
+		}
+
+		/// Stop notifying `location` should the XCM change. Returns an error if there is no existing
+		/// notification set up.
+		fn stop(dest: &MultiLocation) -> XcmResult {
+			let versioned_dest = LatestVersionedMultiLocation(dest);
+			let already = VersionNotifyTargets::<T>::contains_key(XCM_VERSION, versioned_dest);
+			ensure!(already, XcmError::InvalidLocation);
+			VersionNotifyTargets::<T>::remove(XCM_VERSION, versioned_dest);
```
shawntabrizi, line 986: im not convinced this is preferable to just removing the value. the only upside is that there wont be the logical path of a DB write in the case of no value existing, but that will actually not be that expensive since it will not modify the merkle trie when there is nothing to update.

xcm/pallet-xcm/src/lib.rs
```diff
@@ -635,87 +1044,133 @@ pub mod pallet {
...
-											Self::deposit_event(e);
-											// Not much to do with the result as it is. It's up to the parachain to ensure that the
-											// message makes sense.
-											error_and_info.post_info.actual_weight
-										},
-									}
-									.unwrap_or(weight)
-								} else {
-									let e = Event::NotifyDecodeFailed(
+			match (response, Queries::<T>::get(query_id)) {
```
shawntabrizi, line 1047: in terms of weighing this function, it would be helpful to break down this large match statement into sub-functions, which would then allow us to weight the individual logical paths in a more certain and controllable way. If you want, this refactor can happen in #3735

shawntabrizi (commented):
most other stuff looks okay to me

xcm/pallet-xcm/src/lib.rs
shawntabrizi: I get that u64 in this context will never overflow, but i just don't see why we wouldn't use checked math here since it comes at practically zero cost. Your call, hopefully its not too much of a pain that I continue to point out such things

xcm/pallet-xcm/src/lib.rs
shawntabrizi: i saw later in the review that you dont allow sub-locations, which may resolve this point. however, i think it would be better to design this logic only taking N items at a time, where N is some defined upper limit, and handling not being able to go through all the items at once.

xcm/pallet-xcm/src/lib.rs
gavofyork: this must be passed as a type parameter to a `BoundedVec`. would you expect it to be re-wrapped in a `type_parameter` and passed that way?

xcm/pallet-xcm/src/lib.rs
gavofyork: it's a counter for the number of times a send to the destination has been attempted.

gavofyork: docs updated.

xcm/pallet-xcm/src/lib.rs
gavofyork: done.

shawntabrizi (approved):

xcm/pallet-xcm/src/lib.rs
```diff
@@ -230,16 +272,41 @@ pub mod pallet {
...
+			r.truncate(1);
+			self.0.using_encoded(|d| r.extend_from_slice(d));
+			r
+		}
+	}
+
+	#[derive(Copy, Clone, Encode, Decode, Eq, PartialEq, Ord, PartialOrd)]
+	pub enum VersionMigrationStage {
+		MigratateSupportedVersion,
+		MigratateVersionNotifiers,
```
KiChjang, line 296:
```suggestion
		MigrateSupportedVersion,
		MigrateVersionNotifiers,
```

xcm/pallet-xcm/src/lib.rs
```diff
@@ -230,16 +272,41 @@ pub mod pallet {
...
+		NotifyCurrentTargets,
+		MigrateAndNotifyOldTargets,
+	}
+
+	impl Default for VersionMigrationStage {
+		fn default() -> Self {
+			Self::MigratateSupportedVersion
```
KiChjang, line 303:
```suggestion
			Self::MigrateSupportedVersion
```

xcm/pallet-xcm/src/lib.rs
```diff
@@ -230,16 +272,41 @@ pub mod pallet {
...
-	/// Value of a query, must be unique for each query.
-	pub type QueryId = u64;
+	#[derive(Copy, Clone)]
+	pub(crate) struct LatestVersionedMultiLocation<'a>(pub(crate) &'a MultiLocation);
+	impl<'a> EncodeLike<VersionedMultiLocation> for LatestVersionedMultiLocation<'a> {}
+	impl<'a> Encode for LatestVersionedMultiLocation<'a> {
+		fn encode(&self) -> Vec<u8> {
+			let mut r = VersionedMultiLocation::from(MultiLocation::default()).encode();
+			r.truncate(1);
+			self.0.using_encoded(|d| r.extend_from_slice(d));
```
KiChjang, line 288: Huh, if `LatestVersionedMultiLocation` encodes exactly like `VersionedMultiLocation`, what's the reasoning behind this code, i.e. extracting the first byte of a `VersionedMultiLocation` and appending the encoded `MultiLocation` to it?

KiChjang: Oh wait nevermind, `LatestVersionedMultiLocation` is a struct, not an enum like `VersionedMultiLocation`.

xcm/src/v1/mod.rs
```diff
@@ -79,6 +79,8 @@ pub mod prelude {
 pub enum Response {
 	/// Some assets.
 	Assets(MultiAssets),
+	/// An XCM version.
+	Version(super::Version),
 }
```
KiChjang, line 85: No VERSION constant for v1?

xcm/xcm-executor/src/traits/on_response.rs
```diff
@@ -42,3 +42,29 @@ impl OnResponse for () {
 		0
 	}
 }
+
+/// Trait for a type which handles notifying a destination of XCM version changes.
+pub trait VersionChangeNotifier {
+	/// Start notifying `location` should the XCM version of this chain change.
+	///
+	/// When it does, this type should ensure an `QueryResponse` message is sent with the given
```
KiChjang, line 50:
```suggestion
	/// When it does, this type should ensure a `QueryResponse` message is sent with the given
```

xcm/pallet-xcm/src/lib.rs
```diff
@@ -448,9 +638,268 @@ pub mod pallet {
...
+			ensure_root(origin)?;
+			let location = *location;
+			SupportedVersion::<T>::insert(
+				XCM_VERSION,
+				LatestVersionedMultiLocation(&location),
+				xcm_version,
+			);
+			Self::deposit_event(Event::SupportedVersionChanged(location, xcm_version));
+			Ok(())
+		}
```
KiChjang, line 667: Doesn't look like `BadLocation` can ever be returned by this dispatch function?

xcm/pallet-xcm/src/lib.rs
```diff
@@ -448,9 +638,268 @@ pub mod pallet {
...
+		///
+		/// - `origin`: Must be Root.
+		/// - `location`: The location to which we should subscribe for XCM version notifications.
+		#[pallet::weight(100_000_000u64)]
+		pub fn force_subscribe_version_notify(
+			origin: OriginFor<T>,
+			location: Box<VersionedMultiLocation>,
+		) -> DispatchResult {
+			ensure_root(origin)?;
+			let location = (*location).try_into().map_err(|()| Error::<T>::BadLocation)?;
```
KiChjang, line 694: I'm somewhat surprised that you need the explicit dereference here. Does the compiler not know how to auto-deref `location`?

KiChjang: Oh, I'm guessing this is done due to type inference not being smart enough to recognize what the target type is for `try_into`.

xcm/pallet-xcm/src/lib.rs
```diff
@@ -448,9 +638,268 @@ pub mod pallet {
...
+			let todo_vnt_migrate_weight: Weight =
+				T::DbWeight::get().read + T::DbWeight::get().write;
+			let todo_vnt_migrate_fail_weight: Weight =
+				T::DbWeight::get().read + T::DbWeight::get().write;
+			let todo_vnt_notify_migrate_weight: Weight =
+				T::DbWeight::get().read + T::DbWeight::get().write * 3;
+
+			use VersionMigrationStage::*;
+
+			if stage == MigratateSupportedVersion {
```
KiChjang, line 751:
```suggestion
			if stage == MigrateSupportedVersion {
```

xcm/pallet-xcm/src/lib.rs
```diff
@@ -448,9 +638,268 @@ pub mod pallet {
...
+						if let Ok(new_key) = old_key.into_latest() {
+							SupportedVersion::<T>::insert(XCM_VERSION, new_key, value);
+						}
+						weight_used.saturating_accrue(todo_sv_migrate_weight);
+						if weight_used >= weight_cutoff {
+							return (weight_used, Some(stage))
+						}
+					}
+				}
+				stage = MigratateVersionNotifiers;
```
KiChjang, line 765:
```suggestion
				stage = MigrateVersionNotifiers;
```

xcm/pallet-xcm/src/lib.rs
```diff
@@ -448,9 +638,268 @@ pub mod pallet {
...
+			}
+			if stage == MigratateVersionNotifiers {
```
KiChjang, line 767:
```suggestion
			if stage == MigrateVersionNotifiers {
```

xcm/pallet-xcm/src/lib.rs
```diff
@@ -448,9 +638,268 @@ pub mod pallet {
...
+			}
+
+			let xcm_version = T::AdvertisedXcmVersion::get();
+			let response = Response::Version(xcm_version);
+
+			if stage == NotifyCurrentTargets {
+				for (key, mut value) in VersionNotifyTargets::<T>::iter_prefix(XCM_VERSION) {
+					let new_key: MultiLocation =
+						match (value.2 == xcm_version, key.clone().try_into()) {
+							(false, Ok(k)) => k,
```
KiChjang, line 789: I prefer:

```suggestion
						match key.clone().try_into() {
							Ok(k) if value.2 != xcm_version => k,
```

But your call, not going to insist on this.

xcm/pallet-xcm/src/lib.rs
```diff
@@ -448,9 +638,268 @@ pub mod pallet {
...
+					}
+				}
+				stage = NotifyCurrentTargets;
+			}
+
+			let xcm_version = T::AdvertisedXcmVersion::get();
+			let response = Response::Version(xcm_version);
+
+			if stage == NotifyCurrentTargets {
+				for (key, mut value) in VersionNotifyTargets::<T>::iter_prefix(XCM_VERSION) {
```
KiChjang, line 786: Consider destructuring `value` here and give it more descriptive names:

```suggestion
				for (key, (query_id, max_weight, mut xcm_version)) in VersionNotifyTargets::<T>::iter_prefix(XCM_VERSION) {
```

xcm/pallet-xcm/src/lib.rs
```diff
@@ -448,9 +638,268 @@ pub mod pallet {
...
+					let new_key: MultiLocation =
+						match (value.2 == xcm_version, key.clone().try_into()) {
+							(false, Ok(k)) => k,
+							_ => {
+								// We don't early return here since we need to be certain that we make some
+								// progress.
```
KiChjang, line 792: I wonder if the `NotifyCurrentTargets` variant can just store an additional field that tells you which key that it last processed, so that you don't have to start iterating from the beginning of the list every time?

You'd then have `NotifyCurrentTargets(Option<MultiLocation>)`, and you'd extract the `MultiLocation` and start iterating from that key onwards using `iter_prefix_from`.

xcm/pallet-xcm/src/lib.rs
```diff
@@ -448,9 +638,268 @@ pub mod pallet {
...
+					weight_used.saturating_accrue(todo_vnt_notify_weight);
+					if weight_used >= weight_cutoff {
+						return (weight_used, Some(stage))
+					}
+				}
+				stage = MigrateAndNotifyOldTargets;
+			}
+			if stage == MigrateAndNotifyOldTargets {
+				for v in 0..XCM_VERSION {
+					for (old_key, mut value) in VersionNotifyTargets::<T>::drain_prefix(v) {
```
KiChjang, line 823: Same comment as the `NotifyCurrentTargets` stage -- could destructure `value` and give its fields names instead.

KiChjang: In addition, and this may be a dangerous and stupid idea, instead of using `drain_prefix`, we could just make use of `iter_prefix_from`, store the last key that was iterated upon to `MigrateAndNotifyOldTargets` variant, **and update the 2nd key for this target in the `VersionNotifyTargets` map**. We'll need to be extra careful to avoid iterator invalidation here, but in my mind, this algorithm looks to be a bit more efficient and thus consumes less weight.

xcm/pallet-xcm/src/lib.rs
```diff
@@ -635,87 +1147,133 @@ pub mod pallet {
...
+							query_id,
+							Some(responder),
+						));
+						return 0
+					}
+					return match maybe_notify {
+						Some((pallet_index, call_index)) => {
+							// This is a bit horrible, but we happen to know that the `Call` will
+							// be built by `(pallet_index: u8, call_index: u8, QueryId, Response)`.
+							// So we just encode that and then re-encode to a real Call.
```
KiChjang, line 1217: I must have missed this in my previous reviews, but how come we've managed to end up in this state of affairs?

KiChjang: Ah, is it because it's impossible to create a new `Call` struct otherwise?

xcm/src/v1/mod.rs
gavofyork: It's never used in v1 or v0.

xcm/pallet-xcm/src/lib.rs
gavofyork: Premature optimisation is the root of all evil.

xcm/pallet-xcm/src/lib.rs
gavofyork: yup

PR #3753, opened 2021-08-31
Fixes/improvements for disputes

eskimor:
In general things are looking good. I believe, I found one bug though - which might explain the memory leak we were seeing, at the very least it was very wrong: In case we could not find any undisputed block after the target block, we would use the best subchain block instead of the target block.

So disputes would not modify chain selection at all, although it is disputed and we should no longer build upon it ... I don't know the exact mechanisms in place here, but I could imagine that this leads to some code behaving weirdly.

To avoid such mistakes in the future, I adjusted the API to do the right thing, instead of leaving room for error to the caller.

### DIFF
--- a/node/core/dispute-coordinator/src/real/mod.rs
+++ b/node/core/dispute-coordinator/src/real/mod.rs
@@ -882,7 +882,7 @@ async fn issue_local_statement(
 
 	// Do import
 	if !statements.is_empty() {
-		let (pending_confirmation, _rx) = oneshot::channel();
+		let (pending_confirmation, rx) = oneshot::channel();
 		handle_import_statements(
 			ctx,
 			overlay_db,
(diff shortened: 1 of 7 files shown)


node/service/src/relay_chain_selection.rs
```diff
@@ -444,8 +444,7 @@ where
 			let (subchain_number, subchain_head) = rx
 				.await
 				.map_err(Error::OverseerDisconnected)
-				.map_err(|e| ConsensusError::Other(Box::new(e)))?
-				.unwrap_or_else(|| (subchain_number, subchain_head));
```
eskimor, line 448: Here we would previously use subchain_number/subchain_head in case we get `None` from the dispute coordinator.

rphmeier: Got it

node/core/dispute-coordinator/src/real/mod.rs
```diff
@@ -895,6 +895,32 @@ async fn issue_local_statement(
 			pending_confirmation,
 		)
 		.await?;
+		match rx.await {
+			Err(_) => {
+				tracing::error!(
```
rphmeier, line 900: I suspect these will get noisy.

Metrics?

rphmeier (approved):
API change looks much less error-prone.

node/core/dispute-coordinator/src/real/mod.rs
eskimor: actually no, those should not be noisy. The errors I would not expect to happen at all (hence errors) and the trace also only happens on `IssueLocalVote`, which should also not be noisy as this only happens on disputes and we will only vote once per dispute, so this should really be very low volume.

node/core/dispute-coordinator/src/real/mod.rs
```diff
@@ -582,12 +582,12 @@ async fn handle_incoming(
 			.await?;
 		},
 		DisputeCoordinatorMessage::DetermineUndisputedChain {
-			base_number,
+			base: (base_number, base_hash),
```
drahnr, line 585: nit: split this into `base_number` and `base_hash` instead on the messages already, `base` is never used as tuple.

eskimor: They really belong together, so I liked the grouping to highlight that relationship.

drahnr (approved):
Iff the mentioned logic change is sound, LGTM!

PR #3762, opened 2021-09-01
chore: test helper arbitrary ordering for 2

drahnr:
Adds a test helper to allow incoming messages that might come in non-deterministic order. 

Current impl is limited to two patterns.

Anticipated usage

```rust
let (header, zer0bytes) = arbitrary_order!(
     overseer_recv(&mut virtual_overseer).await;
     AllMessages::ChainApi(ChainApiMessage::BlockHeader(
				relay_parent,
				tx,
			)) => {
				assert_eq!(relay_parent, new_leaf);
				tx.send(Ok(Some(header))).unwrap();
				header
			};
AllMessages::ChainApi(ChainApiMessage::Ancestors{..
			}) => {
                               ///...
                               vec![0u32;16]
			}
);
```

### DIFF
--- a/node/subsystem-test-helpers/src/lib.rs
+++ b/node/subsystem-test-helpers/src/lib.rs
@@ -240,7 +240,7 @@ pub struct TestSubsystemContextHandle<M> {
 }
 
 impl<M> TestSubsystemContextHandle<M> {
-	/// Send a message or signal to the subsystem. This resolves at the point in time where the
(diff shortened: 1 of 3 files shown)


eskimor:
Nice! Thanks Bernhard!

eskimor (approved):
Nit: Shouldn't this be in subsystem-test-helpers?

drahnr:
You are absolutely correct.

PR #3766, opened 2021-09-01
XCM v1 version notification stub

gavofyork:
This is a minimally functional `SubscribeVersion`/`UnsubscribeVersion` implementation for XCM v1, which allows XCM v2 networks to query v1 networks' version and be notified when they upgrade to v2.

This doesn't contain any notification-on-upgrade code since it's unneeded: when they upgrade to v2, then #3736 will be in place with its notification code. The only thing this PR does is ensure that the storage item for all notification targets is maintained ready for #3736 to use when it gets upgraded to.

### XCM Config

Ensure that any `xcm_executor::Config` implementations includes a definition for the new type `SubscriptionService`. Generally this should be set to `PalletXcm`, the `pallet_xcm` concrete type. i.e.:

```rust
impl xcm_executor::Config for XcmConfig {
	/* snip */
	type SubscriptionService = XcmPallet;
}
```

### Barrier

An additional barrier `AlllowSubscriptionsFrom` should be included in your `Barriers` tuple. You can probably set the origin filter to `Everything`, unless you want to filter which locations can ask to be notified of your chain's XCM version changes:

```rust
pub type Barrier = (
	TakeWeightCredit,
	/* snip */
	AllowSubscriptionsFrom<Everything>,
);
```


tomusdrw:
@TriplEight please remove `bridges` code from any checks, this is just a `git subtree` of the main respository, and it should not be spellchecked, cargo --fmt checked or anything else.

PR #3780, opened 2021-09-03
comment out bridges runtime modules (broken) and update Rococo chain-spec

rphmeier:
1. Remove all bridge-related pallets from Rococo (there was an _unconditional panic_ in there somewhere)
2. Create new rococo.json


rphmeier:
I've verified that a `rococo-local` testnet works. Will roll out to Rococo as soon as green

rphmeier:
cc @tomusdrw @svyatonik Please prepare a PR to re-enable all bridges code, but with the appropriate migrations so the runtime doesn't panic.

Lldenaurois (approved):
I scanned the top and bottom of each addition removal and am quite confident that everything that was commented out faithfully.

Tests green indicate that the feature was removed from the runtime successfully.

LGTM

PR #3781, opened 2021-09-04
allow config NextFreeParaId in genesis

### LINKED ISSUE #3768: Genesis registered parachains does not increase NextFreeParaId and make impossible to reserve new para id
https://github.com/paritytech/polkadot/blob/ed5613555062a445902acb9e05216d24e3f2f257/runtime/common/src/paras_registrar.rs#L176

This is not updated when parachains are registered at genesis. This means `register.reserve` may try to reserve an existing parachain ID and failed. And make it impossible to reserve new para ID without sudo.

The easy solution is make `NextFreeParaId` configurable at genesis so we can manually set it to 2001

xlc:
Closes #3768
Only added to Rococo & Westend. Let me know if I should also add this to runtimes.

### DIFF
--- a/node/service/src/chain_spec.rs
+++ b/node/service/src/chain_spec.rs
@@ -1027,6 +1027,9 @@ fn rococo_staging_testnet_config_genesis(wasm_binary: &[u8]) -> rococo_runtime::
 		configuration: rococo_runtime::ConfigurationConfig {
 			config: default_parachains_host_configuration(),
 		},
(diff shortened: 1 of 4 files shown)


GopherJ:
@xlc I think it makes sense to add it to westend as well because parachain teams also test with westend-local

shawntabrizi (approved):

rphmeier (approved):

PR #3783, opened 2021-09-04
Add debug logs and make approval voting failable

Lldenaurois:


node/core/dispute-coordinator/src/real/mod.rs
```diff
@@ -542,16 +571,25 @@ async fn handle_incoming(
 		},
 		DisputeCoordinatorMessage::RecentDisputes(rx) => {
 			let recent_disputes = overlay_db.load_recent_disputes()?.unwrap_or_default();
+			tracing::debug!(target: DEBUG_LOG_TARGET, "Recent Disputes {:?}.", recent_disputes,);
 			let _ = rx.send(recent_disputes.keys().cloned().collect());
 		},
 		DisputeCoordinatorMessage::ActiveDisputes(rx) => {
+			// TODO (ladi): Shouldn''t this load active disputes?
```
rphmeier, line 578: I think there's only 'recent disputes' on disk, and then this filters them to be only active

node/core/dispute-participation/src/lib.rs
```diff
@@ -103,7 +104,7 @@ impl Error {
 				tracing::debug!(target: LOG_TARGET, err = ?self)
 			},
 			// it's worth reporting otherwise
-			_ => tracing::warn!(target: LOG_TARGET, err = ?self),
+			_ => tracing::debug!(target: LOG_TARGET, err = ?self),
```
rphmeier, line 107: Why demote this?

node/core/dispute-participation/src/lib.rs
```diff
@@ -119,7 +120,7 @@ where
 		match ctx.recv().await {
 			Err(_) => return,
 			Ok(FromOverseer::Signal(OverseerSignal::Conclude)) => {
-				tracing::info!(target: LOG_TARGET, "Received `Conclude` signal, exiting");
+				tracing::debug!(target: LOG_TARGET, "Received `Conclude` signal, exiting");
```
rphmeier, line 123: And this?

rphmeier: And many others in the same file

node/core/approval-voting/src/lib.rs
```diff
@@ -97,6 +99,7 @@ const APPROVAL_CHECKING_TIMEOUT: Duration = Duration::from_secs(120);
 const APPROVAL_CACHE_SIZE: usize = 1024;
 const TICK_TOO_FAR_IN_FUTURE: Tick = 20; // 10 seconds.
 const LOG_TARGET: &str = "parachain::approval-voting";
+const MALICIOUS_FREQUENCY: usize = 5;
```
rphmeier, line 102: 5 is _really_ low!

node/core/dispute-participation/src/lib.rs
Lldenaurois: It's easier (for now) to just display everything in Debug logs imo. I can change it back if you see it differently.

This PR is not intended to get merged at any point

node/core/approval-voting/src/lib.rs
```diff
@@ -727,20 +744,37 @@ where
...
+									},
+								)
 							)
-						)
-						.collect();
-					actions.append(&mut approvals);
+							.collect();
+						actions.append(&mut approvals);
+					}
+					ApprovalOutcome::Malicious((candidate, session_index)) => {
```
rphmeier, line 763: This seems like a lot of extra machinery to add instead of just sending the message from within the `if coin_flip == 0` block

node/core/approval-voting/src/lib.rs
```diff
@@ -100,7 +101,8 @@ const APPROVAL_CACHE_SIZE: usize = 1024;
 const TICK_TOO_FAR_IN_FUTURE: Tick = 20; // 10 seconds.
 const LOG_TARGET: &str = "parachain::approval-voting";
 const DEBUG_LOG_TARGET: &str = "parachain::ladi-debug-approval-voting";
-const MALICIOUS_FREQUENCY: usize = 5_000;
+const MALICIOUS_MEAN: f64 = 2_500f64;
+const MALICIOUS_SDEV: f64 = 70f64;
```
drahnr, line 105: nit: `.._STDDEV` is the common abbreviation.

node/core/approval-voting/src/lib.rs
```diff
@@ -346,12 +348,15 @@ where
 {
 	fn start(self, ctx: Context) -> SpawnedSubsystem {
 		let backend = DbBackend::new(self.db.clone(), self.db_config);
+		let normal = Normal::new(MALICIOUS_MEAN, MALICIOUS_SDEV);
+		let offset: u64 = normal.sample(&mut rand::thread_rng()).round() as u64;
```
drahnr, line 352: I'd recommend adding `3 * STDDEV` which means 99.7% are not edited, but you avoid hickups due to huge outliers.

node/core/approval-voting/src/lib.rs
```diff
@@ -791,6 +797,7 @@ where
 			&mut approvals_cache,
 			&mut subsystem.mode,
 			actions,
+			if counter.fetch_add(1, Ordering::SeqCst) == offset { true } else { false },
```
drahnr, line 800: the if case can be dropped :) just use the boolean expression

drahnr (commented):
LGTM (for test)

node/core/approval-voting/src/lib.rs
Lldenaurois: Yeah, I wanted MEAN and SDEV to have the same length :)

node/core/approval-voting/src/lib.rs
Lldenaurois: I added logging in these two cases, so for now I'll leave it as an if-statement. I'll keep the logging for deploying to one node but can drop the logging before deploying to all validators.

eskimor (commented):
Looks good overall. I would move the invalid statement generation into the `self.seconded.is_none() && self.issued_statements.contains(...)` check, to avoid another misbehavior of sending out multiple seconded statements. This could interfere with the actual behavior we want to test right now.

eskimor:
I just realized, that I reviewed an ancient version - having another look.

node/core/dispute-participation/src/lib.rs
eskimor: You can match logs with e.g. `level =~"DEBUG|INFO|ERROR|WARN"` ... just for reference.

PR #3784, opened 2021-09-04
Don't drop UMP queue items if weight exhausted

### LINKED ISSUE #3782: UMP sink bug: XCM message would be discarded if ump_service_total_weight reached
The message should be put back to queue if execution failed because of weight exhausted (which is caused by `ump_service_total_weight` limitation in host config). Currently it's discarded, which would lead to fund loss.

https://github.com/paritytech/polkadot/blob/285a35bf16adca556924907a4bbd787a759dc3ee/runtime/parachains/src/ump.rs#L419-L424

Fortunately this hasn't happened *yet*, as `ump_service_total_weight` is set to `100_000_000_000`, which makes it be able to handle 33 orml-xtokens transfer in one block. `max_upward_message_num_per_candidate` is set to 10, and since we have only Karura and Statemine sending ump transfers/teleports for now, this limitation hasn't be reached yet.

But it needs to be fixed ASAP.

gavofyork:
Closes #3782

Takes items from the queue in a two-part "peek/remove" mechanism, which means no need faff around popping/pushing/passing `UpwardMessage`s back and forth.

Since there's actually only one place in on-chain code that `pop_front` was used and since that isn't necessary if we just stick to using cursors, then we can replace the `VecDeque` with a straight `Vec` which should give better performance characteristics. This code also avoids the O(N) `VecDeque::split` call.

In the case that the first item cannot be consumed, this will also avoid a storage write.

Revamps the tests so that this condition can actually be tested for.

### DIFF
--- a/runtime/parachains/src/ump.rs
+++ b/runtime/parachains/src/ump.rs
@@ -14,7 +14,6 @@
 // You should have received a copy of the GNU General Public License
 // along with Polkadot.  If not, see <http://www.gnu.org/licenses/>.
 
-
 use crate::{
 	configuration::{self, HostConfiguration},
 	initializer,
@@ -22,11 +21,7 @@ use crate::{
 use frame_support::pallet_prelude::*;
 use primitives::v1::{Id as ParaId, UpwardMessage};
 use sp_std::{
-	collections::btree_map::BTreeMap,
-	convert::TryFrom,
-	fmt,
-	marker::PhantomData,
-	prelude::*,
+	collections::btree_map::BTreeMap, convert::TryFrom, fmt, marker::PhantomData, prelude::*,
 };
 use xcm::latest::Outcome;
 
@@ -482,7 +477,7 @@ impl QueueCache {
 	}
 
 	/// Returns the message at the front of `para`'s queue, or `None` if the queue is empty.
-	/// 
+	///
 	/// Does not mutate the queue.
 	fn peek_front<T: Config>(&mut self, para: ParaId) -> Option<&UpwardMessage> {
 		let entry = self.ensure_cached::<T>(para);
--- a/runtime/parachains/src/ump.rs
+++ b/runtime/parachains/src/ump.rs
@@ -417,12 +417,14 @@ impl<T: Config> Pallet<T> {
 					},
 					Err((id, required)) => {
 						// we process messages in order and don't drop them if we run out of weight,
-						// so need to break here and requeue the message we took off the queue.
+						// so need to break here without calling `consume_front`.
 						Self::deposit_event(Event::WeightExhausted(id, max_weight, required));
 						break
 					},
 				}
 			} else {
+				// this should never happen, since the cursor should never point to an empty queue.
+				// it is resolved harmlessly here anyway.
 				true
 			};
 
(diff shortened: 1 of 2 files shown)


runtime/parachains/src/ump.rs
```diff
@@ -864,52 +811,60 @@ mod tests {
...
+			assert_eq!(take_processed(), vec![]);
+			assert_storage_consistency_exhaustive();
+		});
+	}
 
-			// 3rd iteration.
-			{
-				let mut probe = Probe::new();
+	#[test]
+	fn dispatch_keeps_message_after_weight_exhausted() {
```
gavofyork, line 839: this is the test that used to fail.

runtime/parachains/src/ump.rs
```diff
@@ -408,22 +404,29 @@ impl<T: Config> Pallet<T> {
 			};
 
 			// dequeue the next message from the queue of the dispatchee
-			let (upward_message, became_empty) = queue_cache.dequeue::<T>(dispatchee);
-			if let Some(upward_message) = upward_message {
+			let maybe_next = queue_cache.peek_front::<T>(dispatchee);
+			let became_empty = if let Some(upward_message) = maybe_next {
 				match T::UmpSink::process_upward_message(
 					dispatchee,
 					&upward_message[..],
```
gilescope, line 411: Now `upward_message` is already a reference we can simplify to:
```suggestion
					upward_message,
```

pepyakin (commented):
I believe the proper solution for this problem is placing these messages into a separate overweight queue, which is implemented by https://github.com/paritytech/polkadot/pull/3575. 

In fact, this PR introduces another problem. Suppose a message that exceeds `ump_service_total_weight` somehow ends up in a para's queue. Each time we `process_pending_upward_messages` we would peek into the queue we left off the last time, discover that the message under the cursor is overweight and bail. Therefore, no progress can be made until we yank the message.

That said, I am in favor of those changes, so I propose we rebase https://github.com/paritytech/polkadot/pull/3575 on top.

gavofyork:
> In fact, this PR introduces another problem. Suppose a message that exceeds `ump_service_total_weight` somehow ends up in a para's queue. Each time we `process_pending_upward_messages` we would peek into the queue we left off the last time, discover that the message under the cursor is overweight and bail. Therefore, no progress can be made until we yank the message.

This PR does not "introduce" any such problem. The problem existed already, and this is strictly a bugfix PR that optimises a number of codepaths.

Certainly there should be an overweight queue.

shawntabrizi (approved):
The changes make sense to me

runtime/parachains/src/ump.rs
```diff
@@ -408,22 +404,25 @@ impl<T: Config> Pallet<T> {
 			};
 
 			// dequeue the next message from the queue of the dispatchee
```
shawntabrizi, line 406: use of "dequeue" here should probably be updated

runtime/parachains/src/ump.rs
```diff
@@ -461,51 +460,63 @@ impl<T: Config> Pallet<T> {
...
-				<Pallet<T> as Store>::RelayDispatchQueueSize::remove(&para);
-			} else {
-				<Pallet<T> as Store>::RelayDispatchQueues::insert(&para, queue);
-				<Pallet<T> as Store>::RelayDispatchQueueSize::insert(&para, (count, total_size));
+				RelayDispatchQueues::<T>::remove(&para);
+				RelayDispatchQueueSize::<T>::remove(&para);
+			} else if entry.consumed_count > 0 {
+				RelayDispatchQueues::<T>::insert(&para, &entry.queue[entry.consumed_count..]);
+				let count = (entry.queue.len() - entry.consumed_count) as u32;
+				let size = (entry.total_size as usize - entry.consumed_size) as u32;
```
shawntabrizi, line 518: saturating math on specifically `total_size` may be safer, since there is nothing in the immediately surrounding code which ensures that this will not underflow

shawntabrizi (approved):

shawntabrizi:
bot merge

PR #3788, opened 2021-09-05
set `DisputesHandler` in initializer on Rococo

rphmeier:
It was previously set to `()` and the `DisputesHandler` is the primary way for the initializer & inherent handler to communicate info to disputes. Without this, the runtime will drop all disputes.


ordian (approved):

rphmeier:
bot merge

runtime/rococo/src/lib.rs
```diff
@@ -585,7 +585,7 @@ impl runtime_parachains::inclusion::RewardValidators for RewardValidators {
 
 impl parachains_inclusion::Config for Runtime {
 	type Event = Event;
-	type DisputesHandler = ();
+	type DisputesHandler = ParasDisputes;
```
eskimor, line 588: Shouldn't this be feature gated as well?

PR #3802, opened 2021-09-07
Introduce XCM Weight Traits

shawntabrizi:
This PR introduces the traits needed to implement weights for all the various XCM instructions.

Basically, each XCM instruction gets it's own function, with all of its parameters, and must output a weight. How exactly that happens is up to the implementor, but this is certainly what is required at the end of the day.

Next up is some of the benchmarks...

### DIFF
--- a/xcm/src/lib.rs
+++ b/xcm/src/lib.rs
@@ -422,7 +422,7 @@ impl WrapVersion for AlwaysV1 {
 	}
 }
 
-/// `WrapVersion` implementation which attempts to always convert the XCM to version 1 before wrapping it.
+/// `WrapVersion` implementation which attempts to always convert the XCM to version 2 before wrapping it.
 pub struct AlwaysV2;
 impl WrapVersion for AlwaysV2 {
 	fn wrap_version<Call>(
@@ -473,4 +473,9 @@ pub mod opaque {
 
 	/// The basic `VersionedXcm` type which just uses the `Vec<u8>` as an encoded call.
 	pub type VersionedXcm = super::VersionedXcm<()>;
+}
+
+// A simple trait to get the weight of some object.
+pub trait GetWeight<W> {
+	fn weight(&self) -> latest::Weight;
 }
--- a/xcm/src/v2/traits.rs
+++ b/xcm/src/v2/traits.rs
@@ -19,7 +19,7 @@
 use core::result;
 use parity_scale_codec::{Decode, Encode};
 
-use super::{MultiLocation, Xcm};
+use super::*;
 
 #[derive(Copy, Clone, Encode, Decode, Eq, PartialEq, Debug)]
 pub enum Error {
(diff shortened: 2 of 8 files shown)


xcm/xcm-builder/src/weight.rs
```diff
@@ -35,29 +35,81 @@ impl<T: Get<Weight>, C: Decode + GetDispatchInfo, M: Get<u32>> WeightBounds<C>
...
+		instruction: &Instruction<C>,
 		instrs_limit: &mut u32,
 	) -> Result<Weight, ()> {
-		Ok(T::get().saturating_add(match message {
-			Transact { require_weight_at_most, .. } => *require_weight_at_most,
-			SetErrorHandler(xcm) | SetAppendix(xcm) => Self::weight_with_limit(xcm, instrs_limit)?,
-			_ => 0,
-		}))
+		T::get()
+			.checked_add(match instruction {
```
shawntabrizi, line 57: note checked math change here from saturating math

xcm/src/lib.rs
```diff
@@ -474,3 +474,8 @@ pub mod opaque {
 	/// The basic `VersionedXcm` type which just uses the `Vec<u8>` as an encoded call.
 	pub type VersionedXcm = super::VersionedXcm<()>;
 }
+
+// A simple trait to get the weight of some object.
+pub trait GetWeight<W> {
+	fn weight(&self) -> latest::Weight;
+}
```
shawntabrizi, line 481: Should I just use `Get<Weight>` here?

xcm/src/v2/traits.rs
```diff
@@ -313,3 +313,63 @@ impl SendXcm for Tuple {
 		Err(SendError::CannotReachDestination(destination, message))
 	}
 }
+
+/// The info needed to weight an XCM.
+// TODO: Automate Generation
```
kianenigma, line 318: you mean a macro that takes `Enum XCM` and generates this?

xcm/xcm-builder/src/currency_adapter.rs
```diff
@@ -142,6 +142,7 @@ impl<
 	}
 
 	fn deposit_asset(what: &MultiAsset, who: &MultiLocation) -> Result {
+		log::trace!("xcm::currency_adapter deposit_asset {:?} {:?}", what, who);
```
kianenigma, line 145: still needed?

xcm/xcm-builder/src/weight.rs
```diff
@@ -35,29 +35,81 @@ impl<T: Get<Weight>, C: Decode + GetDispatchInfo, M: Get<u32>> WeightBounds<C>
...
+
+struct WeightInfoBounds<W, C, M>(PhantomData<(W, C, M)>);
+impl<W, C, M> WeightBounds<C> for WeightInfoBounds<W, C, M>
+where
+	W: XcmWeightInfo<C>,
+	C: Decode + GetDispatchInfo,
+	M: Get<u32>,
+	Instruction<C>: xcm::GetWeight<W>,
+{
+	fn weight(message: &mut Xcm<C>) -> Result<Weight, ()> {
```
kianenigma, line 75: I've been a bit away from XCM, can you briefly explain why the deep/shallow system is gone? It also seems like the order/instruction system is also retired now.

xcm/xcm-builder/src/weight.rs
```diff
@@ -35,29 +35,81 @@ impl<T: Get<Weight>, C: Decode + GetDispatchInfo, M: Get<u32>> WeightBounds<C>
...
+	M: Get<u32>,
+	Instruction<C>: xcm::GetWeight<W>,
+{
+	fn weight_with_limit(message: &Xcm<C>, instrs_limit: &mut u32) -> Result<Weight, ()> {
+		let mut r: Weight = 0;
+		*instrs_limit = instrs_limit.checked_sub(message.0.len() as u32).ok_or(())?;
+		for m in message.0.iter() {
+			r = r.checked_add(Self::instr_weight_with_limit(m, instrs_limit)?).ok_or(())?;
+		}
+		Ok(r)
```
kianenigma, line 97:
```suggestion
		*instrs_limit = instrs_limit.checked_sub(message.0.len() as u32).ok_or(())?;
		message.0.iter().try_fold(0, |acc, x| 
			Self::instr_weight_with_limit(m, instrs_limit).and_then(|inst_weight| {
				acc.checked_add(x)
			})
		).ok_or(())
```

kianenigma (commented):
Looks correct to me so far. It is not done yet though right?

xcm/src/lib.rs
gavofyork: Depends if you need `&self` or not.

xcm/xcm-builder/src/weight.rs
gavofyork: https://medium.com/polkadot-network/xcm-the-cross-consensus-message-format-3b77b1373392

xcm/xcm-builder/src/currency_adapter.rs
shawntabrizi: It was added before, i dont see a problem with log messages

xcm/src/v2/traits.rs
shawntabrizi: yes. it is all boiler plate, and maintaining it is annoying-ish.

xcm/xcm-builder/src/weight.rs
shawntabrizi: I dont find this easier to read

shawntabrizi:
> Looks correct to me so far. It is not done yet though right?

This step is "done". Benchmarking code will be a follow up PR on top of this, but these are all the building blocks into the core XCM system needed for benchmarking.

gavofyork (approved):

shawntabrizi:
bot merge

PR #3806, opened 2021-09-07
Minor update to README.md

csknk:
Add note re: custom `ws-port` for a second node in a 2 node dev setup.


bkchr:
I just realized that you said that the second node is panicking. This is a bug and should not happen. It should just work, even if it can not bind to the port.

bkchr:
I will make sure we fix the underlying bug.

PR #3815, opened 2021-09-08
add parachains pallets to Polkadot runtime

rphmeier:
There are still some TODOs w.r.t. crowdloan deposits, but that can be changed later.


runtime/polkadot/src/lib.rs
```diff
@@ -1130,48 +1277,63 @@ pub type Executive = frame_executive::Executive<
 	frame_system::ChainContext<Runtime>,
 	Runtime,
 	AllPallets,
-	TechnicalMembershipStoragePrefixMigration,
```
rphmeier, line 1133: Did I remove this prematurely or was it good to remove?

runtime/polkadot/src/lib.rs
```diff
@@ -1036,6 +1045,125 @@ impl pallet_proxy::Config for Runtime {
...
+
+impl parachains_scheduler::Config for Runtime {}
+
+impl parachains_initializer::Config for Runtime {
+	type Randomness = pallet_babe::RandomnessFromOneEpochAgo<Runtime>;
+	type ForceOrigin = EnsureRoot<AccountId>;
+}
+
+parameter_types! {
+	pub const ParaDeposit: Balance = 40 * UNITS;
```
pepyakin, line 1095: Doesn't this need to be updated? cc @shawntabrizi

runtime/polkadot/src/lib.rs
```diff
@@ -1036,6 +1045,125 @@ impl pallet_proxy::Config for Runtime {
...
+	type Currency = Balances;
+	type OnSwap = (Crowdloan, Slots);
+	type ParaDeposit = ParaDeposit;
+	type DataDepositPerByte = DataDepositPerByte;
+	type WeightInfo = weights::runtime_common_paras_registrar::WeightInfo<Runtime>;
+}
+
+parameter_types! {
+	// 6 weeks
+	pub const LeasePeriod: BlockNumber = 6 * WEEKS;
```
pepyakin, line 1110: cc @shawntabrizi

runtime/polkadot/src/lib.rs
```diff
@@ -1130,48 +1277,63 @@ pub type Executive = frame_executive::Executive<
...
+			hrmp_max_parathread_outbound_channels: 0,
+			hrmp_open_request_ttl: 72,
+			hrmp_sender_deposit: deposit(1004, 100 * 1024),
+			hrmp_recipient_deposit: deposit(1004, 100 * 1024),
+			hrmp_channel_max_capacity: 1_000,
+			hrmp_channel_max_total_size: 102_400,
+			hrmp_max_parachain_inbound_channels: 10,
+			hrmp_max_parathread_inbound_channels: 0,
+			hrmp_channel_max_message_size: 102_400,
+			code_retention_period: EPOCH_DURATION_IN_SLOTS * 6,
```
pepyakin, line 1315: What's the rationale for this number?

runtime/polkadot/src/lib.rs
```diff
@@ -1130,48 +1277,63 @@ pub type Executive = frame_executive::Executive<
...
-	TechnicalMembershipStoragePrefixMigration,
+	SetInitialHostConfiguration,
 >;
 /// The payload being signed in transactions.
 pub type SignedPayload = generic::SignedPayload<Call, SignedExtra>;
 
-const TECHNICAL_MEMBERSHIP_OLD_PREFIX: &str = "Instance1Membership";
 /// Migrate from `Instance1Membership` to the new pallet prefix `TechnicalMembership`
-pub struct TechnicalMembershipStoragePrefixMigration;
+pub struct SetInitialHostConfiguration;
```
pepyakin, line 1286: Rustdoc should be updated

runtime/polkadot/src/lib.rs
shawntabrizi: On the current Polkadot network (9090), this migration has NOT happened yet.

Will these changes go into v0.9.10? If so, then yeah, we should have both in there.

shawntabrizi: I added this back for now

runtime/polkadot/src/lib.rs
```diff
@@ -1231,6 +1395,60 @@ impl OnRuntimeUpgrade for TechnicalMembershipStoragePrefixMigration {
...
+			validation_upgrade_frequency: 14_400,
+			validation_upgrade_delay: 600,
+			max_pov_size: 5_242_880,
+			max_downward_message_size: 51_200,
+			ump_service_total_weight: 100_000_000_000,
+			hrmp_max_parachain_outbound_channels: 10,
+			hrmp_max_parathread_outbound_channels: 0,
+			_hrmp_open_request_ttl: 72,
+			hrmp_sender_deposit: deposit(1004, 100 * 1024),
+			hrmp_recipient_deposit: deposit(1004, 100 * 1024),
```
shawntabrizi, line 1421: This will be a massive deposit. 20,000 DOT at least (1004 * 20 * DOLLARS), then we also have to add cost of bytes.

How was this determined?

rphmeier: Yeah, that's a TODO - it's the same as the Kusama values which are themselves too large.

runtime/polkadot/src/lib.rs
rphmeier: AFAIK Code needs to be available for the entire dispute period, which is 6 epochs.

shawntabrizi (approved):

PR #3818, opened 2021-09-09
XCM Benchmarks for Asset Transactor w/ Fungible Asset

shawntabrizi:
This PR includes benchmarks for XCMs which use the `AssetTransactor` trait, implemented for an asset which implements the `Fungible` trait.


shawntabrizi:
/benchmark runtime custom --chain=westend-dev --pallet=pallet_xcm_benchmarks::fungible --extrinsic="*" --steps=10 --repeat=10 --template=./xcm/pallet-xcm-benchmarks/template.hbs --output=./

shawntabrizi:
--chain=westend-dev --pallet=pallet_xcm_benchmarks::fungible --extrinsic="*" --steps=10 --repeat=10 --template=./xcm/pallet-xcm-benchmarks/template.hbs --output=./ --execution=wasm --wasm-execution=compiled

shawntabrizi:
/benchmark runtime custom --chain=westend-dev --pallet=pallet_xcm_benchmarks::fungible --extrinsic="*" --steps=10 --repeat=10 --template=./xcm/pallet-xcm-benchmarks/template.hbs --output=./ --execution=wasm --wasm-execution=compiled

shawntabrizi:
/benchmark runtime custom --chain=westend-dev --pallet=pallet_xcm_benchmarks::fungible --extrinsic="*" --steps=50 --repeat=20 --template=./xcm/pallet-xcm-benchmarks/template.hbs --output=./ --execution=wasm --wasm-execution=compiled

runtime/westend/src/lib.rs
```diff
@@ -1529,6 +1566,12 @@ sp_api::impl_runtime_apis! {
 				hex_literal::hex!("26aa394eea5630e07c48ae0c9558cef780d41e5e16056765bc8461851072c9d7").to_vec().into(),
 				// Treasury Account
 				hex_literal::hex!("26aa394eea5630e07c48ae0c9558cef7b99d880ec681799c0cf30e8886371da95ecffd7b6c0f78751baa9d281e0bfa3a6d6f646c70792f74727372790000000000000000000000000000000000000000").to_vec().into(),
+				// Dmp DownwardMessageQueueHeads
```
kianenigma, line 1569: justification for whitelisting these?

runtime/westend/src/weights/xcm/mod.rs
```diff
@@ -0,0 +1,191 @@
...
+use sp_std::prelude::*;
+use xcm::{latest::prelude::*, DoubleEncoded};
+
+use pallet_xcm_benchmarks_fungible::WeightInfo as XcmBalancesWeight;
+use pallet_xcm_benchmarks_generic::WeightInfo as XcmGeneric;
+
+pub enum AssetTypes {
+	Balances,
+	Unknown,
+}
```
kianenigma, line 15:
```suggestion
/// Types of asset supported by the westend runtime. 
pub enum AssetTypes {
	/// An asset backed by `pallet-balances`.
	Balances,
	/// Unknown asset.
	Unknown,
}
```

runtime/westend/src/weights/xcm/mod.rs
```diff
@@ -0,0 +1,191 @@
...
+
+impl From<&MultiAsset> for AssetTypes {
+	fn from(asset: &MultiAsset) -> Self {
+		match asset {
+			MultiAsset { id: Concrete(MultiLocation { parents: 0, interior: Here }), .. } =>
+				AssetTypes::Balances,
+			_ => AssetTypes::Unknown,
```
kianenigma, line 22: shouldn't be `unreachable!`?

runtime/westend/src/weights/xcm/mod.rs
```diff
@@ -0,0 +1,191 @@
...
+}
+
+// TODO wild case
+impl WeighMultiAssets for MultiAssetFilter {
+	fn weigh_multi_assets(&self, balances_weight: Weight) -> Weight {
+		match self {
+			Self::Definite(assets) => assets
+				.inner()
+				.into_iter()
+				.map(|m| <AssetTypes as From<&MultiAsset>>::from(m))
```
kianenigma, line 38:
```suggestion
				.map(From::from)
```

runtime/westend/src/weights/xcm/mod.rs
```diff
@@ -0,0 +1,191 @@
...
+
+pub struct WestendXcmWeight<Call>(core::marker::PhantomData<Call>);
+impl<Call> XcmWeightInfo<Call> for WestendXcmWeight<Call> {
+	fn withdraw_asset(assets: &MultiAssets) -> Weight {
+		assets.weigh_multi_assets(XcmBalancesWeight::<Runtime>::withdraw_asset())
+	}
+	fn reserve_asset_deposited(assets: &MultiAssets) -> Weight {
+		assets.weigh_multi_assets(XcmGeneric::<Runtime>::reserve_asset_deposited())
+	}
+	// TODO none of these need effects
```
kianenigma, line 70:
```suggestion
```

seems stale

runtime/westend/src/weights/xcm/mod.rs
```diff
@@ -0,0 +1,191 @@
...
+		_xcm: &Xcm<()>,
+	) -> Weight {
+		assets.weigh_multi_assets(XcmBalancesWeight::<Runtime>::transfer_reserve_asset())
+	}
+	fn transact(
+		_origin_type: &OriginKind,
+		_require_weight_at_most: &u64,
+		_call: &DoubleEncoded<Call>,
+	) -> Weight {
+		XcmGeneric::<Runtime>::transact()
```
kianenigma, line 92: Transact has no arguments? values seem also mock, what's the scenario of the generic xcm calls?

xcm/pallet-xcm-benchmarks/Cargo.toml
```diff
@@ -0,0 +1,45 @@
...
+pallet-assets = { version = "4.0.0-dev", branch = "master", git = "https://github.com/paritytech/substrate" }
+sp-core = { version = "4.0.0-dev", branch = "master", git = "https://github.com/paritytech/substrate" }
+sp-io = { version = "4.0.0-dev", branch = "master", git = "https://github.com/paritytech/substrate" }
+sp-tracing = { version = "4.0.0-dev", branch = "master", git = "https://github.com/paritytech/substrate" }
+xcm-builder = { path = "../xcm-builder" }
+xcm = { path = ".." }
+# temp
+pallet-xcm = { path = "../pallet-xcm" }
+polkadot-runtime-common = { path = "../../runtime/common" }
+# westend-runtime = { path = "../../runtime/westend", features = ["runtime-benchmarks"] }
```
kianenigma, line 33: TODO remove me later

xcm/pallet-xcm-benchmarks/src/lib.rs
```diff
@@ -0,0 +1,108 @@
...
+/// The overarching call type.
+pub type OverArchingCallOf<T> = <T as frame_system::Config>::Call;
+/// The asset transactor of our executor
+pub type AssetTransactorOf<T> = <<T as Config>::XcmConfig as xcm_executor::Config>::AssetTransactor;
+/// The call type of executor's config. Should eventually resolve to the same overarching call type.
+pub type XcmCallOf<T> = <<T as Config>::XcmConfig as xcm_executor::Config>::Call;
+
+/// The worst case number of assets in the holding.
+const HOLDING_FUNGIBLES: u32 = 99;
+const HOLDING_NON_FUNGIBLES: u32 = 99;
```
kianenigma, line 61: Should go into the config of the corresponding `xcm_benchmark` pallet, right?

xcm/pallet-xcm-benchmarks/src/lib.rs
```diff
@@ -0,0 +1,108 @@
...
+
+pub fn worst_case_holding() -> Assets {
+	let fungibles_amount: u128 = 100; // TODO probably update
```
kianenigma, line 64: your opinion was always sticking to ED and multiplies thereof, which should be possible here if you pass in a `T`.

runtime/westend/src/weights/xcm/mod.rs
```diff
@@ -0,0 +1,191 @@
...
+		_max_assets: &u32, // TODO use max assets?
+		_dest: &MultiLocation,
+	) -> Weight {
+		assets.weigh_multi_assets(XcmBalancesWeight::<Runtime>::deposit_asset())
+	}
+	fn deposit_reserve_asset(
+		assets: &MultiAssetFilter,
+		_max_assets: &u32, // TODO use max assets?
+		_dest: &MultiLocation,
+		_xcm: &Xcm<()>,
```
kianenigma, line 135: these `xcm` are being forwarded, and should be taken into account by multiplying them with something like `generic_xcm_weights::send_xcm()`. A few other similar instances exist.

kianenigma: UPDATE: no `Vec<_>` involved, at most one xcm forwarding, but my comment over all still holds.

xcm/pallet-xcm-benchmarks/src/fungible/benchmarking.rs
```diff
@@ -0,0 +1,221 @@
...
+		let asset = T::get_multi_asset();
+		<AssetTransactorOf<T>>::deposit_asset(&asset, &sender_location).unwrap();
+		let assets: MultiAssets = vec![ asset ].into();
+		assert!(T::TransactAsset::balance(&dest_account).is_zero());
+
+		let mut executor = new_executor::<T>(sender_location);
+		let instruction = Instruction::TransferReserveAsset {
+			assets,
+			dest: dest_location,
+			xcm: Xcm::new()
```
kianenigma, line 98: so the cost of this nested send is already incorporated here, this is much simpler than the old scheme of order and effect.

xcm/pallet-xcm-benchmarks/src/fungible/benchmarking.rs
```diff
@@ -0,0 +1,221 @@
...
+
+		// Add our asset to the holding.
+		holding.subsume(asset.clone());
+
+		// our dest must have no balance initially.
+		let dest_location = T::valid_destination()?;
+		let dest_account = T::AccountIdConverter::convert(dest_location.clone()).unwrap();
+		assert!(T::TransactAsset::balance(&dest_account).is_zero());
+
+		let mut executor = new_executor::<T>(Default::default());
```
kianenigma, line 144: nit: might want to make it `fn new_executor<T: Config>(_: MultiLocation, initial_holding: Assets)`

xcm/pallet-xcm-benchmarks/src/fungible/benchmarking.rs
```diff
@@ -0,0 +1,221 @@
...
+		let asset = T::get_multi_asset();
+		let mut holding = worst_case_holding();
+
+		// Add our asset to the holding.
+		holding.subsume(asset.clone());
+
+		if let Some(checked_account) = T::CheckedAccount::get() {
+			// Checked account starts at zero
+			assert!(T::TransactAsset::balance(&checked_account).is_zero());
+		}
```
kianenigma, line 197:
```suggestion
		// Checked account starts at zero
		assert!(T::CheckedAccount::get().map_or(true, |c| T::TransactAsset::balance(&c).is_zero());
```

xcm/pallet-xcm-benchmarks/src/fungible/mock.rs
```diff
@@ -0,0 +1,180 @@
+// Copyright 2021 Parity Technologies (UK) Ltd.
```
kianenigma, line 1: not super relevant yet since this PR is only about the `fungible`, but once a `fungibles` has been added as well, it would be sad to see this mock almost entirely duplicated. I think one mock runtime that implements both `xcm_benchamrk::fungible::Config` and `xcm_benchamrk::fungibles::Config` should work fine.

kianenigma (approved):
LGTM, most TODOs seem to be related to the next steps, I think we can leave them, try and merge this, to keep the PRs within bite-size pieces.

runtime/westend/src/lib.rs
shawntabrizi: We can assume XCM will be used once every block, and these storage items will be updated pretty much once every block. No need to double count them everywhere.

runtime/westend/src/weights/xcm/mod.rs
shawntabrizi: No, this is where someone sends us an XCM with an asset that we are not familiar with. We will make the XCM invalid, but we def do not want to panic.

runtime/westend/src/weights/xcm/mod.rs
shawntabrizi: Yeah this is a mock, and probably will change once i work on generic benchmark

runtime/westend/src/lib.rs
```diff
@@ -1561,6 +1605,10 @@ sp_api::impl_runtime_apis! {
 			add_benchmark!(params, batches, pallet_utility, Utility);
 			add_benchmark!(params, batches, pallet_vesting, Vesting);
 
+			// XCM Benchmarks
+			// NOTE: Make sure you point to the individual modules below.
```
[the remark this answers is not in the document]
shawntabrizi: No, note that the pallet name is a full path, which is important.

xcm/pallet-xcm-benchmarks/src/fungible/mod.rs
```diff
@@ -0,0 +1,47 @@
...
+		/// The type of `fungible` that is being used under the hood.
+		///
+		/// This is useful for testing and checking.
+		type TransactAsset: frame_support::traits::fungible::Mutate<Self::AccountId>;
+
+		/// The account used to check assets being teleported.
+		type CheckedAccount: Get<Option<Self::AccountId>>;
+
+		/// A trusted location which we allow teleports from, and the asset we allow to teleport.
+		type TrustedTeleporter: Get<Option<(xcm::latest::MultiLocation, xcm::latest::MultiAsset)>>;
```
apopiak, line 39: why introduce an extra asset here?, you already have `get_multi_asset` below?

xcm/pallet-xcm-benchmarks/src/fungible/mock.rs
```diff
@@ -0,0 +1,180 @@
...
+	type Trader = xcm_builder::FixedRateOfFungible<WeightPrice, ()>;
+	type ResponseHandler = DevNull;
+	type AssetTrap = ();
+	type AssetClaims = ();
+	type SubscriptionService = ();
+}
+
+impl crate::Config for Test {
+	type XcmConfig = XcmConfig;
+	type AccountIdConverter = AccountIdConverter;
```
apopiak, line 147: Why introduce an extra `AccountIdConverter` when there is the `OriginConverter` in the `XcmConfig`?

runtime/westend/src/weights/xcm/mod.rs
```diff
@@ -0,0 +1,193 @@
...
+			_ => AssetTypes::Unknown,
+		}
+	}
+}
+
+trait WeighMultiAssets {
+	fn weigh_multi_assets(&self, balances_weight: Weight) -> Weight;
+}
+
+// TODO wild case
```
apopiak, line 34: Todo.
I'm guessing you'll address this in a follow-up PR?

apopiak (commented):
:+1: for XCM weights/benchmarks! :tada: 
Seems to me this has a few rough edges. Unclear to me whether you want to smooth those out now or in a follow up PR.

runtime/westend/src/lib.rs
```diff
@@ -1516,6 +1520,39 @@ sp_api::impl_runtime_apis! {
 			impl pallet_offences_benchmarking::Config for Runtime {}
 			impl frame_system_benchmarking::Config for Runtime {}
 
+			use xcm::latest::MultiAsset;
+
+			impl pallet_xcm_benchmarks::Config for Runtime {
+				type XcmConfig = XcmConfig;
+				type AccountIdConverter = LocationConverter;
+				fn valid_destination() -> Result<MultiLocation, sp_runtime::DispatchError> {
```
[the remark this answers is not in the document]
shawntabrizi: There may need to be some setup required to make a destination valid, for example calling some dispatchables which sets up UMP/DMP settings, or gives a chain funds or anything. Also there may be no valid destination for a chain which doesnt communicate to anyone.

xcm/pallet-xcm-benchmarks/src/fungible/benchmarking.rs
```diff
@@ -0,0 +1,219 @@
...
+		let assets: MultiAssets = vec![ teleportable_asset ].into();
+
+		let mut executor = new_executor::<T>(trusted_teleporter);
+		let instruction = Instruction::ReceiveTeleportedAsset(assets.clone());
+		let xcm = Xcm(vec![instruction]);
+	}: {
+		executor.execute(xcm).map_err(|_| {
+			BenchmarkError::Override(
+				BenchmarkResult::from_weight(T::BlockWeights::get().max_block)
+			)
```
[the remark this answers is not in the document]
shawntabrizi: What about chains who just do not support this particular XCM. it covers that scenario where you can still run benchmarks for everything else.

The resulting number screams pretty loudly.

xcm/pallet-xcm-benchmarks/src/mock.rs
```diff
@@ -0,0 +1,64 @@
...
+	pub UnitWeightCost: Weight = 10;
+	pub WeightPrice: (AssetId, u128) = (Concrete(Here.into()), 1_000_000);
+}
+
+pub struct AllAssetLocationsPass;
+impl FilterAssetLocation for AllAssetLocationsPass {
+	fn filter_asset_location(_: &MultiAsset, _: &MultiLocation) -> bool {
+		true
+	}
+}
```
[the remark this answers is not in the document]
shawntabrizi: why? it is a mock function to enable tests to pass

xcm/pallet-xcm-benchmarks/src/fungible/mod.rs
shawntabrizi: It can be a different asset for teleports versus other xcm messages.

why are you worried about including more information?

apopiak: Just trying to keep the interface/config tidy.

apopiak (approved):
Seems fine if we iron out any kinks in follow ups.

shawntabrizi:
bot merge

shawntabrizi:
https://research.web3.foundation/ is down and causing the CI failure.

PR #3828, opened 2021-09-10
dockerfiles: upgrade to ubuntu:20.04; some chore

TriplEight:
I'm upgrading the CI image from being based on `debian:buster` to `ubuntu:20.04`, some details: https://github.com/paritytech/scripts/pull/309
The binaries produced in our new CI image will be linked by the newer `GLIBC` so won't be executable in debian:
```
/usr/local/bin/adder-collator: /lib/x86_64-linux-gnu/libm.so.6: version `GLIBC_2.29' not found (required by /usr/local/bin/adder-collator)
/usr/local/bin/adder-collator: /usr/lib/x86_64-linux-gnu/libstdc++.so.6: version `GLIBCXX_3.4.26' not found (required by /usr/local/bin/adder-collator)
```

### DIFF
--- a/bridges/deployments/networks/OpenEthereum.Dockerfile
+++ b/bridges/deployments/networks/OpenEthereum.Dockerfile
@@ -1,4 +1,4 @@
-FROM ubuntu:xenial AS builder
+FROM docker.io/library/ubuntu:xenial AS builder
 
 # show backtraces
 ENV RUST_BACKTRACE 1
@@ -60,7 +60,7 @@ WORKDIR /openethereum
 RUN cargo build --release --verbose
 RUN strip ./target/release/openethereum
 
-FROM ubuntu:xenial
+FROM docker.io/library/ubuntu:xenial
 
 # show backtraces
 ENV RUST_BACKTRACE 1
--- a/scripts/docker/release.Dockerfile
+++ b/scripts/docker/release.Dockerfile
@@ -1,4 +1,4 @@
-FROM ubuntu:20.04
+FROM docker.io/library/ubuntu:20.04
 
 # metadata
 ARG VCS_REF
--- a/scripts/docker/release.Dockerfile
+++ b/scripts/docker/release.Dockerfile
@@ -1,4 +1,4 @@
-FROM debian:buster-slim
+FROM ubuntu:20.04
 
 # metadata
 ARG VCS_REF
@@ -8,7 +8,7 @@ ARG POLKADOT_VERSION
 LABEL io.parity.image.authors="devops-team@parity.io" \
 	io.parity.image.vendor="Parity Technologies" \
 	io.parity.image.title="parity/polkadot" \
-	io.parity.image.description="polkadot: a platform for web3" \
+	io.parity.image.description="Polkadot: a platform for web3" \
 	io.parity.image.source="https://github.com/paritytech/polkadot/blob/${VCS_REF}/scripts/docker/Dockerfile" \
 	io.parity.image.revision="${VCS_REF}" \
 	io.parity.image.created="${BUILD_DATE}" \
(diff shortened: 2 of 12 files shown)


bkchr (approved):

bridges/deployments/networks/OpenEthereum.Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM ubuntu:xenial AS builder
```
TriplEight, line 1: @tomusdrw is this image relevant to Polkadot?
Also, Xenial (16.04) is out of standard support since April 2021, our containers policy is not happy.

scripts/docker/release.Dockerfile
```diff
@@ -19,24 +19,25 @@ ENV RUST_BACKTRACE 1
...
-		useradd -m -u 1000 -U -s /bin/sh -d /polkadot polkadot && \
-		gpg --recv-keys --keyserver hkps://keys.mailvelope.com 9D4B2B6EB8F97156D19669A9FF0812D491B96798 && \
-		gpg --export 9D4B2B6EB8F97156D19669A9FF0812D491B96798 > /usr/share/keyrings/parity.gpg && \
-		echo 'deb [signed-by=/usr/share/keyrings/parity.gpg] https://releases.parity.io/deb release main' > /etc/apt/sources.list.d/parity.list && \
-		apt-get update && \
-		apt-get install -y --no-install-recommends polkadot=${POLKADOT_VERSION#?} && \
+	DEBIAN_FRONTEND=noninteractive apt-get install -y --no-install-recommends \
+		libssl1.1 \
+		ca-certificates \
+		curl \
```
chevdor, line 25: Adding `curl` involves potential security risks. Since `curl` is not required for running `polkadot`, could we either:
- remove the curl dependency entirely ?
- remove `curl` once we are done with it ?

scripts/docker/release.Dockerfile
```diff
@@ -19,24 +19,25 @@ ENV RUST_BACKTRACE 1
...
+		gnupg && \
+	useradd -m -u 1000 -U -s /bin/sh -d /polkadot polkadot && \
+# add repo's gpg keys and install the published polkadot binary
+	gpg --recv-keys --keyserver hkps://keys.mailvelope.com 9D4B2B6EB8F97156D19669A9FF0812D491B96798 && \
```
chevdor, line 29: Storing `9D4B2B6EB8F97156D19669A9FF0812D491B96798 ` in an `ARG` would bring benefits:
- less repetition
- more readable
- easier/saver to update when needed

chevdor (changes requested):
Dealing with `curl` in all the images would be a good thing security-wise.

scripts/docker/release.Dockerfile
TriplEight: I do not believe we depend on `curl`, it's there for debug purposes.

TriplEight: happy to remove it

bridges/deployments/networks/OpenEthereum.Dockerfile
tomusdrw: No I don't think so. If possible please avoid touching `bridges` code in this repo, it's only a clone of the `bridges` repository and the changes are better done there first and then synced here to avoid conflicts.

Is it caught by some blanket CI checks? Can we ignore `bridges` sub-dir from any checks of this kind? In the future the plan is to integrate the bridges code more closely with Polkadot/Substrate repositories, but for now we would rather keep focus on making sure the code works correctly in our repo.

scripts/docker/release.Dockerfile
TriplEight: let's merge it as is here, I'll create a separate issue for this.

scripts/docker/release.Dockerfile
TriplEight: let's merge it as is here, I'll create a separate issue for this.

bridges/deployments/networks/OpenEthereum.Dockerfile
TriplEight: yeah, a more general question was "what the entire bridges repo does here?" and now I get it, thanks. I'll make a change in bridges then.

tomusdrw: It's a `git subtree`, because we want to use some of the crates from bridges repo, but keep working on the separately. Other option is referencing crates using git repo, but that makes it cumbersome in case of compilation errors (i.e. companion PR system).

scripts/docker/release.Dockerfile
chevdor: See https://github.com/paritytech/ci_cd/issues/197

scripts/docker/release.Dockerfile
chevdor: See https://github.com/paritytech/ci_cd/issues/197

chevdor (approved):

tomusdrw (approved):

PR #3845, opened 2021-09-14
Fix spelling

chevdor:
This PR is a backport of a commit from the 0.9.10 release allowing to fix the blocking spelling issues.

### DIFF
--- a/bridges/.config/lingua.dic
+++ b/bridges/.config/lingua.dic
@@ -65,6 +65,7 @@ isn
 io
 js
 keccak256/M
+keypair/MS
 KSM/S
 Lane1
 Lane2
@@ -101,6 +102,7 @@ parablock/MS
 parachain/MS
 parameterize/D
 pallet_message_lane
+plancks
 polkadot/MS
 pov-block/MS
 PoA
--- a/bridges/.config/lingua.dic
+++ b/bridges/.config/lingua.dic
@@ -66,6 +66,7 @@ io
 js
 keccak256/M
 keypair/MS
+KeyPair
 KSM/S
 Lane1
 Lane2
--- a/bridges/.config/lingua.dic
+++ b/bridges/.config/lingua.dic
@@ -66,7 +66,6 @@ io
 js
 keccak256/M
 keypair/MS
-KeyPair
 KSM/S
 Lane1
 Lane2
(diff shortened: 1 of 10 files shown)


tomusdrw (approved):
This should be done by a `git subtree` update to avoid conflicts in the future, but let's unblock whatever is blocking you :)

svyatonik (approved):

chevdor:
Thank you @tomusdrw. Please note however that this is NOT blocking me as I have the fix in the release branch anyway. I just wanted to backport it here to ensure it is kept in sync.
So if you see complications merging this PR, we can also drop it.

tomusdrw:
> So if you see complications merging this PR, we can also drop it.

It's alright, should be resolved automatically anyway. I plan to update the bridges code this/next week anyway.

bridges/.config/lingua.dic
```diff
@@ -65,6 +65,8 @@ isn
 io
 js
 keccak256/M
+keypair/MS
+KeyPair
```
drahnr, line 69: This should be code right? So we should not whitelist it, but require it to be `\`KeyPair\`` (working on a shortcut for `cargo-spellcheck` to have this in interactive mode).

bridges/.config/lingua.dic
```diff
@@ -101,6 +103,7 @@ parablock/MS
 parachain/MS
 parameterize/D
 pallet_message_lane
+plancks
```
drahnr, line 106: :+1:

scripts/gitlab/lingua.dic
```diff
@@ -117,6 +117,7 @@ jaeger/MS
 js
 keccak256/M
 keypair/MS
+KeyPair
```
drahnr, line 120: CC

drahnr (approved):
LGTM, we should probably remove `\`KeyPair\`` from the allowed list.

I am also slightly puzzled why bridges has their own spellcheck config :monocle_face:

chevdor:
@drahnr I did not dig into the meaning of those `/MS` but `keypair/MS` appears to accept only `Keypair` and not `KeyPair`. The detected mispelling comes from a comment refering to the type `KeyPair`. If you know how we can modify the dictionnary without the duplication, we can do it now but I think those disctionnary need a major cleanup anyway

bridges/.config/lingua.dic
chevdor: No this is a **comment** refering to the type called `KeyPair`.

drahnr:
bot merge

PR #3851, opened 2021-09-15
fix clock drift for assignments issued before the block

rphmeier:
In our tests, we typically import a block with tick 10 and then import assignments recorded at tick 0 (as the test clock is at tick 0). This often leads to the nonsensical tranche-0 no-show wakeups at tick 20, when they should be at 30 (block-tick + no-show-delay = 10 + 20 = 30).

This breaks a bunch of tests, so those need to be fixed first.


node/core/approval-voting/src/approval_checking.rs
```diff
@@ -843,11 +843,10 @@ mod tests {
 				no_show_duration,
 				needed_approvals,
 			),
-			RequiredTranches::Pending {
-				considered: 2,
-				next_no_show: None,
-				maximum_broadcast: 3,
-				clock_drift: block_tick,
+			RequiredTranches::Exact {
```
rphmeier: Given the name of the test I suspect that this test is wrong if the result is not `Pending`

node/core/approval-voting/src/tests.rs
```diff
@@ -173,6 +173,7 @@ impl MockClockInner {
 	}
 
 	fn has_wakeup(&self, tick: Tick) -> bool {
+		println!("WAKEUPS {:?}", self.wakeups);
```
rphmeier: Stray println

node/core/approval-voting/src/tests.rs
```diff
@@ -2435,9 +2436,9 @@ fn subsystem_assignment_triggered_by_all_with_less_than_threshold() {
 		assignments_to_import: vec![1, 2, 3, 4, 5],
 		approvals_to_import: vec![2, 4],
 		ticks: vec![
-			20, // Check for no shows
+			21, // Check for no shows
 		],
-		should_be_triggered: |_| true,
+		should_be_triggered: |_| false,
```
rphmeier, line 2440: This seems wrong based on the test name

ordian (approved):

Issue #3850, opened 2021-09-15
Benchmark Parachains Runtime: Tracking Issue


shawntabrizi:
This Issue will be used to track end to end benchmarking of the Parachains runtime:

https://github.com/paritytech/polkadot/tree/master/runtime/parachains

- [x] Configuration https://github.com/paritytech/polkadot/pull/3862
- [x] Disputes (not required)
- [x] DMP (not required)
- [x] HRMP: https://github.com/paritytech/polkadot/pull/3876
- [x] Inclusion (not required)
- [x] Initializer #3913
- [x] Paras #3888
- [x] Paras Inherent (https://github.com/paritytech/polkadot/pull/4028)
- [x] Reward Points (not required)
- [x] Scheduler (not required)
- [x] Session Info (not required)
- [x] Shared (not required)
- [x] UMP #3889
- [x] Util (not required)

When checking off an item, please include a PR which includes the benchmark PR.

shawntabrizi:
We are looking for help on this issue, so consider it if you are an external contributor and want to get more involved with benchmarking!

hirschenberger:
> We are looking for help on this issue, so consider it if you are an external contributor and want to get more involved with benchmarking!

I'd like to help, so can I take @KiChjang 's changes as a guide what's to do? Anything that I have to take care of beforehand? So I would start with `Displutes` then.

shawntabrizi:
@hirschenberger happy to have you help. I dont really have a mental model of which pallets are easier than others, so maybe take a look a them and see what you can learn.

The main thing with any benchmarking is:

* reading the pallet and understanding the code
* determining the worst case scenario path for an extrinsic
* creating a benchmark which sets up that worst case scenario, and executes the call
* verifying the final resulting state is what you expect.

https://github.com/paritytech/substrate/tree/master/frame/benchmarking

hirschenberger:
Ok, I read into it and will start with `Dmp`. 

I assume I have to go through some bootstrapping process: 

* Writing the benches
* Compile and run the benches to generate the `weights.rs`
* Add the weights to the exrinsic declarations.
* Commit and the final weigths are generated on the spec-hardware by CI

Is that correct?

kianenigma:
[...]

PR #3862, opened 2021-09-16
Add benchmarking for parachain runtime configuration pallet

KiChjang:
Part of #3850.


runtime/parachains/src/configuration.rs
```diff
@@ -613,7 +663,7 @@ pub mod pallet {
 		}
 
 		/// Sets the soft limit for the phase of dispatching dispatchable upward messages.
-		#[pallet::weight((1_000, DispatchClass::Operational))]
+		#[pallet::weight(T::WeightInfo::set_ump_service_total_weight())]
```
shawntabrizi, line 666: practically speaking, we could probably make one benchmark for all extrinsics which are simply `new: u32`, which will simplify maintenance a lot.

runtime/parachains/src/configuration.rs
```diff
@@ -665,7 +715,7 @@ pub mod pallet {
 
 		/// Sets the amount of funds that the recipient should provide for accepting opening an HRMP
 		/// channel.
-		#[pallet::weight((1_000, DispatchClass::Operational))]
+		#[pallet::weight(T::WeightInfo::set_hrmp_recipient_deposit())]
```
shawntabrizi, line 718: and then a single benchmark for `new: Balance`... etc...

shawntabrizi (approved):
lgtm

adoerr (approved):

KiChjang:
bot merge

PR #3876, opened 2021-09-17
HRMP benchmarks

kianenigma:
as you see, nothing is done, just opening a PR to signal that I will work on this and prevent overlap. 

- [x] Weigh the background-like tasks, and make all of the inter-pallet calls self-weiging. Could be done in a follow-up. All dispatchables area already being weighed. 

part of https://github.com/paritytech/polkadot/issues/3850


runtime/kusama/src/lib.rs
```diff
@@ -1082,22 +1082,22 @@ impl InstanceFilter<Call> for ProxyType {
 			),
 			ProxyType::Governance => matches!(
 				c,
-				Call::Democracy(..) |
-					Call::Council(..) | Call::TechnicalCommittee(..) |
-					Call::PhragmenElection(..) |
-					Call::Treasury(..) | Call::Bounties(..) |
-					Call::Tips(..) | Call::Utility(..)
+				Call::Democracy(..)
+					| Call::Council(..) | Call::TechnicalCommittee(..)
```
kianenigma, line 1086: errrr I have no idea why my vscode is doing this. 

When I use the default formatter in the editor, it does this. It seems like it is using a different version of `cargo fmt` in there. When I do `cargo +nightly fmt` in the root, it is all good. Anyways, will revert later.

runtime/parachains/src/hrmp.rs
```diff
@@ -2008,3 +2015,244 @@ mod tests {
...
+		hrmp_close_channel {
+			let [(sender, sender_origin), (recipient, _)] = establish_para_connection::<T>(1, 2, ParachainSetupStep::Established);
+			let channel_id = HrmpChannelId { sender, recipient };
+		}: _(sender_origin, channel_id.clone())
+		verify {
+			assert_last_event::<T>(Event::<T>::ChannelClosed(sender, channel_id).into());
+		}
+
+		force_clean_hrmp {
+			// TODO: reasonable constant/config for these.
```
kianenigma, line 2159: Proper way to solve this, for my future self: These values come from top level `struct HostConfiguration` and it seems like all of the values are upgradable in the runtime through root. So we can't really make any assumption about them.

We need to put a bound in the upper bound of channel counts in the `configurations` pallet, and never allow even governance to go beyond that. Then we use that bound here as well.

shawntabrizi: yup, we need this :)

runtime/parachains/src/paras.rs
```diff
@@ -850,6 +850,19 @@ impl<T: Config> Pallet<T> {
 		Ok(())
 	}
 
+	#[cfg(feature = "runtime-benchmarks")]
+	pub(crate) fn initialize_para_now(id: ParaId, genesis: ParaGenesisArgs) -> DispatchResult {
+		// first queue this para actions..
+		let _ = Self::schedule_para_initialize(id, genesis)?;
+
+		// .. and immediately apply them.
+		let _ = Self::apply_actions_queue(Self::scheduled_session());
```
shawntabrizi, line 859: i dont think you need `let _ =`

runtime/parachains/src/paras.rs
```diff
@@ -850,6 +850,19 @@ impl<T: Config> Pallet<T> {
...
+
+		// ensure it has become a para.
+		assert_eq!(ParaLifecycles::<T>::get(id), Some(ParaLifecycle::Parachain));
```
shawntabrizi, line 862: might as well let this be a result too

runtime/parachains/src/configuration.rs
```diff
@@ -284,6 +284,18 @@ pub mod pallet {
 	pub trait Config: frame_system::Config + shared::Config {
 		/// Weight information for extrinsics in this pallet.
 		type WeightInfo: WeightInfo;
+
+		/// Maximum value that `hrmp_max_parachain_outbound_channels` or
+		/// `hrmp_max_parathread_outbound_channels` can be set to.
+		///
+		/// This is used for benchmarking sanely bounding relevant storage items.
+		type HrmpMaxOutboundChannelsBound: Get<u32>;
```
kianenigma, line 292: Long term, I don't think this patter is good. We probably will need to add a lot of bound types here, one for each field of `HostConfigurations`. I don't see it within the scope of this PR, but if we go down that road, it is probably worthwhile to abstract all of these types into one new type, maybe called `struct HostConfigurationBounds` and only use that to minimize the footprint in the interface.

runtime/kusama/src/lib.rs
```diff
@@ -1134,6 +1134,8 @@ impl parachains_origin::Config for Runtime {}
 
 impl parachains_configuration::Config for Runtime {
 	type WeightInfo = weights::runtime_parachains_configuration::WeightInfo<Runtime>;
+	type HrmpMaxOutboundChannelsBound = ConstU32<128>;
+	type HrmpMaxInboundChannelsBound = ConstU32<128>;
```
kianenigma, line 1138: rather arbitrary numbers.

runtime/parachains/src/configuration.rs
kianenigma: Also, note that for simplicity, for now I have combined the parachain and parathread variant. This might help us use `BoundedVec` in some places as well, but that didn't work out.

runtime/parachains/src/hrmp.rs
```diff
@@ -2008,3 +2018,251 @@ mod tests {
...
+			let message_size = Configuration::<T>::config().hrmp_channel_max_message_size;
+		}: _(sender_origin, recipient_id, capacity, message_size)
+		verify {
+			assert_last_event::<T>(
+				Event::<T>::OpenChannelRequested(sender_id, recipient_id, capacity, message_size).into()
+			);
+		}
+
+		hrmp_accept_open_channel {
+			let [(sender, _), (recipient, recipient_origin)] =establish_para_connection::<T>(1, 2, ParachainSetupStep::Requested);
```
pepyakin, line 2152:
```suggestion
			let [(sender, _), (recipient, recipient_origin)] = establish_para_connection::<T>(1, 2, ParachainSetupStep::Requested);
```

runtime/parachains/src/hrmp.rs
```diff
@@ -2008,3 +2018,251 @@ mod tests {
...
+		}
+
+		force_clean_hrmp {
+			// ingress channels to a single leaving parachain that need to be closed.
+			let i in 0 .. (<T as configuration::Config>::HrmpMaxInboundChannelsBound::get() - 1);
+			// egress channels to a single leaving parachain that need to be closed.
+			let e in 0 .. (<T as configuration::Config>::HrmpMaxOutboundChannelsBound::get() - 1);
+
+			// we use 10_000 in this benchmark as the prefix of accounts. The components must be
+			//than that to keep accounts sane.
```
pepyakin, line 2173:
```suggestion
			// than that to keep accounts sane.
```

pepyakin (commented):
It looks fine, assuming the formatting changes will be reverted.

PS: For the future, I would really appreciate if a PR didn't include so much formatting. It complicates reviewing as it takes effort to see what actually requires attention.

kianenigma:
> It looks fine, assuming the formatting changes will be reverted.
> 
> PS: For the future, I would really appreciate if a PR didn't include so much formatting. It complicates reviewing as it takes effort to see what actually requires attention.

Yeah, I am having some weird issues with cargo fmt: https://github.com/paritytech/polkadot/pull/3876#discussion_r712371825

runtime/parachains/src/hrmp.rs
```diff
@@ -305,17 +309,17 @@ pub mod pallet {
...
 	pub type HrmpChannels<T: Config> = StorageMap<_, Twox64Concat, HrmpChannelId, HrmpChannel>;
 
-	/// Ingress/egress indexes allow to find all the senders and receivers given the opposite
-	/// side. I.e.
+	/// Ingress/egress indexes allow to find all the senders and receivers given the opposite side.
+	/// I.e.
 	///
-	/// (a) ingress index allows to find all the senders for a given recipient.
-	/// (b) egress index allows to find all the recipients for a given sender.
+	/// (a) ingress index allows to find all the senders for a given recipient. (b) egress index
```
pepyakin, line 315: I would prefer to keep that as a list.

runtime/parachains/src/hrmp.rs
```diff
@@ -339,27 +343,27 @@ pub mod pallet {
...
 		StorageMap<_, Twox64Concat, ParaId, Vec<(T::BlockNumber, Vec<ParaId>)>, ValueQuery>;
 
 	/// Preopen the given HRMP channels.
 	///
-	/// The values in the tuple corresponds to `(sender, recipient, max_capacity, max_message_size)`,
-	/// i.e. similar to `init_open_channel`. In fact, the initialization is performed as if
-	/// the `init_open_channel` and `accept_open_channel` were called with the respective parameters
-	/// and the session change take place.
+	/// The values in the tuple corresponds to `(sender, recipient, max_capacity,
+	/// max_message_size)`, i.e. similar to `init_open_channel`. In fact, the initialization is
```
pepyakin, line 359: IMO, the prior version that does not wrap the tuple is more legible since it avoids scanning the line.

runtime/parachains/src/hrmp.rs
```diff
@@ -261,13 +262,16 @@ pub mod pallet {
...
+	// global state.
 	#[pallet::storage]
 	pub type HrmpOpenChannelRequestsList<T: Config> =
 		StorageValue<_, Vec<HrmpChannelId>, ValueQuery>;
 
 	/// This mapping tracks how many open channel requests are initiated by a given sender para.
-	/// Invariant: `HrmpOpenChannelRequests` should contain the same number of items that has `(X, _)`
-	/// as the number of `HrmpOpenChannelRequestCount` for `X`.
+	/// Invariant: `HrmpOpenChannelRequests` should contain the same number of items that has `(X,
+	/// _)` as the number of `HrmpOpenChannelRequestCount` for `X`.
```
pepyakin, line 274: This became less legible as well.

pepyakin (commented):
Ah, actually, I also assumed that comments underwent some kind of automatic formatting. 

I am really sensitive about the line width and never go over 120. The ones you wrapped looked more like they were 100, which I believe is fine to do occasionally, if that improves reading experience.  In some instances it is fine and I do not feel strongly, but in others (I mentioned them below) I do.

runtime/parachains/src/hrmp.rs
```diff
@@ -2008,3 +2018,251 @@ mod tests {
...
+			crate::paras::ParaGenesisArgs {
+				parachain: true,
+				genesis_head: vec![1].into(),
+				validation_code: vec![1].into(),
+			},
+		));
+		T::Currency::make_free_balance_be(&id.into_account(), balance);
+	}
+
+	fn assert_last_event<T: Config>(generic_event: <T as Config>::Event) {
```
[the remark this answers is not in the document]
kianenigma: a bit of hassle, since the one in mock is not generic.

runtime/parachains/src/hrmp.rs
```diff
@@ -2009,3 +2140,251 @@ mod tests {
 		});
 	}
 }
+
+#[cfg(feature = "runtime-benchmarks")]
+mod benchmarking {
```
KiChjang, line 2145: Why not put this in a separate file instead?

runtime/parachains/src/hrmp.rs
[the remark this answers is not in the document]
KiChjang: The one in mock also can't be used here, since it requires `cfg(test)`.

runtime/parachains/src/hrmp.rs
```diff
@@ -2009,3 +2140,251 @@ mod tests {
...
+
+			for id in 0 .. c {
+				let _ = establish_para_connection::<T>(PREFIX_0 + id, PREFIX_1 + id, ParachainSetupStep::CloseRequested);
+			}
+
+			assert_eq!(HrmpCloseChannelRequestsList::<T>::decode_len().unwrap_or_default() as u32, c);
+		}: _(frame_system::Origin::<T>::Root, c)
+		 verify {
+			 assert_eq!(HrmpCloseChannelRequestsList::<T>::decode_len().unwrap_or_default() as u32, 0);
+		 }
```
KiChjang, line 2365: nit: Extra space at the beginning:

```suggestion
		verify {
			assert_eq!(HrmpCloseChannelRequestsList::<T>::decode_len().unwrap_or_default() as u32, 0);
		}
```

runtime/parachains/src/mock.rs
```diff
@@ -296,6 +304,7 @@ impl inclusion::RewardValidators for TestRewardValidators {
 
 /// Create a new set of test externalities.
 pub fn new_test_ext(state: MockGenesisConfig) -> TestExternalities {
+	sp_tracing::try_init_simple();
```
KiChjang, line 307: Guess this is a debugging artifact, is it still desirable to leave it included?

kianenigma: Yes, I always prefer the _option_ to enable logging when running tests, which this line enables. It won't have any impact if you don't.

runtime/parachains/src/hrmp.rs
kianenigma: looking at the whole crate, it seems like the pattern is one file per pallet (main code + test, and now adding benchmarks), but happy to revamp if you think I am wrong.

pepyakin: Recent migration for the `configuration` changed this.

kianenigma: UPDATE: I see that you've changed this pattern in the `configuration`. Why is there a `configuration` folder and a `configuration.rs`? shouldn't `configuration.rs` be the `mod.rs` of `configuration` folder? I think that's a good pattern, and would like to do it that way for this pallet as well.

KiChjang: That's actually a new format in Rust 2018 -- instead of a `configuration/mod.rs` file, you can have a `configuration.rs` file to replace it and all the files under the `configuration/` directory would work as if they're modules of `configuration.rs`.

runtime/kusama/src/lib.rs
```diff
@@ -1163,6 +1163,8 @@ impl parachains_origin::Config for Runtime {}
 
 impl parachains_configuration::Config for Runtime {
 	type WeightInfo = weights::runtime_parachains_configuration::WeightInfo<Runtime>;
+	type HrmpMaxOutboundChannelsBound = frame_support::traits::ConstU32<128>;
```
kianenigma, line 1166: @pepyakin @rphmeier can either of you confirm that this value is a sensible bound?

kianenigma:
@pepyakin @emostov this should be now in good state to be reviewed.

runtime/polkadot/src/lib.rs
```diff
@@ -1179,6 +1181,7 @@ impl parachains_hrmp::Config for Runtime {
 	type Event = Event;
 	type Origin = Origin;
 	type Currency = Balances;
+	type WeightInfo = parachains_hrmp::TestWeightInfo;
```
emostov, line 1184: I think we should add / list benchmark and use the real weights

runtime/parachains/src/configuration.rs
```diff
@@ -1079,11 +1095,11 @@ mod tests {
 				hrmp_recipient_deposit: 4905,
 				hrmp_channel_max_capacity: 3921,
 				hrmp_channel_max_total_size: 7687,
-				hrmp_max_parachain_inbound_channels: 3722,
-				hrmp_max_parathread_inbound_channels: 1967,
+				hrmp_max_parachain_inbound_channels: 37,
+				hrmp_max_parathread_inbound_channels: 19,
```
emostov, line 1099: Is there any specific justification for the values chosen, or did we just want to have a random value less than the bound?

runtime/parachains/src/hrmp.rs
```diff
@@ -458,8 +515,10 @@ pub mod pallet {
 		///
 		/// If there are pending HRMP open channel requests, you can use this
 		/// function process all of those requests immediately.
-		#[pallet::weight(0)]
-		pub fn force_process_hrmp_open(origin: OriginFor<T>) -> DispatchResult {
+		///
+		/// Total number of opening channels must be provided as witness data.
+		#[pallet::weight(<T as Config>::WeightInfo::force_process_hrmp_open(*_channels))]
```
emostov, line 520: Same questions as above

runtime/parachains/src/hrmp/benchmarking.rs
```diff
@@ -0,0 +1,289 @@
...
+		// NOTE: this is just for expressiveness, otherwise the if-statement is 100% useless.
+		return output
+	}
+
+	output
+}
+
+const PREFIX_0: u32 = 10_000;
+const PREFIX_1: u32 = PREFIX_0 * 2;
+const MAX_UNIQUE_CHANNELS: u32 = 128;
```
emostov, line 123: Where does this number come from?

emostov (approved):
Benchmarks look good. Just a couple questions

runtime/parachains/src/hrmp.rs
```diff
@@ -447,8 +497,15 @@ pub mod pallet {
 		/// you to trigger the cleanup immediately for a specific parachain.
 		///
 		/// Origin must be Root.
-		#[pallet::weight(0)]
-		pub fn force_clean_hrmp(origin: OriginFor<T>, para: ParaId) -> DispatchResult {
+		///
+		/// Number of inbound and outbound channels for `para` must be provided as witness data.
+		#[pallet::weight(<T as Config>::WeightInfo::force_clean_hrmp(*_inbound, *_outbound))]
```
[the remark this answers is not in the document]
kianenigma: the `*` is a deref. This is a bit detailed about how the macros expand. Unlike the body of the function where `_inbound` is `u32`, the weight expression is evaluated in a code similar to: 

```rust 
match call_enum {
	Call::force_clean_hrmp(para, _inbound, _outbound) => { /* weight expression */}
}
```

and this by default borrows all the enum arguments, even if they are `Copy`.

---

Validating this one is optional IMO, since the call is root. But a great catch, the non-root ones need to be validated.

runtime/parachains/src/hrmp.rs
```diff
@@ -470,22 +529,27 @@ pub mod pallet {
...
 		/// The cancellation happens immediately. It is not possible to cancel the request if it is
 		/// already accepted.
-		#[pallet::weight(0)]
+		///
+		/// Total number of opening channels must be provided as witness data.
+		#[pallet::weight(<T as Config>::WeightInfo::hrmp_cancel_open_request(*_channels))]
 		pub fn hrmp_cancel_open_request(
 			origin: OriginFor<T>,
 			channel_id: HrmpChannelId,
+			_channels: u32,
```
kianenigma, line 552: TODO: validate this.

runtime/parachains/src/hrmp/benchmarking.rs
kianenigma: This one does not have a bound anywhere in the codebase, so I left it as a const and u128 should be enough to capture the dependency in the benchmark analysis process. For example, `force_process_hrmp_open ` will iterate over all open channels requests and services them. We know from that

1. it is worse case if all channels have unique sender and receiver.
2. The complexity is indeed `O(2n)` with `n` channels. The main point is to make sure the benchmark system picks up this relation, which it should with any large enough value for `MAX_UNIQUE_CHANNELS`. I reckon if we do 32 or 4096 for it the benchmark should be exactly the same

hirschenberger:
> /benchmark runtime polkadot runtime_parachains::hrmp

I have the same problem running the polkadot benches in the ump branch. Any idea what's causing that?

emostov:
> > /benchmark runtime polkadot runtime_parachains::hrmp
> 
> I have the same problem running the polkadot benches in the ump branch. Any idea what's causing that?

I have had this issue before when I push after starting the bench bot, but not sure what the other causes could be

runtime/kusama/src/lib.rs
pepyakin: Yes, 128 seems sensible upper bound. Each parachain can only open a single channel to another parachain. There is only so many parachains, i.e 100.

runtime/parachains/src/hrmp.rs
kianenigma: Done.

runtime/parachains/src/hrmp.rs
```diff
@@ -470,24 +531,35 @@ pub mod pallet {
...
 		/// already accepted.
-		#[pallet::weight(0)]
+		///
+		/// Total number of open requests (i.e. `HrmpOpenChannelRequestsList`) must be provided as
+		/// witness data.
+		#[pallet::weight(<T as Config>::WeightInfo::hrmp_cancel_open_request(*open_requests))]
 		pub fn hrmp_cancel_open_request(
 			origin: OriginFor<T>,
 			channel_id: HrmpChannelId,
+			open_requests: u32,
```
pepyakin, line 555: This actually may complicate the usage of the API. Those calls are issued by parachains via UMP and they are necessarily less flexible, I imagine. (E.g. issued by governance)

runtime/parachains/src/hrmp/benchmarking.rs
```diff
@@ -0,0 +1,287 @@
...
+
+	hrmp_accept_open_channel {
+		let [(sender, _), (recipient, recipient_origin)] = establish_para_connection::<T>(1, 2, ParachainSetupStep::Requested);
+	}: _(recipient_origin, sender)
+	verify {
+		assert_last_event::<T>(Event::<T>::OpenChannelAccepted(sender, recipient).into());
+	}
+
+	hrmp_close_channel {
+		let [(sender, sender_origin), (recipient, _)] = establish_para_connection::<T>(1, 2, ParachainSetupStep::Established);
```
pepyakin, line 158: Exceeds hard limit for the line width (as well as several other lines in this file)

runtime/parachains/src/hrmp.rs
```diff
@@ -534,31 +606,49 @@ impl<T: Config> Pallet<T> {
...
+			));
 		}
+		w
 	}
 
 	// Go over the HRMP open channel requests and remove all in which offboarding paras participate.
 	//
 	// This will also perform the refunds for the counterparty if it doesn't offboard.
-	fn clean_open_channel_requests(
+	pub fn clean_open_channel_requests(
```
pepyakin, line 648: pub(crate). Could better, could we do `pub(in benchmarking)`? In that case we would need to move `#[cfg(feature = "runtime-benchmarks")]` inside of the module and make it `#![..]`

runtime/parachains/src/hrmp.rs
```diff
@@ -534,31 +606,49 @@ impl<T: Config> Pallet<T> {
...
+	) -> Weight {
+		let w1 = Self::perform_outgoing_para_cleanup(&notification.prev_config, outgoing_paras);
 		Self::process_hrmp_open_channel_requests(&notification.prev_config);
 		Self::process_hrmp_close_channel_requests();
+		w1.saturating_add(<T as Config>::WeightInfo::force_process_hrmp_open(
+			outgoing_paras.len() as u32
+		))
+		.saturating_add(<T as Config>::WeightInfo::force_process_hrmp_close(
+			outgoing_paras.len() as u32
+		))
```
pepyakin, line 618: I find it a little bit weird how it reads here: we use the weight data that was meant to be used (judging by the name) by `force_*` dispatchables.

runtime/parachains/src/hrmp.rs
pepyakin, line 648: ditto

runtime/parachains/src/hrmp.rs
```diff
@@ -1066,6 +1160,7 @@ impl<T: Config> Pallet<T> {
 		{
 			// this should never happen unless the max downward message size is configured to an
 			// jokingly small number.
+			log::error!(target: "runtime::hrmp", "sending 'init_open_channel::notification_bytes' failed.");
```
pepyakin, line 1163: Can we wrap it so that it doesn't exceed the soft limit of 100?

runtime/parachains/src/hrmp.rs
```diff
@@ -1120,6 +1215,7 @@ impl<T: Config> Pallet<T> {
 		{
 			// this should never happen unless the max downward message size is configured to an
 			// jokingly small number.
+			log::error!(target: "runtime::hrmp", "sending 'accept_open_channel::notification_bytes' failed.");
```
pepyakin, line 1218: ditto

runtime/parachains/src/hrmp.rs
```diff
@@ -1194,6 +1290,7 @@ impl<T: Config> Pallet<T> {
 		{
 			// this should never happen unless the max downward message size is configured to an
 			// jokingly small number.
+			log::error!(target: "runtime::hrmp", "sending 'close_channel::notification_bytes' failed.");
```
pepyakin, line 1293: ditto

runtime/parachains/src/hrmp.rs
kianenigma: Understandable, but that's just the way it is, if you want to have accurate (and not overly pessimistic) weights. We do the same in many other substrate APIs, and the transaction submission app (aka. a _UI_) will take care of it for you. For example, to do `withdraw_unbonded`, you have to pass in the number of slashing spans. That is also complicating the API, but it is intention. 

In this case, I believe the parachains can also deal with this extra parameter. If you really insist, the only other way is to put a `const` max upper bound for open request. This won't be trivial either, as the number of open request is not strictly bounded in the code.

runtime/parachains/src/hrmp/benchmarking.rs
kianenigma: err yeah, cargo fmt is not working at all in macros unfortunately..

runtime/parachains/src/hrmp.rs
kianenigma: True, we're reusing the weights of the force dispatchables here since they are almost 100% the same and can be reused. Maybe this extended snipped clearifies my intention further. 

```
// is not weighed anywhere else, therefore we return its weight manually. 
Self::perform_outgoing_para_cleanup(&notification.prev_config, outgoing_paras);
// already weighed as a part of dispatchable, so we reuse it.
Self::process_hrmp_open_channel_requests(&notification.prev_config);
Self::process_hrmp_close_channel_requests();
```

runtime/parachains/src/hrmp.rs
kianenigma: will suffice to just making pub(crate) for now. How about a `#[cfg(feature = "runtime-benchmarks")] pub(crate) fn clean_open_channel_requests_for_benchmarks` wrapper for the private one?

kianenigma:
@pepyakin your comments are mostly addressed.

runtime/parachains/src/hrmp.rs
pepyakin: I think it's just the naming implies that the `force_` functions are central and these internal functions are secondary, whereas it is the other way around: it is just so happens that the `force_` functions are the same as those internal functions.

kianenigma:
@pepyakin or @shawntabrizi, either of you can do a second review?

kianenigma:
> @pepyakin or @shawntabrizi, either of you can do a second review?

Weekly reminder on this 😅.

runtime/parachains/src/hrmp/weights.rs
```diff
@@ -0,0 +1,114 @@
...
+	fn hrmp_cancel_open_request(c: u32) -> Weight {
+		(35_031_000 as Weight)
+			// Standard Error: 0
+			.saturating_add((56_000 as Weight).saturating_mul(c as Weight))
+			.saturating_add(RocksDbWeight::get().reads(3 as Weight))
+			.saturating_add(RocksDbWeight::get().writes(3 as Weight))
+	}
+
+	fn clean_open_channel_requests(_: u32) -> Weight {
+		0
```
pepyakin, line 112: Why is this 0?

runtime/parachains/src/hrmp.rs
```diff
@@ -1036,6 +1128,8 @@ impl<T: Config> Pallet<T> {
 			config.hrmp_sender_deposit.unique_saturated_into(),
 		)?;
 
+		// mutating storage -- shall not bail henceforth.
```
pepyakin, line 1131: I think this comment is misleading since mutating storage happens in reserve.

runtime/parachains/src/paras.rs
```diff
@@ -884,6 +884,22 @@ impl<T: Config> Pallet<T> {
...
+		// .. and immediately apply them.
+		Self::apply_actions_queue(Self::scheduled_session());
+
+		// ensure it has become a para.
+		ensure!(
+			ParaLifecycles::<T>::get(id) == Some(ParaLifecycle::Parachain),
+			"Parachain not created properly"
+		);
+		Ok(())
+	}
```
pepyakin, line 901: I think this is a rather crude way of doing this. Can we initialize the parachain in `register_parachain_with_balance` and then manually move 2 sessions, removing this code from here?

runtime/parachains/src/hrmp.rs
```diff
@@ -470,24 +531,35 @@ pub mod pallet {
 		///
 		/// If there are pending HRMP close channel requests, you can use this
 		/// function process all of those requests immediately.
-		#[pallet::weight(0)]
-		pub fn force_process_hrmp_close(origin: OriginFor<T>) -> DispatchResult {
+		///
+		/// Total number of closing channels must be provided as witness data.
```
pepyakin, line 535: Nit: Total number of closing channels must be provided as witness data for correctly estimating the weight?

(Here and in other places)

runtime/parachains/src/hrmp/weights.rs
kianenigma: actually this file should be removed. We don't need a `impl WeightInfo for ()`.

runtime/parachains/src/hrmp.rs
kianenigma: yeah I see your point. I guess what I am trying to say is that **mutating storage one way or another**. `reserve` is fallible, what follows this comment is no longer fallible.

kianenigma: made it clear now.

runtime/parachains/src/paras.rs
kianenigma: moving forward two sessions in a realistic way requires bringing in a helper function like https://github.com/paritytech/polkadot/blob/d16547d930108945fd4d188677d92efbe6f0be06/runtime/parachains/src/hrmp.rs#L1386. Are you sure this is worth it? I have second doubts about it.

pepyakin: You know what, on the second thought, with #4457 I think it would be better if I take care of it later.

kianenigma:
@pepyakin new year, time to merge or close this one. 

@emostov, is this still an approve from you?

pepyakin:
No, it's fine from my side.

kianenigma:
> No, it's fine from my side.

But you have not approved :p

kianenigma:
well, it needs some major updates. I will do one last push for this and hope for it to finally get a second approve.

emostov:
> well, it needs some major updates given the long time that it has waited in the queue. I will do one last push for this and hope for it to finally get a second approve.

ping me for a re-review when ready

drahnr (approved):
Changes look reasonable to me, I did not do a deep dive on the parameterization logic of the actual benchmarking code.

runtime/parachains/src/hrmp.rs
```diff
@@ -534,31 +607,49 @@ impl<T: Config> Pallet<T> {
...
+		let mut w = Self::clean_open_channel_requests(config, outgoing);
 		for outgoing_para in outgoing {
 			Self::clean_hrmp_after_outgoing(outgoing_para);
+
+			// we need a few extra bits of data to weigh this -- all of this is read internally
+			// anyways, so no overhead.
+			let ingress_count =
+				<Self as Store>::HrmpIngressChannelsIndex::get(outgoing_para).len() as u32;
+			let egress_count =
+				<Self as Store>::HrmpEgressChannelsIndex::get(outgoing_para).len() as u32;
```
pepyakin, line 637: Nit: wouldn't `decode_len` be more appropriate here?

runtime/parachains/src/hrmp.rs
```diff
@@ -458,8 +518,10 @@ pub mod pallet {
 		///
 		/// If there are pending HRMP open channel requests, you can use this
 		/// function process all of those requests immediately.
-		#[pallet::weight(0)]
-		pub fn force_process_hrmp_open(origin: OriginFor<T>) -> DispatchResult {
+		///
+		/// Total number of opening channels must be provided as witness data of weighing.
+		#[pallet::weight(<T as Config>::WeightInfo::force_process_hrmp_open(*_channels))]
```
pepyakin, line 523: This appears to be not checked. Do you think this right? It also applies to other places above and below.

In any case, this looks weird for reviewers/auditors and as such required a comment.

runtime/parachains/src/configuration.rs
```diff
@@ -1005,6 +1020,7 @@ pub mod pallet {
 			new: u32,
 		) -> DispatchResult {
 			ensure_root(origin)?;
+			ensure!(new <= T::HrmpMaxOutboundChannelsBound::get(), Error::<T>::InvalidNewValue);
```
pepyakin, line 1023: ensuring the conditions in setters are now discouraged in favor of checking it centrally in `fn check_consistency`

runtime/parachains/src/hrmp/benchmarking.rs
```diff
@@ -0,0 +1,294 @@
...
+		assert_last_event::<T>(Event::<T>::ChannelClosed(sender, channel_id).into());
+	}
+
+	// NOTE: a single parachain should have the maximum number of allowed ingress and egress
+	// channels.
+	force_clean_hrmp {
+		// ingress channels to a single leaving parachain that need to be closed.
+		let i in 0 .. (<T as configuration::Config>::HrmpMaxInboundChannelsBound::get() - 1);
+		// egress channels to a single leaving parachain that need to be closed.
+		let e in 0 .. (<T as configuration::Config>::HrmpMaxOutboundChannelsBound::get() - 1);
```
pepyakin, line 173: Just out of my inexperience: why do we check it for every variable and just not once for the bound const variable?

runtime/parachains/src/hrmp/benchmarking.rs
```diff
@@ -0,0 +1,294 @@
...
+	if matches!(until, ParachainSetupStep::CloseRequested) {
+		// NOTE: this is just for expressiveness, otherwise the if-statement is 100% useless.
+		return output
+	}
+
+	output
+}
+
+const PREFIX_0: u32 = 10_000;
+const PREFIX_1: u32 = PREFIX_0 * 2;
```
pepyakin, line 122: Nit: could you leave a note what those are?

runtime/parachains/src/hrmp/benchmarking.rs
```diff
@@ -0,0 +1,294 @@
...
+
+	Hrmp::<T>::process_hrmp_open_channel_requests(&Configuration::<T>::config());
+	if matches!(until, ParachainSetupStep::Established) {
+		return output
+	}
+
+	let channel_id = HrmpChannelId { sender, recipient };
+	assert_ok!(Hrmp::<T>::hrmp_close_channel(sender_origin.clone().into(), channel_id));
+	if matches!(until, ParachainSetupStep::CloseRequested) {
+		// NOTE: this is just for expressiveness, otherwise the if-statement is 100% useless.
```
pepyakin, line 114: Nit: This function has an obvious code pattern which makes this comment superfluous. It does not harm though.

runtime/parachains/src/hrmp/benchmarking.rs
```diff
@@ -0,0 +1,294 @@
...
+
+		let para: ParaId = 1u32.into();
+		let para_origin: crate::Origin = 1u32.into();
+		register_parachain_with_balance::<T>(para, deposit);
+		T::Currency::make_free_balance_be(&para.into_account(), deposit * 256u32.into());
+
+		for ingress_para_id in 0..i {
+			// establish ingress channels to `para`.
+			let ingress_para_id = ingress_para_id + PREFIX_0;
+			let _ = establish_para_connection::<T>(ingress_para_id, 1, ParachainSetupStep::Established);
```
pepyakin, line 201: `para`?

runtime/parachains/src/hrmp/benchmarking.rs
```diff
@@ -0,0 +1,294 @@
...
+		}
+
+		// nothing should be left unprocessed.
+		assert_eq!(HrmpOpenChannelRequestsList::<T>::decode_len().unwrap_or_default(), 0);
+
+		for egress_para_id in 0..e {
+			// establish egress channels to `para`.
+			let egress_para_id = egress_para_id + PREFIX_1;
+			let _ = establish_para_connection::<T>(1, egress_para_id, ParachainSetupStep::Established);
```
pepyakin, line 210: `para` instead of 1?

runtime/parachains/src/hrmp/benchmarking.rs
```diff
@@ -0,0 +1,294 @@
...
+		}
+
+		// nothing should be left unprocessed.
+		assert_eq!(HrmpOpenChannelRequestsList::<T>::decode_len().unwrap_or_default(), 0);
+
+		// all in all, we have created this many channels.
+		assert_eq!(HrmpChannels::<T>::iter().count() as u32, i + e);
+	}: _(frame_system::Origin::<T>::Root, para, i, e) verify {
+		// all in all, all of them must be gone by now.
+		assert_eq!(HrmpChannels::<T>::iter().count() as u32, 0);
```
pepyakin, line 220: Here and for other verify:

There is a handy function `assert_storage_consistency_exhaustive` that fails in case any of the storage invariants are invalidated. Really useful to check if something is wrong. 

This is especially important since you are doing shenanigans like `process_hrmp_open_channel_requests` outside of the session change, which should be fine but still not expected use-case.

runtime/parachains/src/hrmp.rs
```diff
@@ -470,24 +532,35 @@ pub mod pallet {
...
+		#[pallet::weight(<T as Config>::WeightInfo::hrmp_cancel_open_request(*open_requests))]
 		pub fn hrmp_cancel_open_request(
 			origin: OriginFor<T>,
 			channel_id: HrmpChannelId,
+			open_requests: u32,
 		) -> DispatchResult {
 			let origin = ensure_parachain(<T as Config>::Origin::from(origin))?;
+			ensure!(
+				<Self as Store>::HrmpOpenChannelRequestsList::decode_len().unwrap_or_default()
+					as u32 == open_requests,
```
emostov, line 561:
```suggestion
					as u32 <= open_requests,
```

Is this helpful since I assume open_requests is just needs to be an upper bound

runtime/parachains/src/hrmp/benchmarking.rs
```diff
@@ -0,0 +1,294 @@
...
+	/// A channel open has been requested and accepted, and a session has passed and is now
+	/// open, and now it has been requested to close down.
+	CloseRequested,
+}
+
+fn establish_para_connection<T: Config>(
+	from: u32,
+	to: u32,
+	until: ParachainSetupStep,
+) -> [(ParaId, crate::Origin); 2]
```
pepyakin, line 65: Nit: You probably know what I am going to say about this.

pepyakin (commented):
A couple of fixes and good to go

emostov (approved):
👍

runtime/parachains/src/hrmp.rs
kianenigma: great catch.

runtime/parachains/src/hrmp.rs
kianenigma: It is not checked because it is coming from root origin and it is assumed to be correct, but I could easily add that, but IMO it is just noise. It will clutter both the dispatch code and the error variant. 

About the comment, if you look around Frame, this a pretty common patter, where some additional data needs to be passed to the dispatch, sometime not really used other than checked for correctness, just to measure the weight of some transaction ahead of time and have efficient pool processing. These are typically called "witness data".

runtime/parachains/src/hrmp/benchmarking.rs
kianenigma: (if I have understood your question correctly:) We want the benchmarking machinery to tell us the dependency on `i` and `e`, for which you need to run the benchmark in a range of values for both `i` and `e`. The final outcome of this would be `fn(i, e) -> weight`.

runtime/parachains/src/hrmp/benchmarking.rs
kianenigma: indeed.

runtime/parachains/src/hrmp/benchmarking.rs
kianenigma: `let para = establish_para_connection::<T>(ingress_para_id, 1, ParachainSetupStep::Established);`? that won't be accurate, because it returns the para id and para origin of both ends of the channel, and I rather just stick to `_` since it is unused.

runtime/parachains/src/hrmp/benchmarking.rs
kianenigma: agreed.

runtime/parachains/src/hrmp/benchmarking.rs
kianenigma: In fact I don't, enlighten me please.

runtime/polkadot/src/weights/runtime_parachains_hrmp.rs
```diff
@@ -0,0 +1,157 @@
...
+			.saturating_add((59_000 as Weight).saturating_mul(c as Weight))
+			.saturating_add(T::DbWeight::get().reads(3 as Weight))
+			.saturating_add(T::DbWeight::get().writes(3 as Weight))
+	}
+	// Storage: Hrmp HrmpOpenChannelRequestsList (r:1 w:1)
+	// Storage: Hrmp HrmpOpenChannelRequests (r:2 w:2)
+	fn clean_open_channel_requests(c: u32, ) -> Weight {
+		(0 as Weight)
+			// Standard Error: 6_000
+			.saturating_add((5_781_000 as Weight).saturating_mul(c as Weight))
```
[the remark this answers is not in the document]
kianenigma: not sure what you mean, c is used in the function and the final weight depends on it (and fwiw this is not a massive number weight-wise 😅)

runtime/parachains/src/configuration.rs
kianenigma: I settled this in https://github.com/paritytech/polkadot/pull/3876/commits/e22a79fd6afe8c955a96d0646071aea22017fc70, LMK what you think.

runtime/parachains/src/hrmp/benchmarking.rs
kianenigma: https://github.com/paritytech/polkadot/pull/3876/commits/a47ea77b84e8b54727a4b8a05d838f596b1f05df

runtime/parachains/src/hrmp/benchmarking.rs
drahnr: Use a tuple instead :)

runtime/parachains/src/configuration.rs
hirschenberger: So, I will also change this in Ump? #3889

runtime/parachains/src/hrmp/benchmarking.rs
kianenigma: What is the rationale? an array of two is perfectly fine in my opinion and practically synonym to a tuple of two in my head.

runtime/parachains/src/hrmp/benchmarking.rs
kianenigma: NOTE: what re-using this (useful) function from the tests will restrict us to only call it in `verify` block of when we are testing benchmarks, which I think is enough, because the tests will run on the (low, mid, high of all benchmarks). Nonetheless, I made it reusable in https://github.com/paritytech/polkadot/pull/3876/commits/745e146f7df89ce7f9c738003b71a1c2c531057e, which we can decide to keep or revert.

runtime/parachains/src/hrmp/benchmarking.rs
pepyakin: Sorry, I assumed too much.

While I think an array of two is definitely uncommon in such cases, and by default I would expect a tuple in cases like this. However, I agree it's more or less synonymous (It seems a bit novel and I even like it a bit compared to a tuple, because you use different brackets for the outer and inner items)

I, however, meant to propose to use a struct instead of the array. The inner tuples are OK since they are at least typed. That has an advantage that you fix names in one place — at the declaration side compared to the each place (which promotes consistency). Names are a bit better in this setting compared to indices: 

What's more clear to refer the recipient, a field named `recipient` or an element at index `1`?

runtime/parachains/src/hrmp/benchmarking.rs
pepyakin: But why do we want us to tell the dependency on those? `i` and `e` are constants anyway?

runtime/parachains/src/configuration.rs
pepyakin: i missed your comment... ༼ ༎ຶ ෴ ༎ຶ༽ 

But that seems to be the same cases

runtime/polkadot/src/weights/runtime_parachains_hrmp.rs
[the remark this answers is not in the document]
pepyakin: >  fwiw this is not a massive number weight-wise

yeah, that's what I meant. Ok then

runtime/parachains/src/hrmp/benchmarking.rs
pepyakin: That makes perfect sense and in fact I also considered the same.

runtime/parachains/src/configuration.rs
```diff
@@ -319,6 +319,10 @@ pub enum InconsistentError<BlockNumber> {
 	},
 	/// `validation_upgrade_delay` is less than or equal 1.
 	ValidationUpgradeDelayIsTooLow { validation_upgrade_delay: BlockNumber },
+	/// Maximum number of HRMP outbound channels exceeded.
+	MaxHrmpOutboundChannelsExceeded,
+	/// Maximum number of HRMP inbound channels exceeded.
+	MaxHrmpInboundChannelsExceeded,
```
pepyakin, line 325: Nit: mention the value for consistency with other similar variants?

pepyakin (approved):

runtime/parachains/src/hrmp/benchmarking.rs
kianenigma: their upper bound is constant and we happen to use that here as well, If you see where `T::WeightInfo::force_clean_hrmp` is being called, these two parameters need to be provided (similar to the other comment, it is a root call and we expect root to give a proper estimate of the witness data)

kianenigma:
bot merge

runtime/parachains/src/paras.rs
pepyakin: As promised: https://github.com/paritytech/polkadot/pull/4934

PR #3878, opened 2021-09-17
Add benchmarking for parachain runtime `dmp` pallet

hirschenberger:
part of #3850


rphmeier:
Thanks! This needs labels.

KiChjang:
@hirschenberger Since the `Dmp` pallet does not contain any dispatch functions, we don't actually need any benchmarking here. I'll redirect you to the next pallets in the list in #3850 if you want to work on the benchmarks. Thanks!

PR #3880, opened 2021-09-17
Docker files chore

alvicsam:
Renamed docker files with unified naming convention, adjusted .gitlab-ci.yml and documentation.
Moved docker folder to script/docker/polkadot, adjusted scripts and documentation.

Closes https://github.com/paritytech/ci_cd/issues/197


TriplEight (changes requested):
- Please add the metadata to https://github.com/paritytech/polkadot/blob/1c0c709a92de263dcf7edadc7c96f2eb847b3c38/scripts/docker/polkadot/manual.Dockerfile and mention a scripts/docker/polkadot/README.md there.
  Metadata should be quite similar to the same in `ci.` and `release.` Dockerfiles, just mention the type of the image in the description and have different links.
- please decide something stable with the naming convention (this is a first time we apply it, and it's not yet really written). So, where to put `ci/manual/release/debug` before the name or not (`collator_ci` vs `ci_collator`), what to use in which case - underscore or hyphen (`ci_debug` vs `staking**_**miner**-**ci`). And it would be a good help to write this naming so it would go to the fresh policy document.

chevdor (changes requested):
I don't think the naming of image is appropriate.
For instance an image called `ci-debug` could belong to any project and do anything...
I commented out the naming in the issue that initiated this PR.

TriplEight:
also commented to the initial issue

scripts/docker/staking-miner/staking_miner_injected.Dockerfile
```diff
@@ -9,7 +9,7 @@ LABEL io.parity.image.authors="devops-team@parity.io" \
 	io.parity.image.vendor="Parity Technologies" \
 	io.parity.image.title="${IMAGE_NAME}" \
 	io.parity.image.description="staking-miner for substrate based chains" \
-	io.parity.image.source="https://github.com/paritytech/polkadot/blob/${VCS_REF}/scripts/docker/staking-miner/staking_miner-injected.Dockerfile" \
+	io.parity.image.source="https://github.com/paritytech/polkadot/blob/${VCS_REF}/scripts/docker/staking-miner/staking_miner-ci.Dockerfile" \
```
chevdor, line 12:
```suggestion
	io.parity.image.source="https://github.com/paritytech/polkadot/blob/${VCS_REF}/scripts/docker/staking-miner/staking_miner_injected.Dockerfile" \
```

chevdor (changes requested):
minor issue with one image name

.github/workflows/publish-docker-release.yml
```diff
@@ -29,7 +29,7 @@ jobs:
         uses: docker/build-push-action@v2
         with:
           push: true
-          file: scripts/docker/release.Dockerfile
+          file: scripts/docker/polkadot_builder_docker.Dockerfile
```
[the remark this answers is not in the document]
chevdor: I am not sure about the `_docker` part... A dockerfile is obviously a docker thing.
IMO that should be: `polkadot_injected_release.Dockerfile`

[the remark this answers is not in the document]
alvicsam: Here I was a little bit confused with naming because from what I understood this file is needed for creating a docker image and publishing it to docker hub. That's why I added `_docker` here as an indicator.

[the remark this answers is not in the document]
chevdor: IMO the name of the file should describe what the file **IS** and not what it is used for.
For instance, one Dockerfile maybe used in CI and no longer tomorrow (or the other way around) so `ci` does not have a place in the name of an image. So the main question is **what is the image about**.

I would not even mention **apt** in the name as this is an implementation detail. Today we may use apt and tomorrow we will do something else (potentially...).

NOTE: If some images are very similar, they may be generated from the same dockerfile, published with the same name but use another TAG. We may consider this for instance for some debug images with more flags.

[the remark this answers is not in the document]
alvicsam: Sounds good. In this case I'll add info about docker into image description.

chevdor:
Sorry for the nits but I think this PR is the place set those.

## Multiple files vs ARG
Do we really want to have `polkadot_injected_release.Dockerfile` and `polkadot_injected_debug.Dockerfile` or could we manage with a single file and an ARG to describe this `debug`/`release` ? If we really have 2 files for polkadot, we probably also will need 2 for the collatator, staking-miner, etc... and that will explode the number of files.

## Naming

I think we still have a mix in the names. For the `staking-miner` for instance, it should be everywhere (file names, doc, etc...) `staking-miner_foobar` and not `staking_miner_foobar`.

TriplEight:
@chevdor this all sounds good, thank you for your help.
I'd like to add that i.e. the `collator_injected.Dockerfile` and `polkadot_injected_debug.Dockerfile` are identical and the difference is only the binary name, we could definitely put it into `ARG`.
The question here is where to draw the line. We could
- put a binary name into `ARG` and remove every second Dockerfile (i.e. polkadot/collator/staking_miner ones)
- put a conditional script into `ARG` and tell it to run an "injection" logic or "a docker builder pattern". This would leave us with just a couple of Dockerfiles and a script, so no need to maintain pairs of similar files. From this point it becomes increasingly complicated and less readable.
-  go all-in into templating Dockerfiles and even end up removing all of them in favour of some templating language

I suggest we take the first approach and see if it's good enough.

scripts/docker/polkadot/polkadot_builder.Dockerfile
```diff
@@ -0,0 +1,40 @@
+FROM docker.io/paritytech/ci-linux:production as builder
+LABEL io.parity.image.authors="devops-team@parity.io" \
+	io.parity.image.vendor="Parity Technologies" \
+	io.parity.image.title="builder" \
+	io.parity.image.description="This is the build stage for Polkadot. Here we create the binary." \
+	io.parity.image.source="https://github.com/paritytech/polkadot/blob/master/scripts/docker/polkadot/polkadot_builder.Dockerfile" \
+	io.parity.image.documentation="https://github.com/paritytech/polkadot/scripts/docker/polkadot/README.md"
+
+ARG PROFILE=release
```
chevdor, line 9: Not a blocker but I ended up removing this ARG in other images as I don't think it makes much sense to build anything but a release version.

scripts/docker/polkadot/polkadot_builder.Dockerfile
```diff
@@ -0,0 +1,40 @@
...
+WORKDIR /polkadot
+
+COPY . /polkadot
+
+RUN cargo build --$PROFILE
```
chevdor, line 14:
```suggestion
RUN cargo build --$PROFILE --locked
```

scripts/docker/polkadot/polkadot_builder.Dockerfile
```diff
@@ -0,0 +1,40 @@
...
+
+RUN useradd -m -u 1000 -U -s /bin/sh -d /polkadot polkadot && \
+	mkdir -p /polkadot/.local/share && \
+	mkdir /data && \
+	chown -R polkadot:polkadot /data && \
+	ln -s /data /polkadot/.local/share/polkadot && \
+	rm -rf /usr/bin /usr/sbin
+
+USER polkadot
+EXPOSE 30333 9933 9944
```
chevdor, line 37:
```suggestion
EXPOSE 30333 9933 9944 9615
```

scripts/docker/polkadot/polkadot_builder.Dockerfile
```diff
@@ -0,0 +1,40 @@
...
+VOLUME ["/data"]
+
+CMD ["/usr/local/bin/polkadot"]
```
chevdor, line 40: I suggest aligning the way we use this image with how the injected image works.

```suggestion
ENTRYPOINT ["/usr/local/bin/polkadot"]
```

scripts/docker/polkadot/polkadot_builder.Dockerfile
```diff
@@ -0,0 +1,40 @@
...
+
+RUN cargo build --$PROFILE
+
+# ===== SECOND STAGE ======
+
+FROM docker.io/library/ubuntu:20.04
+LABEL io.parity.image.authors="devops-team@parity.io" \
+	io.parity.image.vendor="Parity Technologies" \
+	io.parity.image.title="builder" \
+	io.parity.image.description="This is the 2nd stage: a very small image where we copy the Polkadot binary." \
```
chevdor, line 22: The part:
```
This is the 2nd stage: a very small image where we copy the Polkadot binary.
```
can be kept as comment but this does not really describe the image.
```suggestion
	io.parity.image.description="Polkadot: a platform for web3. This is a self-buit multistage image." \
```

scripts/docker/collator_injected.Dockerfile
```diff
@@ -10,7 +10,7 @@ LABEL io.parity.image.authors="devops-team@parity.io" \
 	io.parity.image.vendor="Parity Technologies" \
 	io.parity.image.title="${IMAGE_NAME}" \
 	io.parity.image.description="adder-collator image" \
```
chevdor, line 12:
```suggestion
	io.parity.image.description="Injected adder-collator Docker image" \
```

scripts/docker/polkadot/polkadot_builder.Dockerfile
```diff
@@ -0,0 +1,40 @@
+FROM docker.io/paritytech/ci-linux:production as builder
+LABEL io.parity.image.authors="devops-team@parity.io" \
+	io.parity.image.vendor="Parity Technologies" \
+	io.parity.image.title="builder" \
+	io.parity.image.description="This is the build stage for Polkadot. Here we create the binary." \
```
chevdor, line 5: I don't think the builder images need all those labels. Those images are transient and will never be published unlike the second stage. We probably can cut that down to keep the image more readable.
@TriplEight any take on that?

scripts/docker/polkadot_injected_release.Dockerfile
```diff
@@ -4,12 +4,13 @@ FROM docker.io/library/ubuntu:20.04
...
 ARG BUILD_DATE
 ARG POLKADOT_VERSION
+ARG POLKADOT_GPGKEY=9D4B2B6EB8F97156D19669A9FF0812D491B96798
 
 LABEL io.parity.image.authors="devops-team@parity.io" \
 	io.parity.image.vendor="Parity Technologies" \
 	io.parity.image.title="parity/polkadot" \
-	io.parity.image.description="Polkadot: a platform for web3" \
-	io.parity.image.source="https://github.com/paritytech/polkadot/blob/${VCS_REF}/scripts/docker/Dockerfile" \
+	io.parity.image.description="Publish Polkadot Docker image for new releases" \
```
chevdor, line 12:
```suggestion
	io.parity.image.description="Polkadot: a platform for web3. This is the official Parity image with an injected binary." \
```

scripts/docker/polkadot_injected_release.Dockerfile
```diff
@@ -22,12 +23,11 @@ RUN apt-get update && \
 	DEBIAN_FRONTEND=noninteractive apt-get install -y --no-install-recommends \
 		libssl1.1 \
 		ca-certificates \
-		curl \
 		gnupg && \
 	useradd -m -u 1000 -U -s /bin/sh -d /polkadot polkadot && \
 # add repo's gpg keys and install the published polkadot binary
-	gpg --recv-keys --keyserver hkps://keys.mailvelope.com 9D4B2B6EB8F97156D19669A9FF0812D491B96798 && \
-	gpg --export 9D4B2B6EB8F97156D19669A9FF0812D491B96798 > /usr/share/keyrings/parity.gpg && \
+	gpg --recv-keys --keyserver hkps://keys.mailvelope.com ${POLKADOT_GPGKEY} && \
```
chevdor, line 29: I think it could be useful to extract an ARG for the key server. Keyservers do go down and if the one we use here does go down, we will no longer be able to build the image. With an ARG, we can use the same default but that allows an external override (from CI for instance) without having to touch the image.

scripts/docker/staking-miner/staking-miner_builder.Dockerfile
```diff
@@ -20,7 +20,7 @@ LABEL io.parity.image.authors="devops-team@parity.io" \
 	io.parity.image.vendor="Parity Technologies" \
 	io.parity.image.title="${IMAGE_NAME}" \
 	io.parity.image.description="staking-miner for substrate based chains" \
```
chevdor, line 22:
```suggestion
	io.parity.image.description="${IMAGE_NAME} for substrate based chains" \
```

scripts/docker/staking-miner/staking-miner_builder.Dockerfile
```diff
@@ -20,7 +20,7 @@ LABEL io.parity.image.authors="devops-team@parity.io" \
...
-	io.parity.image.source="https://github.com/paritytech/polkadot/blob/${VCS_REF}/scripts/docker/staking-miner/staking_miner-builder.Dockerfile" \
+	io.parity.image.source="https://github.com/paritytech/polkadot/blob/${VCS_REF}/scripts/docker/staking-miner/staking-miner_builder.Dockerfile" \
```
chevdor, line 23:
```suggestion
	io.parity.image.source="https://github.com/paritytech/polkadot/blob/${VCS_REF}/scripts/docker/${IMAGE_NAME}/${IMAGE_NAME}_builder.Dockerfile" \
```

scripts/docker/staking-miner/staking-miner_injected.Dockerfile
```diff
@@ -9,7 +9,7 @@ LABEL io.parity.image.authors="devops-team@parity.io" \
 	io.parity.image.vendor="Parity Technologies" \
 	io.parity.image.title="${IMAGE_NAME}" \
 	io.parity.image.description="staking-miner for substrate based chains" \
```
chevdor, line 11:
```suggestion
	io.parity.image.description="${IMAGE_NAME} for substrate based chains" \
```

scripts/docker/staking-miner/staking-miner_injected.Dockerfile
```diff
@@ -9,7 +9,7 @@ LABEL io.parity.image.authors="devops-team@parity.io" \
...
-	io.parity.image.source="https://github.com/paritytech/polkadot/blob/${VCS_REF}/scripts/docker/staking-miner/staking_miner-injected.Dockerfile" \
+	io.parity.image.source="https://github.com/paritytech/polkadot/blob/${VCS_REF}/scripts/docker/staking-miner/staking-miner_injected.Dockerfile" \
```
chevdor, line 12:
```suggestion
	io.parity.image.source="https://github.com/paritytech/polkadot/blob/${VCS_REF}/scripts/docker/${IMAGE_NAME}/${IMAGE_NAME}_injected.Dockerfile" \
```

chevdor (changes requested):
I made quite a bunch of suggestion to help with the maintenance of those images in the future.

chevdor (approved):
LGTM, thanks for the big work !

scripts/docker/polkadot/polkadot_builder.Dockerfile
TriplEight: Yeah, totally, I think it was there more for explicity.

scripts/docker/polkadot/polkadot_builder.Dockerfile
TriplEight: The idea behind it was a policy-as-a-code in future that would check that every Dockerfile has metadata. But good point on readability, makes sense to make an exclusion for the files with the specific `*builder*` in the name for such policy.

TriplEight: It's OK to have simple explanation comments instead of the full metadata in such files.

TriplEight (changes requested):

scripts/docker/polkadot/polkadot_builder.Dockerfile
chevdor: I initially used that when I made the first images thinking `debug` image could be useful but that never happened to be really the case.

TriplEight (approved):

TriplEight:
now you should wait until https://github.com/paritytech/polkadot/pull/3807 gets merged

chevdor:
As a side note (I can open another issue if you prefer), it would be super convenient to have an injected docker image where the binary is built with all features. That would help with `try-runtime` for instance.

TriplEight:
@chevdor in another PR please

chevdor:
New issue about a new image with `all-features`: #3925

.gitlab-ci.yml
```diff
@@ -210,7 +210,7 @@ test-build-linux-stable:
     - echo "Polkadot version = ${VERSION} (EXTRATAG = ${EXTRATAG})"
     - echo -n ${VERSION} > ./artifacts/VERSION
     - echo -n ${EXTRATAG} > ./artifacts/EXTRATAG
-    - cp -r scripts/docker/* ./artifacts
+    - cp scripts/docker/* ./artifacts
```
chevdor, line 213: Are you sure about that one ?
I see we have already the staking miner images under [its own folder](https://github.com/paritytech/polkadot/tree/master/scripts/docker/staking-miner).

Since we expect to have at least 2 images (injected + builder) pro binary, we will likely have only folders under `scripts/docker` so the `-r` will become useful.

alvicsam: >Are you sure about that one ?
Yes and no :)
So currently in .gitlab-ci.yml we use only files in scripts/docker/ folder (excluding subfolders). I wasn't sure about staking-miner but I couldn't find any steps related to it. But I don't know about future intensions.
Also need @TriplEight opinion.

chevdor: The staking-miner is rather new, so are its images. But those should be built/published as well just like polkadot:
- simple test build of the 'builder' image
- build + push on release for the 'injected' image

PR #3884, opened 2021-09-17
chore: update `builder` image

chevdor:
This PR refreshes the multistage `builder` Docker image and aligns it with the `injected` parity image.

⚠️ BREAKING: This image introduces a breaking change to align with the `injected` image behavior: `polkadot` is now the entrypoint. So the image is no longer used as `docker run -it <image> polkadot --version` but as `docker run -it <image> --version` (the binary `polkadot` should not longer be provided).

- [x] add metadata
- [x] use ubuntu:20.04 as base
- [x] add label `io.parity.image.type="builder"`
- [x] remove labels not relevant to images not built via CI
- [x] align the entrypoint with the `injected` image
- [x] align the volumes with the `injected` image
- [x] fix exposed ports
- [x] rename to a more specific dockerfile name

fix #9765


docker/polkadot_builder.Dockerfile
```diff
@@ -0,0 +1,42 @@
...
+# This is the 2nd stage: a very small image where we copy the Polkadot binary."
+FROM docker.io/library/ubuntu:20.04
+
+LABEL description="Multistage Docker image for Polkadot: a platform for web3" \
+    io.parity.image.type="builder" \
+    io.parity.image.authors="chevdor@gmail.com, devops-team@parity.io" \
+	io.parity.image.vendor="Parity Technologies" \
+	io.parity.image.description="Polkadot: a platform for web3" \
+	io.parity.image.source="https://github.com/paritytech/polkadot/blob/${VCS_REF}/docker/Dockerfile" \
+	io.parity.image.documentation="https://github.com/paritytech/polkadot/"
```
TriplEight, line 22: different padding

docker/docker-compose-local.yml
```diff
@@ -7,6 +7,7 @@ services:
       - "30333:30333"
       - "9933:9933"
       - "9944:9944"
+      - "9615:9615"
     image: chevdor/polkadot:latest
```
TriplEight, line 11: would be great to refer to the latest official release here

docker/docker-compose-local.yml
```diff
@@ -22,6 +23,7 @@ services:
       - "30344:30344"
       - "9935:9935"
       - "9945:9945"
+      - "9615:9615"
     image: chevdor/polkadot:latest
```
TriplEight, line 27: would be great to refer to the latest official release here

TriplEight (changes requested):
CC another container chore at the same time https://github.com/paritytech/polkadot/pull/3880

docker/polkadot_builder.Dockerfile
chevdor: fixed

docker/docker-compose-local.yml
chevdor: fixed

docker/docker-compose-local.yml
chevdor: fixed

scripts/dockerfiles/polkadot/docker-compose-local.yml
```diff
@@ -1,35 +1,39 @@
...
+      - "9615:9615"
+    image: parity/polkadot:latest
     volumes:
       - "polkadot-data-alice:/data"
-    command: polkadot --chain=polkadot-local --alice -d /data --node-key 0000000000000000000000000000000000000000000000000000000000000001
+    command: |
+      --chain=polkadot-local
+      --alice
+      -d /data
+      --node-key 0000000000000000000000000000000000000000000000000000000000000001
```
TriplEight, line 16: does it assume the binary?

scripts/dockerfiles/polkadot/docker-compose-local.yml
```diff
@@ -1,35 +1,39 @@
...
     volumes:
       - "polkadot-data-bob:/data"
     links:
       - "node_alice:alice"
-    command: polkadot --chain=polkadot-local --bob -d /data --port 30344 --rpc-port 9935 --ws-port 9945 --bootnodes '/ip4/172.28.1.1/tcp/30333/p2p/QmRpheLN4JWdAnY7HGJfWFNbfkQCb6tFf4vvA6hgjMZKrR'
+    command: |
+      --chain=polkadot-local
+      --bob
+      -d /data
+      --bootnodes '/ip4/172.28.1.1/tcp/30333/p2p/QmRpheLN4JWdAnY7HGJfWFNbfkQCb6tFf4vvA6hgjMZKrR'
```
TriplEight, line 36: does it assume the binary?

scripts/dockerfiles/polkadot/docker-compose.yml
```diff
@@ -1,16 +1,23 @@
...
-    image: chevdor/polkadot:latest
+    image: parity/polkadot:latest
     volumes:
       - "polkadot-data:/data"
-    command: polkadot
+    command: |
+      --unsafe-rpc-external
+      --unsafe-ws-external
+      --rpc-cors all
+      --prometheus-external
```
TriplEight, line 14: does it assume the binary?

scripts/dockerfiles/polkadot/polkadot_builder.Dockerfile
```diff
@@ -1,27 +1,35 @@
...
+	mkdir -p /data /polkadot/.local/share && \
 	chown -R polkadot:polkadot /data && \
-	ln -s /data /polkadot/.local/share/polkadot && \
-	rm -rf /usr/bin /usr/sbin
+	ln -s /data /polkadot/.local/share/polkadot
 
+# check if executable works in this container
+RUN /usr/local/bin/polkadot --version
+
+RUN	rm -rf /usr/bin /usr/sbin
```
TriplEight, line 30: why do you want to remove `/usr/bin` from the image that's intended to be used in the local experiments?

TriplEight (changes requested):

scripts/dockerfiles/polkadot/polkadot_builder.Dockerfile
TriplEight: also padding

scripts/dockerfiles/polkadot/docker-compose-local.yml
chevdor: there is an [entrypoint](https://github.com/paritytech/polkadot/blob/be7dbb2c036d0ba6a979ac22f6b4b3f352ad9fb3/scripts/dockerfiles/polkadot/polkadot_builder.Dockerfile#L36)

scripts/dockerfiles/polkadot/docker-compose-local.yml
chevdor: there is an [entrypoint](https://github.com/paritytech/polkadot/blob/be7dbb2c036d0ba6a979ac22f6b4b3f352ad9fb3/scripts/dockerfiles/polkadot/polkadot_builder.Dockerfile#L36)

scripts/dockerfiles/polkadot/polkadot_builder.Dockerfile
chevdor: I had this removal initially and decided to keep it but I did consider removing (.. the rm).

> from the image that's intended to be used in the local experiments?

Who says that? Users may use the image for `subkey` for instance.

Here are the pros/cons or the removal:

## pros
- safer
- (slightly) smaller image
- the image is made to use the binaries we provide, not to fiddle around with bash, curl & Co.
- safe by default but users can comment out the line and build an image where those are not removed

## cons
- cannot bash into the container
- may surprise some users
- admitedly, this container is not running 24/7 and will be short lived so we may not need the extra safety

As conclusion, I think it is better to go safer by default and avoid making assumptions about **how** users will use the image. Still users can easily get more freedom by commenting out the line.

I would be fine removing this `rm` if you can provide a sample use case where we'd need something in `/usr/bin` while using this image.

TriplEight: Wait, I thought `builder_*.Dockerfile` are designed for the developers to build a binary in a reproducible env to play with it in the first place. To not have to setup the exact Rust versions and other deps. Not to run as a validator, etc.
I realize that actually we can't tell or control its' usage, it's just the main usecase is dependent on how we advertise it.

chevdor: > are designed for the developers to build a binary in a reproducible env to play with it in the first place.

I definitely don't think this the only use case. I also see the case of validators who want to build their own binary + image themselves.

TriplEight: OK, sounds good to me. Then let's just do the following.

scripts/dockerfiles/polkadot/docker-compose-local.yml
TriplEight: ah missed the line

scripts/dockerfiles/polkadot/polkadot_builder.Dockerfile
```diff
@@ -1,27 +1,35 @@
...
+	mkdir -p /data /polkadot/.local/share && \
 	chown -R polkadot:polkadot /data && \
-	ln -s /data /polkadot/.local/share/polkadot && \
-	rm -rf /usr/bin /usr/sbin
+	ln -s /data /polkadot/.local/share/polkadot
 
+# check if executable works in this container
+RUN /usr/local/bin/polkadot --version
+
+RUN rm -rf /usr/bin /usr/sbin
```
TriplEight, line 30:
```suggestion
	ln -s /data /polkadot/.local/share/polkadot && \
# unclutter and minimize the attack surface
	rm -rf /usr/bin /usr/sbin && \
# check if executable works in this container
	/usr/local/bin/polkadot --version

```
Less layers, better test

TriplEight (changes requested):

TriplEight (approved):

PR #3888, opened 2021-09-17
Add benchmarking for parachain runtime paras pallet

KiChjang:
Part of #3850.


runtime/parachains/src/paras/benchmarking.rs
```diff
@@ -0,0 +1,82 @@
...
+	let events = frame_system::Pallet::<T>::events();
+	let system_event: <T as frame_system::Config>::Event = generic_event.into();
+	// compare to the last event record
+	let frame_system::EventRecord { event, .. } = &events[events.len() - 1];
+	assert_eq!(event, &system_event);
+}
+
+benchmarks! {
+	force_set_current_code {
+		let c in 1024 .. 2048;
```
shawntabrizi, line 34: I think the range should be in MB not KB.

And more specifically, the top of the range should come from the MAX_CODE_SIZE constant

runtime/parachains/src/paras/benchmarking.rs
```diff
@@ -0,0 +1,82 @@
...
+		let new_code = ValidationCode(vec![0; c as usize]);
+		let para_id = ParaId::from(c as u32);
+	}: _(RawOrigin::Root, para_id, new_code)
+	verify {
+		assert_last_event::<T>(Event::CurrentCodeUpdated(para_id).into());
+	}
+	force_set_current_head {
+		let new_head = HeadData(vec![0]);
```
shawntabrizi, line 42: Should be using `MAX_HEAD_SIZE` const here

runtime/parachains/src/paras/benchmarking.rs
shawntabrizi: This benchmark is also missing complexity with the binary search and insert in: `note_past_code`

There is a vec, which probably we should start by bounding, and then filling before the benchmark is executed.

runtime/westend/src/weights/runtime_parachains_paras.rs
```diff
@@ -0,0 +1,102 @@
...
+pub struct WeightInfo<T>(PhantomData<T>);
+impl<T: frame_system::Config> runtime_parachains::paras::WeightInfo for WeightInfo<T> {
+	// Storage: Paras CurrentCodeHash (r:1 w:1)
+	// Storage: Paras CodeByHashRefs (r:1 w:1)
+	// Storage: Paras PastCodeMeta (r:1 w:1)
+	// Storage: Paras PastCodePruning (r:1 w:1)
+	// Storage: Paras PastCodeHash (r:0 w:1)
+	// Storage: Paras CodeByHash (r:0 w:1)
+	fn force_set_current_code(c: u32, ) -> Weight {
+		(0 as Weight)
```
KiChjang, line 54: Ok, I've no idea what's going on here, this weight is worsened, but `force_note_new_head` is clearly doing better.

runtime/parachains/src/paras/benchmarking.rs
```diff
@@ -29,17 +29,31 @@ fn assert_last_event<T: Config>(generic_event: <T as Config>::Event) {
...
 		let new_code = ValidationCode(vec![0; c as usize]);
 		let para_id = ParaId::from(c as u32);
+		generate_disordered_pruning::<T>();
 	}: _(RawOrigin::Root, para_id, new_code)
 	verify {
 		assert_last_event::<T>(Event::CurrentCodeUpdated(para_id).into());
 	}
 	force_set_current_head {
-		let new_head = HeadData(vec![0]);
+		let max_head_data_size = configuration::Pallet::<T>::config().max_head_data_size as usize;
```
shawntabrizi, line 55: Not sure this is going to work because this is a storage item, and can be modified on chain.

We need something like a configurable constant which also sets the upper limit for what this storage value can be.

Does that make sense?

runtime/parachains/src/paras/benchmarking.rs
```diff
@@ -0,0 +1,96 @@
...
+	let system_event: <T as frame_system::Config>::Event = generic_event.into();
+	// compare to the last event record
+	let frame_system::EventRecord { event, .. } = &events[events.len() - 1];
+	assert_eq!(event, &system_event);
+}
+
+fn generate_disordered_pruning<T: Config>() {
+	let mut needs_pruning = Vec::new();
+
+	for i in 0..1000 {
```
thiolliere, line 35: the 1000 is pretty arbirary maybe better to put it in a const with a good comment on its value, I think 1000 is fine but still better to know why.

runtime/parachains/src/paras/benchmarking.rs
```diff
@@ -0,0 +1,96 @@
...
+	}: _(RawOrigin::Root, para_id, new_head)
+	verify {
+		assert_last_event::<T>(Event::CurrentHeadUpdated(para_id).into());
+	}
+	force_schedule_code_upgrade {
+		let c in 1 .. MAX_CODE_SIZE;
+		let new_code = ValidationCode(vec![0; c as usize]);
+		let para_id = ParaId::from(c as u32);
+		let block = T::BlockNumber::from(c);
+	}: _(RawOrigin::Root, para_id, new_code, block)
```
thiolliere, line 66: this does a binary search in UpcomingUpgrades UpgradeCooldowns, and they don't seem populated, they are bounded by the number of parachain and their cost can be negligeable though.

runtime/parachains/src/paras/benchmarking.rs
```diff
@@ -0,0 +1,96 @@
...
+		let config = HostConfiguration::<T::BlockNumber>::default();
+		generate_disordered_pruning::<T>();
+		Pallet::<T>::schedule_code_upgrade(para_id, ValidationCode(vec![0]), expired, &config);
+	}: _(RawOrigin::Root, para_id, new_head)
+	verify {
+		assert_last_event::<T>(Event::NewHeadNoted(para_id).into());
+	}
+	force_queue_action {
+		let para_id = ParaId::from(1000);
+	}: _(RawOrigin::Root, para_id)
```
thiolliere, line 85: Note: this do a binary search in ActionsQueue, which is not populated in this benchmark.

thiolliere (commented):
some benchmark seems not completly accurate, maybe it is negligible or it should be addressed.

runtime/parachains/src/paras/benchmarking.rs
KiChjang: I should probably make it 1024 since the time complexity for a binary search is O(log(2, n)).

runtime/parachains/src/paras/benchmarking.rs
```diff
@@ -0,0 +1,133 @@
...
+use super::*;
+use crate::{configuration::HostConfiguration, shared};
+use frame_benchmarking::{benchmarks, impl_benchmark_test_suite};
+use frame_system::RawOrigin;
+use primitives::v1::{HeadData, Id as ParaId, ValidationCode, MAX_CODE_SIZE, MAX_HEAD_DATA_SIZE};
+use sp_runtime::traits::{One, Saturating};
+
+// 2 ^ 10, because binary search time complexity is O(log(2, n)) and n = 1024 gives us a
+// big and round number.
+const SAMPLE_SIZE: u32 = 1024;
```
thiolliere, line 26: maybe we should say that due to the limited number of parachains, the number of pruning, upcoming upgrades and cooldowns shouldn't exceed this number ?

runtime/westend/src/weights/runtime_parachains_paras.rs
thiolliere: I think it is fine, if the length of code is 0 it shouldn't cost that much more than the storage read and writes I think no ?

thiolliere (approved):
this nit: https://github.com/paritytech/polkadot/pull/3888/files#r712791711

But looks good to me

shawntabrizi (approved):

KiChjang:
bot merge

primitives/src/v1/mod.rs
```diff
@@ -237,6 +237,13 @@ pub const ASSIGNMENT_KEY_TYPE_ID: KeyTypeId = KeyTypeId(*b"asgn");
 /// * when detecting a code decompression bomb in the client
 pub const MAX_CODE_SIZE: u32 = 3 * 1024 * 1024;
 
+/// Maximum head data size we support right now.
+///
+/// Used for:
+/// * initial genesis for the Parachains configuration
+/// * checking updates to this stored runtime configuration do not exceed this limit
+pub const MAX_HEAD_DATA_SIZE: u32 = 1 * 1024 * 1024;
```
emostov, line 245: How was this value chosen?

KiChjang: From places like this: https://github.com/paritytech/polkadot/blob/e9c4bc1eb1908829322337def9bbe9f7d5c4beba/runtime/common/src/integration_tests.rs#L241

runtime/parachains/src/paras/benchmarking.rs
```diff
@@ -0,0 +1,135 @@
...
+		let config = HostConfiguration::<T::BlockNumber>::default();
+		generate_disordered_pruning::<T>();
+		Pallet::<T>::schedule_code_upgrade(para_id, ValidationCode(vec![0]), expired, &config);
+	}: _(RawOrigin::Root, para_id, new_head)
+	verify {
+		assert_last_event::<T>(Event::NewHeadNoted(para_id).into());
+	}
+	force_queue_action {
+		let para_id = ParaId::from(1000);
+		generate_disordered_actions_queue::<T>();
```
emostov, line 123: `generate_disordered_actions_queue` fills up the actions queue with para id 1000, meaning a binary search for `1000` would be O(1) right? Would it be better to fill up the actions queue with anything but the id we are adding so we have to do a full binary search?

PR #3889, opened 2021-09-18
Add benchmarking for parachain runtime `ump` pallet

hirschenberger:
Part of #3850

@KiChjang Can you add me to the tracking list?


runtime/parachains/src/ump.rs
```diff
@@ -188,6 +212,9 @@ pub mod pallet {
 
 		/// Origin which is allowed to execute overweight messages.
 		type ExecuteOverweightOrigin: EnsureOrigin<Self::Origin>;
+
+		/// Weight information for extrinsics in this pallet
```
kianenigma, line 216:
```suggestion
		/// Weight information for extrinsics in this pallet.
```

runtime/parachains/src/ump/benchmarking.rs
```diff
@@ -0,0 +1,105 @@
...
+		origin_type: OriginKind::SovereignAccount,
+		require_weight_at_most: weight,
+		call: call.encode().into(),
+	}]))
+	.encode()
+}
+
+frame_benchmarking::benchmarks! {
+	process_upward_message {
+		let s in 0..configuration::ActiveConfig::<T>::get().max_upward_message_size;
```
kianenigma, line 65: this is basically using the value of the `ActiveConfig` in genesis, not the real on on-chain.

I feel like coding it like this is actually misleading. A better approach would be to hardcode some maximum value for it, and ensuring that when setting it, we don't go over it. See an example here: https://github.com/paritytech/polkadot/pull/3876/files

runtime/parachains/src/ump/benchmarking.rs
```diff
@@ -0,0 +1,105 @@
...
+		let para = ParaId::from(1978);
+		let host_conf = configuration::ActiveConfig::<T>::get();
+		let weight = host_conf.ump_max_individual_weight + 1;
+		let data = create_message::<T>(weight, s);
+	}: {
+		assert!(T::UmpSink::process_upward_message(para, &data[..], Weight::MAX).is_ok());
+	}
+
+	service_overweight {
+		let host_conf = configuration::ActiveConfig::<T>::get();
```
kianenigma, line 75: same

kianenigma (commented):
a minor note, otherwise LGTM

runtime/parachains/src/ump/benchmarking.rs
hirschenberger: I see, but how can I determine this upper bound? 
Isn't it a delicate point bounding on-chain values, which in principle stand under the control of democracy.

kianenigma: Some reasonable defaults should be enough, and no if some upper bounds exists even the governance should not (easily) be able to set higher values. @pepyakin might be able to chime in here.

hirschenberger: I meant it the other way around. Governance  should not be **limited** by bounds set in code by a single developer.

But I'm still a noob and have not so much insight in the delicate gears that make a chain safe and governance truly  democratic. It was just a feeling and an open question.

kianenigma: > I meant it the other way around. Governance should not be limited by bounds set in code by a single developer.

The governance is not and will not be bound at the end of the day by any of these because by the same token that they can set this parameter, they can also change the code that contains this parameter.

I agree that the value should not be arbitrary, therefore I mentioned someone more knowledgable about parachains.

runtime/parachains/src/mock.rs
```diff
@@ -138,6 +138,7 @@ impl crate::ump::Config for Test {
 	type UmpSink = TestUmpSink;
 	type FirstMessageFactorPercent = FirstMessageFactorPercent;
 	type ExecuteOverweightOrigin = frame_system::EnsureRoot<AccountId>;
+	type WeightInfo = ();
```
kianenigma, line 141: Please call this `TestWeightInfo` instead of `()`.

runtime/parachains/src/ump.rs
```diff
@@ -162,6 +170,22 @@ impl fmt::Debug for AcceptanceCheckErr {
 	}
 }
 
+pub trait WeightInfo {
+	fn service_overweight() -> Weight;
+	fn process_upward_message(s: u32) -> Weight;
+}
+
+// fallback implementation
+impl WeightInfo for () {
```
kianenigma, line 179:
```suggestion
impl WeightInfo for TestWeightInfo {
```

runtime/parachains/src/ump.rs
```diff
@@ -100,25 +103,30 @@ impl<XcmExecutor: xcm::latest::ExecuteXcm<C::Call>, C: Config> UmpSink for XcmSi
 			xcm::MAX_XCM_DECODE_DEPTH,
 			&mut &data[..],
 		)
-		.map(Xcm::<C::Call>::try_from);
+		.map(|xcm| {
+			(
+				Xcm::<C::Call>::try_from(xcm),
+				<C as Config>::WeightInfo::process_upward_message(data.len() as u32),
```
kianenigma, line 109: This benchmark is a bit wrong IMO. `WeightInfo::process_upward_message` is measuring the execution of this entire code path, including the execution of the xcm itself. Yet, you are summing the weight used in the execution of the xcm with this weight. 

Ideally, the xcm operation itself should be something as close as possible to noop in the benchmark (`Transact` with `remark` might actually _be_ good for that) and then you could add them up.

That being said though, this all is kinda fine since you are over-couting the weight, which is less of a big deal.

kianenigma (commented):
aside from my other note, needs to benchmark `initializer_on_new_session` as well and return its weight. I will do the same for #3876 as well. Everything else looks good to me.

runtime/parachains/src/ump.rs
hirschenberger: I do exactly this, the xcm operation is a Transact-remark with different sizes, to take the decoding effort into account but having a minimal xcm-execution overhead. https://github.com/paritytech/polkadot/blob/f88e7ee92e9c915139d70332c4e4f7e199b3499b/runtime/parachains/src/ump/benchmarking.rs#L55-L59

I can also just benchmark the xcm-decoding https://github.com/paritytech/polkadot/blob/f88e7ee92e9c915139d70332c4e4f7e199b3499b/runtime/parachains/src/ump.rs#L102

Or is there a even a benchmarked weight for this somewhere? If not it may make sense to add one and take decoding time into account in all places where it's used.

runtime/parachains/src/ump.rs
```diff
@@ -332,20 +365,20 @@ impl<T: Config> Pallet<T> {
...
 	/// Iterate over all paras that were noted for offboarding and remove all the data
 	/// associated with them.
-	fn perform_outgoing_para_cleanup(outgoing: &[ParaId]) {
-		for outgoing_para in outgoing {
-			Self::clean_ump_after_outgoing(outgoing_para);
-		}
+	fn perform_outgoing_para_cleanup(outgoing: &[ParaId]) -> Weight {
+		outgoing
+			.iter()
+			.fold(0, |w, p| w.saturating_add(Self::clean_ump_after_outgoing(p)))
```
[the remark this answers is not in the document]
hirschenberger: I'm using a FP-language in my fulltime job, so this seems more natural to me than a loop. But I'm not religious about that.

runtime/parachains/src/ump.rs
```diff
@@ -762,7 +797,7 @@ pub(crate) mod tests {
 		}
 	}
 
-	fn queue_upward_msg(para: ParaId, msg: UpwardMessage) {
+	pub fn queue_upward_msg(para: ParaId, msg: UpwardMessage) {
```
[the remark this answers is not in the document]
hirschenberger: You're right, this is a relic from an older version.

runtime/parachains/src/ump.rs
```diff
@@ -162,6 +170,28 @@ impl fmt::Debug for AcceptanceCheckErr {
 	}
 }
 
+pub trait WeightInfo {
```
kianenigma, line 173:
```suggestion
/// Weight information of this pallet. 
pub trait WeightInfo {
```

runtime/parachains/src/ump.rs
```diff
@@ -162,6 +170,28 @@ impl fmt::Debug for AcceptanceCheckErr {
...
+	fn service_overweight() -> Weight;
+	fn process_upward_message(s: u32) -> Weight;
+	fn perform_outgoing_para_cleanup() -> Weight;
+}
+
+// fallback implementation
```
kianenigma, line 179:
```suggestion
/// fallback implementation
```

runtime/parachains/src/ump.rs
kianenigma: Yeah this is really tricky. I am fine with it given

1. This is an overestimate and is therefore safe.
2. please make `// NOTE: ` and explain what is being done, and why it is not accurate.

kianenigma: ideally, I would break down the function in ways that it can be easily weighed, but a story for another day.

runtime/parachains/src/ump/benchmarking.rs
kianenigma: This is not fixed.

hirschenberger: I need a hint for sensible bound values.

hirschenberger: @kianenigma @pepyakin Can someone give me a hint for reasonable default values? I really like to get this PR landed before christmas 😄

pepyakin: Sorry.

Currently on kusama and polkadot this value is set to 50 KiB. I think this is a sensible value. I think it may be possible we decrease this value even. (Decreasing would mean that the value would go out of sync with the benchmarking but that's OK as long as it goes into the conservative direction)

hirschenberger: @kianenigma I added a max-bound as suggested with the value @pepyakin proposed. Can I have another review please?

runtime/parachains/src/configuration.rs
```diff
@@ -784,6 +789,7 @@ pub mod pallet {
 		))]
 		pub fn set_max_upward_message_size(origin: OriginFor<T>, new: u32) -> DispatchResult {
 			ensure_root(origin)?;
+			ensure!(new <= T::MaxUpwardMessageSizeBound::get(), Error::<T>::InvalidNewValue);
```
pepyakin, line 792: Perhaps, this should also be reflected in `check_consistency`?

hirschenberger: @pepyakin As I see it, I have to change the generic parameter of `HostConfiguration` from `Config::BlockNumber` to `Config` make get access to `MaxUpwardMessageSizeBound`. Is this ok?

hirschenberger: Hey @pepyakin When I moved the checks to `check_consistency`, running any chain fails because in `node/service/src/chain_spec.rs` and `bridges/bin/rialto/node/src/chain_spec.rs` the max-size is set to `1024*1024` and that is way higher than the 50KiB you proposed. What do you suggest to solve this?

pepyakin: Ah, good to hear that `check_consistency` gets its job done. I say we decrease those numbers to 50 KiB.

hirschenberger: Hey @pepyakin or @kianenigma I'm stuck here with an issue and maybe you can help me. ATM the `service_overweight` bench fails bc. I'm unable to create an overweight message. When I limit the message size to 50KiB I won't reach the current weight limit to make it overweight.

Ideally one can create an arbitrary-weighted call by using `fill_block( req_weight / max_weight )` I tried it but it seems the execution of `fill_block` in `XcmExecutor::execute_xcm` does not consume any weight at all and so will not issue a `XcmError::WeightLimitReached`.

My idea was to create a `pallet_utilities::batch` with as many `remark` calls to reach overweight, but I ran into compiler issues with `T::Config` vs. `pallet_utilities::Config`. Do you have a better idea how to create a heavyweight call?
If the later approach is ok to you, I'd try to fix the issues (or ask more specific questions).

bridges/bin/rialto/node/src/chain_spec.rs
```diff
@@ -265,7 +265,7 @@ fn testnet_genesis(
 				// an import since that's a made up number and should be replaced with a constant
 				// obtained by benchmarking anyway.
 				ump_service_total_weight: 4 * 1_000_000_000,
-				max_upward_message_size: 1024 * 1024,
+				max_upward_message_size: 50 * 1024,
```
emostov, line 268: Is there any specific rational for this change?

hirschenberger: @pepyakin Proposed that value here: https://github.com/paritytech/polkadot/pull/3889#discussion_r758584658

runtime/parachains/src/ump/benchmarking.rs
```diff
@@ -0,0 +1,129 @@
...
+
+frame_benchmarking::benchmarks! {
+	// NOTE: We are overestimating slightly here.
+	// The benchmark is timing this whole function with different message sizes and a NOOP extrinsic to
+	// measure the size-dependent weight. But as we use the weight funtion **in** the benchmarked function we
+	// are taking call and control-flow overhead into account twice.
+	process_upward_message {
+		let s in 0..MAX_UPWARD_MESSAGE_SIZE_BOUND;
+		let para = ParaId::from(1978);
+		let host_conf = configuration::ActiveConfig::<T>::get();
```
emostov, line 75: Is this getting used anywhere?

emostov (approved):
This makes sense to me. Would be good if @KiChjang , @pepyakin , or @kianenigma could take another look

runtime/parachains/src/ump/benchmarking.rs
```diff
@@ -0,0 +1,129 @@
...
+		let data = create_message_size::<T>(s);
+	}: {
+		assert!(T::UmpSink::process_upward_message(para, &data[..], Weight::MAX).is_ok());
+	}
+
+	perform_outgoing_para_cleanup {
```
emostov, line 81: I don't think its a big deal, but would it make more sense to name this benchmark `clean_ump_after_outgoing`?

hirschenberger: This is indeed confusing, as I first benchmarked `perform_outgoing_para_cleanup` and switched it later to only bench it's inner function `clean_ump_after_outgoing`, as the weight used by this function is summed up and returned in `perform_outgoing_para_cleanup`.

runtime/parachains/src/ump.rs
```diff
@@ -100,25 +109,34 @@ impl<XcmExecutor: xcm::latest::ExecuteXcm<C::Call>, C: Config> UmpSink for XcmSi
...
+			(
+				Xcm::<C::Call>::try_from(xcm),
+				// NOTE: We are overestimating slightly here.
+				// The benchmark is timing this whole function with different message sizes and a NOOP extrinsic to
+				// measure the size-dependent weight. But as we use the weight funtion **in** the benchmarked funtion we
+				// are taking call and control-flow overhead into account twice.
+				<C as Config>::WeightInfo::process_upward_message(data.len() as u32),
+			)
+		});
 		match maybe_msg {
```
kianenigma, line 122: bad name -- should be called `maybe_msg_and_weight` or something of that sort.

runtime/parachains/src/ump.rs
```diff
@@ -714,7 +762,7 @@ pub(crate) mod tests {
 	use frame_support::{assert_noop, assert_ok, weights::Weight};
 	use std::collections::HashSet;
 
-	struct GenesisConfigBuilder {
+	pub(super) struct GenesisConfigBuilder {
```
drahnr, line 765: the `mod tests` has been made a separate file, so moving the changeset to `runtime/parachains/src/ump/tests.rs` should resolve the conflict.

drahnr (commented):
Generally LGTM, I defer the final review to @pepyakin who knows the intricacies and possible side effects better, and hence can judge if the benchmarks measure what the ought to measure :)

pepyakin (approved):
Looks good. I am OK with merging it as is, possibly following up with nit fixes. Thanks for your work and I am sorry about it took so much time to get the review.

runtime/parachains/src/ump.rs
```diff
@@ -336,20 +380,22 @@ impl<T: Config> Pallet<T> {
...
 		for outgoing_para in outgoing {
-			Self::clean_ump_after_outgoing(outgoing_para);
+			weight = weight.saturating_add(Self::clean_ump_after_outgoing(outgoing_para));
 		}
+		weight
 	}
 
 	/// Remove all relevant storage items for an outgoing parachain.
-	fn clean_ump_after_outgoing(outgoing_para: &ParaId) {
+	pub(crate) fn clean_ump_after_outgoing(outgoing_para: &ParaId) -> Weight {
```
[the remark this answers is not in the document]
hirschenberger: When I do `pub(in crate::ump::benchmarking)` the compiler complains: `error[E0742]: visibilities can only be restricted to ancestor modules`

drahnr:
Let's merge this after 0.9.17 is cut.

runtime/parachains/src/ump.rs
```diff
@@ -165,6 +183,29 @@ impl fmt::Debug for AcceptanceCheckErr {
...
+	fn service_overweight() -> Weight {
+		BlockWeights::default().max_block
+	}
+
+	fn process_upward_message(_msg_size: u32) -> Weight {
+		BlockWeights::default().max_block
+	}
+
+	fn clean_ump_after_outgoing() -> Weight {
+		BlockWeights::default().max_block
```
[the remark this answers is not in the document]
hirschenberger: Yes, `Weight::MAX` seems a better value

runtime/parachains/src/ump/benchmarking.rs
```diff
@@ -0,0 +1,127 @@
...
+	let len = msg.len() as u32;
+	let msgs = vec![msg];
+	Ump::<T>::check_upward_messages(host_conf, para, &msgs).unwrap();
+	let _ = Ump::<T>::receive_upward_messages(para, msgs);
+	assert_last_event_type::<T>(Event::UpwardMessagesReceived(para, 1, len).into());
+}
+
+fn create_message_size<T: Config>(size: u32) -> Vec<u8> {
+	// fill the remark but leave some bytes for the encoding of the message
+	let size = size.saturating_sub(20) as usize;
```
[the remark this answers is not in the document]
hirschenberger: I changed the function name to indicate that the created message will have at least `size` bytes.

I also changed the code to create an empty message (without remark payload) to determine the encoded size and use this to create the actual message as close as possible (but always bigger) to `size`.  Ensuring that with an assertion.

emostov:
> Let's merge this after 0.9.17 is cut.

@drahnr is this ok to merge now?

drahnr:
Yes, feel free to merge

kianenigma:
bot merge

hirschenberger:
> @hirschenberger can you put your DOT or KSM address in your PR description? Then we can propose a tip for you!

Oh nice, thank you. I'm glad I got this PR finally landed. My DOT address is:
`1ZApdbwrNnT1Hey4xeTAH6SZPDgbA9ZVftNuu9jTRbkc2Bg`

PR #3899, opened 2021-09-21
Companion to Substrate #9829 (Add `put_in_front_of` extrinsic to bags-list)

emostov:
Companion for https://github.com/paritytech/substrate/pull/9829


runtime/kusama/src/weights/pallet_bags_list.rs
```diff
@@ -62,4 +62,7 @@ impl<T: frame_system::Config> pallet_bags_list::WeightInfo for WeightInfo<T> {
 			.saturating_add(T::DbWeight::get().reads(7 as Weight))
 			.saturating_add(T::DbWeight::get().writes(5 as Weight))
 	}
+	fn put_in_front_of() -> Weight {
+		0
+	}
```
thiolliere, line 67: so I guess we will upgrade in a follow up or before runtime upgrade, but we should be careful that we actually upgrade it.
Because it happens that system isn't updated by our release process (https://github.com/paritytech/polkadot/issues/4290) , so we should keep a careful eye on it.

thiolliere (approved):
looks good to me, as long as we keep an eye on it so weight is for sure updated in next runtime upgrade

runtime/kusama/src/weights/pallet_bags_list.rs
emostov: Yup planning on doing a follow up immediatley after this was merged

shawntabrizi (approved):

shawntabrizi:
bot merge

PR #3904, opened 2021-09-22
always broadcast tranche 0 assignments and add a delay before approval

rphmeier:


burdges:
Although not before importing the block I presume.

drahnr (approved):
LGTM

PR #3907, opened 2021-09-22
Move artifacts states into memory in PVF validation host

### LINKED ISSUE #3720: PVF validation host: move artifacts states into memory
Initially we designed PVF validation host so that it is persistent. That is, results of validation is always persisted on disk be it a compiled artifact or an error status. One example is that the prepare queue doesn't even communicate the result back to the host directly but instead does it through a file, when execute worker pick it up it will read the result and return it. Even if it is not the most efficient way it is a simple one.

However, after deploying initial version of pvf validation host we pivoted to clear the cache on the startup and refill it. That made the current code structure suboptimal. Moreover, bugs like https://github.com/paritytech/polkadot/issues/3581 or https://github.com/paritytech/polkadot/issues/3044 become possible.

We can restructure the system so that the results of preparation will be stored directly in the memory. That is, instead of storing the success flag or error status in the file we will store it in the artifacts map.

However, since the compiled artifacts can be rather heavy we have to continue storing them in files.

From 10k ft view, it looks like we could do it by doing the following steps:

1. the prepare worker will return detailed outcome of the compilation. It will continue saving compiled artifacts into a file
2. the prepare pool will pass the outcome to the queue via `FromPool::Concluded`
3. in the prepare queue we handle `handle_worker_concluded` and pass the outcome to the host via `FromQueue::Prepared`
4. in host's `handle_prepare_done` we save the prepared variant and supply the outcome there.
5. in host's `handle_execute_pvf` we will have to check the artifacts preparation outcome and act on it eagerly.

slumber:
Resolves #3720


node/core/pvf/src/prepare/worker.rs
```diff
@@ -288,31 +281,43 @@ pub fn worker_entrypoint(socket_path: &str) {
...
+
+					tracing::debug!(
+						target: LOG_TARGET,
+						worker_pid = %std::process::id(),
+						"worker: writing artifact to {}",
+						dest.display(),
+					);
+					async_std::fs::write(&dest, &artifact_bytes).await?;
+
+					vec![1u8]
```
pepyakin, line 304: I think sending `1u8` to indicate success doesn't work any longer.

To understand why, look at the `Artifact::Error(err)` case. `err` is of type `PrepareError` which has roughly the following shape:

```rust
#[derive(Debug, Clone, Encode, Decode)]
pub enum PrepareError {
	Prevalidation(String),
	Preparation(String),
	DidNotMakeIt,
}
```

`PrepareError` implements `Decode` via deriving and for enum that is very simple: each variant is assigned a special one byte tag value. Thus, encoding `PrepareError::Preparation` will start with a `1`.

I hope you can see the problem now.

Besides that, on the receiving side the code also looks a bit clunky. Can we just put a proper typed value on the wire, such as `Result<(), PrepareError>`? That would simultaneously solve this issue and also clean up the `match` on the receiving side.

pepyakin (commented):
Had a brief look.

node/core/pvf/src/host.rs
```diff
@@ -493,9 +498,21 @@ async fn handle_prepare_done(
...
+			never!("the artifact is already processed unsuccessfully: {:?}", artifact_id);
+			return Ok(())
+		},
 		Some(state @ ArtifactState::Preparing) => state,
 	};
 
+	// Don't send failed artifacts to the execution's queue.
+	if let Err(error) = result {
+		*state = ArtifactState::FailedToProcess(error);
+		return Ok(())
```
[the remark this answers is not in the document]
slumber: That's certainly true, thanks!

node/core/pvf/src/prepare/worker.rs
slumber: [agreed](https://github.com/paritytech/polkadot/pull/3907/commits/8bdc7d3791aec014a3f1dd26b2291add569bc07a)

node/core/pvf/src/prepare/worker.rs
```diff
@@ -288,31 +284,43 @@ pub fn worker_entrypoint(socket_path: &str) {
...
-			async_std::fs::write(&dest, &artifact_bytes).await?;
+			let result = match prepare_artifact(&code) {
+				Artifact::Error(err) => {
+					// Serialized error will be written into the socket.
+					Err(err)
+				},
+				Artifact::Compiled(compiled_artifact) => {
+					// Write the serialized artifact into a temp file.
+					// Since a compiled artifact can be heavy, we send an empty
+					// `Ok` to indicate the success.
```
pepyakin, line 296: This is a bit misleading. In reality, we do not put the artifact on UDS since the host is not expected to hold them in memory. That is by design since the artifacts can be quite hefty. The difference is subtle, but it is important to not oversimplify.

Here, if the reader already knows about this high-level design point, then they will be just distracted and no benefit will be gained. OTOH, if the reader is not familiar with it, then I do not think this comment will help understand it.

Generally, as we write comments, we should strive to be as useful to the reader as possible. We should not forget that comments have costs too: they require maintenance when things change and they require attention when read. 

I see two alternatives here:

1. Remove the whole thing
2. Describe what we are doing here instead of why.

Issue #3193, opened 2021-06-07
Ability to query correct Xcm destination weights


jacogr:
Currently on the asset teleports, we need to specify the destination weight. (Effectively calculated from the BaseXcmWeight) From a use perspective, this is currently hardcoded and already has some issues between different chains (e.g. Westend vs Kusama) since the parameters are different.

It would be optimal and much less error prone if we were to expose an rpc along the lines of `queryXcmWeight(Xcm) -> Weight` (copy-and-pasted from Gav's suggestion) that can be used to estimate this value. Since we pass in the encoded Xcm call, we can use it to estimate any Xcm message, not just the one I wish to use it for in the short term.

This can be used for -

- the dest weight
- calculations with the WeightToFee (the RPC could already return this)

Obviously getting this value 100% correct is quite important here, getting it wrong can lead to sending assets via teleport into a black hole.

PR #3911, opened 2021-09-23
update BaseXcmWeight to match Kusama

xlc:
Why they are different? This makes testing impossible because testing environment is not same as production environment (Kusama).

Code works on Westend / Rococo means nothing if it is going to work on Kusama.

There are no way to access those value from parachain runtime so we need to hardcode the number and this just makes everything harder for parachains with zero benefits.

### DIFF
--- a/runtime/rococo/src/lib.rs
+++ b/runtime/rococo/src/lib.rs
@@ -625,7 +625,7 @@ type LocalOriginConverter = (
 );
 
 parameter_types! {
-	pub const BaseXcmWeight: Weight = 100_000;
+	pub const BaseXcmWeight: Weight = 1_000_000_000;
 }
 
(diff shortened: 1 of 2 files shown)


shawntabrizi (approved):
This is fine, but I will say, that probably whatever end to end experiences you are building, you will need to have some system to actually read the weight of the corresponding XCM functions, since they may change over time.

So hard coding any values may not be the right long term thing.

But I agree, no reason for this to be different in the context of the different runtimes.

xlc:
I think Gav have some plan on exposing some weight crates? But otherwise currently I don't see how we can do that. We definitely need a better long term solution.

jacogr:
There is this issue logged for an xcm-like payment_queryInfo - https://github.com/paritytech/polkadot/issues/3193

apopiak (approved):

shawntabrizi:
bot merge

PR #3913, opened 2021-09-23
Add benchmarking for parachain runtime initializer pallet

KiChjang:
Part of #3850.


thiolliere (approved):

runtime/westend/src/lib.rs
```diff
@@ -805,7 +805,7 @@ impl pallet_proxy::Config for Runtime {
 impl parachains_origin::Config for Runtime {}
 
 impl parachains_configuration::Config for Runtime {
-	type WeightInfo = parachains_configuration::weights::WeightInfo<Runtime>;
+	type WeightInfo = weights::runtime_parachains_configuration::WeightInfo<Runtime>;
```
emostov, line 808: I think #3914 should take of this

runtime/parachains/src/initializer/benchmarking.rs
```diff
@@ -0,0 +1,37 @@
...
+benchmarks! {
+	force_approve {}: _(RawOrigin::Root, One::one())
+	verify {
+		assert_eq!(
+			<frame_system::Pallet<T>>::digest().logs.last().unwrap(),
+			&<DigestItemOf<T>>::from(ConsensusLog::ForceApprove(One::one())),
+		);
+	}
+}
+
```
emostov, line 32: Essentially I think the only storage op here is an `<Digest<T>>::append(item);` - without looking at the code I am guessing it has to decode the vec, push the item, and then encode it (But maybe it it is more optimized?). I am wondering if for something like this we should benchmark with a large vec to assume worst case?

My naive guess though is that practically we will never have enough logs for the append op to be that expensive + this is from a force origin so it doesn't look like a significant DoS vector.

emostov (approved):
generally looks good to me. Maybe worth looking into if filling up the vec of digest logs would get a better worst case benchmark.

runtime/parachains/src/initializer/benchmarking.rs
KiChjang: I looked through the code, and while it doesn't decode the entire vec (it only decodes the first 4 bytes to get the encoded length), it does however copy the entire vec to a new one whenever the new encoded length is different from the previous (which is pretty much always).

I'll most likely need to add in a variable and see how the complexity varies with the digest vec length.

KiChjang:
bot merge

PR #3924, opened 2021-09-24
collect included disputes from on-chain

drahnr:
TODO:

* [x] hookup all runtimes to (roughly) `runtime_impl::imported_disputes()`-fns.
* [x] Assure graceful handling of `imported_disputes` on the node side, so older runtime versions that do not provide the call, do not take down the node.
* [x] add runtime tests
* [x] refactor
* [x] :bike: :house:


primitives/src/v1/mod.rs
```diff
@@ -1017,6 +1030,9 @@ sp_api::decl_runtime_apis! {
 
 		/// Get the validation code from its hash.
 		fn validation_code_by_hash(hash: ValidationCodeHash) -> Option<ValidationCode>;
+
+		/// Scrape dispute relevant from on-chain, backing votes and resolved disputes.
+		fn imported_on_chain_disputes() -> Option<ScrapedImportDisputesAndBackingVotes<H>>;
```
rphmeier, line 1035: I recommend a rename - maybe just 'on_chain_votes'?

runtime/polkadot/src/lib.rs
```diff
@@ -2068,7 +2068,7 @@ mod test_fees {
 		// much it is going to cost.
 		use pallet_elections_phragmen::WeightInfo;
 
-		// Loser candidate lose a lot of money; sybil attack by candidates is even more expensive,
+		// Loser candidate lose a lot of money; sibyl attack by candidates is even more expensive,
```
rphmeier, line 2071: https://www.sciencedirect.com/topics/computer-science/sybil-attack

drahnr: I put too much trust in the `en_US.dic` :facepalm:

drahnr, line 2071:
```suggestion
		// Loser candidate lose a lot of money; sybil attack by candidates is even more expensive,
```

node/core/dispute-coordinator/src/real/mod.rs
```diff
@@ -499,10 +503,182 @@ async fn handle_new_activations(
...
+			Ok(Ok(Some(val))) => val,
+			Ok(Ok(None)) => {
+				tracing::trace!(
+					target: LOG_TARGET,
+					relay_parent = ?new_leaf,
+					"No on chain votes stored for relay chain leaf");
+				return Ok(())
+			},
+			Ok(Err(e)) => {
+				tracing::trace!(
```
rphmeier, line 543: debug logs seem fine here

node/core/dispute-coordinator/src/real/mod.rs
```diff
@@ -499,10 +503,182 @@ async fn handle_new_activations(
...
 
+	// Obtain the session info, for sake of `ValidatorId`s
+	// either from the rolling session window or query from
+	// the on chain state directly
+	let session_info: SessionInfo =
+		if let Some(session_info) = state.rolling_session_window.session_info(session) {
+			session_info.clone()
+		} else {
+			let (tx, rx) = oneshot::channel();
+			ctx.send_message(RuntimeApiMessage::Request(
```
rphmeier, line 573: As `scrape_on_chain_votes` is invoked after the session window update, this branch is redundant and feels too defensive. I recommend removing it and adding a comment on why the rolling session window should always contain the right session.

node/core/dispute-coordinator/src/real/mod.rs
```diff
@@ -499,10 +503,182 @@ async fn handle_new_activations(
...
+		};
+
+	// Scraped on-chain backing votes for the candidates with
+	// the new active leaf as if we received them via gossip.
+	for (candidate_receipt, backers) in backing_validators {
+		let candidate_hash = candidate_receipt.hash();
+		let statements = backers.into_iter().filter_map(|(validator_idx, validator_signature)| {
+			let validator_public: ValidatorId =
+				session_info.validators.get(validator_idx.0 as usize)?.clone();
+			let signed_dispute_statement = SignedDisputeStatement::new_checked(
```
rphmeier, line 592: Sig verification here shouldn't be a bottleneck, but FWIW we can trust that the runtime has checked them already.

node/core/dispute-coordinator/src/real/mod.rs
```diff
@@ -499,10 +503,182 @@ async fn handle_new_activations(
...
+	}
+
+	if disputes.is_empty() {
+		return Ok(())
+	}
+
+	// Import concluded disputes from on-chain, this already went through a vote so it's assumed
+	// as verified. This will only be stored, gossiping it is not necessary.
+
+	// First try to obtain all the backings which ultimately contain the candidate
```
rphmeier, line 631: This seems wrong. It seems that the goal is to get the full candidate receipt, but I don't think that the candidate backing subsystem is the way to do it - the candidate has likely been backed at least a few blocks ago and the relevant job (which would be `candidate_receipt.relay_parent`, not `new_leaf`) has been pruned a while ago.

runtime/parachains/src/paras_inherent.rs
```diff
@@ -254,13 +256,24 @@ pub mod pallet {
...
 				<scheduler::Pallet<T>>::group_validators,
 			)?;
 
+			// The number of disputes included in a block is
+			// limited by the weight as well as the number of candidate blocks.
+			OnChainVotes::<T>::put(ScrapedOnChainVotes::<<T::Header as HeaderT>::Hash> {
+				session: current_session,
+				backing_validators: candidate_receipt_with_backing_validator_indices,
+				disputes,
+			});
```
eskimor, line 275: I am a bit confused that we need to store this - shouldn't the required information be already available on chain?

node/core/dispute-coordinator/src/real/mod.rs
```diff
@@ -499,10 +503,182 @@ async fn handle_new_activations(
...
+			}
+		};
+
+	// Scraped on-chain backing votes for the candidates with
+	// the new active leaf as if we received them via gossip.
+	for (candidate_receipt, backers) in backing_validators {
+		let candidate_hash = candidate_receipt.hash();
+		let statements = backers.into_iter().filter_map(|(validator_idx, validator_signature)| {
+			let validator_public: ValidatorId =
+				session_info.validators.get(validator_idx.0 as usize)?.clone();
```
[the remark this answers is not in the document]
drahnr: It needs to be handled somehow and I wasn't sure enough add an expect, and the handling was sufficiently easy.

runtime/parachains/src/inclusion.rs
```diff
@@ -565,17 +588,28 @@ impl<T: Config> Pallet<T> {
...
+							let mut backer_idx_and_signature =
+								Vec::<(ValidatorIndex, ValidatorSignature)>::with_capacity(
+									backed_candidate.validator_indices.count_ones(),
+								);
+							for ((bit_idx, _), attestation) in backed_candidate
 								.validator_indices
 								.iter()
 								.enumerate()
 								.filter(|(_, signed)| **signed)
+								.zip(backed_candidate.validity_votes.iter())
```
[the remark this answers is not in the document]
drahnr: The creation in i.e `table_attested_to_backed` guarantees this, we should improve the documentation of `BackedCandidate` to provide a integrity maintaining interface to it's interior.

runtime/parachains/src/paras_inherent.rs
eskimor: New to runtime programming, so maybe a stupid question: Assuming we need to store this, should we also take care of freeing that storage once we are done with it?

eskimor: > New to runtime programming, so maybe a stupid question: Assuming we need to store this, should we also take care of freeing that storage once we are done with it?

Bernhard already pointed out - that's a single entry that gets overwritten every time - so short answer: No.

node/core/dispute-coordinator/src/real/mod.rs
drahnr: There are two options here: Either spend the time to verify and maintaining our invariant of `SignedDisputeStatement`, or returning to by _convention_ and introducing a `fn from_on_chain_included`.

node/core/dispute-coordinator/src/real/mod.rs
```diff
@@ -499,10 +503,182 @@ async fn handle_new_activations(
...
+
+	if backing_validators.is_empty() && disputes.is_empty() {
+		return Ok(())
 	}
 
+	// Obtain the session info, for sake of `ValidatorId`s
+	// either from the rolling session window or query from
+	// the on chain state directly
+	let session_info: SessionInfo =
+		if let Some(session_info) = state.rolling_session_window.session_info(session) {
```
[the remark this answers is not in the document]
drahnr: That's in line with Rob's comment

runtime/parachains/src/paras_inherent.rs
drahnr: We need the candidate receipts for the backing votes somehow, now if we have to query around anyways, there is little reason to not just utilize the `Disputes` storage. TBD

node/core/dispute-coordinator/src/real/mod.rs
drahnr: Added `new_unchecked_from_trusted_source`

node/core/dispute-coordinator/src/real/mod.rs
```diff
@@ -499,10 +503,182 @@ async fn handle_new_activations(
...
+	          + overseer::SubsystemContext<Message = DisputeCoordinatorMessage>),
+	overlay_db: &mut OverlayedBackend<'_, impl Backend>,
+	state: &mut State,
+	new_leaf: Hash,
+	now: u64,
+	metrics: &Metrics,
+) -> Result<(), Error> {
+	// obtain the concluded disputes as well as the candidate backing votes
+	// from the new leaf
+	let ScrapedOnChainVotes { session, backing_validators, disputes } = {
```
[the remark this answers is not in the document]
drahnr: At the time of block import / inclusion, we don't have any just yet. So that part has to rely on `ApprovalDist` and if we miss those, eventually - given other validators can cast their vote on a dispute - a dispute resolution.

node/core/dispute-coordinator/src/real/mod.rs
drahnr: That is unfortunately correct. Added a draft on how this might work, pending a sanity check with a fresh pair of eyes.

node/core/dispute-coordinator/src/real/mod.rs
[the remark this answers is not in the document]
rphmeier: Approval votes don't appear on-chain, but we do harvest all off-chain ones already.

runtime/parachains/src/inclusion.rs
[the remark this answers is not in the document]
rphmeier: File an issue?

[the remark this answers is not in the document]
drahnr: https://github.com/paritytech/polkadot/issues/3995

node/core/dispute-coordinator/src/real/mod.rs
```diff
@@ -499,10 +502,212 @@ async fn handle_new_activations(
...
+						relay_parent = ?new_leaf,
+						"Missing public key for validator {:?}",
+						&validator_index);
+						None
+					})
+					.cloned()?;
+				let signed_dispute_statement =
+					SignedDisputeStatement::new_unchecked_from_trusted_source(
+						// TODO verify this is the correct statement
+						DisputeStatement::Valid(ValidDisputeStatementKind::BackingValid(new_leaf)),
```
rphmeier, line 601: Likely not correct. Both `BackingValid` and `BackingSeconded` statements can appear on-chain.

node/core/dispute-coordinator/src/real/mod.rs
```diff
@@ -499,10 +502,212 @@ async fn handle_new_activations(
...
+						validator_signature,
+					);
+				Some((signed_dispute_statement, validator_index))
+			});
+		let import_result = handle_import_statements(
+			ctx,
+			overlay_db,
+			state,
+			candidate_hash,
+			MaybeCandidateReceipt::AssumeBackingVotePresent,
```
rphmeier, line 614: Isn't the candidate receipt unconditionally present at this point? It's part of the `backing_validators` object that was iterated over.

rphmeier: The only point I can think of when the candidate receipt wouldn't be present is if we're scraping a remote dispute from the chain and haven't seen any of:
1. a fork where the block was backed
2. an off-chain backing message
3. the dispute messages off-chain

node/core/dispute-coordinator/src/real/mod.rs
```diff
@@ -499,10 +502,212 @@ async fn handle_new_activations(
...
+				let session_info: SessionInfo = if let Some(session_info) =
+					state.rolling_session_window.session_info(session)
+				{
+					session_info.clone()
+				} else {
+					tracing::warn!(
+					target: LOG_TARGET,
+					relay_parent = ?new_leaf,
+					?session,
+					"Could not retrieve session info from rolling session window for concluded dispute");
```
rphmeier, line 656:
```suggestion
					"Could not retrieve session info from rolling session window for recently concluded dispute");
```

node/core/dispute-coordinator/src/real/mod.rs
```diff
@@ -499,10 +502,212 @@ async fn handle_new_activations(
...
+					validator_index,
+				))
+			})
+			.collect::<Vec<_>>();
+		let import_result = handle_import_statements(
+			ctx,
+			overlay_db,
+			state,
+			candidate_hash,
+			MaybeCandidateReceipt::AssumeBackingVotePresent,
```
rphmeier, line 691: This seems correct, yeah.

node/core/dispute-coordinator/src/real/mod.rs
```diff
@@ -648,21 +861,36 @@ async fn handle_import_statements(
...
 
 	let n_validators = validators.len();
 
 	let supermajority_threshold = polkadot_primitives::v1::supermajority_threshold(n_validators);
 
-	let mut votes = overlay_db
+	// In case we are not provided with a candidate receipt
+	// we operate under the assumption, that a previous vote
+	// which included a `CandidateReceipt` was seen.
+	// This holds since every block is preceeded by the `Backing`-phase.
```
rphmeier, line 875: Unfortunately not, but it's _really, really unlikely_ that we haven't seen the candidate receipt. It would require a pretty complex and fairly arbitrary block withholding attack that serves no purpose because this branch is only hit when disputes are already concluded and therefore already have 34% honest participation.

node/core/dispute-coordinator/src/real/mod.rs
drahnr: We might piggyback on the disputes storage items instead, which contain the `DisputeStatement`.

runtime/parachains/src/inclusion.rs
```diff
@@ -565,17 +588,28 @@ impl<T: Config> Pallet<T> {
...
 							{
-								let val_idx =
-									group_vals.get(bit_idx).expect("this query done above; qed");
+								let val_idx = group_vals
+									.get(bit_idx)
+									.expect("this query succeeded above; qed");
```
rphmeier, line 604: aye

runtime/parachains/src/inclusion.rs
```diff
@@ -565,17 +588,28 @@ impl<T: Config> Pallet<T> {
...
+
+								backer_idx_and_signature
+									.push((*val_idx, attestation.signature().clone()));
```
rphmeier, line 607: it should probably store the attestation type as well, re: https://github.com/paritytech/polkadot/pull/3924/files#r720683514

runtime/parachains/src/paras.rs
```diff
@@ -457,6 +457,12 @@ pub mod pallet {
 	pub(super) type CodeByHash<T: Config> =
 		StorageMap<_, Identity, ValidationCodeHash, ValidationCode>;
 
+	/// Scraped on chain data for extracting resolved disputes as well as backing votes.
+	#[pallet::storage]
+	#[pallet::getter(fn on_chain_votes)]
+	pub(crate) type OnChainVotes<T: Config> =
```
rphmeier, line 463: `Paras` really doesn't feel like the right module for this. `ParasInherent` would make more sense.

node/core/dispute-coordinator/src/real/mod.rs
drahnr: I'll extend your comment, and it's dealt with by a graceful early return

runtime/parachains/src/paras_inherent.rs
```diff
@@ -24,6 +24,7 @@
 use crate::{
 	disputes::DisputesHandler,
 	inclusion,
+	paras::OnChainVotes,
```
rphmeier, line 27: yeah, reaching directly into the storage of another module is not best practice. As mentioned in https://github.com/paritytech/polkadot/pull/3924/files#r720684852 this storage item should be in `ParasInherent`.

node/core/dispute-coordinator/src/real/mod.rs
rphmeier: When we revert a dispute we also typically revert to a point after it was already backed so unless an attacker is chaining reverts it's almost certain that the validators have all seen the candidate receipt.

node/core/dispute-coordinator/src/real/mod.rs
```diff
@@ -648,21 +861,36 @@ async fn handle_import_statements(
...
+		None =>
+			if let MaybeCandidateReceipt::Provides(candidate_receipt) = candidate_receipt {
+				CandidateVotes { candidate_receipt, valid: Vec::new(), invalid: Vec::new() }
+			} else {
+				tracing::warn!(
+					target: LOG_TARGET,
+					session,
+					"Missing info for session which has an active dispute",
+				);
+				return Ok(ImportStatementsResult::InvalidImport)
```
rphmeier, line 890: I don't really like that this discards the votes, but as mentioned above it's very difficult to reach this code path. An ideal solution would be for `CandidateVotes` to tolerate not having the candidate receipt, but that we'd only participate once we have the candidate receipt.

rphmeier (changes requested):
A few issues & questions but generally looks correct.

node/core/dispute-coordinator/src/real/mod.rs
rphmeier: This is not a blocker but I would like a follow-up issue for this at the very least.

node/core/dispute-coordinator/src/real/mod.rs
rphmeier: No, that doesn't help because that'd only store things once they're already disputed and this PR is about making sure that the validators all have the necessary 'valid' votes on bad blocks to begin a dispute during approval-checking.

rphmeier: i.e. the attack scraping backing votes solves is about:
1. Malicious validator group submits backing votes to the chain without gossiping them
2. Approval checkers find that the block is bad, but don't have any 'valid' votes in their dispute DB to take the other side of the dispute
3. Approval checkers fail to initiate a dispute as they see it as one-sided and the bad block survives.

This PR inserts a step 1a: "all honest validators (including approval checkers) scrape the backing votes from the chain to ensure they have the opposing votes", and this makes step 2 of the attack impossible.

drahnr: I see, thanks for clarifying.

node/core/dispute-coordinator/src/real/mod.rs
drahnr: https://github.com/paritytech/polkadot/issues/4011

primitives/src/v0.rs
```diff
@@ -747,6 +748,18 @@ pub enum ValidityAttestation {
 }
 
 impl ValidityAttestation {
+	/// Produce the underlying signed payload of the attestation, given the hash of the candidate,
+	/// which should be known in context.
+	pub fn to_compact_statement(&self, candidate_hash: CandidateHash) -> CompactStatement {
+		// Explicit and implicit map directly from
+		// `ValidityVote::Valid` and `ValidityVote::Issued`, and hence there is a
+		// `1:1` relationshow which enables the conversion.
```
drahnr, line 756: Reviewers note: There is a guide section which states that there is a `1:1` mapping between `ValidityAttestation::Explicit <-> Vote` and `ValidityAttestation::Implicit <-> Seconded/Candidate` which was cross checked with the usage, so the above should be correct.

node/primitives/src/disputes/mod.rs
```diff
@@ -66,6 +66,35 @@ impl CandidateVotes {
...
+	///
+	/// Attention: Not to be used other than with guaranteed fetches.
+	pub fn new_unchecked_from_trusted_source(
+		dispute_statement: DisputeStatement,
+		candidate_hash: CandidateHash,
+		session_index: SessionIndex,
+		validator_public: ValidatorId,
+		validator_signature: ValidatorSignature,
+	) -> Self {
+		assert_eq!(
```
rphmeier, line 80: Codebase policy is to prove that panickers cannot fail. I think that `new_unchecked_from_trusted_source` should simply not do this check, or it should return an error.

primitives/src/v0.rs
rphmeier: Yep, that is correct.

primitives/src/v1/mod.rs
```diff
@@ -945,6 +945,22 @@ pub struct SessionInfo {
 	pub needed_approvals: u32,
 }
 
+/// Scraped runtime backing votes and resolved disputes.
+#[derive(Clone, Encode, Decode, RuntimeDebug, TypeInfo)]
+#[cfg_attr(feature = "std", derive(PartialEq, Default, MallocSizeOf))]
+pub struct ScrapedOnChainVotes<H: Encode + Decode = Hash> {
+	/// The session in which the block was included.
+	pub session: SessionIndex,
+	/// Set of backing validators for each candidate, represented by it's candidate
```
rphmeier, line 954:
```suggestion
	/// Set of backing validators for each candidate, represented by its candidate
```

runtime/parachains/src/inclusion.rs
```diff
@@ -2410,6 +2470,56 @@ mod tests {
 				vec![CoreIndex::from(0), CoreIndex::from(1), CoreIndex::from(2)]
 			);
 
+			// Transform the votes into the setup we expect
+			let expected = {
+				let mut intermediate = std::collections::HashMap::<
```
rphmeier, line 2475: In the runtime, we should use `BTreeMap` as randomness is not properly available for hash maps and this can lead to attacks. It can also lead to mismatches in consensus.

runtime/parachains/src/inclusion.rs
```diff
@@ -2410,6 +2470,56 @@ mod tests {
...
+				backed_candidates.into_iter().for_each(|backed_candidate| {
+					let candidate_receipt_with_backers = intermediate
+						.entry(backed_candidate.hash())
+						.or_insert_with(|| (backed_candidate.receipt(), Vec::new()));
+
+					assert_eq!(
+						backed_candidate.validity_votes.len(),
+						backed_candidate.validator_indices.count_ones()
+					);
+					dbg!(&backed_candidate.validator_indices);
```
rphmeier, line 2488: stray dbg

runtime/parachains/src/inclusion.rs
rphmeier: (ah, this is test code)

runtime/parachains/src/paras_inherent.rs
```diff
@@ -79,6 +80,12 @@ pub mod pallet {
 	#[pallet::storage]
 	pub(crate) type Included<T> = StorageValue<_, ()>;
 
+	/// Scraped on chain data for extracting resolved disputes as well as backing votes.
+	#[pallet::storage]
+	#[pallet::getter(fn on_chain_votes)]
+	pub(crate) type OnChainVotes<T: Config> =
```
rphmeier, line 86: Implementers guide changes for this and the other runtime changes are needed.

rphmeier (commented):
Looks good. Only needs implementers guide changes.

rphmeier (approved):

eskimor (approved):

node/core/runtime-api/src/cache.rs
```diff
@@ -41,6 +42,7 @@ const SESSION_INFO_CACHE_SIZE: usize = 64 * 1024;
 const DMQ_CONTENTS_CACHE_SIZE: usize = 64 * 1024;
 const INBOUND_HRMP_CHANNELS_CACHE_SIZE: usize = 64 * 1024;
 const CURRENT_BABE_EPOCH_CACHE_SIZE: usize = 64 * 1024;
+const ON_CHAIN_VOTES_CACHE_SIZE: usize = 3 * 1024;
```
ordian, line 45: any reason for choosing 3KiB for cache size? the lowest we have is 64KiB, but it's not really backed by data

node/core/dispute-coordinator/src/real/mod.rs
```diff
@@ -499,10 +502,217 @@ async fn handle_new_activations(
...
+			Ok(Ok(Some(val))) => val,
+			Ok(Ok(None)) => {
+				tracing::trace!(
+					target: LOG_TARGET,
+					relay_parent = ?new_leaf,
+					"No on chain votes stored for relay chain leaf");
+				return Ok(())
+			},
+			Ok(Err(e)) => {
+				tracing::debug!(
```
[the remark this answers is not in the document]
eskimor: Fair, but should we make it `warn` once the runtime got upgraded? Follow up issue?

node/core/runtime-api/src/cache.rs
drahnr: No particular reason. We simply won't use it that much, so this could be an optimization to remove it.

drahnr:
bot merge

PR #3940, opened 2021-09-27
XCM Benchmarks for Generic Instructions

shawntabrizi:
This PR introduces benchmarks for all "generic" XCM messages, that is, anything which does not use an `AssetTransactor`.

skip check-dependent-cumulus


xcm/pallet-xcm-benchmarks/src/generic/benchmarking.rs
```diff
@@ -0,0 +1,355 @@
...
+	query_holding {
+		let holding = worst_case_holding();
+
+		let mut executor = new_executor::<T>(Default::default());
+		executor.holding = holding;
+
+		let instruction = Instruction::<XcmCallOf<T>>::QueryHolding {
+			query_id: Default::default(),
+			dest: T::valid_destination()?,
+			assets: Wild(WildMultiAsset::All), // TODO is worst case filter?
```
KiChjang, line 40: Actually no, this is in fact the best case filter -- while processing the `QueryHolding` instruction, the executor calls the [`Assets::min` function](https://github.com/paritytech/polkadot/blob/f26005b999acb8936bda66df10daa0d75c85ce62/xcm/xcm-executor/src/assets.rs#L397-L431), which looks to have the worst case that's based on the number of assets that you if you have a `MultiAssetFilter::Definite`. If instead you have a `WildNonFungible`, then it would depend upon the number of non-fungible asset classes that exist in holding.

Either way, I don't think there's a way to not break the weight calculation into separate cases based on the `assets` parameter, sadly.

xcm/pallet-xcm-benchmarks/src/generic/benchmarking.rs
```diff
@@ -0,0 +1,355 @@
...
+
+	// This benchmark does not use any additional orders or instructions. This should be managed
+	// by the `deep` and `shallow` implementation.
+	buy_execution {
+		let holding = worst_case_holding();
+
+		let mut executor = new_executor::<T>(Default::default());
+		executor.holding = holding;
+
+		let fee_asset = Concrete(Here.into());
```
KiChjang, line 61: This may be the same case as `QueryHolding` in that the weight depends on what kind of assets that you put here. It ultimately calls [`Assets::general_take`](https://github.com/paritytech/polkadot/blob/f26005b999acb8936bda66df10daa0d75c85ce62/xcm/xcm-executor/src/assets.rs#L244-L324) to match the wildcard against the holding register to see which ones to take out for weight payment.

xcm/pallet-xcm-benchmarks/src/generic/benchmarking.rs
shawntabrizi: Fixed. Worst case as I can see is using `Definite(all_assets_explicitly)`.

xcm/pallet-xcm-benchmarks/src/generic/benchmarking.rs
```diff
@@ -0,0 +1,311 @@
...
+	} : {
+		let result = executor.execute(xcm)?;
+	} verify {
+		assert_eq!(executor.total_surplus, 1337);
+		assert_eq!(executor.total_refunded, 1337);
+	}
+
+	set_error_handler {
+		let mut executor = new_executor::<T>(Default::default());
+		let instruction = Instruction::<XcmCallOf<T>>::SetErrorHandler(Xcm(vec![]));
```
KiChjang, line 151: It looks to me that the worst complexity of the `SetErrorHandler` instruction is dependent upon the Weigher trying to weight the error handler's XCM instructions, and so having a 0 length vec is the best case.

xcm/pallet-xcm-benchmarks/src/generic/benchmarking.rs
```diff
@@ -0,0 +1,311 @@
...
+	} : {
+		executor.execute(xcm)?;
+	} verify {
+		assert_eq!(executor.error_handler, Xcm(vec![]));
+	}
+
+	set_appendix {
+		let mut executor = new_executor::<T>(Default::default());
+		let appendix = Xcm(vec![]);
+		let instruction = Instruction::<XcmCallOf<T>>::SetAppendix(appendix);
```
KiChjang, line 162: Same comment as `SetErrorHandler`.

xcm/pallet-xcm-benchmarks/src/generic/benchmarking.rs
```diff
@@ -0,0 +1,311 @@
...
+		let xcm = Xcm(vec![instruction]);
+	} : {
+		match executor.execute(xcm) {
+			Err(error) if error.xcm_error == XcmError::Trap(10) => {
+				// This is the success condition
+			},
+			_ => return Err(BenchmarkError::Stop("xcm trap did not return the expected error"))
+		};
+	} verify {
+		// Verification is done above.
```
KiChjang, line 255: I wonder why you didn't assign the result of `executor.execute(xcm)` to a variable and verify it here in the verify block?

xcm/pallet-xcm-benchmarks/src/generic/benchmarking.rs
```diff
@@ -0,0 +1,311 @@
...
+		let instruction = Instruction::UnsubscribeVersion;
+		let xcm = Xcm(vec![instruction]);
+	} : {
+		executor.execute(xcm)?;
+	} verify {
+		assert!(!<T::XcmConfig as xcm_executor::Config>::SubscriptionService::is_subscribed(&origin));
+	}
+
+	initiate_reserve_withdraw {
+		let assets = MultiAssetFilter::Wild(All);
```
KiChjang, line 294: Same issue with `BuyExecution`, it calls `Assets::general_take`, which doesn't necessarily have the worst case if you have an `All` wildcard.

KiChjang: This one's a bit more annoying to track down, because it also depends on the kind of assets that already exist in holding, and the executor has to [reanchor these assets](https://github.com/paritytech/polkadot/blob/c198d26d3fdb4b3e26599059335a711c4bc9e3b1/xcm/xcm-executor/src/lib.rs#L410), i.e. _where_ the assets originated matters in the worst case as well.

xcm/pallet-xcm-benchmarks/src/generic/benchmarking.rs
shawntabrizi: Weight calculation itself should be trivial to calculate (minimal operations, no storage reads), so practically speaking, this should be fine I think.

KiChjang:
Definitely a good step towards the right direction, but one thing that's sort of bugging me is that some XCM instructions depend on how the chain has configured its XCM executor, e.g. the `SubscribeVersion` XCM -- this depends on what the subscription mechanism is for the chain. In our case, we know it's the XCM pallet that's performing all such work, so firstly I wonder if we can just leverage the benchmarked weights on the XCM pallet to help us here; and more importantly, how "generic" this benchmark can be when some of the instructions are dependent on configuration.

xcm/pallet-xcm-benchmarks/src/generic/benchmarking.rs
shawntabrizi: Same response.

xcm/pallet-xcm-benchmarks/src/generic/benchmarking.rs
shawntabrizi: done. a little ugly, but yeah probably more clear.

shawntabrizi:
> Definitely a good step towards the right direction, but one thing that's sort of bugging me is that some XCM instructions depend on how the chain has configured its XCM executor, e.g. the `SubscribeVersion` XCM -- this depends on what the subscription mechanism is for the chain. In our case, we know it's the XCM pallet that's performing all such work, so firstly I wonder if we can just leverage the benchmarked weights on the XCM pallet to help us here; and more importantly, how "generic" this benchmark can be when some of the instructions are dependent on configuration.

Yeah, the way things are configured, we can def write more specific benchmarks in Pallet XCM and pipe that into the weight calculations for XCM. Unfortunately XCM itself is just way too customizable. What I attempt to do here is something that is good enough for the most basic scenarios. I think that as people develop more custom XCM configurations or even potentially custom executors, they will also need to write custom benchmarks.

So for the purposes of this review, lets ensure that these benchmarks are fair to judge the current state of Westend, Kusama, Polkadot, Statemint, Statemine, and Westmint.

xcm/pallet-xcm-benchmarks/src/generic/benchmarking.rs
shawntabrizi: added worst case holding and added a multiassetfilter where we look for every asset in holding explicitly.

xcm/pallet-xcm-benchmarks/src/generic/benchmarking.rs
shawntabrizi: For all the chains we care about right now, the only way you can pay fees is with the base currency

KiChjang: Okay, we may need to revisit sometime in the future when NFTs can be used to pay for weights, e.g. coupons.

xcm/pallet-xcm-benchmarks/src/generic/benchmarking.rs
```diff
@@ -0,0 +1,316 @@
...
+		// In order to access result in the verification below, it needs to be defined here.
+		let mut _result = Ok(());
+	} : {
+		_result = executor.execute(xcm);
+	} verify {
+		match _result {
+			Err(error) if error.xcm_error == XcmError::Trap(10) => {
+				// This is the success condition
+			},
+			_ => panic!("xcm trap did not return the expected error")
```
KiChjang, line 256: Do we panic here or are we supposed to return a `BenchmarkError`?

KiChjang (approved):
Don't think I have much else to add other than that small comment I added above.

xcm/pallet-xcm-benchmarks/src/generic/benchmarking.rs
shawntabrizi: panic is okay. shouldnt let benchmarking keep going if this check fails.

xcm/pallet-xcm-benchmarks/src/generic/benchmarking.rs
```diff
@@ -0,0 +1,316 @@
...
+
+	claim_asset {
+		let (origin, ticket, assets) = T::claimable_asset()
+			.ok_or(BenchmarkError::Skip)?;
+
+		// We place some items into the asset trap to claim.
+		let mut executor = new_executor::<T>(origin.clone());
+		executor.holding = assets.clone().into();
+		executor.execute(Xcm(vec![]))?;
+		executor.post_execute(0);
```
apopiak, line 231:
```suggestion
		<T as Config>::XcmConfig::AssetTrap::drop_assets(origin, assets);
```
mmh, if you're going below the abstractions why not call `<T as Config>::XcmConfig::AssetTrap::drop_assets(origin, assets)` directly?

xcm/pallet-xcm-benchmarks/src/generic/benchmarking.rs
```diff
@@ -0,0 +1,316 @@
...
+			Err(error) if error.xcm_error == XcmError::Trap(10) => {
+				// This is the success condition
+			},
+			_ => panic!("xcm trap did not return the expected error")
+		};
+	}
+
+	subscribe_version {
+		use xcm_executor::traits::VersionChangeNotifier;
+		let origin = T::transact_origin().ok_or(BenchmarkError::Skip)?;
```
apopiak, line 262: nitpick: I don't like this line.
Are you sure you want to reuse the transact origin for subscribe version here?
Are you sure you want to just skip the benchmark if it's not present?

xcm/pallet-xcm-benchmarks/src/generic/benchmarking.rs
```diff
@@ -0,0 +1,316 @@
...
+	unsubscribe_version {
+		use xcm_executor::traits::VersionChangeNotifier;
+		// First we need to subscribe to notifications.
+		let origin = T::transact_origin().ok_or(BenchmarkError::Skip)?;
+		let query_id = Default::default();
+		let max_response_weight = Default::default();
+		let mut executor = new_executor::<T>(origin.clone());
+		let instruction = Instruction::SubscribeVersion { query_id, max_response_weight };
+		let xcm = Xcm(vec![instruction]);
+		executor.execute(xcm)?;
```
apopiak, line 283: weaker case than above, but you might still consider using `SubscriptionService::start(&origin, query_id, max_response_weight)` directly

xcm/pallet-xcm-benchmarks/src/generic/benchmarking.rs
```diff
@@ -0,0 +1,316 @@
...
+	} verify {
+		assert!(!<T::XcmConfig as xcm_executor::Config>::SubscriptionService::is_subscribed(&origin));
+	}
+
+	initiate_reserve_withdraw {
+		let holding = worst_case_holding();
+		let assets: MultiAssets = holding.clone().into();
+		let assets_filter = MultiAssetFilter::Definite(assets);
+		let reserve = T::valid_destination().map_err(|_| BenchmarkError::Skip)?;
```
apopiak, line 299: I can't shake the feeling that these Skips will turn out to be a footgun for parachain teams. Maybe log a warning in the error case here before returning Skip?

xcm/pallet-xcm-benchmarks/src/generic/benchmarking.rs
```diff
@@ -0,0 +1,316 @@
...
+		let mut executor = new_executor::<T>(Default::default());
+		executor.holding = holding;
+		let instruction = Instruction::InitiateReserveWithdraw { assets: assets_filter, reserve, xcm: Xcm(vec![]) };
+		let xcm = Xcm(vec![instruction]);
+	}:{
+		executor.execute(xcm)?;
+	} verify {
+		// The execute completing successfully is as good as we can check.
```
apopiak, line 307: You could maybe check that the `XcmSender` has queued the outgoing message.

apopiak (commented):
yay for XCM weights! :tada:

xcm/pallet-xcm-benchmarks/src/generic/benchmarking.rs
```diff
@@ -0,0 +1,316 @@
...
+			assets: Definite(assets),
+			max_response_weight: u64::MAX,
+		};
+
+		let xcm = Xcm(vec![instruction]);
+
+	} : {
+		executor.execute(xcm)?;
+	} verify {
+		// The assert above is enough to validate this is completed.
```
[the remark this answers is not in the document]
shawntabrizi: updated comment

xcm/pallet-xcm-benchmarks/src/generic/benchmarking.rs
```diff
@@ -0,0 +1,316 @@
...
+			id: Abstract(i.encode()),
+			fun: Fungible(i as u128),
+
+		}).collect::<vec::Vec<_>>();
+		let multiassets: MultiAssets = assets.into();
+
+		let instruction = Instruction::ReserveAssetDeposited(multiassets.clone());
+		let xcm = Xcm(vec![instruction]);
+	}: {
+		executor.execute(xcm).map_err(|_| BenchmarkError::Skip)?;
```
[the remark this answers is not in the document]
shawntabrizi: This means that the output file will not have this benchmark.

Addressing your other comments about not liking Skip, this is basically the only thing I can think of to handle XCM benchmarks. Someones XCM config may not allow certain XCM Instructions to execute, and when that happens, we can either stop the whole benchmarking process (meaning there would be no easy way to generate benchmarks for XCM) or we can simply skip this benchmark.

As for detecting this, if any code depends on this benchmark, then the outputted file will not have that benchmark and the compilation will fail. So I think this is a fine way to handle the situation. Of course open to feedback if there are other ideas.

xcm/pallet-xcm-benchmarks/src/generic/benchmarking.rs
```diff
@@ -0,0 +1,316 @@
...
+		};
+		let xcm = Xcm(vec![instruction]);
+
+		let num_events = frame_system::Pallet::<T>::events().len();
+	}: {
+		executor.execute(xcm)?;
+	} verify {
+		// TODO make better assertion?
+		let num_events2 = frame_system::Pallet::<T>::events().len();
+		assert_eq!(num_events + 1, num_events2);
```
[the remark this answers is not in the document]
shawntabrizi: Yeah the better idea i had was an extrinsic which modifies storage, but `System::set_storage` is a root function, and then we wouldn't always be able to pass the origin test.

xcm/pallet-xcm-benchmarks/src/generic/benchmarking.rs
```diff
@@ -0,0 +1,316 @@
...
+	}
+
+	refund_surplus {
+		let mut executor = new_executor::<T>(Default::default());
```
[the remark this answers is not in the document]
shawntabrizi: done. thanks!

xcm/pallet-xcm-benchmarks/src/generic/benchmarking.rs
```diff
@@ -0,0 +1,316 @@
...
+	claim_asset {
+		let (origin, ticket, assets) = T::claimable_asset()
+			.ok_or(BenchmarkError::Skip)?;
+
+		// We place some items into the asset trap to claim.
+		let mut executor = new_executor::<T>(origin.clone());
+		executor.holding = assets.clone().into();
+		executor.execute(Xcm(vec![]))?;
+		executor.post_execute(0);
+		// Assets should be in the trap now.
```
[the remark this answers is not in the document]
shawntabrizi: trap is not something that can be easily inspected. I don't think I can do this assert, but also, the XCM would simply fail if the assert would fail here.

xcm/pallet-xcm-benchmarks/src/generic/benchmarking.rs
shawntabrizi: done!

xcm/pallet-xcm-benchmarks/src/generic/benchmarking.rs
shawntabrizi: I added a `subscription_origin` config, but note my comments above about skip.

xcm/pallet-xcm-benchmarks/src/generic/benchmarking.rs
shawntabrizi: done!

xcm/pallet-xcm-benchmarks/src/generic/benchmarking.rs
shawntabrizi: check out my note above

xcm/pallet-xcm-benchmarks/src/generic/benchmarking.rs
shawntabrizi: also not easy to inspect the XcmSender, not something I think I can add, at least not in this PR

xcm/pallet-xcm-benchmarks/src/generic/benchmarking.rs
shawntabrizi: i updated to be an error

xcm/pallet-xcm-benchmarks/src/generic/benchmarking.rs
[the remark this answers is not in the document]
apopiak: Would the following suggestion work?
Have two modes distinguished by a command line flag.
+ By default: trigger an error on skip. This allows running tests and noticing where things are missing.
+ with `--skip-missing-xcm-benchmarks` (or whatever) flag: Skip the benchmarks and don't generate the weights.

(of course you can also invert and thus skip by default an error on sth like `--benchmark-skips-are-errors`)
I just want to make sure that as a parachain dev I have an easy way to notice skips that shouldn't be there, especially in development/while testing (while still supporting your very legitimate point of intentionally not supporting all XCMs everywhere).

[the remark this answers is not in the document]
shawntabrizi: yes, this is a reasonable feature, and I can add it to substrate benchmarking CLI.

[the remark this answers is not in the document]
shawntabrizi: https://github.com/paritytech/substrate/issues/10396

shawntabrizi:
@apopiak we should be good to go here

runtime/westend/src/weights/xcm/pallet_xcm_benchmarks_generic.rs
```diff
@@ -1,73 +1,132 @@
-// this whole file is temp, and will be replaced in the future TODO
+// Copyright 2017-2021 Parity Technologies (UK) Ltd.
```
kianenigma, line 1:
```suggestion
// Copyright 2021 Parity Technologies (UK) Ltd.
```

xcm/pallet-xcm-benchmarks/src/generic/benchmarking.rs
```diff
@@ -0,0 +1,319 @@
...
+	}
+
+	clear_origin {
+		let mut executor = new_executor::<T>(Default::default());
+		let instruction = Instruction::ClearOrigin;
+		let xcm = Xcm(vec![instruction]);
+	} : {
+		executor.execute(xcm)?;
+	} verify {
+		assert_eq!(executor.origin, None);
```
kianenigma, line 204: what is this value before the benchmark?

xcm/pallet-xcm-benchmarks/src/generic/mock.rs
```diff
@@ -0,0 +1,171 @@
...
+
+	fn transact_origin() -> Option<MultiLocation> {
+		Some(Default::default())
+	}
+
+	fn subscribe_origin() -> Option<MultiLocation> {
+		Some(Default::default())
+	}
+
+	fn claimable_asset() -> Option<(MultiLocation, MultiLocation, MultiAssets)> {
```
kianenigma, line 146: Why is this not returning `BenchmarkError` directly? You usually do `.map_err(|_| BenchmarkError::Skip)` inside the benchmark code. This seems like a decision that should be made by the top level runtime.

kianenigma: As in, for some cases I might want to do `Override` and for other `Skip`.

kianenigma (approved):
https://github.com/paritytech/polkadot/pull/3940/files#r759965208 might be a worthwhile consideration, otherwise LGTM

xcm/pallet-xcm-benchmarks/src/generic/mock.rs
apopiak: :+1: If that's possible I really like that idea.

apopiak (approved):
:+1: 

Though I like Kian's suggestion of giving the parachain devs control over the benchmark error directly.

xcm/pallet-xcm-benchmarks/src/generic/mock.rs
shawntabrizi: yup, great suggestion

shawntabrizi:
bot merge

shawntabrizi:
bot merge

PR #3943, opened 2021-09-27
Rococo: ability to programatically assign slots to teams

### LINKED ISSUE #3942: Rococo: ability to programatically assign slots to teams
Add features to the Rococo runtime to be able to programatically assign slots to teams to test their parachain on Rococo.
We envision two categories of slot assignments to accommodate for different contexts:

1. **Long lasting connections** permanent slots reserved for teams that are connected as parachains on Kusama, or teams that have proven extensive contributions to the ecosystem.
2. **Short-period testing connections** temporary slots for teams that need to test their WASMs before committing them on chain and starting their crowdloans.

Proposed approach is to add sudo calls to the `parasSudoWrapper` pallet to assign permanent / temporary slots, and implement logic to go round-robin on teams that don't have a permanent connection, and assigns them slots every X amount of team (downgrading from / upgrading to parachain as needed).

stiiifff:
Resolves #3942 

Added new `assigned_slots` pallet with calls to assign permanent or temporary parachain slots on Rococo:

```rust
pub enum SlotLeasePeriodStart {
  Current,
  Next,
}

/// Assign a permanent parachain slot
pub fn assign_perm_parachain_slot(origin: OriginFor<T>, id: ParaId) -> DispatchResult {}

/// Assign a temporary parachain slot
pub fn assign_temp_parachain_slot(origin: OriginFor<T>, id: ParaId, lease_period_start: SlotLeasePeriodStart) -> DispatchResult {}

/// Unassign a permanent or temporary parachain slot
pub fn unassign_parachain_slot(origin: OriginFor<T>, id: ParaId) -> DispatchResult {}
```

**Flow**
- After a para was initialized with `parasSudoWrapper.sudoScheduleParaInitialize`) and onboarded as a parathread.
- Assign a temporary parachain slot with `assignedSlots.assignTempParachainSlot`, specifying the para Id and lease period start (Current lease period or next one).
- This will create a lease and the parathread will be upgraded to a parachain when its start lease period begins.
- When a slot's lease ends, the parachain is automatically downgraded to a parathread.
- The `assigned_slots` pallet keeps track of temporary slots and distributes turns fairly between temp slots (best effort).
- You can also assign a permanent (or long-running) slot with `assignedSlots.assignPermParachainSlot` or unassign a slot (permanent or temporary) with `assignedSlots.unassignParachainSlot` (this clears any lease which downgrades the para back to a parathread).
- The implementation (in the `assigned_slots` pallet) relies on the existing slot lease mechanism in the `slots` pallet.
- The following parameters have to be configured (with suggested values for Rococo):
```rust
pub const LeasePeriod: BlockNumber = 7 * DAYS;
// The number of lease periods a permanent parachain slot lasts
pub const PermanentSlotLeasePeriodLength: u32 = 26; 
// The number of lease periods a temporary parachain slot lasts
pub const TemporarySlotLeasePeriodLength: u32 = 1;
// The max number of permanent slots that can be registered
pub const MaxPermanentSlots: u32 = 5;
// The max number of temporary slots that can be registered
pub const MaxTemporarySlots: u32 = 20;
// The max number of temporary slots to be scheduled per lease periods
pub const MaxTemporarySlotPerLeasePeriod: u32 = 5;
```


al3mart (approved):
Looks good to me

stiiifff:
Still need to write tests, but requesting early feedback.

runtime/common/src/assigned_slots.rs
```diff
@@ -0,0 +1,1086 @@
+// Copyright 2020 Parity Technologies (UK) Ltd.
```
apopiak, line 1:
```suggestion
// Copyright 2021 Parity Technologies (UK) Ltd.
```

runtime/common/src/assigned_slots.rs
```diff
@@ -0,0 +1,1086 @@
...
+};
+use scale_info::TypeInfo;
+use sp_runtime::{
+	traits::{One, Saturating, Zero},
+	ArithmeticError,
+};
+use sp_std::prelude::*;
+
+#[derive(Encode, Decode, Clone, Copy, Eq, PartialEq, RuntimeDebug, TypeInfo)]
+pub enum SlotLeasePeriodStart {
```
apopiak, line 47: some docs?

runtime/common/src/assigned_slots.rs
```diff
@@ -0,0 +1,1086 @@
...
+	Current,
+	Next,
+}
+
+#[derive(Encode, Decode, Clone, PartialEq, Eq, Default, RuntimeDebug, TypeInfo)]
+pub struct ParachainTemporarySlot<AccountId, LeasePeriod> {
```
apopiak, line 53: some docs?

runtime/common/src/assigned_slots.rs
```diff
@@ -0,0 +1,1086 @@
...
+			} else {
+				0
+			}
+		}
+	}
+
+	#[pallet::call]
+	impl<T: Config> Pallet<T> {
+		/// Assign a permanent parachain slot
+		#[pallet::weight((1_000, DispatchClass::Operational))]
```
apopiak, line 167: this is not benchmarked because it will not end up in production?

runtime/common/src/assigned_slots.rs
```diff
@@ -0,0 +1,1086 @@
...
+						SlotLeasePeriodStart::Next => current_lease_period + One::one(),
+					},
+					period_count: T::TemporarySlotLeasePeriodLength::get().into(),
+					last_lease: None,
+					lease_count: 0,
+				}),
+			);
+
+			if lease_period_start == SlotLeasePeriodStart::Current {
+				Self::allocate_temporary_slot_leases(current_lease_period)?;
```
apopiak, line 263: This can fail even though you've changed storage above. Are you sure this should not be atomic?

runtime/common/src/assigned_slots.rs
```diff
@@ -0,0 +1,1086 @@
...
+				TemporarySlots::<T>::remove(id);
+			}
+
+			// Clean any para lease
+			Self::clear_slot_leases(origin.clone(), id)?;
+
+			// Force downgrade to parathread (if needed) before end of lease period
+			if paras::Pallet::<T>::lifecycle(id) == Some(ParaLifecycle::Parachain) {
+				runtime_parachains::schedule_parachain_downgrade::<T>(id)
+					.map_err(|_| Error::<T>::CannotDowngrade)?;
```
apopiak, line 297: Similar note here about atomicity

runtime/common/src/assigned_slots.rs
```diff
@@ -0,0 +1,1086 @@
...
+			Ok(())
+		}
+	}
+}
+
+impl<T: Config> Pallet<T> {
+	fn allocate_temporary_slot_leases(lease_period_index: LeasePeriodOf<T>) -> DispatchResult {
+		let mut active_temp_slots = 0u32;
+		let mut pending_temp_slots = Vec::new();
+		TemporarySlots::<T>::iter().for_each(|(para, maybe_slot)| {
```
apopiak, line 309: This is only constrained to the number of parathreads in the system (see note above) and could thus become very heavy.

runtime/common/src/assigned_slots.rs
```diff
@@ -0,0 +1,1086 @@
...
+				a.1.lease_count
+					.cmp(&b.1.lease_count)
+					.then_with(|| a.1.last_lease.cmp(&b.1.last_lease))
+					.then_with(|| a.0.cmp(&b.0))
+			});
+
+			let slots_to_be_upgraded = pending_temp_slots
+				.iter()
+				.take((T::MaxTemporarySlotPerLeasePeriod::get() - active_temp_slots) as usize)
+				.collect::<Vec<_>>();
```
apopiak, line 351: this collect seems unnecessary to me, you just iterate by reference anyway

runtime/common/src/assigned_slots.rs
```diff
@@ -0,0 +1,1086 @@
...
+	}
+
+	fn lease_period() -> LeasePeriodOf<T> {
+		T::Leaser::lease_period()
+	}
+
+	fn manage_lease_period_start(lease_period_index: LeasePeriodOf<T>) -> Weight {
+		// Note: leases that have ended in previous lease period,
+		// should have been cleaned in slots pallet.
+		let _ = Self::allocate_temporary_slot_leases(lease_period_index);
```
apopiak, line 412: This drops any errors.
I would add logging, e.g. sth lik:
```suggestion
		let res = Self::allocate_temporary_slot_leases(lease_period_index);
		res.map_err(|e| log::error!(target: "assigned_slots", "Allocating slots failed for lease period {}, with: {}", lease_period_index, e));
```

runtime/common/src/assigned_slots.rs
```diff
@@ -0,0 +1,1086 @@
...
+					manager,
+					period_begin: match lease_period_start {
+						SlotLeasePeriodStart::Current => current_lease_period,
+						SlotLeasePeriodStart::Next => current_lease_period + One::one(),
+					},
+					period_count: T::TemporarySlotLeasePeriodLength::get().into(),
+					last_lease: None,
+					lease_count: 0,
+				}),
+			);
```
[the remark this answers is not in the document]
apopiak: Consider adding a `TemporarySlotsCount` similar to the permanent slots one.

runtime/common/src/assigned_slots.rs
```diff
@@ -0,0 +1,1086 @@
...
+// Polkadot is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with Polkadot.  If not, see <http://www.gnu.org/licenses/>.
+
+//! This pallet allows to assign permanent (long-lived) or temporary
+//! (short-lived) parachain slots to paras, leveraging the existing
```
apopiak, line 18: nitpick: the permanent leases are not permanent, but just long-lived :stuck_out_tongue:

runtime/common/src/assigned_slots.rs
```diff
@@ -0,0 +1,1086 @@
...
+//! parachain slot lease mechanism. Temporary slots are given turns
+//! in a fair (though best-effort) manner.
+//! The dispatchables must be called from the configured origin
+//! (typically `Sudo` or a governance origin).
+//! This pallet is mostly to be used on test relay chain (e.g. Rococo).
```
apopiak, line 23: At least as currently written it should only ever be used in testnets.

runtime/common/src/assigned_slots.rs
```diff
@@ -0,0 +1,1086 @@
...
+					.map_err(|_| Error::<T>::CannotDowngrade)?;
+			}
+
+			Ok(())
+		}
+	}
+}
+
+impl<T: Config> Pallet<T> {
+	fn allocate_temporary_slot_leases(lease_period_index: LeasePeriodOf<T>) -> DispatchResult {
```
apopiak, line 306: docs?

apopiak (commented):
I'm probably being too harsh for a pallet that will only end up on a test chain, but I don't like that atomicity isn't kept and that a weakly bounded storage map is iterated.
Apart from that :+1: for making this happen.

runtime/common/src/assigned_slots.rs
stiiifff: Added weight of the delegated call on the `slots` pallet e.g. `1_000 + <T as slots::Config>::WeightInfo::force_lease()`. I think it's a reasonable approximation.

runtime/common/src/assigned_slots.rs
stiiifff: Add a `MaxTemporarySlots` param which constraints the nbr of temp slots.

runtime/common/src/assigned_slots.rs
stiiifff: Not a biggie I think.

runtime/common/src/assigned_slots.rs
stiiifff: Good point, reworked it.

runtime/common/src/assigned_slots.rs
stiiifff: Fair enough .. they're permanent until they're unassigned (and for as long the `PermanentSlotLeasePeriodLength` lasts). It's just less of a mouthful than long-lived (or short-lived for temporary).

runtime/common/src/assigned_slots.rs
stiiifff: Rephrased to be more categorical.

runtime/common/src/assigned_slots.rs
```diff
@@ -303,6 +394,18 @@ pub mod pallet {
 }
 
 impl<T: Config> Pallet<T> {
+	/// Allocate temporary slot leases up to `MaxTemporarySlotPerLeasePeriod` per lease period.
+	/// Beyong the already active temporary slot leases, this function will activate more leases
```
apopiak, line 398:
```suggestion
	/// Beyond the already active temporary slot leases, this function will activate more leases
```

runtime/common/src/assigned_slots.rs
apopiak: sure :shrug:

apopiak (approved):

runtime/common/src/assigned_slots.rs
stiiifff: But you're right, I did the change.

runtime/common/src/assigned_slots.rs
```diff
@@ -0,0 +1,1280 @@
...
+type LeasePeriodOf<T> =
+	<<T as Config>::Leaser as Leaser<<T as frame_system::Config>::BlockNumber>>::LeasePeriod;
+
+#[frame_support::pallet]
+pub mod pallet {
+	use super::*;
+
+	#[pallet::pallet]
+	#[pallet::generate_store(pub(super) trait Store)]
+	pub struct Pallet<T>(_);
```
shawntabrizi, line 84:
```suggestion
	#[pallet::generate_store(pub(super) trait Store)]
	#[pallet::generate_storage_info]
	pub struct Pallet<T>(_);
```

runtime/common/src/assigned_slots.rs
```diff
@@ -0,0 +1,1280 @@
...
+
+	#[pallet::config]
+	#[pallet::disable_frame_system_supertrait_check]
```
shawntabrizi, line 87: why though?

stiiifff: AFAIU the `runtime_parachains::configuration`'s Config trait already derives from frame_system's Config trait, so there is no need to derive explicitly from it again.

runtime/common/src/assigned_slots.rs
```diff
@@ -0,0 +1,1281 @@
...
+
+			// We didn't return early above, so we didn't do anything.
+			0
+		}
+	}
+
+	#[pallet::call]
+	impl<T: Config> Pallet<T> {
+		/// Assign a permanent parachain slot and immediately create a lease for it.
+		#[pallet::weight((1_000 + <T as slots::Config>::WeightInfo::force_lease(), DispatchClass::Operational))]
```
shawntabrizi, line 205: If you don't have real benchmarks, better to set this to something like 1/10th the total block weight, which is certainly an over-estimate.

shawntabrizi: and add a todo

runtime/common/src/assigned_slots.rs
```diff
@@ -0,0 +1,1281 @@
...
+					id,
+					current_lease_period,
+					// Check current lease & next one
+					current_lease_period.saturating_add(
+						T::BlockNumber::from(2u32)
+							.saturating_mul(T::PermanentSlotLeasePeriodLength::get().into())
+					)
+				),
+				Error::<T>::OngoingLeaseExists
+			);
```
shawntabrizi, line 230: You are already placing a hard dependency on the `slots` pallet, so probably you can do a more efficient check here directly with the storage struct.

runtime/common/src/assigned_slots.rs
```diff
@@ -0,0 +1,1281 @@
...
+			<PermanentSlotCount<T>>::try_mutate(|count| -> DispatchResult {
+				// Permanent slot assignment fails if a lease cannot be created
+				Self::configure_slot_lease(
+					id,
+					manager,
+					current_lease_period,
+					T::PermanentSlotLeasePeriodLength::get().into(),
+				)
+				.map_err(|_| Error::<T>::CannotUpgrade)?;
+				*count = count.checked_add(One::one()).ok_or(ArithmeticError::Overflow)?;
```
shawntabrizi, line 246: configure_slot_lease appears to write to storage, thus you are not allowed to fail here on the `checked_add`.

Either do the checked add before the storage write, or use saturating, or anything but return an error after a storage write.

runtime/common/src/assigned_slots.rs
```diff
@@ -0,0 +1,1281 @@
...
+				PermanentSlots::<T>::insert(id, Some(current_lease_period));
```
shawntabrizi, line 247: You should be tracking the beginning and end lease period. You may choose to update `T::PermanentSlotLeasePeriodLength::get()` at some point, and then you wont have correct data on chain about when the person actually ends.

And whenever trying to calculate the final slot of a parachain, you should use the storage item rather than `T::PermanentSlotLeasePeriodLength::get()` added to the stored lease period.

runtime/common/src/assigned_slots.rs
```diff
@@ -0,0 +1,1281 @@
...
+
+	/// Assigned temporary slots.
+	#[pallet::storage]
+	#[pallet::getter(fn temporary_slots)]
+	pub type TemporarySlots<T: Config> = StorageMap<
+		_,
+		Twox64Concat,
+		ParaId,
+		Option<ParachainTemporarySlot<T::AccountId, LeasePeriodOf<T>>>,
+		ValueQuery,
```
shawntabrizi, line 143: This doesn't make sense as is.

```suggestion
		ParachainTemporarySlot<T::AccountId, LeasePeriodOf<T>>,
		OptionQuery,
```

runtime/common/src/assigned_slots.rs
shawntabrizi: Inserting `Some` here was a trigger that something was funny with your storage config, and above I made a suggestion to use an `OptionQuery` rather than a `ValueQuery + Option`

runtime/common/src/assigned_slots.rs
```diff
@@ -0,0 +1,1281 @@
...
+		pub fn assign_temp_parachain_slot(
+			origin: OriginFor<T>,
+			id: ParaId,
+			lease_period_start: SlotLeasePeriodStart,
+		) -> DispatchResult {
+			T::AssignSlotOrigin::ensure_origin(origin)?;
+
+			let manager = T::Registrar::manager_of(id).ok_or(Error::<T>::ParaDoesntExist)?;
+
+			ensure!(T::Registrar::is_parathread(id), Error::<T>::NotParathread,);
```
shawntabrizi, line 267:
```suggestion
			ensure!(T::Registrar::is_parathread(id), Error::<T>::NotParathread);
```

runtime/common/src/assigned_slots.rs
```diff
@@ -0,0 +1,1281 @@
...
+						temp_slot.period_count,
+					) {
+						Ok(_) => {
+							temp_slot.last_lease = Some(temp_slot.period_begin);
+							temp_slot.lease_count += 1;
+							ActiveTemporarySlotCount::<T>::try_mutate::<_, ArithmeticError, _>(
+								|count| {
+									*count = count
+										.checked_add(One::one())
+										.ok_or(ArithmeticError::Overflow)?;
```
shawntabrizi, line 325: cant fail here

runtime/common/src/assigned_slots.rs
```diff
@@ -0,0 +1,1086 @@
...
+			!pending_temp_slots.is_empty()
+		{
+			// Sort by lease_count, favoring slots that had no or less turns first
+			// (then by last_lease index, and then Para ID)
+			pending_temp_slots.sort_by(|a, b| {
+				a.1.lease_count
+					.cmp(&b.1.lease_count)
+					.then_with(|| a.1.last_lease.cmp(&b.1.last_lease))
+					.then_with(|| a.0.cmp(&b.0))
+			});
```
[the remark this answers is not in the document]
shawntabrizi: +1

I think a lot of this code would be more readable with tmp variables

shawntabrizi:
Where is the programatic part of this?

You will run some off-chain bot that can call these functions?

Also, might be worth just implementing a "permanent" lease into the main `slots` pallet. Something we are seeming to need anyway.

stiiifff:
Thx @shawntabrizi, gonna revise the code as per your comments.

> Where is the programatic part of this? You will run some off-chain bot that can call these functions?

That was the phrasing of the initial issue, I realise now it is confusing, thank you for pointing it out.
The idea for now is to have a simple mechanism to assign perm/temp slots manually (sudo or governance) as mentioned here: https://polkadot.network/blog/rococo-revamp-becoming-a-community-parachain-testbed/.
It might be automated at a later stage (no idea how as of now).

> Also, might be worth just implementing a "permanent" lease into the main `slots` pallet. Something we are seeming to need anyway.

Fair enough. I took the approach of doing it in a separate pallet, which is only active on Rococo / Westend, so as to avoid borking the slots pallet in this first iteration. I'd rather get this out in this form, and then move perm slots to the slots pallet in a separare PR, what you think ?

shawntabrizi:
Is the end result much different than `force_lease`?

https://github.com/paritytech/polkadot/blob/master/runtime/common/src/slots.rs#L164

Could just change `root` to some origin like you have

stiiifff:
Not much indeed. Then I'd remove the permanent -related stuff in favour of just using `slots.force_lease` (w/ a configurable origin) for any arbitrarily assigned long-term slot.
The new pallet would then only contain logic for short-term slots assigned in a rolling manner (could rename the pallet to `rolling_slots`).

runtime/common/src/assigned_slots.rs
stiiifff: > You should be tracking the beginning and end lease period. You may choose to update `T::PermanentSlotLeasePeriodLength::get()` at some point, and then you wont have correct data on chain about when the person actually ends.
> 
> And whenever trying to calculate the final slot of a parachain, you should use the storage item rather than `T::PermanentSlotLeasePeriodLength::get()` added to the stored lease period.

Probably not needed anymore if (re)moving the permanent slots -related logic to the `slots` pallet.

runtime/common/src/assigned_slots.rs
stiiifff: Using `contains-key` on the leases storage map would not be enough .. then I'd have to re-implement some of the logic that's nicely encapsulated (and tested) in `T::Leaser::already-leased` function. Also this call shouldn't be called that often, so I'd go for correctness over further optimisation, what you think ?

shawntabrizi:
In general i don't see that much issue with this PR, especially if it is only going on Rococo, but the "automation" part of this PR is quite questionable :)

Feels like all of this can be automated with more flexibility using a permissioned origin and some off-chain application.

stiiifff:
One could argue anything can be automated with more flexibility using a permissioned origin and some off-chain app 😉
Jokes aside, I think the current logic fits the current needs for Rococo (scratch my last comment).
Also, as recommended, I've added the ForceOrigin to the slots pallet, which adds flexibility for other chains as well.
Let me know what you think.

stiiifff:
@shawntabrizi Anything else you'd like to see revised ?

stiiifff:
@shawntabrizi ?

shawntabrizi:
@stiiifff please remember to update the Substrate docs on the new process you and your team has established: https://docs.substrate.io/tutorials/v3/cumulus/rococo/

https://github.com/substrate-developer-hub/substrate-docs/issues/665

stiiifff:
Thx ! I had not seen that issue, I will update the docs.

PR #3946, opened 2021-09-27
Silence some alerts due to overly verbose warnings.

eskimor:


node/network/availability-distribution/src/requester/fetch_task/mod.rs
```diff
@@ -344,7 +344,7 @@ impl RunningTask {
 				Err(TaskError::PeerError)
 			},
 			Err(RequestError::NetworkError(err)) => {
-				tracing::warn!(
+				tracing::info!(
```
bkchr, line 347: What is the benefit of decreasing the severity of this? They will still be printed.

This question applies to all of these changes ;)

eskimor: Yes, that was the intention. Main focus was to not make anyone alerted by this (which happens for warnings as far as I understand), but getting lots of these is still a good indicator that something is wrong with the network, so I did not want them to be completely silent, as seeing them immediately would speed up debugging of issues.

eskimor: Ok, I see now how the title of the PR is misleading - should read silence alerts due to verbose warnings.

eskimor: That being said, happy to make it debug even, if you guys feel that this would be more appropriate.

bkchr: I mean these messages correspond to real errors that should be investigated. Decreasing the severity will hide them, but not fix them :P 

I'm for solving the underlying issue instead of hiding it.

eskimor: Well, network errors are something that can happen without there being any real issue apart from congestion. The connectivity issues on Kusama, which make these errors more frequent, should be fixed, but that's a different issue.

bkchr: Yeah for sure, but when they are irrelevant, you can directly decrease them to debug messages. 

`info` doesn't solve here anything, it will still be printed.

eskimor: So the issue was not that logs became too spammy, the issue was that people suspected something is wrong because they saw them (which is kind of true on Kusama, but that's a different story).

bkchr: The logging string even contains the word "error". People will still report this ;)

bkchr: And it is also weird to log anything that is not "info" as info. An error should either be printed as warn/error or hidden in as debug

eskimor: Fair :+1: . Ok, will need to make up my mind whether those should stay `warn` or should be `debug`. :thinking:

If `debug` then probably together with introducing some metrics with adequate alerts.

eskimor:
Closing for now.

eskimor:
info -> debug

bkchr (approved):

node/network/statement-distribution/src/error.rs
```diff
@@ -87,8 +87,12 @@ pub enum NonFatal {
 	#[error("Relay parent could not be found in active heads")]
 	NoSuchHead(Hash),
 
+	/// Received message from actually disconnected peer.
+	#[error("Message from not connected peer")]
+	NoSuchPeer(PeerId),
```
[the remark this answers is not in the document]
eskimor: It should not be, but it is a representable state - so if it happened we need to do something -> report it & fix the bug.

PR #3960, opened 2021-09-29
add slot offset for slots

rphmeier:
In progress; for offsetting lease periods by a preconfigured amount of blocks.


rphmeier:
Superseded

PR #3964, opened 2021-09-29
assure disputes are checked too

drahnr:
Currently the disputes feature is disabled in CI, but that's our focus of work now.

This adds an extra command to verify the `"disputes"` enabled `disputes-coordinator` is functional.


drahnr:
Duplicate of #3962

PR #3967, opened 2021-09-29
Update RELEASE.md to reflect reality

s3krit:


RELEASE.md
```diff
@@ -4,15 +4,20 @@ Polkadot Release Process
...
+  
 ### Notes
 * The release-candidate branch *must* be made in the paritytech/polkadot repo in
 order for release automation to work correctly
 * Any new pushes/merges to the release-candidate branch (for example,
-refs/heads/release-v0.8.26) will result in the rc index being bumped (e.g., v0.8.26-rc1
-to v0.8.26-rc2) and new wasms built.
+  refs/heads/release-v0.8.26) will result in the rc index being bumped (e.g., v0.8.26-rc1
+  to v0.8.26-rc2) and new wasms built. If you need to merge several changes into
+a new release-candidate, it's advised to do this as a single PR against the rc branch
```
chevdor, line 14: There is no **rc branch** per say but a **release** branch generating new RCs (tags).

RELEASE.md
```diff
@@ -4,15 +4,20 @@ Polkadot Release Process
...
+* There is a step on the auto-created github issue checklist that specifies we should
```
chevdor, line 15: Would that now be better to update the mentioned step instead of having contradictory info here ?

RELEASE.md
```diff
@@ -38,12 +43,9 @@ automated and require no human action.
...
   or all of the manual tests again.
-7. Once happy with the release-candidate, perform the release using the release
-  script located at `scripts/release.sh` (or perform the steps in that script
-  manually):
-  - `./scripts/release.sh v0.8.26`
-8. NOACTION: The HEAD of the `release` branch will be tagged with `v0.8.26`,
-  and a final draft release will be created on Github.
+7. Once happy with the release-candidate (i.e., the checks in the auto-created
+  checklist are completed) tag the tip of the release-candidate branch:
+  - `git tag -s -m 'v0.8.26' v0.8.26; git push --tags`
```
chevdor, line 48: I would advise against using `git push --tags` (you don't want to know the extra tags I have on my machine :)).
It would be safer to specify the tag and remote explicitely and using a VAR prevents typos.

Untested (can be collapsed to a single line):
```
VERSION=0.8.26
REMOTE=origin
git tag -s -m "v${VERSION}" v${VERSION}; git push "v${VERSION}" ${REMOTE}
```

It may be a good option to bring back the `release.sh` with the content above.
We can use `REMOTE=origin` as default and complain if version is not set.

In the script, I would suggest to:
- print a summary of what will be done (so we can stop typical errors such as being about to create tag `vv1.2.3` instead of `v1.2.3` because we passed the wrong `VERSION`
- pause with a read and wait for confirmation to move on. Could be done at start or just before the push
- apply

RELEASE.md
```diff
@@ -52,5 +54,6 @@ released version of Polkadot, without taking *every* change to `master` since
 the last release. For example, in the event of a security vulnerability being
 found, where releasing a fixed version is a matter of some expediency. In cases
 like this, the fix should first be merged with master, cherry-picked to a branch
-forked from `release`, tested, and then finally merged with `release`. A
-sensible versioning scheme for changes like this is `vX.Y.Z-1`.
+forked from the last release-candidate branch, tested, and then finally merged
+with the release-candidate branch. A sensible versioning scheme for changes like
+this is `vX.Y.Z-1`, though this can cause problems with the RPM packaging.
```
chevdor, line 59: So we remain with this potential issue or do you think we can work around ?
If we expect problems, we should point to a solutioN/workaround.

chevdor (changes requested):

PR #3980, opened 2021-10-01
Allow an Offset to Lease Periods

shawntabrizi:
Supersedes: https://github.com/paritytech/polkadot/pull/3960

This PR allows the entire crowdloan, slots, auction system to be delayed by a constant block number to do fine adjustments to when the lease period cycle should start. We do this by extending lease period 0 by `LeaseOffset` blocks, which offsets all lease periods by the same amount.

A few core level API changes have been introduced which ensures that no existing APIs break due to this new feature. By doing this, we ensure there is a single point of calculation of the current lease period, and every other API just reads from it. So other APIs should not be exposed to any concept of the lease period offset, and the underlying behavior of calculating the current lease period is completely opaque.

* `fn lease_period()` was renamed to `fn lease_period_length()`, returns the offset in addition to the length, and is removed from the public api. We only use it in benchmarking, so we make sure to hide it behind a feature flag so that no external logic would use it unexpectedly.
* `fn lease_period_index()` use to always return the CURRENT lease period for the current block number. We updated the API so that it always accepts an input of the block number we want to query, and it will return both the current lease period for that block, and if that was the first block of a new lease period.
* In order to support the traits which span the different pallets, we had to expose `BlockNumber` as a generic argument, slightly changing the signature of the `Leaser` and `Auctioneer` traits. No big deal.
* Docs updated for these changed traits.
* A new test was added to verify `fn lease_period_index()` behaves correct with and without an offset.
* An end-to-end integration test was updated to verify everything behaves as expected with a bunch of different offset values.


runtime/common/src/slots.rs
```diff
@@ -427,20 +431,27 @@ impl<T: Config> Leaser for Pallet<T> {
...
-		<frame_system::Pallet<T>>::block_number() / T::LeasePeriod::get()
+	fn lease_period_index(b: T::BlockNumber) -> (Self::LeasePeriod, bool) {
+		// Note that lease lease period 0 is artificially extended by the lease offset.
+		let offset_block_now = b.saturating_sub(T::LeaseOffset::get());
+
+		let lease_period = offset_block_now / T::LeasePeriod::get();
+		let first_block = (offset_block_now % T::LeasePeriod::get()).is_zero();
+
+		(lease_period, first_block)
 	}
```
shawntabrizi, line 447: Here is the core change. This is the one source of truth for calculating the current lease period index given some offset. All other APIs must use this to do stuff.

runtime/common/src/slots.rs
```diff
@@ -427,20 +431,28 @@ impl<T: Config> Leaser for Pallet<T> {
 			.unwrap_or_else(Zero::zero)
 	}
 
-	fn lease_period() -> Self::LeasePeriod {
-		T::LeasePeriod::get()
+	#[cfg(any(feature = "runtime-benchmarks", test))]
+	fn lease_period_length() -> (T::BlockNumber, T::BlockNumber) {
+		(T::LeasePeriod::get(), T::LeaseOffset::get())
```
shawntabrizi, line 436: This API is only needed for benchmarks, so we hide it from the public API, and therefore eliminate any chance that a user may try to calculate this on their own.

runtime/common/src/slots.rs
```diff
@@ -427,20 +431,28 @@ impl<T: Config> Leaser for Pallet<T> {
...
-	fn lease_period_index() -> Self::LeasePeriod {
-		<frame_system::Pallet<T>>::block_number() / T::LeasePeriod::get()
+	fn lease_period_index(b: T::BlockNumber) -> (Self::LeasePeriod, bool) {
+		// Note that lease lease period 0 is artificially extended by the lease offset.
+		let offset_block_now = b.saturating_sub(T::LeaseOffset::get());
+
+		let lease_period = offset_block_now / T::LeasePeriod::get();
+		let first_block = (offset_block_now % T::LeasePeriod::get()).is_zero();
+
+		(lease_period, first_block)
```
shawntabrizi, line 446: Here is the core change. This is the one source of truth for calculating the current lease period index given some offset. All other APIs must use this to do stuff.

runtime/common/src/slots.rs
```diff
@@ -427,20 +431,28 @@ impl<T: Config> Leaser for Pallet<T> {
...
 	}
 
 	fn already_leased(
 		para_id: ParaId,
 		first_period: Self::LeasePeriod,
 		last_period: Self::LeasePeriod,
 	) -> bool {
-		let current_lease_period = Self::lease_period_index();
+		let now = frame_system::Pallet::<T>::block_number();
+		let current_lease_period = Self::lease_period_index(now).0;
```
shawntabrizi, line 455: In most places where we just called `lease_period_index` before, we now need to pass the current block number. This is the same logic and complexity as before, just explicit now.

runtime/common/src/slots.rs
```diff
@@ -832,7 +848,7 @@ mod tests {
 	}
 
 	#[test]
-	fn lease_out_current_lease_period() {
+	fn lease_out_current_lease_period_length() {
```
shawntabrizi, line 851:
```suggestion
	fn lease_out_current_lease_period() {
```

runtime/common/src/integration_tests.rs
```diff
@@ -298,171 +300,176 @@ fn last_event() -> Event {
...
-			None,
-		));
-		let crowdloan_account = Crowdloan::fund_account_id(ParaId::from(para_2));
-
-		// Auction ending begins on block 100, so we make a bid before then.
-		run_to_block(90);
-
-		Balances::make_free_balance_be(&10, 1_000_000_000);
-		Balances::make_free_balance_be(&20, 1_000_000_000);
+	for offset in [0u32, 50, 100, 200].iter() {
```
shawntabrizi, line 307: This test is basically unchanged except introducing `offset` into the timeline of the setup.

runtime/common/src/slots.rs
```diff
@@ -427,20 +431,34 @@ impl<T: Config> Leaser for Pallet<T> {
...
+
+		// Special logic to handle lease period 0 being extended by the offset.
+		let first_block = if lease_period.is_zero() {
+			// Lease period 0 has their first block on only block 0.
+			b.is_zero()
+		} else {
+			(offset_block_now % T::LeasePeriod::get()).is_zero()
+		};
+
+		(lease_period, first_block)
```
shawntabrizi, line 452: Here is the core change. This is the one source of truth for calculating the current lease period index given some offset. All other APIs must use this to do stuff.

crystalin:
@shawntabrizi , isn't this feature going to bring even more complexity to the parachain lease computations? I still have today to explain people why Moonriver is not ending 48 weeks after it started :p

I'm curious, what is the use case for this change ?

rphmeier:
@crystalin It's very useful for a relay chain's governance to decide the exact date at which parachains go live.

rphmeier (approved):

runtime/common/src/auctions.rs
```diff
@@ -779,12 +788,18 @@ mod tests {
...
-			10
+		fn lease_period_length() -> (BlockNumber, BlockNumber) {
+			(10, 0)
 		}
 
-		fn lease_period_index() -> Self::LeasePeriod {
-			(System::block_number() / Self::lease_period()).into()
+		fn lease_period_index(b: BlockNumber) -> (Self::LeasePeriod, bool) {
+			let (lease_period_length, offset) = Self::lease_period_length();
+			let b = b.saturating_sub(offset);
```
gavofyork, line 797: Do we want lease period 0 to be [0, offset+length)? It made more sense to me to have it be [offset, offset+length) and return an `Option<Self::LeasePeriod>` which is `None` if `now < offset`.

runtime/common/src/slots.rs
```diff
@@ -427,20 +431,34 @@ impl<T: Config> Leaser for Pallet<T> {
...
+	#[cfg(any(feature = "runtime-benchmarks", test))]
+	fn lease_period_length() -> (T::BlockNumber, T::BlockNumber) {
+		(T::LeasePeriod::get(), T::LeaseOffset::get())
 	}
 
-	fn lease_period_index() -> Self::LeasePeriod {
-		<frame_system::Pallet<T>>::block_number() / T::LeasePeriod::get()
+	fn lease_period_index(b: T::BlockNumber) -> (Self::LeasePeriod, bool) {
+		// Note that lease lease period 0 is artificially extended by the lease offset.
+		let offset_block_now = b.saturating_sub(T::LeaseOffset::get());
```
gavofyork, line 441: See previous comment. I would have expected this to be:

```suggestion
		let offset_block_now = b.checked_sub(T::LeaseOffset::get())?;
```

emostov (approved):
Code LGTM

w.r.t. Gav's suggestion I think it could be nice for a relay chain to have fine grained control over the start of lease period 0. But given that the target use case is already past that point I don't see a compelling reason to add the little bit of complexity in handling an `Option` from `lease_period_index`.

runtime/common/src/auctions.rs
shawntabrizi: let me look how this will affect the API and any error handling.

shawntabrizi:
@rphmeier take a look at the last changes made and if it is good for you, feel free to merge.

PR #4002, opened 2021-10-03
Use background tasks properly in candidate-validation

rphmeier:
Also does some refactoring of error handling and runtime API requests.

### DIFF
--- a/node/core/candidate-validation/src/lib.rs
+++ b/node/core/candidate-validation/src/lib.rs
@@ -35,7 +35,7 @@ use polkadot_node_subsystem::{
 		CandidateValidationMessage, RuntimeApiMessage, RuntimeApiRequest, ValidationFailed,
 	},
 	overseer, FromOverseer, OverseerSignal, SpawnedSubsystem, SubsystemContext, SubsystemError,
-	SubsystemResult,
+	SubsystemResult, SubsystemSender,
 };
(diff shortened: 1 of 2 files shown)


ordian (approved):
Looks reasonable.

I wonder if we should be using `spawn_blocking` because of wasm decompression thing might be CPU heavy, but OTOH we don't want to spawn too many threads.

node/core/candidate-validation/src/lib.rs
```diff
@@ -163,50 +167,68 @@ where
...
+	receiver
+		.await
+		.map_err(|_| {
+			tracing::debug!(target: LOG_TARGET, ?relay_parent, "Runtime API request dropped");
+
+			RuntimeRequestFailed
+		})
+		.and_then(|res| {
+			res.map_err(|e| {
+				tracing::debug!(
```
pepyakin, line 222: What's the reasoning behind making those two `debug!`?

pepyakin (approved):

node/core/candidate-validation/src/lib.rs
rphmeier: They are unexpected but non-fatal - trace! is too low, but warn! is too high.

rphmeier:
My reasoning behind not using spawn-blocking was that the heaviest part of it is the actual execution which happens in a separate task. I don't expect the decompression to be very heavy.

Issue #4003, opened 2021-10-04
Flaky test `bitfield_checks`


bkchr:
```
failures:
---- inclusion::tests::bitfield_checks stdout ----
thread 'inclusion::tests::bitfield_checks' panicked at 'assertion failed: ParaInclusion::process_bitfields(expected_bits(), vec![signed.into()],\n                                 &core_lookup).is_err()', runtime/parachains/src/inclusion.rs:1416:17
stack backtrace:
   0: rust_begin_unwind
             at /rustc/c8dfcfe046a7680554bf4eb612bad840e7631c4b/library/std/src/panicking.rs:515:5
   1: core::panicking::panic_fmt
             at /rustc/c8dfcfe046a7680554bf4eb612bad840e7631c4b/library/core/src/panicking.rs:92:14
   2: core::panicking::panic
             at /rustc/c8dfcfe046a7680554bf4eb612bad840e7631c4b/library/core/src/panicking.rs:50:5
   3: std::thread::local::LocalKey<T>::with
   4: core::ops::function::FnOnce::call_once
   5: core::ops::function::FnOnce::call_once
             at /rustc/c8dfcfe046a7680554bf4eb612bad840e7631c4b/library/core/src/ops/function.rs:227:5
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
failures:
    inclusion::tests::bitfield_checks
```

bkchr:
```
			// non-pending bit set.
			{
				let mut bare_bitfield = default_bitfield();
				*bare_bitfield.0.get_mut(0).unwrap() = true;
				let signed = block_on(sign_bitfield(
					&keystore,
					&validators[0],
					ValidatorIndex(0),
					bare_bitfield,
					&signing_context,
				));

				assert!(ParaInclusion::process_bitfields(
					expected_bits(),
					vec![signed.into()],
					&core_lookup,
				)
				.is_err());
			}
```

Shouldn't that have been removed in: https://github.com/paritytech/polkadot/commit/9e4ee69a00f974d99e06a836f0cf808d73dedf4d ?

It is also weird that locally I can only reproduce this when building the test in release, in debug it works...

bkchr:
Okay, I found out while debug vs release returns different results..

PR #4004, opened 2021-10-04
Fix unoccupied bitfields

bkchr:
If there is an unoccupied bitfield set, we should just ignore it and not
keep it for the rest of the logic in `process_bitfields`.

Fixes: https://github.com/paritytech/polkadot/issues/4003

### DIFF
--- a/runtime/parachains/src/inclusion.rs
+++ b/runtime/parachains/src/inclusion.rs
@@ -1416,6 +1416,27 @@ mod tests {
 				.is_err());
 			}
 
+			// non-pending bit set.
+			{
+				let mut bare_bitfield = default_bitfield();
+				*bare_bitfield.0.get_mut(0).unwrap() = true;
+				let signed = block_on(sign_bitfield(
+					&keystore,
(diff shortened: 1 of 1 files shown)


runtime/parachains/src/inclusion.rs
```diff
@@ -289,6 +297,13 @@ impl<T: Config> Pallet<T> {
 					Error::<T>::ValidatorIndexOutOfBounds,
 				);
 
+				// If there is a bit set that shouldn't bet set, we ignore it.
```
rphmeier, line 300: We should ignore the bit, not the entire bitfield. That was the previous behavior, but #3999 was meant to change it to allow those bitfields and ignore.

rphmeier: We're supposed to have a check later on in the bitwise loop. But I guess it doesn't work right - it should probably just look into the `occupied_bitmask` instead.

rphmeier (approved):

rphmeier:
bot merge

PR #4017, opened 2021-10-05
Implement dependent check for Cumulus

joao-paulo-parity:
Work in progress

:exclamation: After this PR is merged, the required checks in the Github Repository Settings should be updated:

- `continuous-integration/gitlab-check-dependent-cumulus` was added (should be required)

For more context see https://github.com/paritytech/ci_cd/issues/227


.gitlab-ci.yml
```diff
@@ -136,445 +136,27 @@ default:
       vault:                       cicd/gitlab/$CI_PROJECT_PATH/SSH_PRIVATE_KEY@kv
       file:                        false
 
-#### stage:                        test
```
joao-paulo-parity, line 139: Those removals are for testing purposes only. I'll revert them before merge.

PR #4026, opened 2021-10-06
Allow `Root` to register a para

joepetrowski:


bkchr (commented):
AFAIK root means that all filters will be ignored. But let's wait for @shawntabrizi.

shawntabrizi:
yeah, root bypasses all filters, so this shouldnt be needed

joepetrowski:
Not needed

Issue #3989, opened 2021-10-01
Refactor inherent filtering to be contained completely in the runtime


rphmeier:
At the moment, the provisioner does some filtering of candidates and bitfields to provide them to the runtime. However, it's not tightly integrated into the runtime which makes filtering difficult to do accurately and on a per-candidate or per-bitfield basis, and also for it to account for weight. 

For disputes, the filtering is offloaded completely to the runtime's `ProvideInherent` implementation for the `ParasInherent`. This is invoked off-chain by the block author, and the provisioner just supplies all dispute votes and expects the runtime to filter it down to an acceptable set of disputes. It allows filtering out just a single dispute.

This pattern should be extended to cover both candidates and bitfields, and also apply filtering based on the disputes that are included. Basically, any kind of filtering needed in order for the end result to pass all checks in `ParasInherent::enter` on a fresh state. Additionally, the filtering done in the provisioner can be discarded, as it'll all be done by the runtime.

Concretely:
* Filter out _specific_ bitfields that reference a core which would be dropped as a result of a concluded dispute
* Filter out _specific_ candidates that have lost a dispute
* Filter out _specific_ bitfields which don't pass other checks.
* Filter out arbitrary bitfields according to weight. The `InherentData` could be altered to contain a random value provided by the authoring node as a means of seeding an RNG within the runtime.
* Filter out _specific_ candidates that build on top of the wrong parent head, according to the current state + anything that might become available as a result of the remaining set of bitfields.
* Filter out _specific_ candidates that don't pass general validity checks.
* Filter out arbitrary candidates according to weight. The `InherentData` could be altered to contain a random value provided by the authoring node as a means of seeding an RNG within the runtime.

Most of these checks are already done by the provisioner, but some are new.

PR #4028, opened 2021-10-06
move paras inherent filtering to runtime

drahnr:
* [x] add filter methods for `bitfields` and `candidates`
* [x] provide randomness via inherent, and use it for candidate selection
* [x] add tests
* [x] <del>remove duplicate provisioner checks</del> defered to  a separate PR
Ref #3989 

Contains changes of #3924 

follow up work:
-  #4172


rphmeier:
Also this change is substantial enough that it needs a Rococo burn-in before merging.

roadmap/implementers-guide/src/runtime/inclusion.md
```diff
@@ -45,13 +45,38 @@ PendingAvailabilityCommitments: map ParaId => CandidateCommitments;
...
+
+* `sanitize_backed_candidates<const EARLY_RETURN: bool>(
+	relay_parent: Hash,
+	mut backed_candidates: Vec<BackedCandidate>,
+	candidate_has_concluded_invalid_dispute: Fn(CandidateHash) -> bool,
+	scheduled: &[CoreAssignment],
+)`
+  1. if `EARLY_RETURN` is `true`, return an error when encountering a backed candidates that does not pass the checks, if `false`, drop the backed candidates from the set that will be returned.
+  1. check all backed candidates have no concluded invalid dispute by means of the provided closure `candidate_has_concluded_invalid_dispute`
+  1. check all backed candidates have the matchin `relay_parent`
```
emostov, line 77:
```suggestion
  1. check all backed candidates have the matching `relay_parent`
```

runtime/parachains/src/runtime_api_impl/v1.rs
```diff
@@ -332,6 +332,6 @@ pub fn validation_code_by_hash<T: paras::Config>(
 }
 
 /// Disputes imported via means of on-chain imports.
-pub fn on_chain_votes<T: paras_inherent::Config>() -> Option<ScrapedOnChainVotes<T::Hash>> {
+pub fn on_chain_votes<T: paras_inherent::pallet::Config>() -> Option<ScrapedOnChainVotes<T::Hash>> {
```
emostov, line 335: this diff shouldn't be neccesary since we re-export contents of pallet

runtime/parachains/src/paras_inherent.rs
```diff
@@ -295,15 +520,262 @@ pub mod pallet {
...
 }
 
+macro_rules! ensure2 {
+	($condition:expr, $err:expr, $action:ident $(, $alt:expr)? $(,)?) => {
+		let condition = $condition;
+		if !condition {
+			if $action {
+				ensure!(condition, $err);
+			} else {
+				$($alt)?
```
emostov, line 540: Just to clarify, what is the expected behaviour if there is no alt?

A little comment above this function explaining the expected behaviour might be nice

drahnr: If there is no `alt`, it will do nothing, as in the `else {}`-block will be empty.

runtime/parachains/src/runtime_api_impl/v1.rs
drahnr: true, residue of a duplicate `Config` re-export where `rustc` failed to resolve the right one accoring to `trait` bounds. Rust issue pending.

runtime/parachains/Cargo.toml
```diff
@@ -48,7 +49,7 @@ hex-literal = "0.3.3"
 keyring = { package = "sp-keyring", git = "https://github.com/paritytech/substrate", branch = "master" }
 frame-support-test = { git = "https://github.com/paritytech/substrate", branch = "master" }
 sc-keystore = { git = "https://github.com/paritytech/substrate", branch = "master" }
-
+assert_matches = "1"
```
emostov, line 52: dq: what is this used for?

drahnr: Only for test cases, ignore

runtime/parachains/src/paras_inherent.rs
```diff
@@ -105,32 +181,154 @@ pub mod pallet {
...
+
+			T::DisputesHandler::filter_multi_dispute_data(&mut disputes);
+
+			let scheduled: Vec<CoreAssignment> = <scheduler::Pallet<T>>::scheduled();
+
+			let (freed_cores, concluded_invalid_disputes) =
+				frame_support::storage::with_transaction(|| {
+					// we don't care about fresh or not disputes
+					// this writes them to storage, so let's query it via those means
+					// if this fails for whatever reason, that's ok
```
emostov, line 220: I assume if anything returns an error this closure returns `TransactionOutcome::Rollback` ? Its just not immediatley obvious to me how we go from a normal error to `TransactionOutcome::Rollback` but I assume there is a From blanket impl or something for errors?

drahnr: There are no `?` in that closure. All errors are just printed.

runtime/parachains/src/paras_inherent.rs
```diff
@@ -24,30 +24,106 @@
...
+		.saturating_add((10_000_000 as Weight).saturating_mul(n_disputes as Weight))
+		.saturating_add(weights.reads(10 as Weight))
+		.saturating_add(weights.reads((20 as Weight).saturating_mul(n_bitfields as Weight)))
+		.saturating_add(weights.writes(10 as Weight))
+		.saturating_add(weights.writes((20 as Weight).saturating_mul(n_bitfields as Weight)))
+}
+
+/// Extract the static weight.
+fn static_weight<T: Config>() -> Weight {
+	paras_inherent_total_weight::<T>(0, 0, 0)
```
emostov, line 95: @Lldenaurois I think if we input 0 to a benchmark that does everything we can get a good value here

runtime/parachains/src/paras_inherent.rs
```diff
@@ -295,15 +519,265 @@ pub mod pallet {
...
+	});
+	ensure2!(backed_candidates.len() == n, Error::<T>::CandidateConcludedInvalid, EARLY_RETURN);
+
+	// Assure the backed candidate's `ParaId`'s core is free.
+	// This holds under the assumption that `Scheduler::schedule` is called _before_.
+	// Also checks the candidate references the correct relay parent.
+	backed_candidates.retain(|backed_candidate| {
+		let desc = backed_candidate.descriptor();
+		desc.relay_parent == relay_parent &&
+			scheduled.iter().any(|core| core.para_id == desc.para_id)
```
emostov, line 773: Can we make `scheduled` a set? If it just O(scheduled.len()) to collect into the set and then each look up is O(1) should be better than doing a O(backed_candidate.len() * scheduled.len())

drahnr: We would go from `O(backed_candidate.len() * scheduled.len())` to `O(backed_candidate.len() + scheduled.len())`, which is still a fair improvement, since we expect more than two cores and backed candidates.

runtime/parachains/src/paras_inherent.rs
```diff
@@ -105,32 +181,154 @@ pub mod pallet {
...
+				return None
+			}
+
+			let current_session = <shared::Pallet<T>>::session_index();
+			let expected_bits = <scheduler::Pallet<T>>::availability_cores().len();
+			let validator_public = shared::Pallet::<T>::active_validator_keys();
+
+			T::DisputesHandler::filter_multi_dispute_data(&mut disputes);
+
+			let scheduled: Vec<CoreAssignment> = <scheduler::Pallet<T>>::scheduled();
```
emostov, line 214: To recap what was discussed offline: reading the `scheduled` here will lead to unnecessarily filtering out valid backed candidates because the schedule is updated in `enter` with newly freed cores (from concluding disputes and fully available blocks). Instead we are going to re-create much of the logic from `enter` within the `with_transaction` closure and then read `scheduled`, guaranteeing that the block author won't have an early exit when sanitizing backed candidates.

Here is a test that I just added that currently fails but should work once the aforementioned changes are implemented: https://github.com/paritytech/polkadot/blob/zeke-create-enter-benchmarks-5/runtime/parachains/src/paras_inherent.rs#L845-L907 . (It uses the scenario builder from my benchmark branch so I didn't add it directly to this branch).

runtime/parachains/src/disputes.rs
```diff
@@ -281,6 +298,7 @@ pub mod pallet {
 	///
 	/// The i'th entry of the vector corresponds to the i'th validator in the session.
 	#[pallet::storage]
+	#[pallet::getter(fn spam_slots)]
```
emostov, line 301: Is this used anywhere?

runtime/parachains/src/inclusion.rs
```diff
@@ -272,19 +295,34 @@ impl<T: Config> Pallet<T> {
 		signed_bitfields: UncheckedSignedAvailabilityBitfields,
 		disputed_bitfield: DisputedBitfield,
 		core_lookup: impl Fn(CoreIndex) -> Option<ParaId>,
+		is_create_inherent: bool,
 	) -> Result<Vec<(CoreIndex, CandidateHash)>, DispatchError> {
 		let validators = shared::Pallet::<T>::active_validator_keys();
 		let session_index = shared::Pallet::<T>::session_index();
```
drahnr, line 301: Make this use a const generic as well

emostov:
/benchmark runtime westend runtime_parachains::paras_inherent

runtime/parachains/src/paras_inherent.rs
```diff
@@ -23,30 +23,157 @@
...
+		.map(|d| {
+			<<T as Config>::WeightInfo as WeightInfo>::enter_variable_disputes(
+				d.statements.len() as u32
+			)
+		})
+		.fold(0, |acc, x| acc.saturating_add(x))
+}
+
+fn signed_bitfields_weight<T: Config>(bitfields_len: usize) -> Weight {
+	<<T as Config>::WeightInfo as WeightInfo>::enter_bitfields() * bitfields_len as Weight
```
emostov, line 141:
```suggestion
	<<T as Config>::WeightInfo as WeightInfo>::enter_bitfields().saturating_mul(bitfields_len as Weight)
```

shawntabrizi:
/benchmark runtime westend runtime_parachains::paras_inherent

primitives/src/v1/mod.rs
```diff
@@ -633,12 +638,12 @@ pub fn check_candidate_backing<H: AsRef<[u8]> + Clone + Encode>(
...
 			signed += 1;
 		} else {
-			return Err(())
+			return Err(());
 		}
 	}
 
 	if signed != backed.validity_votes.len() {
-		return Err(())
+		return Err(());
```
drahnr, line 646: There is quite a bit of formatting bullshit in here, which hints that the usage of fmt either changed in this PR or the CI container bumped the nightly while this PR was open. Double check.

bkchr:
/benchmark runtime westend runtime_parachains::paras_inherent

bkchr:
/benchmark runtime westend runtime_parachains::paras_inherent

drahnr:
/benchmark runtime westend runtime_parachains::paras_inherent

runtime/parachains/src/paras_inherent.rs
```diff
@@ -290,239 +459,1658 @@ pub mod pallet {
...
-/// even that we can't trust the provisioner to provide a fair / random ordering of candidates.
-fn limit_backed_candidates<T: Config>(
-	mut backed_candidates: Vec<BackedCandidate<T::Hash>>,
-) -> Vec<BackedCandidate<T::Hash>> {
-	const MAX_CODE_UPGRADES: usize = 1;
+/// Otherwise tries to include all bitfields, and fills in the remaining weight with candidates.
+///
+/// If even the bitfields are too large to fit into the `max_weight` limit, bitfields are randomly
+/// picked and _no_ candidates will be included.
+fn apply_weight_limit<T: Config + inclusion::Config>(
```
drahnr, line 724: @Lldenaurois we should probably move the disputes weight limiting in here as well, to make it an all-in-one weight limiting, so there cannot be ordering problems. It also de-duplicates some of the `RNG` creation boilerplate.

runtime/parachains/src/paras_inherent.rs
```diff
@@ -290,239 +459,1658 @@ pub mod pallet {
...
+		// Due to the binary search predicate above, the index computed will constitute the beginning
+		// of the remote disputes sub-array
+		let remote_disputes = disputes.split_off(idx);
+
+		// Select disputes in-order until the remaining weight is attained
+		disputes.retain(|d| {
+			let dispute_weight = <<T as Config>::WeightInfo as WeightInfo>::enter_variable_disputes(
+				d.statements.len() as u32,
+			);
+			if remaining_weight > dispute_weight {
```
drahnr, line 957:
```suggestion
			if remaining_weight >= dispute_weight {
```

drahnr: I would prefer this to use `checked_sub()`, but I am not sure this makes it any more legible, but it would prevent us from having above issues.

runtime/parachains/src/paras_inherent.rs
```diff
@@ -149,11 +275,43 @@ pub mod pallet {
 		}
 	}
 
+	/// Collect all freed cores based on storage data.
+	///
+	/// The paramter `freed_concluded` contains all core indicies that became
+	/// free due to candidate that became available.
+	pub(crate) fn collect_all_freed_cores<T, I>(
```
drahnr, line 282: bike-shedding appreciated.

drahnr (commented):
Gave it another pass, a few inaccuracies, mostly looking ok

runtime/parachains/src/scheduler.rs
```diff
@@ -90,6 +90,7 @@ impl ParathreadClaimQueue {
 }
 
 /// Reasons a core might be freed
+#[derive(Clone, Copy)]
```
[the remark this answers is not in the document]
drahnr: `Copy` requires `Clone` - https://doc.rust-lang.org/std/marker/trait.Copy.html

shawntabrizi:
/benchmark runtime westend runtime_parachains::paras_inherent

shawntabrizi:
/benchmark runtime westend runtime_parachains::paras_inherent

shawntabrizi:
/benchmark runtime westend runtime_parachains::paras_inherent

shawntabrizi:
/benchmark runtime westend runtime_parachains::paras_inherent

runtime/parachains/src/paras_inherent/benchmarking.rs
```diff
@@ -0,0 +1,206 @@
...
+			inclusion::PendingAvailabilityCommitments::<T>::iter().count(),
+			cores_with_backed.len()
+		);
+		assert_eq!(
+			inclusion::PendingAvailability::<T>::iter().count(),
+			cores_with_backed.len()
+		);
+	}
+
+	enter_backed_candidate_code_upgrade {
```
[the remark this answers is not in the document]
Lldenaurois: The plan is to make use of this in weighing the code upgrades for backed candidates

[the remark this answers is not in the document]
Lldenaurois: To be specific: It's going to be `enter_backed_candidates_variable(c.commitments.validity_votes.len()) + enter_backed_candidate_code_upgrade()`

runtime/parachains/src/paras_inherent.rs
```diff
@@ -743,14 +744,21 @@ fn apply_weight_limit<T: Config + inclusion::Config>(
...
-			remaining_weight,
-		);
+		let (acc_candidate_weight, indices) =
+			random_sel::<BackedCandidate<<T as frame_system::Config>::Hash>, _>(
+				&mut rng,
+				candidates.clone(),
+				|c| {
+					<<T as Config>::WeightInfo as WeightInfo>::enter_backed_candidates_variable(
+						c.validity_votes.len() as u32,
+					) + if c.candidate.commitments.new_validation_code.is_some() {
```
emostov, line 754: needs saturating arithmetic

runtime/parachains/src/disputes.rs
```diff
@@ -129,6 +129,9 @@ pub trait DisputesHandler<BlockNumber> {
 		included_in: BlockNumber,
 	);
 
+	/// Retrieve the included state of a given candidate in a particular session.
```
emostov, line 132:
```suggestion
/// Retrieve the included state of a given candidate in a particular session. If it
/// returns `Some`, then we have a local dispute for the given `candidate_hash`.
```

runtime/parachains/src/inclusion.rs
```diff
@@ -746,6 +767,7 @@ impl<T: Config> Pallet<T> {
 		}
 
 		// enact the messaging facet of the candidate.
+		// TODO check how to account for these
```
emostov, line 770:
```suggestion
```

runtime/parachains/src/inclusion.rs
```diff
@@ -255,76 +286,34 @@ impl<T: Config> Pallet<T> {
 		for _ in <AvailabilityBitfields<T>>::drain() {}
 	}
 
-	/// Process a set of incoming bitfields.
+	/// Extract the freed cores based on cores tht became available.
```
drahnr, line 289:
```suggestion
	/// Extract the freed cores based on cores that became available.
```

runtime/parachains/src/inclusion.rs
drahnr: @emostov please create follow up issue to track this and further discuss

roadmap/implementers-guide/src/runtime/inclusion.md
```diff
@@ -78,6 +103,9 @@ All failed checks should lead to an unrecoverable error making the block invalid
   1. call `Hrmp::prune_hrmp` with the para id of the candiate and the candidate's `hrmp_watermark`.
   1. call `Hrmp::queue_outbound_hrmp` with the para id of the candidate and the list of horizontal messages taken from the commitment,
   1. Call `Paras::note_new_head` using the `HeadData` from the receipt and `relay_parent_number`.
+
+* `fn sanitize_bitfields`
```
drahnr, line 107: FIXME this is a stub and needs some attention

runtime/parachains/src/paras_inherent.rs
```diff
@@ -166,20 +325,29 @@ pub mod pallet {
...
 				disputes,
 			} = data;
+			let total_weight =
+				paras_inherent_total_weight::<T>(&backed_candidates, &signed_bitfields, &disputes);
+
+			// Abort if the total weight of the block exceeds the max block weight
+			ensure!(
+				total_weight <= <T as frame_system::Config>::BlockWeights::get().max_block,
+				Error::<T>::InherentOverweight
+			);
```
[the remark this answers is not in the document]
ordian: > pick randomly from local disputes

we're using specific disputes ordering when handling them in #4134, it should be the same here `(relay_parent_number, candidate_hash)`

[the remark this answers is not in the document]
emostov: I haven't read that PR, but how would we enforce ordering from the node in the runtime? Malicious block author can always pass data with bad ordering

primitives/Cargo.toml
```diff
@@ -57,5 +57,4 @@ std = [
 ]
 runtime-benchmarks = [
 	"polkadot-parachain/runtime-benchmarks",
-	"frame-system/runtime-benchmarks",
```
emostov: did you mean to remove this?

runtime/parachains/src/paras_inherent.rs
```diff
@@ -321,19 +332,33 @@ pub mod pallet {
...
+			if candidate_weight.saturating_add(bitfields_weight).saturating_add(disputes_weight) > max_block_weight {
+				backed_candidates.clear();
+				candidate_weight = 0;
+				signed_bitfields.clear();
+				bitfields_weight = 0;
+			}
+
+			let total_weight = if disputes_weight > max_block_weight {
+				let parent_hash = <frame_system::Pallet<T>>::parent_hash();
+				let entropy = compute_entropy::<T>(parent_hash);
```
emostov, line 357: we can compute the entropy once at the top of enter so we don't hav to compute it multiple times

runtime/parachains/src/paras_inherent.rs
```diff
@@ -749,21 +761,7 @@ fn apply_weight_limit<T: Config + inclusion::Config>(
...
-						.map(|_code| {
-							<<T as Config>::WeightInfo as WeightInfo>::enter_backed_candidate_code_upgrade()
-						})
-						.unwrap_or(0);
-					<<T as Config>::WeightInfo as WeightInfo>::enter_backed_candidates_variable(
-						backed_candidate.validity_votes.len() as u32,
-					)
-					.saturating_add(code_weight)
-				},
+				|c| backed_candidate_weight::<T>(c),
```
emostov, line 764: why remove the code_upgrade weight here?

runtime/parachains/src/paras_inherent.rs
```diff
@@ -749,21 +761,7 @@ fn apply_weight_limit<T: Config + inclusion::Config>(
...
-						.new_validation_code
-						.as_ref()
-						.map(|_code| {
-							<<T as Config>::WeightInfo as WeightInfo>::enter_backed_candidate_code_upgrade()
-						})
-						.unwrap_or(0);
-					<<T as Config>::WeightInfo as WeightInfo>::enter_backed_candidates_variable(
-						backed_candidate.validity_votes.len() as u32,
-					)
-					.saturating_add(code_weight)
```
emostov: also, we should max code_weight and output of `enter_backed_candidates_variable` since code_weight assumes worst case validity votes anyways

runtime/parachains/src/paras_inherent.rs
```diff
@@ -142,15 +142,26 @@ fn signed_bitfields_weight<T: Config>(bitfields_len: usize) -> Weight {
...
 }
 
+fn backed_candidate_weight<T: frame_system::Config + Config>(
+	candidate: &BackedCandidate<T::Hash>,
+) -> Weight {
+	<<T as Config>::WeightInfo as WeightInfo>::enter_backed_candidates_variable(
+		candidate.validity_votes.len() as u32,
+	)
+	.saturating_add(if candidate.candidate.commitments.new_validation_code.is_some() {
+		<<T as Config>::WeightInfo as WeightInfo>::enter_backed_candidate_code_upgrade()
```
emostov, line 152: We should max the two, not add since `enter_backed_candidate_code_upgrade` takes into account max validity votes

drahnr: @Lldenaurois

runtime/parachains/src/paras_inherent.rs
drahnr: We should create the `Rng` only once and re-use that.

emostov: i am gonna make a little pr here with this refactor rn

runtime/parachains/src/paras_inherent.rs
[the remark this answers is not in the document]
drahnr: To my understanding, it's up to the block producer to include the votes. And hence, if a an overweight block hits the `fn ParaInherent::enter` we must choose randomly. For the `fn create_inherent`, we might prefer the first `n` ones instead of random selection.

runtime/parachains/src/paras_inherent.rs
emostov: https://github.com/paritytech/polkadot/pull/4287

runtime/parachains/src/paras_inherent.rs
emostov: will be closed by https://github.com/paritytech/polkadot/pull/4287

primitives/Cargo.toml
Lldenaurois: I believe we had suspected that this could cause the benchmarks to fail...

Do you know if we need it?

runtime/parachains/src/builder.rs
```diff
@@ -0,0 +1,596 @@
...
+
+		assert_eq!(
+			inclusion::PendingAvailabilityCommitments::<T>::iter().count(),
+			used_cores as usize,
+		);
+		assert_eq!(inclusion::PendingAvailability::<T>::iter().count(), used_cores as usize,);
+
+		// Mark all the use cores as occupied. We expect that their are `backed_and_concluding_cores`
+		// that are pending availability and that there are `non_spam_dispute_cores` which are about
+		// to be disputed.
```
emostov, line 573:
```suggestion

		// Mark all the use cores as occupied. We expect that their are `backed_and_concluding_cores`
		// that are pending availability and that there are `used_cores - backed_and_concluding_cores `
		// which are about to be disputed.
```

runtime/parachains/src/paras_inherent.rs
```diff
@@ -149,122 +279,190 @@ pub mod pallet {
...
 			DispatchClass::Mandatory,
 		))]
 		pub fn enter(
 			origin: OriginFor<T>,
 			data: ParachainsInherentData<T::Header>,
 		) -> DispatchResultWithPostInfo {
+			ensure_none(origin)?;
+
+			ensure!(!Included::<T>::exists(), Error::<T>::TooManyInclusionInherents);
+			Included::<T>::set(Some(()));
```
emostov, line 328: TODO: double check that making Included as Some here is ok. 

This inherent can still error later on, but from what I can tell that is ok?

runtime/parachains/src/paras_inherent.rs
```diff
@@ -149,122 +279,190 @@ pub mod pallet {
...
-				disputes,
+				mut disputes,
 			} = data;
 
-			ensure_none(origin)?;
-			ensure!(!Included::<T>::exists(), Error::<T>::TooManyInclusionInherents);
-
-			// Check that the submitted parent header indeed corresponds to the previous block hash.
 			let parent_hash = <frame_system::Pallet<T>>::parent_hash();
+			// Check that the submitted parent header indeed corresponds to the previous block hash.
```
emostov, line 338:
```suggestion
			// Check that the submitted parent header indeed corresponds to the previous block hash.
			let parent_hash = <frame_system::Pallet<T>>::parent_hash();
```

runtime/parachains/src/paras_inherent.rs
emostov: if do this we should have this function call other functions like `limit_bitfields`, `limit_candidates`, and `limit_disputes`, so each can be unit tested.

runtime/parachains/src/paras_inherent.rs
```diff
@@ -290,239 +488,1609 @@ pub mod pallet {
...
+	remaining_weight: Weight,
+	rng: &mut rand_chacha::ChaChaRng,
+) -> Weight {
+	let mut remaining_weight = remaining_weight;
+	let disputes_weight = dispute_statements_weight::<T>(&disputes);
+	if disputes_weight > remaining_weight {
+		// Sort the dispute statements according to the following prioritization:
+		//  1. Prioritize local disputes over remote disputes.
+		//  2. Prioritize older disputes over newer disputes.
+		disputes.sort_by(|a, b| {
```
emostov, line 928: Looks like we can use `sort_unstable_by` here (https://doc.rust-lang.org/std/vec/struct.Vec.html#method.sort_unstable_by), which is faster according to the docs

runtime/parachains/src/paras_inherent.rs
```diff
@@ -149,122 +279,190 @@ pub mod pallet {
...
 			} = data;
 
-			ensure_none(origin)?;
-			ensure!(!Included::<T>::exists(), Error::<T>::TooManyInclusionInherents);
-
-			// Check that the submitted parent header indeed corresponds to the previous block hash.
 			let parent_hash = <frame_system::Pallet<T>>::parent_hash();
+			// Check that the submitted parent header indeed corresponds to the previous block hash.
 			ensure!(
 				parent_header.hash().as_ref() == parent_hash.as_ref(),
```
drahnr, line 340: This is one of the basics and will remain checked in the provisioningsubsystem. This would indicate a bug rather anything else iiuc. So in reference to the above statement, this is fine.

runtime/parachains/src/paras_inherent.rs
```diff
@@ -274,7 +472,7 @@ pub mod pallet {
 			} = <inclusion::Pallet<T>>::process_candidates(
 				parent_storage_root,
 				backed_candidates,
-				<scheduler::Pallet<T>>::scheduled(),
+				scheduled,
 				<scheduler::Pallet<T>>::group_validators,
 			)?;
```
drahnr, line 477: This is the second and last occasion the block can error, but given our latest impl it will not. Have to double check.

runtime/parachains/src/paras_inherent.rs
drahnr: Indeed, very good point

runtime/parachains/src/paras_inherent.rs
emostov: `process_candidates` can still error. I briefly mentioned this yesterday, but I am wondering if we should also modify process candidates so it will just skip candidates and not error if there are any issue

emostov: I think modifying it might be out of scope of this PR, but from my understanding of things it could make the liveliness of the relay chain more resilient if we simply skip candidates that have any error instead of error'ing out of the whole `enter` inherent

runtime/parachains/src/paras_inherent.rs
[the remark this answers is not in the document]
drahnr: Talking to Robert, we want to sort, and only taking local ones is fine. We should prefer older over newer dispute messages.

runtime/parachains/src/paras_inherent.rs
drahnr: This is ok.

runtime/parachains/src/paras_inherent.rs
shawntabrizi: no please: https://github.com/paritytech/substrate/pull/6754

emostov: Ok @shawntabrizi pointed out this is not worth the undefined behavior and can lead to consensus issues https://github.com/paritytech/substrate/pull/6754

emostov (approved):

emostov:
bot merge

Issue #4033, opened 2021-10-07
Warn in case on chain scraping does not work


eskimor:
Once the runtime is upgraded and fetch on chain votes is supposed to work, we should get a warning once it stops (e.g. due to a runtime upgrade). So [this](https://github.com/paritytech/polkadot/blob/50b0658c1da5c1833dc20bb4fee2c9461fcd5dd9/node/core/dispute-coordinator/src/real/mod.rs#L542) should become `warn` once we upgraded the runtime.

This is a follow up for #3924.

bkchr:
Runtime api supports versioning for exactly these kind of use cases. A proper fix would be to bump the runtime api version and check if the required one is already on chain and otherwise inform the caller that the method isn't yet available.

sandreim:
@eskimor looks like we can close this one, wdyt ?

PR #4054, opened 2021-10-11
Runtime API for retrieving included state of a candidate

eskimor:
This will be used for disputes in order to prioritize candidates that have been seen included on some chain.


eskimor:
> Also, runtime API versioning as discussed in #4033 ?

Yep. I was planning on adding this in a separate PR. But let's see, I might add it right here as well. Just wanted to get the actual runtime changes out.

PR #4112, opened 2021-10-20
WrapVersion for unsafe v1 conversions

gavofyork:


gavofyork:
I guess this is unneeded now.

PR #4108, opened 2021-10-19
v0.9.12-rc2

s3krit:
- [x] Cherry-pick https://github.com/paritytech/polkadot/pull/4107
- [x] Cherry-pick https://github.com/paritytech/polkadot/pull/4093
- [x] Apply benchmarks

PR #4113, opened 2021-10-20
Update weights

s3krit:
Also being applied here https://github.com/paritytech/polkadot/pull/4108/files


thiolliere:
somehow this doesn't update the weight of frame_system like the file runtime/polkadot/src/weights/frame_system.rs isn't touched

chevdor:
Closing as no longer relevant.

PR #4123, opened 2021-10-21
PVF host prechecking support v2

### LINKED ISSUE #3857: Validation host support for pre-checking
Add a function into the validation host that will provide a way to pre-check a given wasm binary. (See https://github.com/paritytech/polkadot/issues/3211 for details on what pre-checking is).

The shape of the function would be:

```rust 
pub async fn precheck_pvf(
    &mut self,
    pvf: Pvf,
    result_tx: oneshot::Sender<PrecheckResult>,
) -> Result<(), ..>
```

When called the function should return immediately if there is no back-pressure. When pre-checking is finished, the response is sent through `result_tx`.

`PrecheckResult` indicates on of the following outcomes:

1. succeeded - meaning we are pretty sure that compilation of the given PVF finishes producing a compiled artifact within the certain memory and time limits.
2. failed - we are pretty sure that compilation exceeds the certain memory and time limits.
3. internal error - this potentially can be returned if there are problems claiming a worker process to do the pre-checking job, or other spurious failures.

The execution happens on the validation host async task. When executed, it will check if the given PVF already compiled. If it is, then it will respond immediately.

The compilation should be performed with turned off parallel compilation. See https://github.com/paritytech/substrate/issues/9605 details.

If the PVF is being compiled right now then we should kill the current process or wait until it finishes on its own. Then the process should be restarted under the required conditions. The actual compilation should not differ from what we currently have, with the exception of custom settings. As an alternative, we may want to compile any PVFs in such a mode that ensures the least variance at all times.

slumber:
Make every prepare request single-threaded

Resolves #3857 
Based on https://github.com/paritytech/polkadot/pull/3907

NOTE(@pepyakin): this PR disables parallel compilation for preparation. I conducted a couple of tests on different machines. It takes ~0.7-0.8ms to compile the code with `parallel_compilation` on and around 3.5s without it.


rphmeier:
checks failing

Issue #4126, opened 2021-10-22
ump_service_total_weight is by default too small


GopherJ:
currently it's hard to do xcm transact because this value is too small and the max_weight used is equal to `ump_service_total_weight`

PR #4127, opened 2021-10-22
increase ump_service_total_weight's default value

GopherJ:
close: #4126 

now by default `FirstMessageFactorPercent` is 100, when the first ump message comes, the `max_weight` will be `ump_service_total_weight` which is `4_000_000_000`, this is too small

https://github.com/paritytech/polkadot/blob/master/runtime/parachains/src/ump.rs#L463-L476


GopherJ:
@pepyakin  can you review again?

pepyakin (approved):

PR #4172, opened 2021-10-28
Weights for enacting candidates in paras inherent

emostov:
This PR 
 - bubbles up the weight from the `enact_candidate` call and adds it to the returned weight of paras inherent `enter`. Previously we where not tracking the weight from enacting candidates. 
 - creates `enact_candidate_weight`, which gives us the worst case weight which we can then use prior to calling enter.

cc @pepyakin 

Notes for reviewers:
- for people familiar with the FRAME storage apis: please focus on making sure the hardcoded storage read / write numbers look correct
- for people familiar with the parachains runtime: please focus on making sure the way the weights from enact_candidate are used is correct

relates to paritytech/polkadot#4044

skip check-dependent-cumulus

### DIFF
--- a/runtime/parachains/src/paras.rs
+++ b/runtime/parachains/src/paras.rs
@@ -990,6 +990,7 @@ impl<T: Config> Pallet<T> {
 	pub(crate) fn note_new_head_weight() -> Weight {
 		T::DbWeight::get().reads_writes(6, 6)
 	}
+
 	/// Returns the current lifecycle state of the para.
 	pub fn lifecycle(id: ParaId) -> Option<ParaLifecycle> {
 		ParaLifecycles::<T>::get(&id)


runtime/parachains/src/paras_inherent.rs
```diff
@@ -153,7 +154,21 @@ pub mod pallet {
...
+			{
+				let c = <configuration::Pallet<T>>::config();
+				let enact_candidate_weight = <inclusion::Pallet<T>>::enact_candidate_weight(
+					c.hrmp_max_message_num_per_candidate,
+					c.max_upward_message_num_per_candidate,
+					c.hrmp_max_parachain_inbound_channels,
+					c.hrmp_max_parathread_inbound_channels,
+				)
+				// NOTE: this will need to updated if the max number of cores changes.
+				* 40u64;
```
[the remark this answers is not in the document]
emostov: Added this suggestion in b8a12f20095f2553d473b2731b275928a4f17155

runtime/parachains/src/hrmp.rs
```diff
@@ -908,6 +908,19 @@ impl<T: Config> Pallet<T> {
 		weight
 	}
 
+	/// Worst case weight for `prune_hrmp`.
+	pub(crate) fn prune_hrmp_weight(
```
kianenigma, line 912: I think if we have #3876 already in, we can automatically measure this function.

emostov:
cc @drahnr @Lldenaurois this is the PR I was discussing yesterday that addressed that stray TODO

runtime/parachains/src/inclusion.rs
```diff
@@ -1394,7 +1423,8 @@ pub(crate) mod tests {
 						vec![signed.into()],
 						DisputedBitfield::zeros(expected_bits()),
 						&core_lookup,
-					),
+					)
+					.0,
```
kianenigma, line 1427: honestly, I would make these function self-weighing to prevent such breakage. Any advantage in returning the weight?

kianenigma (approved):
Code looks good, but I am not sure why this approach of manual counting is chose here. For example, the same code paths where their DB access has been manually counted in this PR are measured automatically in #3876. 

But still, don't see a problem in rolling with this and modifying along the way.

runtime/parachains/src/inclusion.rs
emostov: @kianenigma the main advantage I see is explicitness when reading at the top level. Otherwise it may not be clear to the reader of `enter` that this weighing is going on. 

But I think self weighing sounds like it might be a good approach. How would I go about doing that?

kianenigma: > How would I go about doing that?

Instead of returning anything, each function calls into `frame_system::Pallet::register_extra_weight` itself. 

I don't strongly prefer this, chose whichever approach suits your need best. Self-weighing is def. less noise, but that noise could be for a good reason.

runtime/parachains/src/paras_inherent.rs
```diff
@@ -115,16 +115,37 @@ impl WeightInfo for TestWeightInfo {
...
 ) -> Weight {
-	backed_candidates_weight::<T>(backed_candidates, config)
+	backed_candidates_weight::<T>(backed_candidates)
 		.saturating_add(signed_bitfields_weight::<T>(bitfields.len()))
 		.saturating_add(dispute_statements_weight::<T>(disputes))
+		// in order to get a constant time calculation we get the votes in the first bitfield
+		// and assume the worst case for enacting that many candidates. This may not always be
+		// perfect but should be close enough most of the time for getting an estimate when queuing
+		// extrinsics for block authoring.
+		.saturating_add(enact_candidates_weight::<T>(bitfields_count_ones(bitfields), config))
```
emostov, line 148: Its a bit tricky to get a great estimate upfront because we need to predict which candidates will voted on successfully by looking at the availability bitfields. To be precise, we could iterate through the bitfields, track and the votes and see which will will likely be enacted based on pending available. However that adds a bit of complexity and we already are tight on time from what I understand. My proposal is just to check the votes in the first bitfield and assume the rest are the same. From my understanding this should be ok (assuming the first bitfield is not malicious, which I don't think we can but _hopefully_ that doesn't happen very often). 

Not sure though if I am missing a detail here though

runtime/parachains/src/paras_inherent.rs
```diff
@@ -115,16 +115,37 @@ impl WeightInfo for TestWeightInfo {
 	}
 }
 
+// Count the ones of the first bitfield in a slice. Careful, since this only counts the ones of
+// the first bitfield, it may be misleading as to how many ones the other bitfields contain. It
+// should only be used in circumstances where an estimate is ok.
```
drahnr, line 120: Shouldn't this OR all bitfields?

runtime/parachains/src/paras_inherent.rs
drahnr: We should make sure the bitfields are verified before we do the checking.

This is part of a larger refactor, to integrate weight calculation, caching passed checks, and then doing a random pick all in one `fn`. This is out of scope for this PR.

runtime/parachains/src/paras_inherent.rs
```diff
@@ -927,6 +924,9 @@ fn apply_weight_limit<T: Config + inclusion::Config>(
 		rng,
 		bitfields.clone(),
 		vec![],
+		// bitfields can lead to enacting a candidate; however we don't have a good way of
+		// accounting for that when tracking individual bitfield weight, thus the weight here likely
+		// ends up being an underestimate.
```
drahnr, line 930: Underestimates are generally preferable. I am leaning towards doing the accumulation of bitfields and actually detecting which one is going to be enacted, given reasonable additional cost.

drahnr (changes requested):
Using the first bitfield seems unreasonable. We should at least `OR` all of them to get all potential candidates to be enacted. It would be even better to have a counter for each and assure the required threshold is reached before accounting for the weight if feasible.

runtime/parachains/src/paras_inherent.rs
emostov: The main issue I see with this approach is that we do PendingAvailaibility + new bitfields (in other words pre-existing validity votes + new validity votes) to calculate which candidates to enact.

Specifically, prior to accounting for the new bitfields we get the pending availability of every para id (in other words availability votes from prior blocks):

https://github.com/paritytech/polkadot/blob/64e1616fcc03896bc4eb0ff213dd226e8a1c3f4b/runtime/parachains/src/inclusion.rs#L314-L319

then for each new bitfield, we increment the availability votes for the corresponding para candidate:

https://github.com/paritytech/polkadot/blob/64e1616fcc03896bc4eb0ff213dd226e8a1c3f4b/runtime/parachains/src/inclusion.rs#L345-L350

and finally if the accumulation of availability votes put a candidate over the threshold we enact it:

https://github.com/paritytech/polkadot/blob/64e1616fcc03896bc4eb0ff213dd226e8a1c3f4b/runtime/parachains/src/inclusion.rs#L367

emostov: What we could do is: if the new votes add up to x% of the threshold for it to be enacted, then we include the enactment weight because we think its highly likely to be enacted. If we expect most candidates to get all their votes in one block this should be ok for the average case. However I still see an issue for the edge case where in block N we get `threshold - x% + 1` votes and then in block N+1 we get `x% - r` votes - so the candidate gets enacted in block N+1 but we didn't think it would

runtime/parachains/src/paras_inherent.rs
emostov: Do you suggest removing this from the total weight calc for now? Or leave as is and create a follow up issue to address the fact that the bitfields are unchecked?

emostov: (Also in `paras_inherent_total_weight` we need the same data from the unchecked bitfields)

runtime/parachains/src/paras_inherent.rs
emostov: >  I am leaning towards doing the accumulation of bitfields and actually detecting which one is going to be enacted, given reasonable additional cost.

Given my previous comment, do you think its reasonable to try the above suggestion in this PR?

runtime/parachains/src/paras_inherent.rs
```diff
@@ -113,14 +116,38 @@ impl WeightInfo for TestWeightInfo {
...
+fn bitfields_count_ones(
+	bitfields: &[UncheckedSignedAvailabilityBitfield],
+	expected_bits: usize,
+) -> usize {
+	bitfields
+		.iter()
+		.take(MAX_UNCHECKED_BITFIELD_ITERATIONS)
+		.fold(
+			BitVec::<BitOrderLsb0, u8>::repeat(false, expected_bits),
+			|acc: BitVec<BitOrderLsb0, u8>, cur| acc | cur.unchecked_payload().0.clone(),
```
emostov, line 131: TODO: make sure this OR cannot panic

runtime/parachains/src/paras_inherent.rs
emostov: 75c8b96

emostov:
615e788 adds some checks to `bitfield_count_ones` to try and minimize the amount of bogus bitfields that end up affecting weight.

14faefb tries to optimize counting the votes in bitfields by tracking them in `sanitize_bitfields` (after the checks are passed). This change lets us avoid calling `bitfield_count_ones` in `create_inherent`. However we still need to call `bitfield_count_ones` in the weight annotation for `enter` and when doing initial weight limiting in `enter` because we do not want to check signatures in the runtime when we do not yet know if the input could be overweight

runtime/parachains/src/inclusion/mod.rs
```diff
@@ -402,24 +406,24 @@ impl<T: Config> Pallet<T> {
...
 	/// Returns a `Vec` of `CandidateHash`es and their respective `AvailabilityCore`s that became available,
-	/// and cores free.
+	/// and cores free. Additionally returns the weight consumed by enacted candidates.
 	pub(crate) fn process_bitfields(
 		expected_bits: usize,
 		signed_bitfields: UncheckedSignedAvailabilityBitfields,
 		disputed_bitfield: DisputedBitfield,
 		core_lookup: impl Fn(CoreIndex) -> Option<ParaId>,
-	) -> Vec<(CoreIndex, CandidateHash)> {
+	) -> (Vec<(CoreIndex, CandidateHash)>, Weight) {
```
drahnr, line 421: From a design perspective and writing test cases, it'd make more sense to return a `Vec<CandidateHash>`, and the caller can do the weight calc.

runtime/parachains/src/inclusion/tests.rs
```diff
@@ -1,19 +1,3 @@
...
-// (at your option) any later version.
-
-// Polkadot is distributed in the hope that it will be useful,
-// but WITHOUT ANY WARRANTY; without even the implied warranty of
-// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
-// GNU General Public License for more details.
-
-// You should have received a copy of the GNU General Public License
-// along with Polkadot.  If not, see <http://www.gnu.org/licenses/>.
-
```
drahnr, line 16: Should remain :)

runtime/parachains/src/paras.rs
```diff
@@ -1005,11 +1005,16 @@ impl<T: Config> Pallet<T> {
 
 				let (reads, writes) = Self::increase_code_ref(&new_code_hash, &new_code);
 				FutureCodeHash::<T>::insert(&id, new_code_hash);
-				T::DbWeight::get().reads_writes(2 + reads, 3 + writes)
+				T::DbWeight::get().reads_writes(3 + reads, 5 + writes)
```
drahnr, line 1008: How can I verify these?

runtime/parachains/src/paras_inherent/mod.rs
```diff
@@ -297,6 +313,7 @@ impl<T: Config> Pallet<T> {
 				candidate_weight = 0;
 				signed_bitfields.clear();
 				bitfields_weight = 0;
+				max_enact_candidates_weight = 0;
```
drahnr, line 316: :+1:

runtime/parachains/src/paras_inherent/mod.rs
```diff
@@ -277,18 +283,28 @@ impl<T: Config> Pallet<T> {
...
 		let now = <frame_system::Pallet<T>>::block_number();
+		let expected_bits = <scheduler::Pallet<T>>::availability_cores().len();
 
 		let mut candidate_weight = backed_candidates_weight::<T>(&backed_candidates);
 		let mut bitfields_weight = signed_bitfields_weight::<T>(signed_bitfields.len());
 		let disputes_weight = dispute_statements_weight::<T>(&disputes);
+		let mut max_enact_candidates_weight = enact_candidates_weight::<T>(
+			bitfields_count_ones(
+				&signed_bitfields,
+				expected_bits,
```
drahnr, line 294: It feels odd that the check for bitfield lengths moved in here

runtime/parachains/src/paras_inherent/mod.rs
```diff
@@ -599,12 +624,15 @@ impl<T: Config> Pallet<T> {
 
 		// Assure the maximum block weight is adhered.
 		let max_block_weight = <T as frame_system::Config>::BlockWeights::get().max_block;
+		// TODO subtract enactment weight - return sanitized bitfields from enactment
```
drahnr, line 627: There has to be additional logic, when selecting a `availablity_bitfield`, if that has consequences regarding weight consumed.

runtime/parachains/src/paras_inherent/mod.rs
```diff
@@ -708,17 +736,24 @@ fn apply_weight_limit<T: Config + inclusion::Config>(
...
 
+	// we include the worst case weight of enacting the candidates voted for by the bitfields
 	let total_bitfields_weight = signed_bitfields_weight::<T>(bitfields.len());
 
-	let total = total_bitfields_weight.saturating_add(total_candidates_weight);
+	let total_enact_candidates_weight = enact_candidates_weight::<T>(cores_with_votes, config);
+
+	let total = total_bitfields_weight
+		.saturating_add(total_candidates_weight)
+		.saturating_add(total_enact_candidates_weight);
```
drahnr, line 756: :+1:

runtime/parachains/src/paras_inherent/mod.rs
```diff
@@ -761,8 +799,11 @@ fn apply_weight_limit<T: Config + inclusion::Config>(
 		rng,
 		bitfields.clone(),
 		vec![],
+		// bitfields can lead to enacting a candidate; however we don't have a good way of
+		// accounting for that when tracking individual bitfield weight, thus the weight here likely
+		// ends up being an underestimate.
```
drahnr, line 804: We would need to check if we reached the 2/3 quorum after each bitfield pick, and only then account for the weight, with enactment or not.

runtime/parachains/src/paras_inherent/mod.rs
```diff
@@ -836,46 +878,74 @@ pub(crate) fn sanitize_bitfields<T: crate::inclusion::Config>(
...
 		}
 
 		last_index = Some(validator_index);
 	}
-	bitfields
+
+	(bitfields, cores_with_votes.count_ones())
+}
+
+// Returns `true` iff the checks pass.
```
drahnr, line 903: :+1:

runtime/parachains/src/paras_inherent/tests.rs
```diff
@@ -1,19 +1,3 @@
...
-// (at your option) any later version.
-
-// Polkadot is distributed in the hope that it will be useful,
-// but WITHOUT ANY WARRANTY; without even the implied warranty of
-// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
-// GNU General Public License for more details.
-
-// You should have received a copy of the GNU General Public License
-// along with Polkadot.  If not, see <http://www.gnu.org/licenses/>.
-
```
drahnr, line 16: CC

runtime/parachains/src/paras_inherent/mod.rs
emostov: I would not say its "moved" in here, but instead we do the check here in addition to the places we already did the check. In other words we abstract out some checks to `cheap_bitfield_checks` and do them in `bitfield_count_ones` and then later on in `process_bitfields`. 

Now that I think about it though, we could do some "lightweight" sanitization upfront where we just return the bitfields that pass `cheap_bitfield_checks`

runtime/parachains/src/inclusion/mod.rs
emostov: So right now the path the weight gets bubbled up is `enact_candidate
 -> update_pending_availability_and_get_freed_cores -> process_bitfields`. In `enact_candidate` we go through and track the actual weights consumed (opposed to the worst case estimates we use prior).

With the current function having a seperate functionw would be tricky and require some duplicate logic. Basically we would need to duplicate the logic that gets us to here: https://github.com/paritytech/polkadot/blob/888162259348c7fa3e9ad7454e8056235565efa6/runtime/parachains/src/inclusion/mod.rs#L370-L402 . At the moment we are actually modifying storage right there, so we essentially would need get the weight first of enacting candidates without side effects and then call `update_pending_availability_and_get_freed_cores`, which would duplicate the filtering logic but actually enact the candidate and write to storage

emostov: In an `enact_candidate_weight` function we would need to take the weight calculations done within these calls https://github.com/paritytech/polkadot/blob/888162259348c7fa3e9ad7454e8056235565efa6/runtime/parachains/src/inclusion/mod.rs#L748-L772 and abstract them to their own weight only functions

drahnr (approved):

runtime/parachains/src/inclusion/mod.rs
emostov: While I see the general design merit of having the weight tracked by weight only functions - in this particular case there is a lot of branching logic to determine how much weight actually gets used and duplicating much of that logic in the weight only functions since harder to keep track of then tracking the weight as the actual execution takes place.

drahnr (changes requested):
Previous review was moot, only saw part of the diff.

runtime/parachains/src/paras.rs
emostov: Example: https://docs.rs/frame-support/latest/frame_support/storage/trait.StorageValue.html#tymethod.mutate

public mutate: https://github.com/paritytech/substrate/blob/0e6cc5668d9ee8d852a3aa3f85a2ab5fcb4c75a1/frame/support/src/storage/types/value.rs#L134-L136

which calls private mutate: https://github.com/paritytech/substrate/blob/master/frame/support/src/storage/generator/value.rs#L104-L106

which calls try_mutate: https://github.com/paritytech/substrate/blob/0e6cc5668d9ee8d852a3aa3f85a2ab5fcb4c75a1/frame/support/src/storage/generator/value.rs#L108-L119

which calls get https://github.com/paritytech/substrate/blob/0e6cc5668d9ee8d852a3aa3f85a2ab5fcb4c75a1/frame/support/src/storage/generator/value.rs#L109 and calls either put: https://github.com/paritytech/substrate/blob/0e6cc5668d9ee8d852a3aa3f85a2ab5fcb4c75a1/frame/support/src/storage/generator/value.rs#L114 or kill https://github.com/paritytech/substrate/blob/0e6cc5668d9ee8d852a3aa3f85a2ab5fcb4c75a1/frame/support/src/storage/generator/value.rs#L115

The `get` counts as a read. The `kill` or `put` counts as a write

runtime/parachains/src/paras_inherent/mod.rs
emostov: The comment is a bit stale; my solution to this was to take into account the worst case enactment wheight whenever we include the bitfields.

emostov: Hence why I changed it to

```
	let total = total_bitfields_weight
		.saturating_add(total_candidates_weight)
		.saturating_add(total_enact_candidates_weight);
```

runtime/parachains/src/paras_inherent/mod.rs
emostov: So I think that would also lead to an underestimate since we don't take into account the edge case when a candidate already has pending availibility votes from the prior block. I tried to write about it in a little more detail with these comments: https://github.com/paritytech/polkadot/pull/4172#discussion_r758835875 , https://github.com/paritytech/polkadot/pull/4172#discussion_r758839477 , & https://github.com/paritytech/polkadot/pull/4172#discussion_r758865567

emostov: The current solution is to take into account the worst case enactment weight (`total_enact_candidates_weight`) where we assume all the candidates with any votes are enacted, and make sure that is taken into account before selecting any bitfields

runtime/parachains/src/inclusion/tests.rs
emostov: a54939745a6cf86a585e01e64add770716741093

runtime/parachains/src/paras_inherent/mod.rs
emostov: Actually realizing it would open up DoS vector to do any sanitization prior to weight checks. The reason I consider `bitfields_count_ones` iteration relatively safe to a DoS attack is that we have a hard cap on the number of bitfields with iterate with `MAX_UNCHECKED_BITFIELD_ITERATIONS`:

https://github.com/paritytech/polkadot/blob/9efb72f826ce12721d8575a2e15cb7c7ade131ce/runtime/parachains/src/paras_inherent/weights.rs#L89-L92

emostov: I made sure to add a note to `bitfield_count_ones` explaining this limitation https://github.com/paritytech/polkadot/blob/9efb72f826ce12721d8575a2e15cb7c7ade131ce/runtime/parachains/src/paras_inherent/weights.rs#L81-L82

In theory a malicious validator could craft a block where the `bitfields` is padded with `MAX_UNCHECKED_BITFIELD_ITERATIONS` count of bogus bitfields to skew the weight calc. But that seems like an ok trade off because while it will skew weights it would prevent some insane unbounded iteration

runtime/parachains/src/paras.rs
```diff
@@ -990,44 +990,6 @@ impl<T: Config> Pallet<T> {
...
-	/// in the context of the given relay-chain height. A second block number parameter may be used
-	/// to tell the lookup to proceed as if an intermediate parablock has been with the given
-	/// relay-chain height as its context. This may return the hash for the past, current, or
-	/// (with certain choices of `assume_intermediate`) future code.
-	///
-	/// `assume_intermediate`, if provided, must be before `at`. This will return `None` if the validation
-	/// code has been pruned.
-	///
-	/// To get associated code see [`Self::validation_code_at`].
-	pub(crate) fn validation_code_hash_at(
```
drahnr: Removed since it is unused, correct?

runtime/parachains/src/inclusion/mod.rs
drahnr: I see that this is not feasible in the remaining time, could you create an issue outlining the merits and problems with such an approach?

runtime/parachains/src/paras.rs
drahnr: Thank you for exercising this all the way through :)

runtime/parachains/src/paras_inherent/mod.rs
drahnr: This makes sense, we should also have an upper bound on the node side.

runtime/parachains/src/paras_inherent/mod.rs
drahnr: I see.

runtime/parachains/src/paras_inherent/mod.rs
drahnr: This also mandates a follow up revisit, so creating an issue with this text and the previous one would suffice :)

drahnr (approved):
Two follow up issues needed, other than that, LGTM! :+1: Thank you!

runtime/parachains/src/inclusion/mod.rs
emostov: https://github.com/paritytech/polkadot/issues/4511

runtime/parachains/src/paras_inherent.rs
emostov: https://github.com/paritytech/polkadot/issues/4513

runtime/parachains/src/paras_inherent/mod.rs
emostov: https://github.com/paritytech/polkadot/issues/4513

runtime/parachains/src/paras_inherent/mod.rs
emostov: https://github.com/paritytech/polkadot/issues/4512

runtime/parachains/src/paras.rs
emostov: This got removed when I merge in master. Its not in the diff anymore

runtime/parachains/src/paras.rs
```diff
@@ -981,6 +986,10 @@ impl<T: Config> Pallet<T> {
 		}
 	}
 
+	/// Worst case weight for `note_new_head`.
+	pub(crate) fn note_new_head_weight() -> Weight {
+		T::DbWeight::get().reads_writes(6, 6)
+	}
```
emostov, line 992:
```suggestion
	}

```

emostov:
Can this be closed?

drahnr:
I think it's still desirable to account for the enacted candidates, but it depends how much work that'd entail to update vs reimpl?

PR #4177, opened 2021-10-29
Introduce prechecking into the configuration runtime

### LINKED ISSUE #4010: Add pre-checking into the configuration runtime module
Into the configuration pallet we should add several new fields for https://github.com/paritytech/polkadot/issues/3211. Here is a rough list:

- PVF pre-checking bypass. A boolean. If enabled then PVF is accepted immediately without pre-checking process.
- PVF soaking period. The minimum number of blocks validation code should be stored in the storage before being usable. This may or may not be the same as `validation_upgrade_delay`, thus maybe renaming would do.
- Minimum upgrade delay. If a parachain upgrades to new code which has already soaked, we should still introduce a minimum delay specified by this field.
- PVF voting TTL. The number of session changes a PVF pre-checking voting can observe. If it observed this number of session changes it is treated as unviable and the PVF gets rejected.

This change requires migration. Refer to https://github.com/paritytech/polkadot/pull/3575 as an example

slumber:
Resolves #4010

### DIFF
--- a/runtime/parachains/src/configuration.rs
+++ b/runtime/parachains/src/configuration.rs
@@ -295,6 +295,7 @@ pub trait WeightInfo {
 	fn set_config_with_weight() -> Weight;
 	fn set_config_with_balance() -> Weight;
 	fn set_hrmp_open_request_ttl() -> Weight;
+	fn set_config_with_bool() -> Weight;
 }
 
 pub struct TestWeightInfo;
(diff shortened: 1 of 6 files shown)


runtime/parachains/src/configuration.rs
```diff
@@ -77,6 +77,20 @@ pub struct HostConfiguration<BlockNumber> {
 	/// The delay, in blocks, before a validation upgrade is applied.
 	pub validation_upgrade_delay: BlockNumber,
 
+	// TODO: Which section do these belong to?
```
pepyakin, line 80: I think we do not need to expose those to the parachains, at least not yet.

The closest is the `pvf_upgrade_delay`, but the idea is that parachains will look at the go-ahead signal anyway for synchronizing of the upgrades, so it is not needed as well.

pepyakin:
I am going to take over this issue: will modify the pre-checking paras code and then extract it into a separate PR. Closing this.

PR #4181, opened 2021-10-30
companion for pallet order fix.

thiolliere:
companion for https://github.com/paritytech/substrate/pull/10043

Before the hook were executed in reverse order, now with this PR we can choose the order. This PR choose to execute in declared order, we need to review if the new order is a valid implementation for polkadot.


runtime/common/src/integration_tests.rs
```diff
@@ -288,12 +288,12 @@ fn run_to_block(n: u32) {
 	assert!(System::block_number() < n);
 	while System::block_number() < n {
 		let block_number = System::block_number();
-		AllPallets::on_finalize(block_number);
+		AllPalletsWithSystem::on_finalize(block_number);
 		System::on_finalize(block_number);
 		System::set_block_number(block_number + 1);
 		System::on_initialize(block_number + 1);
```
shawntabrizi, line 294: system is being triggered twice here now

thiolliere: yes indeed, fixed it.
note that both on_finalize and on_initialize are no-op so the reordering shouldn't matter here.

runtime/common/src/integration_tests.rs
```diff
@@ -288,12 +288,10 @@ fn run_to_block(n: u32) {
 	assert!(System::block_number() < n);
 	while System::block_number() < n {
 		let block_number = System::block_number();
-		AllPallets::on_finalize(block_number);
-		System::on_finalize(block_number);
```
thiolliere, line 292: System::on_finalize is inside AllPalletsWithSystem now

runtime/common/src/integration_tests.rs
```diff
@@ -288,12 +288,10 @@ fn run_to_block(n: u32) {
...
+		AllPalletsWithSystem::on_finalize(block_number);
 		System::set_block_number(block_number + 1);
-		System::on_initialize(block_number + 1);
```
thiolliere, line 294: System::on_initialize is now inside AllPalletsWithSystem. And it is no-op so having it after is fine.

thiolliere:
note: if there is some constraint on the order of pallet or that we don't want to take any risk we can instead keep the old order and use `AllPalletsReversedWithSystemFirst`.
But as far as I can see, there is no constraint apart from Babe having to be before Session, which I don't know the reason yet.

shawntabrizi (approved):

thiolliere:
Actually there might some more constraint in the order of the pallet:
pallet authorhsip will do note the author.
when the author is noted, staking give reward point to the author, so it must be in the correct era. and im-online note that the validator has produced a block for the current session, so it must be in the correct session.

Actually in the current order (reversed) it seems the author is noted after pallet session is initialized, thus after the session are rotated. I'm trying to understand the current relation and if the order is respected and matter.

thiolliere:
I added some comment on the constraint which seems relevant to me, I guess we should write tests for them, but I'm a bit lost how.

runtime/kusama/src/lib.rs
```diff
@@ -1594,7 +1596,7 @@ pub type Executive = frame_executive::Executive<
 	Block,
 	frame_system::ChainContext<Runtime>,
 	Runtime,
-	AllPallets,
+	AllPalletsWithSystem,
```
kianenigma, line 1599: So here we have switched from _reverse_ order, to _not-reversed_ order. I was expecting this to cause some items in construct_runtime to be swapped. How so?

thiolliere: the only constraint I have found are those I put into comment, and switching to no-reversed order actually fix them.
I believe than currently the block producer of the last block of an era doesn't get points (except if he is elected in the next era) and also the block producer of last block of a session cannot tell he was online.

shawntabrizi (approved):

KiChjang (approved):

shawntabrizi:
bot merge

shawntabrizi:
bot merge

PR #4184, opened 2021-10-30
Logging & metrics improvements

rphmeier:
* Add an extra backing log
* Fixes dispute concluding metrics (they previously only fired when a dispute concluded at the same time it opened which is never)
* Remove broken dispute distribution failure logs & downgrade to trace - we can re-enable when there's a sensible non-noisy log

### DIFF
--- a/node/core/dispute-coordinator/src/real/mod.rs
+++ b/node/core/dispute-coordinator/src/real/mod.rs
@@ -900,6 +900,8 @@ async fn handle_import_statements(
 			},
 	};
 	let candidate_receipt = votes.candidate_receipt.clone();
+	let was_concluded_valid = votes.valid.len() >= supermajority_threshold;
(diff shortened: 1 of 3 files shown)


Lldenaurois (approved):

node/core/dispute-coordinator/src/real/mod.rs
```diff
@@ -1016,13 +1018,14 @@ async fn handle_import_statements(
...
-			}
-			if concluded_invalid {
-				metrics.on_concluded_invalid();
-			}
+		if !was_concluded_valid && concluded_valid {
+			metrics.on_concluded_valid();
+		}
+
+		if !was_concluded_invalid && concluded_invalid {
+			metrics.on_concluded_invalid();
```
drahnr, line 1028: :+1:

drahnr (approved):
LGTM!

PR #4186, opened 2021-10-30
XCM v3: Bridge infrastructure

gavofyork:
XCM protocol elements needed for bridging.

Introduces a conceptual _Universal Ancestor_ under which all global consensus systems exist. Ancestry of a consensus system (chain) should now go back to this Universal Ancestor. A new junction type is introduced meant for global consensus systems which sit directly under the Universal Ancestor. `NetworkId` is reworked to remove the `Any` which allows it to be used within the new `Junction::GlobalConsensus`.

_Local Consensus_ is defined as all locations under the same Global Consensus system as the local location. _Non-Local Consensus_ is defined as all other locations. Bridges exist as a means of communication between systems which view each other as being under Non-Local Consensus. 

## Usage

There should be a standard router (`NonLocalConsensus`) for bridge-routing. Chains routers will include this in their router tuple. This will be configured with the XCM location(s) of the Bridge hub(s). This router will check for non-local consensus destinations (i.e. those which ascend to the Universal Ancestor and descend into a consensus which is not local). If a message is sent to such a destination and it is supported by a bridge, it will lookup the correct XCM location for the bridge then encompass the message in a `Forward` instruction, providing payment with a prefixed `WithdrawAsset`/`BuyExecution` if needed.

When a message is routed into a bridge, the bridge will check the origin and if a descendent of the overall consensus (which it should be representing), then a `DescendOrigin` will be prefixed. This message will then be queued for sending over the bridge together with the destination location relative to the destination consensus which the remote side of the bridge represents.

Bridge hubs will be configured as UniversalAliases for all networks that they bridge over to. When a message comes over a bridge they will send the message on to the destination consensus chain using intra-consensus transport, prefixed with a `UniversalOrigin` instruction with the source consensus system.

- Rework `NetworkId` into `Option<NetworkId>` and remove `NetworkId::Any`.
- New junction: `Junction::GlobalConsensus(NetworkId)`
- New instruction: `Instruction::UniversalOrigin`
- New `xcm_executor::Config` item: `type UniversalAliases`

CC @tomusdrw 

## TODO 
- [x] New instruction: `ExportMessage` + `xcm_executor::Config` type item to configure it.
- [ ] New `SendXcm` impl which detects destinations beyond our consensus and extracts the remote network and location within it. If a known (configurable by trait parameters) bridge exists for the remote network it prepends and packs the message and sends to the bridge, e.g.: `[WithdrawAsset, BuyExecution, ExportMessage { network: remote_network, destination: remote_destination, xcm: Xcm(vec![UniversalOrigin(local_network), DescendOrigin(local_origin), { msg }]) }]`. The type which configures this should be the same type which configures the `UniversalAliases` type.
- [ ] Tests


xcm/src/v3/mod.rs
```diff
@@ -78,6 +79,24 @@ impl<Call> Xcm<Call> {
 		self.0.len()
 	}
 
+	/// Return a reference to the inner value.
+	pub fn inner(&self) -> &Vec<Instruction<Call>> { &self.0 }
+
+	/// Return a mutable reference to the inner value.
+	pub fn inner_mut(&mut self) -> &mut Vec<Instruction<Call>> { &mut self.0 }
```
KiChjang, line 86: Prefer array references rather than hardcoding a `Vec` to make it more generic:

```suggestion
	pub fn inner(&self) -> &[Instruction<Call>] { &self.0 }

	/// Return a mutable reference to the inner value.
	pub fn inner_mut(&mut self) -> &mut [Instruction<Call>] { &mut self.0 }
```

xcm/src/v3/mod.rs
```diff
@@ -78,6 +79,24 @@ impl<Call> Xcm<Call> {
...
+	pub fn into_inner(self) -> Vec<Instruction<Call>> { self.0 }
+
+	/// Return an iterator over references to the items.
+	pub fn iter(&self) -> impl Iterator<Item=&Instruction<Call>> { self.0.iter() }
+
+	/// Return an iterator over mutable references to the items.
+	pub fn iter_mut(&mut self) -> impl Iterator<Item=&mut Instruction<Call>> { self.0.iter_mut() }
+
+	/// Consume and return an iterator over the items.
+	pub fn into_iter(self) -> impl Iterator<Item=Instruction<Call>> { self.0.into_iter() }
```
KiChjang, line 98: We can just implement `IntoIterator` for the `Xcm` type instead of this being an inherent method. It will give us a nicer `for` loop syntax such as `for msg in xcm` without the need to call `into_iter()`.

gavofyork: Sure - feel free to make the change.

gavofyork:
Closed in favour of https://github.com/paritytech/polkadot/pull/4681 after accidentally merging `master`.

PR #4190, opened 2021-10-31
Create enter benchmarks

Lldenaurois:
Opening PR to run benchmarks on bench machine


emostov:
once the kusama benchmarks finish we should run rococo so we have valid weights for the burn in

emostov:
@Lldenaurois don't push here until the benchmark finishes running - these will take awhile. If we push the benchmark bot won't commit the generated file. Just make a new branch for now and we can merge it back in here later

PR #4195, opened 2021-11-01
Enable full use of pallet-bags-list in westend and kusama runtimes

kianenigma:
This pallet was already deployed, but for staking election iterations we used the nominator map just to be safe. I've done quite a bit of testing and everything seems fine. We can ditch the usage of nominator map `::iter` and use the new bags-list pallet. 

Next, once this is well tested on westend and kusama, we make the same change for polkadot.


kianenigma:
> So all nominators are sorted in the bag list, we don't need any migration isn't it ?

correct, migration has already happened.

thiolliere (approved):

kianenigma:
@shawntabrizi @emostov PTAL? the new version should be on its way soon.

emostov (approved):

kianenigma:
bot merge

PR #4202, opened 2021-11-01
Weights for para inherent

emostov:
target branch: bernhard-inherent-filtering
target pr: #4028


runtime/parachains/src/builder.rs
```diff
@@ -0,0 +1,535 @@
...
+			Default::default(), // digest,
+		)
+	}
+
+	fn relay_parent_number(&self) -> u32 {
+		(self.block_number - One::one())
+			.try_into()
+			.map_err(|_| ())
+			.expect("self.block_number is u32")
+	}
```
emostov, line 85: Is there a better way to get these unwrap values?

runtime/parachains/src/builder.rs
```diff
@@ -0,0 +1,535 @@
...
+	) {
+		let candidate_availability = Self::candidate_availability_mock(
+			group_idx,
+			core_idx,
+			candidate_hash,
+			availability_votes,
+		);
+		// TODO notes: commitments does not include any data that would lead to heavy code
+		// paths in `enact_candidate`. But enact_candidates does return a weight which will get
+		// taken into account.
```
emostov, line 146: This functionality is added in https://github.com/paritytech/polkadot/pull/4172

runtime/parachains/src/builder.rs
```diff
@@ -0,0 +1,535 @@
...
+						.benchmark_signature();
+
+						ValidityAttestation::Explicit(sig.clone())
+					})
+					.collect();
+
+				BackedCandidate::<T::Hash> {
+					candidate,
+					validity_votes,
+					validator_indices: bitvec::bitvec![bitvec::order::Lsb0, u8; 1; group_validators.len()],
```
emostov, line 418: TODO come back to this and make sure it makes sense

runtime/parachains/src/builder.rs
```diff
@@ -440,6 +443,8 @@ impl<T: paras_inherent::Config> BenchBuilder<T> {
 					})
 					.collect();
 
+				let _ = validity_votes.drain(*num_votes as usize..);
```
emostov, line 446: We can avoid making the signatures in the first place by just doing `group_validators.iter().take(num_votes).map(..` above

runtime/parachains/src/paras_inherent.rs
```diff
@@ -80,24 +81,13 @@ impl WeightInfo for TestWeightInfo {
...
-			<T as Config>::WeightInfo::enter_bitfields().saturating_mul(n_bitfields as Weight),
-		)
-}
-
-/// Extract the weight that is added _per_ bitfield.
-fn bitfield_weight<T: Config>() -> Weight {
-	<T as Config>::WeightInfo::enter_bitfields()
+	backed_candidates_weight::<T>(backed_candidates) +
+	signed_bitfields_weight::<T>(bitfields.len()) +
+	dispute_statements_weight::<T>(disputes)
```
emostov, line 90: Always use `saturating_add` for this stuff like this so we don't panic

runtime/parachains/src/paras_inherent.rs
```diff
@@ -107,9 +97,18 @@ fn minimal_inherent_weight<T: Config>() -> Weight {
 		.min(<T as Config>::WeightInfo::enter_backed_candidates_variable(0))
 }
 
-/// Extract the weight that is adder _per_ backed candidate.
-fn backed_candidate_weight<T: Config>(validity_votes_len: u32) -> Weight {
-	<T as Config>::WeightInfo::enter_backed_candidates_variable(validity_votes_len as u32)
+fn dispute_statements_weight<T: Config>(disputes: &[DisputeStatementSet]) -> Weight {
+    disputes.iter().map(|d| <<T as Config>::WeightInfo as WeightInfo>::enter_variable_disputes(d.statements.len() as u32)).sum()
```
emostov, line 101: We need to use `.saturating_add` and sum uses normal additions (see https://doc.rust-lang.org/src/core/iter/traits/accum.rs.html#45), so I would reccomend using a `.fold` with saturating add to sum these

runtime/parachains/src/paras_inherent.rs
```diff
@@ -371,9 +377,6 @@ pub mod pallet {
 				Vec::new()
 			});
 
-			let backed_candidates = limit_backed_candidates::<T>(backed_candidates);
```
emostov, line 490: Why is `limit_backed_candidates` removed?

runtime/parachains/src/paras_inherent.rs
```diff
@@ -252,8 +251,15 @@ pub mod pallet {
...
-			let bitfields_len = signed_bitfields.len() as u32;
+			let total_weight = paras_inherent_total_weight::<T>(
+				&backed_candidates,
+				&signed_bitfields,
+				&disputes,
+			);
+
+			if total_weight > <T as frame_system::Config>::BlockWeights::get().max_block {
+				return Ok(Some(minimal_inherent_weight::<T>()).into());
+			}
```
emostov, line 262: This might not be necessary. I think the weight annotations above the functions would keep it from getting queued if overweight.

runtime/parachains/src/paras_inherent.rs
```diff
@@ -404,12 +407,7 @@ pub mod pallet {
 			// And track that we've finished processing the inherent for this block.
 			Included::<T>::set(Some(()));
 
-			Ok(Some(paras_inherent_total_weight::<T>(
-				backed_candidates_len,
-				bitfields_len,
-				disputes_len,
-			))
-			.into())
+			Ok(Some(total_weight).into())
```
emostov, line 410: This isn't necessary. With the current logic this will be the same weight as what we give to `pallet::weight(`. We can just return `Ok()` here and the macro will use the weight from the `pallet::weight(` annotation. Returning the weight at the end of the call is only helpful when we used less weight than expected.

runtime/parachains/src/paras_inherent.rs
```diff
@@ -659,10 +658,10 @@ fn apply_weight_limit<T: Config + inclusion::Config>(
 	entropy: [u8; 32],
 	max_weight: Weight,
 ) -> (Weight, Vec<BackedCandidate<<T>::Hash>>, UncheckedSignedAvailabilityBitfields) {
-	let total_bitfields_weight = (bitfields.len() as Weight).saturating_mul(bitfield_weight::<T>());
+	let total_bitfields_weight = (bitfields.len() as Weight).saturating_mul(signed_bitfields_weight::<T>(bitfields.len()));
```
emostov, line 661: This looks wrong, I think we want

```
let total_bitfields_weight = signed_bitfields_weight::<T>(bitfields.len());
```

runtime/parachains/src/paras_inherent.rs
```diff
@@ -659,10 +658,10 @@ fn apply_weight_limit<T: Config + inclusion::Config>(
...
 
 	let total_candidates_weight = (candidates.len() as Weight)
-		.saturating_mul(backed_candidate_weight::<T>(MAX_ACTIVE_VALIDATORS));
+		.saturating_mul(backed_candidates_weight::<T>(candidates.as_slice()));
```
emostov, line 664:
```
let total_candidates_weight = backed_candidates_weight::<T>(candidates.as_slice());
```

runtime/parachains/src/paras_inherent.rs
```diff
@@ -868,169 +867,243 @@ fn sanitize_backed_candidates<
...
-fn limit_backed_candidates<T: Config>(
-	mut backed_candidates: Vec<BackedCandidate<T::Hash>>,
-) -> Vec<BackedCandidate<T::Hash>> {
-	const MAX_CODE_UPGRADES: usize = 1;
+fn limit_disputes<T: Config>(
+	disputes: &mut MultiDisputeStatementSet,
+) -> Weight {
+	let mut remaining_weight = <T as frame_system::Config>::BlockWeights::get().max_block;
+	let disputes_weight = dispute_statements_weight::<T>(&disputes);
+	if disputes_weight > remaining_weight {
```
emostov, line 875: This is prioritizing disputes over backed candidates and bitfields - is that what we want?

runtime/parachains/src/paras_inherent.rs
```diff
@@ -868,169 +867,243 @@ fn sanitize_backed_candidates<
...
+			let a_local_block = T::DisputesHandler::included_state(a.session, a.candidate_hash);
+			let b_local_block = T::DisputesHandler::included_state(b.session, b.candidate_hash);
+			match (a_local_block, b_local_block) {
+				// Prioritize local disputes over remote disputes.
+				(None, Some(_)) => Ordering::Greater,
+				(Some(_), None) => Ordering::Less,
+				// For local disputes, prioritize those that occur at an earlier height.
+				(Some(a_height), Some(b_height)) => a_height.cmp(&b_height),
+				// Prioritize earlier remote disputes using session as rough proxy.
+				(None, None) => a.session.cmp(&b.session),
```
emostov, line 889: Do we not have access to block number? If we are far away from the last session boundary I am not sure how relevant this is, especially if we are going to timeout old disputes

runtime/parachains/src/paras_inherent.rs
```diff
@@ -868,169 +867,243 @@ fn sanitize_backed_candidates<
...
+				0 as Weight
+			};
 
-				code_upgrades += 1;
+			if inclusion_weight > 0 {
+				remaining_weight -= dispute_weight;
 			}
 
-			true
+			inclusion_weight > 0
```
emostov, line 905: This logic can be simplified to about:

```rust
let dispute_weight = <<T as Config>::WeightInfo as WeightInfo>::enter_variable_disputes(d.statements.len() as u32);

if remaining_weight > dispute_weight {
    remaining_weight -= dispute_weight
    true
} else {
    false
}
``

File: runtime/parachains/src/paras_inherent.rs [hunk 15]
```diff
@@ -868,169 +867,243 @@ fn sanitize_backed_candidates<
...
+				dispute_statements,
+				dispute_sessions,
+				backed_and_concluding,
+				num_validators_per_core,
+				includes_code_upgrade,
+			}: TestConfig,
+		) -> Bench<Test> {
+			let scenario = BenchBuilder::<Test>::new()
+				.set_max_validators((dispute_sessions.len() as u32) * num_validators_per_core)
+				.set_max_validators_per_core(num_validators_per_core) // 5 validators, 1 validator per core => 5 cores.
```
emostov, line 946: delete comments here

runtime/parachains/src/builder.rs
Lldenaurois: Great point!

runtime/parachains/src/paras_inherent.rs
Lldenaurois: Nice catch

runtime/parachains/src/paras_inherent.rs
Lldenaurois: TODO: Make sure we can remove

runtime/parachains/src/paras_inherent.rs
Lldenaurois: We limit backed candidates in create_inherent, not in enter now.

runtime/parachains/src/paras_inherent.rs
Lldenaurois: Fixed in my latest commit

runtime/parachains/src/paras_inherent.rs
Lldenaurois: Yes

runtime/parachains/src/paras_inherent.rs
```diff
@@ -1061,24 +1053,91 @@ mod tests {
...
-				assert!(limit_para_inherent_data != expected_para_inherent_data);
+				let multi_dispute_inherent_data = Pallet::<Test>::create_inherent_inner(&inherent_data.clone()).unwrap();
+				assert!(multi_dispute_inherent_data != expected_para_inherent_data);
+
+				assert_eq!(multi_dispute_inherent_data.disputes.len(), 2);
 
 				assert_eq!(
-					limit_para_inherent_data.disputes.len(),
-					2usize,
+					&multi_dispute_inherent_data.disputes[..2],
```
emostov, line 1069: Can this just be `&multi_dispute_inherent_data.disputes` since it looks like it should only have 2 disputes?

runtime/parachains/src/paras_inherent.rs
Lldenaurois: I do not think we have access to block number if this is a remote dispute, so we simply sort by session to prioritize disputes from the previous session if need be

runtime/parachains/src/paras_inherent.rs
```diff
@@ -1061,24 +1053,91 @@ mod tests {
...
+				assert_eq!(<scheduler::Pallet<Test>>::scheduled(), vec![]);
+
+				assert_eq!(Pallet::<Test>::on_chain_votes(), None);
+				// Call enter with our 2 backed candidates
+				assert_ok!(Pallet::<Test>::enter(
+					frame_system::RawOrigin::None.into(),
+					multi_dispute_inherent_data,
+				));
+
+				assert_err!(
```
emostov, line 1084: Maybe a comment that testing `TooManyInclusionInherents` is not directly related to the point of this unit test, but just included here because its easy to add

runtime/parachains/src/paras_inherent.rs
```diff
@@ -1061,24 +1053,91 @@ mod tests {
...
+						expected_para_inherent_data,
+					),
+					DispatchError::from(Error::<Test>::TooManyInclusionInherents),
 				);
 
 				assert_eq!(
-					expected_para_inherent_data.disputes[2].session,
-					1,
+					// The length of this vec is equal to the number of candidates, so we know our 2
+					// backed candidates did not get filtered out
```
emostov, line 1094: comment needs update

runtime/parachains/src/paras_inherent.rs
Lldenaurois: Fixed in my latest commit

runtime/parachains/src/paras_inherent.rs
```diff
@@ -868,169 +867,243 @@ fn sanitize_backed_candidates<
...
+				let dispute_statements = BTreeMap::new();
+
+				let mut backed_and_concluding = BTreeMap::new();
+				backed_and_concluding.insert(0, 1);
+				backed_and_concluding.insert(1, 1);
+
+				let scenario = make_inherent_data(
+					TestConfig {
+						dispute_statements,
+						dispute_sessions: vec![0, 0,],
```
emostov, line 970: Why do we want session 0? Isn't the scenario at session 1? Is it just that it does not matter?

runtime/parachains/src/paras_inherent.rs
```diff
@@ -868,169 +867,243 @@ fn sanitize_backed_candidates<
...
+					// The length of this vec is equal to the number of candidates, so we know our 2
+					// backed candidates did not get filtered out
+					Pallet::<Test>::on_chain_votes().unwrap().backing_validators_per_candidate.len(),
+					2
+				);
 			});
 		}
 
 		#[test]
-		fn ignores_subsequent_code_upgrades() {
```
emostov, line 891: I am a little confused by this diff, did this test get deleted?

runtime/parachains/src/paras_inherent.rs
```diff
@@ -966,6 +966,8 @@ mod tests {
 		}
 
 		#[test]
+		// Validate that if we create 2 candidates and schedule them to 2 cores
+		// we obtain 2 backed candidates for each of the candidates in the inherent data
```
emostov, line 970: What about: 
```rust 
		// Validate that if we create 2 backed candidates which are assigned to 2 cores that will be freed via
        // becoming fully available, the backed candidates will not be filtered out in `create_inherent` and
        // will not cause `enter` to exit early.
```

runtime/parachains/src/builder.rs
```diff
@@ -489,7 +497,7 @@ impl<T: paras_inherent::Config> BenchBuilder<T> {
 
 				DisputeStatementSet {
 					candidate_hash: candidate_hash.clone(),
-					session: self.target_session,
+					session: session,
```
drahnr: Should be just `session,` right?

runtime/parachains/src/paras_inherent.rs
```diff
@@ -1114,29 +1110,27 @@ mod tests {
 		}
 
 		#[test]
+		// Ensure that when dispute data establishes an overlength block that we adequately
+		// filter out disputes according to our prioritization rule
```
emostov, line 1114: `overlength` should be `overweight`

runtime/parachains/src/paras_inherent.rs
```diff
@@ -1164,12 +1159,13 @@ mod tests {
...
 
 				assert_eq!(Pallet::<Test>::on_chain_votes(), None);
-				// Call enter with our 2 backed candidates
+				// Call enter with our 2 disputes
 				assert_ok!(Pallet::<Test>::enter(
 					frame_system::RawOrigin::None.into(),
 					limit_inherent_data,
 				));
 
+				// Ensure that calling enter with 3 disputes would cause an overlength block
```
emostov, line 1168: This is not correct - this error is because `enter` is already called for this block. If you want to check that `Ensure that calling enter with 3 disputes would cause an overlength block` it would be easisest to setup the scenario again in another integration test

To get around just this error you could clear the `Included` storage item , but then we will have other issues because the previous call to `enter` already modified storage, hence why I think the easiest/best way to test this is by creating a new scenario

runtime/parachains/src/paras_inherent.rs
```diff
@@ -728,7 +727,7 @@ fn apply_weight_limit<T: Config + inclusion::Config>(
 	let (total, indices) = random_sel::<UncheckedSignedAvailabilityBitfield, _>(
 		&mut rng,
 		bitfields.clone(),
-		|_| bitfield_weight::<T>(),
+    	|_| <<T as Config>::WeightInfo as WeightInfo>::enter_bitfields(),
```
drahnr: This is called per bitfield, so I am not sure this is what you want, the ellipsized arg would be the bitfield to meadure

runtime/parachains/src/paras_inherent.rs
emostov: `overlength` -> `over weight`

runtime/parachains/src/paras_inherent.rs
```diff
@@ -1191,33 +1186,33 @@ mod tests {
...
 
 				let mut backed_and_concluding = BTreeMap::new();
+				// 2 backed candidates shall be scheduled
 				backed_and_concluding.insert(0, 2);
 				backed_and_concluding.insert(1, 2);
 
 				let scenario = make_inherent_data(TestConfig {
 					dispute_statements,
-					dispute_sessions: vec![0, 0, 2, 2, 1],
+					dispute_sessions: vec![0, 0, 2, 2, 1], // 2 backed candidates, 3 disputes at sessions 2, 1 and 1 respectively
```
emostov, line 1204: this comment should be moved above this assignment - confusing why its right here

runtime/parachains/src/paras_inherent.rs
drahnr: We should prioritize local over remote ones, but also  randomize the remaining picks to not become gamable.

runtime/parachains/src/paras_inherent.rs
```diff
@@ -1234,21 +1229,31 @@ mod tests {
 					Pallet::<Test>::create_inherent_inner(&inherent_data.clone()).unwrap();
 				assert!(limit_inherent_data != expected_para_inherent_data);
 
+				// Three disputes is overlength (see previous test), so we expect to only see 2 disputes
 				assert_eq!(limit_inherent_data.disputes.len(), 2);
+				// Ensure disputes are filtered as expected
 				assert_eq!(limit_inherent_data.disputes[0].session, 1);
 				assert_eq!(limit_inherent_data.disputes[1].session, 2);
+				// Ensure all bitfields are included as these are still not overlength
```
emostov, line 1237: could be 
```
// Ensure all bitfields are included as they do not make the block overweight
```

runtime/parachains/src/paras_inherent.rs
```diff
@@ -1234,21 +1229,31 @@ mod tests {
...
 
 				assert_eq!(Pallet::<Test>::on_chain_votes(), None);
-				// Call enter with our 2 backed candidates
+				// Call enter with our 2 disputes
 				assert_ok!(Pallet::<Test>::enter(
 					frame_system::RawOrigin::None.into(),
 					limit_inherent_data,
 				));
 
+				// Ensure that calling enter with 3 disputes and 2 candidates is overlength
```
emostov, line 1256: I think this test is not doing what you want it to do - see https://github.com/paritytech/polkadot/pull/4202/commits/58fc1126a4d79bc07202f24b0a47b5c38dda79e5#r743589515

runtime/parachains/src/paras_inherent.rs
emostov: This should be correct, we assume `<<T as Config>::WeightInfo as WeightInfo>::enter_bitfields()` will always return the weight of processing one bitfield. We don't need to know anything about the bitfield itself though

runtime/parachains/src/paras_inherent.rs
emostov: Ok so basically we could

- iterate through disputes creating two new vecs: sorted_local_disputes & remote_disputes
- select local disputes until weight limit
- randomly select from `remote_disputes` until weight limit is reached

Lldenaurois: Disputes are already sorted to prioritize local disputes in `fn limit_disputes`.

I can add the logic for selecting the remote disputes at random

runtime/parachains/src/paras_inherent.rs
Lldenaurois: I tried to return Ok(), but the compiler would not allow it

runtime/parachains/src/paras_inherent.rs
Lldenaurois: Yes, but I think it's more explicit to slice it on the same bound

runtime/parachains/src/paras_inherent.rs
Lldenaurois: Well, it ensures that the prior data would have led to an overlength block, which ensures that we should indeed have filtered.

runtime/parachains/src/paras_inherent.rs
Lldenaurois: Yes, sessions are skipped for backed candidates and only applied to disputes. This should be modified to not necessitate the garbage vec entries.

runtime/parachains/src/paras_inherent.rs
Lldenaurois: Yes, these tests are no longer valid and they did not test the proper things to begin with.

runtime/parachains/src/paras_inherent.rs
Lldenaurois: You mean because the weight has already been applied based on the above enter function?

emostov: I think we discussed on call but just to clarify all the storage like AvailabilityCores, Included etc gets mutated by `enter`, so calling it twice in a row will not have the intended affect

runtime/parachains/src/paras_inherent.rs
```diff
@@ -368,12 +268,17 @@ pub mod pallet {
...
 			} = data;
-			let disputes_len = disputes.len() as u32;
-			let bitfields_len = signed_bitfields.len() as u32;
+			let total_weight =
+				paras_inherent_total_weight::<T>(&backed_candidates, &signed_bitfields, &disputes);
+
+			// Abort if the total weight of the block exceeds the max block weight
+			ensure!(
+				total_weight <= <T as frame_system::Config>::BlockWeights::get().max_block,
+				Error::<T>::TooManyInclusionInherents
```
emostov, line 277: We need a different error message here.

runtime/parachains/src/paras_inherent.rs
```diff
@@ -789,111 +893,735 @@ fn sanitize_backed_candidates<
...
+
+				// The current schedule is empty prior to calling `create_inherent_enter`.
+				assert_eq!(<scheduler::Pallet<Test>>::scheduled(), vec![]);
+
+				assert_err!(
+					Pallet::<Test>::enter(
+						frame_system::RawOrigin::None.into(),
+						expected_para_inherent_data,
+					),
+					DispatchError::from(Error::<Test>::InherentOverweight),
```
emostov, line 1503: Is there a way we can show this is overweight due to quantity of bitfields and not just because of the added disputes?

emostov: It would be good to show how this is different than `limit_dispute_data_ignore_backed_candidates_failure`

emostov:
relevant changes have been brought into #4028

PR #4209, opened 2021-11-02
PVF prechecking

slumber:
NOTE(pep): This is a PR for the integration branch. It is likely this branch won't be landed as is within this PR. It exists solely to enable CI runs and coordination for now.

Merge of PVF host [changes](https://github.com/paritytech/polkadot/pull/4123) with runtime [implementation](https://github.com/paritytech/polkadot/compare/pep-pvfcheck-runtime?expand=1) by @pepyakin

For additional details see https://github.com/paritytech/polkadot/issues/3211


node/core/runtime-api/src/lib.rs
```diff
@@ -396,10 +386,12 @@ where
 			query!(InboundHrmpChannelsContents, inbound_hrmp_channels_contents(id), sender),
 		Request::CurrentBabeEpoch(sender) => query!(CurrentBabeEpoch, current_epoch(), sender),
 		Request::FetchOnChainVotes(sender) => query!(FetchOnChainVotes, on_chain_votes(), sender),
-		Request::SubmitPvfCheckStatement(stmt, signature, sender) =>
-			query!(@NoCache, submit_pvf_check_statement(stmt, signature), sender),
-		Request::PvfsRequirePrecheck(sender) =>
-			query!(PvfsRequirePrecheck, pvfs_require_precheck(), sender),
+		Request::SubmitPvfCheckStatement(stmt, signature, sender) => {
+			query!(SubmitPvfCheckStatement, submit_pvf_check_statement(stmt, signature), sender)
```
pepyakin, line 390: Can you elaborate on those changes?

slumber: The fact that you don't cache the result is handled later in `store_cache`, returning `None` here would mean dropping the API request result. 
The pipeline is as follows:
1) In `query_cache` we propagate the request since it couldn't be cached
2) No essential changes in `make_runtime_api_request`
3) Do nothing in `store_cache`

The old behavior led to dropping the request immediately

node/network/collator-protocol/src/validator_side/mod.rs
```diff
@@ -1455,7 +1455,7 @@ async fn poll_collation_response(
 				);
 
 				CollationFetchResult::Error(COST_WRONG_PARA)
-			}
+			},
```
pepyakin, line 1458: These changes seem unrelated, can you revert them?

I suspect that your rustfmt might be outdated, you may check that by running `cargo fmt` on master.

slumber: I use the latest rust version, these changes are present in master, the branch is simply outdated

pepyakin:
We are the stage where we can proceed landing the same commits.

PR #3353, opened 2021-06-23
Add XCM Tracing

apopiak:
This PR adds `log::trace` logging for a selection of XCM methods.
This includes:
+ `log::debug` (should this also be trace?) the generic `execute_xcm` implementation
+ trace the relevant methods in the xcm-executor
+ trace the tuple implementations of various traits in case of failure/filter/miss

The logging choices are tentative and open for feedback. Let me know what should be added/removed.

#### Usage
Add `-lxcm=trace` to any Polkadot node.

PR #4211, opened 2021-11-03
Add more XCM tracing

apopiak:
This PR adds more tracing to different parts of XCM.

Feedback wanted on:
+ missing traces/logs
+ logs that should have a different level
+ logs that you think are unnecessary

#### Usage
As with https://github.com/paritytech/polkadot/pull/3353 add `-lxcm=trace` to any Polkadot node. With the new traces this might get noisy, so you can reduce to `-lxcm=debug` if you would like fewer logs.


xcm/xcm-executor/src/lib.rs
```diff
@@ -83,15 +83,44 @@ impl<Config: config::Config> ExecuteXcm<Config::Call> for XcmExecutor<Config> {
 		);
 		let xcm_weight = match Config::Weigher::weight(&mut message) {
 			Ok(x) => x,
-			Err(()) => return Outcome::Error(XcmError::WeightNotComputable),
+			Err(()) => {
+				log::debug!(
```
KiChjang, line 87: y u no trace! here?

apopiak: Cause I was thinking that it would be nice to get the most important subset of the logging via `-lxcm=debug`

KiChjang (approved):
it gud

shawntabrizi (approved):

apopiak:
bot merge

PR #4213, opened 2021-11-03
introduce trusted teleporter Encointer for Rococo and Westend

brenzi:
After discussion with @gavofyork, we have adjusted our common good parachain governance model. As desired for common good, there will be no root origin on the parachain and with this PR we'd like to be included in the trusted teleport set.

We believe this is in line with Kusama council's decision to accept Encointer as a common good:
Discussion: https://kusama.polkassembly.io/post/610
Executed signalling motion:  https://kusama.polkassembly.io/motion/357

As discussed with @joepetrowski, we'd test teleport on rococo first and wait for this PR to enter a release before testing teleport and set_code on westmint 

@joepetrowski may I ask for your review? Should this work?

### DIFF
--- a/runtime/rococo/src/lib.rs
+++ b/runtime/rococo/src/lib.rs
@@ -671,6 +671,7 @@ parameter_types! {
 	pub const RococoForTrack: (MultiAssetFilter, MultiLocation) = (Rococo::get(), Parachain(120).into());
 	pub const RococoForStatemine: (MultiAssetFilter, MultiLocation) = (Rococo::get(), Parachain(1000).into());
 	pub const RococoForCanvas: (MultiAssetFilter, MultiLocation) = (Rococo::get(), Parachain(1002).into());
+	pub const RococoForEncointer: (MultiAssetFilter, MultiLocation) = (Rococo::get(), Parachain(1003).into());
 	pub const MaxInstructions: u32 = 100;
 }
 pub type TrustedTeleporters = (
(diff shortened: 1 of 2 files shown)


joepetrowski:
Looks OK to me. `set_code` can be tested independently of teleportation.

brenzi:
was able to teleport from rococo-local (this PR) to our parachain master: https://github.com/encointer/encointer-parachain
backwards didn't work because of a decoding failure in js/apps

KiChjang (approved):
I don't really have a strong opinion with test networks; anyone should be able to just configure themselves be a trusted teleporter.

KiChjang:
Just one thing though: Westend doesn't allow for unpaid executions from anything other than a system parachain, whereas in this PR, you've explicitly whitelisted Encointer for the barrier. Is this sufficient for your testing purposes?

brenzi:
@KiChjang Our parachain comes without a root origin and starts with zero token supply (apart from the ED for the treasury account). What we need for testing is:
* teleport relaychain tokens to our parachain. Otherwise we can't send extrinsics at all
* runtime upgrade through relaychain
That would be enough ATM

KiChjang:
Okay, just making sure that you're aware that Westend has a different barrier condition from Rococo, i.e. if you're ever getting a `Barrier` error on Westend, try testing on Rococo and see if it resolves your issues.

brenzi:
rebased on polkadot-0.9.13

brenzi:
Now that Encointer has been ratified to become a common good parachain on Kusama, can we merge this for the next runtime upgrade?
see #4401 too

joepetrowski (approved):

PR #4220, opened 2021-11-04
Increase maximum chunk size to adjust for small networks.

eskimor:
And warn on invalid merkle proofs in availability recovery.

### DIFF
--- a/node/network/availability-recovery/src/lib.rs
+++ b/node/network/availability-recovery/src/lib.rs
@@ -393,7 +393,7 @@ impl RequestChunksPhase {
 						metrics.on_chunk_request_invalid();
 						self.error_count += 1;
 
-						tracing::warn!(
+						tracing::debug!(
 							target: LOG_TARGET,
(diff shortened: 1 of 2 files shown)


drahnr (approved):
LGTM!

node/network/availability-recovery/src/lib.rs
```diff
@@ -393,7 +393,7 @@ impl RequestChunksPhase {
 						metrics.on_chunk_request_invalid();
 						self.error_count += 1;
 
-						tracing::debug!(
+						tracing::warn!(
```
rphmeier, line 396: Warning on invalid behavior can be extremely noisy. Nodes misbehaving is abnormal but within expectations, so we typically don't 'warn' in such cases.

rphmeier: Still unaddressed

eskimor: Yep, alert on `on_chunk_request_invalid` would probably be better.

PR #4231, opened 2021-11-05
fix(staking miner): use `StorageKey` in rpc call getStorage

niklasad1:
Use `StorageKey` in the rpc call `state_getStorage` instead of `[u8;32]`, the reason is that StorageKey is serialized as `hex string` and rejects types that is serialized as array.


emostov (approved):

kianenigma (approved):

emostov:
bot merge

PR #4246, opened 2021-11-09
Fix parachain runtime upgrades on local networks

bkchr:
On local networks we often use runtime upgrade delays for parachains of
1 block. However, with the new go-ahead signal, this is broken. The
problem is that when we enact a block in relay chain block N, the
upgrade block X will already be
in the past and the go-ahead signal will be set in the next relay chain
block N + 1. However, the next parachain block that is build and enacted
on N, will not have seen the go-ahead signal, but it is required to
enact the runtime upgrade. The solution to this problem is to directly
set the go-ahead signal when we schedule a runtime upgrade and the
expected block is in the past.

### DIFF
--- a/runtime/parachains/src/paras.rs
+++ b/runtime/parachains/src/paras.rs
@@ -33,7 +33,10 @@ use primitives::v1::{
 };
 use scale_info::TypeInfo;
 use sp_core::RuntimeDebug;
-use sp_runtime::{traits::One, DispatchResult, SaturatedConversion};
+use sp_runtime::{
+	traits::{BlockNumberProvider, One},
+	DispatchResult, SaturatedConversion,
+};
 use sp_std::{prelude::*, result};
 
 #[cfg(feature = "std")]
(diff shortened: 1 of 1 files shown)


runtime/parachains/src/paras.rs
```diff
@@ -966,18 +969,34 @@ impl<T: Config> Pallet<T> {
...
+				// (calculated from the point the parachain block was build on).
 				let expected_at = relay_parent_number + cfg.validation_upgrade_delay;
 				let next_possible_upgrade_at =
 					relay_parent_number + cfg.validation_upgrade_frequency;
+				// The block number of the current block.
+				let now = frame_system::Pallet::<T>::current_block_number();
+
+				// If the expected upgrade block is in the past, we need
+				// to move this up to the current block. Aka when a parachain block
+				// will be enacted that was build on this or any descendent block the
```
pepyakin, line 982:
```suggestion
				// will be enacted that was built on this or any descendent block the
```

runtime/parachains/src/paras.rs
```diff
@@ -966,18 +969,34 @@ impl<T: Config> Pallet<T> {
...
+				// runtime upgrade should be enacted.
+				let expected_at = if expected_at <= now { now } else { expected_at };
```
pepyakin, line 984: Nit: probably worth updating for that case in the docs for this method.

runtime/parachains/src/paras.rs
```diff
@@ -966,18 +969,34 @@ impl<T: Config> Pallet<T> {
...
+				if expected_at <= now {
+					UpgradeGoAheadSignal::<T>::insert(&id, UpgradeGoAhead::GoAhead);
+				} else {
+					<Self as Store>::UpcomingUpgrades::mutate(|upcoming_upgrades| {
+						let insert_idx = upcoming_upgrades
+							.binary_search_by_key(&expected_at, |&(_, b)| b)
+							.unwrap_or_else(|idx| idx);
+						upcoming_upgrades.insert(insert_idx, (id, expected_at));
+					});
+				}
```
pepyakin, line 999: The weight profile of this function was changed, so we probably should update it.

rphmeier:
Implementers' Guide changes needed

PR #4280, opened 2021-11-14
Gracefully handle filtering bitfields & candidates

emostov:
target pr: #4028
target branch: bernhard-inherent-filtering

This updates the logic for `sanitize_bitfields` and `sanitize_backed_candidates` to never error when there is an issue, but instead to simply skip the problematic items. 

We are not worried about unbounded iteration due to malicious inputs in either case because we will totally drop all bitfields and backed candidates if all the inherent data is overweight prior to calling either function.


runtime/parachains/src/paras_inherent.rs
```diff
@@ -355,11 +350,12 @@ pub mod pallet {
...
 					.collect();
 
+				// Note that `provide_multi_dispute_data` will iterate, verify, and import each
+				// dispute; so the input here must be reasonably bounded.
 				let _ = T::DisputesHandler::provide_multi_dispute_data(disputes.clone())?;
 				if T::DisputesHandler::is_frozen() {
 					// The relay chain we are currently on is invalid. Proceed no further on parachains.
-					Included::<T>::set(Some(()));
-					return Ok(Some(minimal_inherent_weight::<T>()).into())
+					return Ok(Some(dispute_statements_weight::<T>(&disputes)).into())
```
emostov, line 358: If this does not get merged today I will pull out this change to its own  mini PR since its not related to the other changes

runtime/parachains/src/paras_inherent.rs
```diff
@@ -326,17 +318,20 @@ pub mod pallet {
...
+			ensure_none(origin)?;
+			ensure!(!Included::<T>::exists(), Error::<T>::TooManyInclusionInherents);
+			// Once we are sure we can proceed, mark as included
+			Included::<T>::set(Some(()));
+
 			let total_weight =
 				paras_inherent_total_weight::<T>(&backed_candidates, &signed_bitfields, &disputes);
-
 			// Abort if the total weight of the block exceeds the max block weight
+			// TODO this should just limit disputes
```
drahnr, line 329: Actually limit disputes, but remove everything else.

runtime/parachains/src/paras_inherent.rs
```diff
@@ -793,39 +792,34 @@ pub(crate) fn sanitize_bitfields<T: crate::inclusion::Config, const EARLY_RETURN
...
 			continue
 		}
 
 		let validator_index = unchecked_bitfield.unchecked_validator_index();
 
 		let validator_public = &validators[validator_index.0 as usize];
 
-		// only check the signatures when returning early
+		// Only check the signatures when returning early
 		if EARLY_RETURN {
```
drahnr, line 816: Renaming this is probably a good idea :)

runtime/parachains/src/paras_inherent.rs
emostov: https://github.com/paritytech/polkadot/pull/4281

runtime/parachains/src/inclusion.rs
```diff
@@ -398,7 +398,7 @@ impl<T: Config> Pallet<T> {
 		signed_bitfields: UncheckedSignedAvailabilityBitfields,
 		disputed_bitfield: DisputedBitfield,
 		core_lookup: impl Fn(CoreIndex) -> Option<ParaId>,
-	) -> Result<Vec<(CoreIndex, CandidateHash)>, DispatchError> {
+	) -> Vec<(CoreIndex, CandidateHash)> {
```
drahnr, line 401: We should consider returning `Result<Vec<(CoreIndex)>, Vec<(CoreIndex)>>` to be able to easily evaluate if the sanitization caused any changes and be able to warn outside this fn call and also keep the `assert_noop!` storage check macro.

runtime/parachains/src/inclusion.rs
```diff
@@ -1364,6 +1363,19 @@ pub(crate) mod tests {
...
+				for (p_id, _) in paras {
+					PendingAvailability::<Test>::insert(
+						p_id,
+						CandidatePendingAvailability {
+							availability_votes: default_availability_votes(),
+							..Default::default()
+						},
+					)
+				}
+			};
```
drahnr, line 1377: :+1:

runtime/parachains/src/inclusion.rs
```diff
@@ -1398,48 +1410,80 @@ pub(crate) mod tests {
...
+					1
 				);
+
+				// clean up
+				PendingAvailability::<Test>::remove_all(None);
 			}
 
 			// out of order.
 			{
-				let bare_bitfield = default_bitfield();
```
drahnr, line 1437: :+1:

runtime/parachains/src/inclusion.rs
```diff
@@ -1448,20 +1492,41 @@ pub(crate) mod tests {
...
 					ParaInclusion::process_bitfields(
 						expected_bits(),
 						vec![signed_1, signed_0],
 						DisputedBitfield::zeros(expected_bits()),
 						&core_lookup,
-					),
-					Error::<Test>::BitfieldDuplicateOrUnordered
+					)
+					.len(),
+					0
```
drahnr, line 1518: nit: `.is_empty()`

runtime/parachains/src/paras_inherent.rs
```diff
@@ -646,22 +622,6 @@ impl<T: Config> Pallet<T> {
...
-		if !condition {
-			if $action {
-				ensure!(condition, $err);
-			} else {
-				$($alt)?
-			}
-		}
-	};
-}
-
```
drahnr, line 664: :+1: I feel a lot better seeing this removed, it added quite some mental overhead combined with the const generic

runtime/parachains/src/paras_inherent.rs
```diff
@@ -803,81 +763,68 @@ fn apply_weight_limit<T: Config + inclusion::Config>(
 /// they were actually checked and filtered to allow using it in both
 /// cases, as `filtering` and `checking` stage.
 ///
-/// `EARLY_RETURN` determines the behavior.
-/// `false` assures that all inputs are filtered, and invalid ones are filtered out.
-/// It also skips signature verification.
-/// `true` returns an `Err(_)` on the first check failing.
-pub(crate) fn sanitize_bitfields<T: crate::inclusion::Config, const EARLY_RETURN: bool>(
+/// `CHECK_SIGS` determines if validator signatures are checked. If true, bitfields that have an
```
drahnr, line 766: Something that is lingering in the back of my head:
```
const VERIFY_SIGS: bool = true;
const SKIP_SIG_VERIFY: bool = false;
```
so they could be used instead of `true` and `false` in the const generic calls.

runtime/parachains/src/paras_inherent.rs
```diff
@@ -2016,83 +1973,49 @@ mod tests {
...
+			// happy path
+			assert_eq!(
+				sanitize_backed_candidates::<Test, _>(
+					relay_parent,
+					backed_candidates.clone(),
+					has_concluded_invalid,
+					scheduled
+				)
+				.len(),
+				2
```
drahnr, line 1985: Verify the content is identical.

drahnr (approved):
Mostly nits, one design decission worth discussing. Other than that, LGTM! Thank you!

runtime/parachains/src/inclusion.rs
emostov: We can use `assert_storage_noop!` if we need to check for storage root changes https://crates.parity.io/frame_support/macro.assert_storage_noop.html

What storage changes though do you want the system to keep track of? It seems like the calling logic doesn't have any handling based on if storage changes are made here

runtime/parachains/src/inclusion.rs
emostov: 8b239cf

runtime/parachains/src/paras_inherent.rs
emostov: a414390fe6ed023e1b9741540d446643cfebf03b

runtime/parachains/src/paras_inherent.rs
emostov: 7d330e8

runtime/parachains/src/inclusion.rs
drahnr: I can't follow. The `Ok()/Err()` would be independent of the `storage_noop` and we should verify that no storage was written in case we do error (which from the tests is never the case), which I missed. So please discard that part :]

runtime/parachains/src/paras_inherent.rs
emostov: e1c32e6a82b85010dc9dfc25740b4cbdaed7be8a

PR #4287, opened 2021-11-15
Refactor inherent data weight limiting logic

emostov:
target PR: #4028

This PR
- refactors limiting functions so `rng` can be reused
- excludes remote disputes when filtering in enter if the disputes are overweight (not sure if we still want this though)

### DIFF
--- a/runtime/parachains/src/paras_inherent.rs
+++ b/runtime/parachains/src/paras_inherent.rs
@@ -1410,7 +1410,7 @@ mod tests {
 				assert_ok!(Pallet::<Test>::enter(
 					frame_system::RawOrigin::None.into(),
 					expected_para_inherent_data,
-				),);
+				));
 
 				assert_eq!(
 					// The length of this vec is equal to the number of candidates, so we know
@@ -1552,7 +1552,7 @@ mod tests {
 				assert_ok!(Pallet::<Test>::enter(
 					frame_system::RawOrigin::None.into(),
 					expected_para_inherent_data,
-				),);
+				));
 
 				assert_eq!(
 					// The length of this vec is equal to the number of candidates, so we know
@@ -1623,7 +1623,7 @@ mod tests {
 				assert_ok!(Pallet::<Test>::enter(
 					frame_system::RawOrigin::None.into(),
 					limit_inherent_data,
-				),);
+				));
 
 				assert_eq!(
 					// The length of this vec is equal to the number of candidates, so we know our 2
@@ -1675,7 +1675,7 @@ mod tests {
 				assert_ok!(Pallet::<Test>::enter(
 					frame_system::RawOrigin::None.into(),
 					expected_para_inherent_data,
-				),);
+				));
 
 				assert_eq!(
 					// The length of this vec is equal to the number of candidates, so we know our 2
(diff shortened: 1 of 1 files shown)


runtime/parachains/src/paras_inherent.rs
```diff
@@ -349,20 +349,39 @@ pub mod pallet {
...
 
-			// Abort if the total weight of the block exceeds the max block weight
-			if candidate_weight.saturating_add(bitfields_weight).saturating_add(disputes_weight) > max_block_weight {
-				backed_candidates.clear();
-				candidate_weight = 0;
-				signed_bitfields.clear();
-				bitfields_weight = 0;
-			}
+			// Potentially trim inherent data to ensure processing will be within weight limits
+			let total_weight = {
```
emostov, line 353: hopefully not controversial, i just thought it was easier to understand when we grouped this logic into a block

runtime/parachains/src/paras_inherent.rs
```diff
@@ -981,29 +1010,28 @@ fn limit_disputes<T: Config>(disputes: &mut MultiDisputeStatementSet, entropy: [
 			}
 		});
 
-		// Compute the statements length of all remote disputes
-		let d = remote_disputes.iter().map(|d| d.statements.len() as u32).collect::<Vec<u32>>();
+		if INCLUDE_REMOTE && !remaining_weight.is_zero() {
```
emostov, line 1013: This is the main logic change that needs close review

runtime/parachains/src/paras_inherent.rs
```diff
@@ -141,7 +143,7 @@ fn backed_candidate_weight<T: frame_system::Config + Config>(
 	<<T as Config>::WeightInfo as WeightInfo>::enter_backed_candidates_variable(
 		candidate.validity_votes.len() as u32,
 	)
-	.saturating_add(if candidate.candidate.commitments.new_validation_code.is_some() {
+	.max(if candidate.candidate.commitments.new_validation_code.is_some() {
```
drahnr, line 146: I don't quite get the `max` here? It would at least mandate a comment.

emostov: I can probably write the code in a more idiomatic way - but basically if there is `new_validation_code` we have a separate weight that we can use. So to get the worst case weight for the candidate we just max two. 

In other words adding them does not make sense because that would give the weight of a `weight_backed_candidate + weigh_backed_candidate_with_validation_code` but we actually just want the worst case weight for the one candidate

emostov: I changed the logic to a simple if else 9dc1abbf096889c2bbe1f4cdabb818551a9a2834

runtime/parachains/src/paras_inherent.rs
drahnr: If it survives `cargo +nightly fmt`, :+1:

runtime/parachains/src/paras_inherent.rs
drahnr: I see, makes sense :+1:

runtime/parachains/src/paras_inherent.rs
drahnr: This needs additional thought, if there is still weight, what's the rationale of throwing away remote disputes? We can defer this discussion, but we should make an explicit statement about the rationale.

emostov: If I recall correctly the rationale we discussed was that remote disputes are the main type that can be created as spam. But on further thought I am gonna remove this since I as long as we are within weight limit we should still include even if they end up being spam so the validator crafting the malicious disputes has a better chance of getting slashed

Issue #4300, opened 2021-11-16
PVF validation host livelock


pepyakin:
There is a potential livelock in the PVF validation host code.

In order to trigger it the following set of conditions need to take place:

1. there is a request for preparation of a certain PVF.
1. the preparation worker process dies.
1. approximately at the same time, the pool receives a message which then leads to calling of `purge_dead` clean up routine.
1. this leads to a race between `purge_dead` and I/O error originating from a read call on the UDS socket that connects the worker and the validation host. (NOTE that the race itself is not unforeseen and was an acceptable part of the design)
1. rip message is sent back to the queue
1. the queue will react by re-adding the message back into the execution queue optionally spawning an additional worker.
1. then when the worker is spawned or freed and picks up that job the cycle starts all over.

So in order to trigger it, those things that should take place:

1. the preparation worker dies.
1. the preparation pool receives a message in the narrow time window that triggers `purge_dead`
1. on top of that `purge_dead` wins the race to the read I/O error.

The first condition may be not easy to trigger, but it is possible. Either the node itself is under heavy load (esp. memory-wise) or the attacker crafted a PVF that can lead to panics in the preparation process.

The second and the third condition seem to be very unlikely. The preparation needs to be requested just in time between the exploited worker dies and but before the kernel notified the polkadot process that the pipe is closed and the async runtime picked up that change.

PR #4304, opened 2021-11-16
prepare worker: Catch unexpected unwinds

pepyakin:
Even though a panic is not expected, we catch it anyway. 

Related: #4209 #4300

### DIFF
--- a/node/core/pvf/src/prepare/worker.rs
+++ b/node/core/pvf/src/prepare/worker.rs
@@ -30,7 +30,7 @@ use async_std::{
 };
 use parity_scale_codec::{Decode, Encode};
 use sp_core::hexdisplay::HexDisplay;
-use std::{sync::Arc, time::Duration};
+use std::{any::Any, panic, sync::Arc, time::Duration};
 
 const NICENESS_BACKGROUND: i32 = 10;
 const NICENESS_FOREGROUND: i32 = 0;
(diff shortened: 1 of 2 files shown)


bkchr (approved):

node/core/pvf/src/prepare/worker.rs
```diff
@@ -318,13 +318,31 @@ pub fn worker_entrypoint(socket_path: &str) {
...
+/// Attempt to convert an opaque panic payload to a string.
+///
+/// This is a best effort, and is not guaranteed to provide the most accurate value.
+fn stringify_panic_payload(payload: Box<dyn Any + Send + 'static>) -> String {
+	match payload.downcast::<&'static str>() {
+		Ok(msg) => msg.to_string(),
+		Err(payload) => match payload.downcast::<String>() {
+			Ok(msg) => *msg,
+			// At least we tried...
+			Err(_) => "Box<Any>".to_string(),
```
slumber, line 345: Does it make sense to replace it with "Unknown" or "Unknown reason" for readability in logs?

slumber (approved):

node/core/pvf/src/prepare/worker.rs
pepyakin: Good call!

PR #4312, opened 2021-11-17
Add Statemine as parachain 1000 to rococo trusted teleporters

apopiak:
Fixing a weird historical thing where Statemint-like chains are at id 1001 on Rococo (and thus you cannot teleport funds to Statemint if it is registered at id 1000).
We can remove 1001 later if we want.


KiChjang:
I'm confused, why would you need Statemine on Rococo when you already have Rockmine? Wouldn't it be better if you just moved Rockmine to 1000?

NachoPal:
> I'm confused, why would you need Statemine on Rococo when you already have Rockmine? Wouldn't it be better if you just moved Rockmine to 1000?

Rockmine was never added. Some context here -> https://github.com/paritytech/cumulus/pull/684

KiChjang:
Okay, does it make more sense to just remove the Rockmine config here then? No reason to keep two assets parachains in Rococo right?

bkchr:
> Okay, does it make more sense to just remove the Rockmine config here then? No reason to keep two assets parachains in Rococo right?

Sounds reasonable to me.

apopiak:
Makes the migration a little harder, but I can remove it, sure.

bkchr (approved):

Issue #4289, opened 2021-11-15
Add Statemint as `TrustedTeleporter` to Polkadot


joepetrowski:
When Shell upgrades to Statemint, it should be a `TrustedTeleporter`, similar to [Statemine on Kusama](https://github.com/paritytech/polkadot/blob/v0.9.12/runtime/kusama/src/lib.rs#L1328L1351). Adding this prior to the Statemint upgrade will prevent the need to migrate a reserve account.

PR #4313, opened 2021-11-17
Add XCM pallet and config to Polkadot runtime

### LINKED ISSUE #4309: Add pallet-xcm to Polkadot
In order to trigger the upgrade from Shell to Statemint we will need governance to send an XCM to Shell.
For that to happen we need to include pallet-xcm (as well as some of the other XCM machinery like the executor) into the Polkadot runtime.

NachoPal:
This PR adds support for sending XCMs in the Polkadot runtime by adding `pallet-xcm`.

The configuration is mostly similar to Kusama with these differences:
+ AdvertisedXcmVersion is set to 2.
+ Child System parachain related configs are removed.

This PR also adds Statemint as a trusted teleporter for DOTs.

closes https://github.com/paritytech/polkadot/issues/4289
closes #4309


runtime/polkadot/src/lib.rs
```diff
@@ -1280,6 +1293,154 @@ impl auctions::Config for Runtime {
...
+	type XcmExecuteFilter = Nothing;
+	type XcmExecutor = XcmExecutor<XcmConfig>;
+	type XcmTeleportFilter = Everything;
+	type XcmReserveTransferFilter = Everything;
+	type Weigher = FixedWeightBounds<BaseXcmWeight, Call, MaxInstructions>;
+	type LocationInverter = LocationInverter<Ancestry>;
+	type Origin = Origin;
+	type Call = Call;
+	const VERSION_DISCOVERY_QUEUE_SIZE: u32 = 100;
+	type AdvertisedXcmVersion = pallet_xcm::CurrentXcmVersion;
```
KiChjang, line 1441: This means every time we release a new XCM version, we'd advertise to every other chain that we're on a newer version. Is this intended? It doesn't sound like a good idea given that we may want to let the new XCM version sit for a while before things stabilize...

runtime/polkadot/src/lib.rs
```diff
@@ -1280,6 +1293,154 @@ impl auctions::Config for Runtime {
...
+/// The means that we convert an the XCM message origin location into a local dispatch origin.
+type LocalOriginConverter = (
+	// A `Signed` origin of the sovereign account that the original location controls.
+	SovereignSignedViaLocation<SovereignAccountOf, Origin>,
+	// A child parachain, natively expressed, has the `Parachain` origin.
+	ChildParachainAsNative<parachains_origin::Origin, Origin>,
+	// The AccountId32 location type can be expressed natively as a `Signed` origin.
+	SignedAccountId32AsNative<PolkadotNetwork, Origin>,
+	// A system child parachain, expressed as a Superuser, converts to the `Root` origin.
+	ChildSystemParachainAsSuperuser<ParaId, Origin>,
```
apopiak, line 1345: I wonder whether we might want to not add this, given that there are no system parachains, yet.

runtime/polkadot/src/lib.rs
```diff
@@ -1280,6 +1293,154 @@ impl auctions::Config for Runtime {
...
+impl pallet_xcm::Config for Runtime {
+	type Event = Event;
+	type SendXcmOrigin = xcm_builder::EnsureXcmOrigin<Origin, LocalOriginToLocation>;
+	type XcmRouter = XcmRouter;
+	// Anyone can execute XCM messages locally...
+	type ExecuteXcmOrigin = xcm_builder::EnsureXcmOrigin<Origin, LocalOriginToLocation>;
+	// ...but they must match our filter, which rejects all.
+	type XcmExecuteFilter = Nothing;
+	type XcmExecutor = XcmExecutor<XcmConfig>;
+	type XcmTeleportFilter = Everything;
```
apopiak, line 1434: I wonder whether we might want to limit this to Statemint.

KiChjang: I was a bit confused that there's apparently two filters for teleports: `XcmTeleportFilter` here in `pallet_xcm::Config`, and `IsTeleporter` in `xcm_executor::Config`. This `XcmTeleportFilter` here specifically checks if the origin is allowed to _send_ assets via `teleport_assets` on the XCM pallet, whereas `IsTeleporter` controls whether or not we trust the sender's origin enough to _receive_ assets from them.

So, going by that logic, I don't think we should just limit `XcmTeleportFilter` to the point where we can only send assets to Statemint; that would mean we'd have to always perform a runtime upgrade every time we add a new trusted teleport destination. If the concern here is that users might accidentally send assets to an unsupported location, then I believe the UI should limit teleport destination choices, rather than changing the config here banning all possible teleport locations except for Statemint.

runtime/polkadot/src/lib.rs
```diff
@@ -1280,6 +1293,157 @@ impl auctions::Config for Runtime {
...
+pub type XcmRouter = (
+	// Only one router so far - use DMP to communicate with child parachains.
+	xcm_sender::ChildParachainRouter<Runtime, XcmPallet>,
+);
+
+parameter_types! {
+	pub const Polkadot: MultiAssetFilter = Wild(AllOf { fun: WildFungible, id: Concrete(DotLocation::get()) });
+	pub const PolkadotForStatemint: (MultiAssetFilter, MultiLocation) = (Polkadot::get(), Parachain(1000).into());
+}
+pub type TrustedTeleporters = (xcm_builder::Case<PolkadotForStatemint>,);
```
apopiak, line 1367: @joepetrowski this addresses https://github.com/paritytech/polkadot/issues/4289

runtime/polkadot/src/lib.rs
gavofyork: Can leave it out for now

runtime/polkadot/src/lib.rs
gavofyork: But `Xcm*Filter` is effectively a filter on the dispatch origin (post-conversion to a `MultiLocation` via `ExecuteXcmOrigin`) for which location/asset combinations are allowed to use the teleporting/reserve asset transfer dispatchables. I don't see a big problem with using `Everything` here.

apopiak: Thinking about what I think our policy should be, a filter limiting teleports to parachains with id < 2000 seems like the right choice to me.

gavofyork (approved):

apopiak (approved):

runtime/polkadot/Cargo.toml
```diff
@@ -158,6 +163,10 @@ std = [
 	"beefy-primitives/std",
 	"pallet-mmr-primitives/std",
 	"frame-election-provider-support/std",
+  "pallet-xcm/std",
```
xlc, line 166: tab or spaces, this is the question

PR #4318, opened 2021-11-18
Improve paras runtime `BenchBuilder` api

emostov:
This PR primarily improves the `BenchBuilder::build` API by not requiring the user to left pad the `dispute_sessions` input to account for the cores used for `backed_and_concluding`.


runtime/parachains/src/paras_inherent.rs
```diff
@@ -1217,12 +1220,12 @@ mod tests {
 			new_test_ext(MockGenesisConfig::default()).execute_with(|| {
 				// Create the inherent data for this block
 				let dispute_statements = BTreeMap::new();
-				// No backed and concluding cores, so all cores will be fileld with disputesw
+				// No backed and concluding cores, so all cores will be filld with disputes.
```
drahnr, line 1223: If this was `cargo-spellcheck` not doing the correcting right, please file an issue there.

runtime/parachains/src/paras_inherent/benchmarking.rs
```diff
@@ -60,7 +60,7 @@ benchmarks! {
 				.collect();
 
 		let scenario = BenchBuilder::<T>::new()
-			.build(cores_with_backed, &[1], None);
+			.build(cores_with_backed, &[], None);
```
drahnr, line 63: I would be nice to remove the cryptic slice `&[]` of integers from the `build` param and add additional builder methods.

emostov: 4a4d03c

Lldenaurois (approved):

runtime/parachains/src/builder.rs
```diff
@@ -86,24 +104,40 @@ impl<T: paras_inherent::Config> BenchBuilder<T> {
...
+	/// Set `self.backed_and_concluding_cores`.
+	pub(crate) fn set_backed_and_concluding_cores(
+		mut self,
+		backed_and_concluding_cores: BTreeMap<u32, u32>,
+	) -> Self {
+		self.backed_and_concluding_cores = backed_and_concluding_cores;
+		self
+	}
+
+	pub(crate) fn set_code_upgrade(mut self, code_upgrade: Option<u32>) -> Self {
```
drahnr, line 136: nit: use `impl Into<u32>` since `Option<T>` provides an `impl From<T> for T {..}` https://doc.rust-lang.org/std/option/enum.Option.html#impl-From%3CT%3E

runtime/parachains/src/builder.rs
```diff
@@ -537,46 +594,51 @@ impl<T: paras_inherent::Config> BenchBuilder<T> {
...
+	/// * `set_backed_and_concluding_cores`: Set a map from core/para id seed to number of validity
+	/// votes.
+	/// * `set_dispute_sessions`: Set the session index of for each dispute. Index of slice
+	/// corresponds to a core, which is offset by the number of entries for
+	/// `backed_and_concluding_cores`. I.E. if `backed_and_concluding_cores` cores has 3 entries,
+	/// the first index of `dispute_sessions` will correspond to core index 3. There must be an
+	/// entry for each core with a dispute statement set.
+	/// *`set_code_upgrade`: Set to `Some` to include a code upgrade for all backed candidates.
+	/// The value within `Some` will be the byte length of the code.
+	pub(crate) fn build(self) -> Bench<T> {
```
drahnr, line 613: wild idea, feel free to discard: return `Result<Bench<T>, Bench<T>>` and determine integrity of the bench setup, the user can then determine via `unwrap_err` and `unwrap` if that is desired or not. Not sure how feasabile this is and what kind of checks we can make with reasonable effort.

runtime/parachains/src/builder.rs
```diff
@@ -86,24 +104,40 @@ impl<T: paras_inherent::Config> BenchBuilder<T> {
...
 			dispute_statements: BTreeMap::new(),
+			dispute_sessions: Default::default(),
+			backed_and_concluding_cores: Default::default(),
+			code_upgrade: None,
 			_phantom: sp_std::marker::PhantomData::<T>,
 		}
 	}
 
+	/// Set `self.dispute_sessions`.
+	pub(crate) fn set_dispute_sessions(mut self, dispute_sessions: Vec<u32>) -> Self {
```
drahnr, line 122: nit: `Vec<T>` as well as `[T]` and `&[T]` impl https://doc.rust-lang.org/std/vec/struct.Vec.html#impl-AsRef%3C%5BT%5D%3E , https://doc.rust-lang.org/std/primitive.slice.html#impl-AsRef%3C%5BT%5D%3E https://doc.rust-lang.org/std/primitive.array.html#impl-AsRef%3C%5BT%5D%3E and hence using `impl AsRef<[T]>` is preferable and avoids all the `vec![..]` clutter.

runtime/parachains/src/builder.rs
```diff
@@ -537,46 +594,51 @@ impl<T: paras_inherent::Config> BenchBuilder<T> {
...
+	///
+	/// * `set_backed_and_concluding_cores`: Set a map from core/para id seed to number of validity
+	/// votes.
+	/// * `set_dispute_sessions`: Set the session index of for each dispute. Index of slice
+	/// corresponds to a core, which is offset by the number of entries for
+	/// `backed_and_concluding_cores`. I.E. if `backed_and_concluding_cores` cores has 3 entries,
+	/// the first index of `dispute_sessions` will correspond to core index 3. There must be an
+	/// entry for each core with a dispute statement set.
+	/// *`set_code_upgrade`: Set to `Some` to include a code upgrade for all backed candidates.
+	/// The value within `Some` will be the byte length of the code.
```
drahnr, line 612: Shouldn't this documentation go on the individual builder `fn`s?

drahnr: Could you rephrase the documentation of `dispute_sessions`, it's not clear to my why it's named `dispute_sessions`.

drahnr (approved):
A few nits, mostly about docs and ergonomic API - other than that LGTM, it's going to be a big improvement!

runtime/parachains/src/builder.rs
emostov: I think this is a cool idea, my main concern though is that its there is so many corner cases to consider that we are bound to miss some and instead it would just be better for the user to be diligent

drahnr: Maybe issue-ify this and come back in the future.

rphmeier (approved):

runtime/parachains/src/builder.rs
emostov: https://github.com/paritytech/polkadot/issues/4412

runtime/parachains/src/builder.rs
emostov: I might be missing some basic understanding here, but using `impl AsRef<[T]>` here seems to make things more complicated because we eventually need the builder struct to own the type. If we have some pointer type instead of a Vec then we need to add lifetimes to the builder struct, which seems unnecessarily complicated. To make this function signature more flexible though I changed it to accept a slice in de60558

runtime/parachains/src/builder.rs
emostov: I think this is what you mean: 0dbe497

runtime/parachains/src/builder.rs
emostov: d81c612

runtime/parachains/src/builder.rs
drahnr: Priority for test code is, that it's low visual pain. If we can avoid `vec!` in a dozend places, it's worth the internal allocation of `as_ref().to_vec()`.

drahnr:
Resolved the last nits.

bot merge

drahnr:
bot merge

drahnr (commented):
bot merge

emostov:
bot merge

PR #4322, opened 2021-11-18
Use magic code instead of default code in benchbuilder

pepyakin:
This PR makes the validation code used for testing non-empty. This is important because as part of #3211 / #4209 we are considering adding invariant that paras cannot be registered with empty validation code.


pepyakin:
bad bot

drahnr (approved):

PR #4332, opened 2021-11-19
Add `fast-runtime` Cargo Feature for Quick Test Runs

apopiak:
This is meant as a first stab at making certain runtime configuration values configurable via feature flag.
This PR:
+ introduces the `fast-runtime` feature flag to toggle at build time whether the production or fast/testing values should be used for configuration values like the session time.
+ introduces the `prod_or_fast` macro to easily define config values taking advantage of the `fast-runtime` flag.
+ adjusts a lot of durations down to 2 minutes = 20 blocks, 1 minute = 10 blocks or 1 block (if the feature is selected).
+ also introduces environment variables that can be set at build time to specify custom periods (if the feature is selected).

### Usage
So to take advantage of the features introduced here you might call cargo like so:
```bash
KSM_EPOCH_DURATION=40 KSM_LAUNCH_PERIOD=3 cargo build --release --features=fast-runtime
```
or with cargo remote:
```bash
cargo remote -c release/polkadot -b 'KSM_LAUNCH_PERIOD=2; DOT_LAUNCH_PERIOD=2' -- build --release --features=fast-runtime
```
Kusama values are prefixed with `KSM`, Polkadot values with `DOT`.

### Motivation
This allows for easier local testing of the interaction between the production chains with parachains (e.g. Kusama and Statemine).

aims to make progress towards https://github.com/paritytech/polkadot/issues/4333


runtime/westend/Cargo.toml
```diff
@@ -248,3 +248,6 @@ try-runtime = [
 # runtime without clashing with the runtime API exported functions
 # in WASM.
 disable-runtime-api = []
+
+# Set timing constants (e.g. session period) to faster versions to speed up testing.
+fast-runtime = []
```
kianenigma, line 253:
```suggestion
fast-runtime = []

```

apopiak:
I tried parsing a const out of an env variable with `core::option_env!("ENV").map(|s| s.parse().ok()).flatten().unwrap_or(default)` but failed because `parse` is not `const`. Open to ideas on how we could something equivalent.

kianenigma:
> I tried parsing a const out of an env variable with
> 
> ```
> core::option_env!("ENV")
>   .map(|s| s.parse().ok())
>   .flatten()
>   .unwrap_or(default)
> ```
> 
> but failed because `parse` is not `const`. Open to ideas on how we could add something equivalent.

If you need const for `parameter types`, it does support non-const as well.

apopiak:
> If you need const for `parameter types`, it does support non-const as well.

Mmh, but would we be comfortable with making it non-const in production for the test usecase?

kianenigma:
> > If you need const for `parameter types`, it does support non-const as well.
> 
> 
> 
> Mmh, but would we be comfortable with making it non-const in production for the test usecase?

I am pretty comfortable with both, as long as there's no unreasonable shenanigans going on. Somethings are hard to make 'const', event though they _are_ constant in realisty. For example, see Type BlockWeights in system.

runtime/kusama/src/lib.rs
```diff
@@ -558,13 +558,33 @@ impl pallet_staking::Config for Runtime {
...
+	}
+}
+
 parameter_types! {
-	pub const LaunchPeriod: BlockNumber = 7 * DAYS;
-	pub const VotingPeriod: BlockNumber = 7 * DAYS;
-	pub const FastTrackVotingPeriod: BlockNumber = 3 * HOURS;
+	pub LaunchPeriod: BlockNumber = prod_or_test!(7 * DAYS, 1, "KSM_LAUNCH_PERIOD");
+	pub VotingPeriod: BlockNumber = prod_or_test!(7 * DAYS, 1 * MINUTES, "KSM_VOTING_PERIOD");
+	pub FastTrackVotingPeriod: BlockNumber = prod_or_test!(3 * HOURS, 1 * MINUTES, "KSM_FAST_TRACK_VOTING_PERIOD");
```
apopiak, line 584: What do you think of this?
@kianenigma @shawntabrizi 
Allows specifying parameter types via env variable in case the `fast-runtime` feature is activated.

kianenigma: looks good to me.

all in all, I think this is a good change. I deal with this personally as well at times and always have to manually tweak the parameters. That being said, I think we shouldn't over do it and only apply this feature to a small number of obvious ones. The reason would be that it is hard to choose a good number for "what is fast" for every parameter.

apopiak: That's why I figured the env variables would be good, because they allow for easy tweaking at build time.

runtime/common/src/lib.rs
```diff
@@ -182,6 +182,37 @@ impl<T: pallet_session::Config> OneSessionHandler<T::AccountId>
...
+///
+/// Usage:
+/// ```Rust
+/// parameter_types! {
+/// 	// Note that the env variable version parameter cannot be const.
+/// 	pub LaunchPeriod: BlockNumber = prod_or_test!(7 * DAYS, 1, "KSM_LAUNCH_PERIOD");
+/// 	pub const VotingPeriod: BlockNumber = prod_or_test!(7 * DAYS, 1 * MINUTES);
+/// }
+#[macro_export]
+macro_rules! prod_or_test {
```
shawntabrizi, line 203: would maybe call it `prod_or_fast`, seems more clear.

apopiak:
We might want to run `cargo check --features=fast-runtime` in CI to make sure the project builds I imagine?

shawntabrizi:
> We might want to run cargo check --features=fast-runtime in CI to make sure the project builds I imagine?

I would tend to say no because i dont want build times to be longer, but you can add it to the CI yourself in this PR if you want.

runtime/rococo/Cargo.toml
```diff
@@ -203,3 +203,6 @@ try-runtime = [
 	"runtime-common/try-runtime",
 	"pallet-multisig/try-runtime",
 ]
+
+# Set timing constants (e.g. session period) to faster versions to speed up testing.
+fast-runtime = []
```
shawntabrizi, line 208:
```suggestion
fast-runtime = []

```

shawntabrizi: configure your editor to add a trailing newline?

shawntabrizi (approved):

runtime/polkadot/src/lib.rs
```diff
@@ -119,7 +120,7 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
 	impl_name: create_runtime_str!("parity-polkadot"),
 	authoring_version: 0,
 	spec_version: 9130,
-	impl_version: 0,
+	impl_version: 1,
```
bkchr, line 123: No need to bump this or any other `impl_version`s.

runtime/common/src/lib.rs
```diff
@@ -189,6 +189,34 @@ impl pallet_staking::BenchmarkingConfig for StakingBenchmarkingConfig {
 	type MaxNominators = ConstU32<1000>;
 }
 
+/// Macro to set a value (e.g. when using the `parameter_types` macro) to either a production value
+/// or to an environment variable or testing value (in case the `fast-runtime` feature is selected).
```
bkchr, line 193: Maybe you should make more clear that the env variable needs to be set at compile time.

shawntabrizi:
I think another alternative implementation here is to just have the macro apply some constant reduction to all the constants than defining both.

So having `fast-runtime` divide any macro wrapped value by `1_000` or an environment variable amount.

Not at flexible, but probably just as useful with less code.

runtime/kusama/Cargo.toml
```diff
@@ -264,3 +264,6 @@ disable-runtime-api = []
 on-chain-release-build = [
 	"sp-api/disable-logging",
 ]
+
+# Set timing constants (e.g. session period) to faster versions to speed up testing.
+fast-runtime = []
```
kianenigma, line 269:
```suggestion
fast-runtime = []

```

runtime/kusama/src/lib.rs
```diff
@@ -368,8 +373,17 @@ impl pallet_session::historical::Config for Runtime {
 
 parameter_types! {
 	// phase durations. 1/4 of the last session for each.
-	pub const SignedPhase: u32 = EPOCH_DURATION_IN_SLOTS / 4;
-	pub const UnsignedPhase: u32 = EPOCH_DURATION_IN_SLOTS / 4;
+	// in testing: 1min or half of the session for each
+	pub SignedPhase: u32 = prod_or_fast!(
+		EPOCH_DURATION_IN_SLOTS / 4,
+		(1 * MINUTES).min(EpochDuration::get().saturated_into::<u32>() / 2),
```
kianenigma, line 379:
```suggestion
		(1 * MINUTES).max(EpochDuration::get().saturated_into::<u32>() / 2),
```

runtime/kusama/src/lib.rs
```diff
@@ -368,8 +373,17 @@ impl pallet_session::historical::Config for Runtime {
...
+		"KSM_SIGNED_PHASE"
+	);
+	pub UnsignedPhase: u32 = prod_or_fast!(
+		EPOCH_DURATION_IN_SLOTS / 4,
+		(1 * MINUTES).min(EpochDuration::get().saturated_into::<u32>() / 2),
```
kianenigma, line 384:
```suggestion
		(1 * MINUTES).max(EpochDuration::get().saturated_into::<u32>() / 2),
```

runtime/kusama/src/lib.rs
```diff
@@ -556,12 +570,12 @@ impl pallet_staking::Config for Runtime {
 }
 
 parameter_types! {
-	pub const LaunchPeriod: BlockNumber = 7 * DAYS;
-	pub const VotingPeriod: BlockNumber = 7 * DAYS;
-	pub const FastTrackVotingPeriod: BlockNumber = 3 * HOURS;
+	pub LaunchPeriod: BlockNumber = prod_or_fast!(7 * DAYS, 1, "KSM_LAUNCH_PERIOD");
```
kianenigma, line 573:
```suggestion
	pub LaunchPeriod: BlockNumber = prod_or_fast!(7 * DAYS, 1 * DAYS, "KSM_LAUNCH_PERIOD");
```

runtime/kusama/src/lib.rs
```diff
@@ -556,12 +570,12 @@ impl pallet_staking::Config for Runtime {
...
+	pub VotingPeriod: BlockNumber = prod_or_fast!(7 * DAYS, 1 * MINUTES, "KSM_VOTING_PERIOD");
+	pub FastTrackVotingPeriod: BlockNumber = prod_or_fast!(3 * HOURS, 1 * MINUTES, "KSM_FAST_TRACK_VOTING_PERIOD");
 	pub const MinimumDeposit: Balance = 100 * CENTS;
-	pub const EnactmentPeriod: BlockNumber = 8 * DAYS;
-	pub const CooloffPeriod: BlockNumber = 7 * DAYS;
+	pub EnactmentPeriod: BlockNumber = prod_or_fast!(8 * DAYS, 1, "KSM_ENACTMENT_PERIOD");
```
kianenigma, line 577: sure 1 is correct?

runtime/polkadot/Cargo.toml
```diff
@@ -251,3 +251,6 @@ disable-runtime-api = []
 on-chain-release-build = [
 	"sp-api/disable-logging",
 ]
+
+# Set timing constants (e.g. session period) to faster versions to speed up testing.
+fast-runtime = []
```
kianenigma, line 256:
```suggestion
fast-runtime = []

```

runtime/polkadot/src/lib.rs
```diff
@@ -407,8 +412,17 @@ impl pallet_session::historical::Config for Runtime {
 
 parameter_types! {
 	// phase durations. 1/4 of the last session for each.
-	pub const SignedPhase: u32 = EPOCH_DURATION_IN_SLOTS / 4;
-	pub const UnsignedPhase: u32 = EPOCH_DURATION_IN_SLOTS / 4;
+	// in testing: 1min or half of the session for each
+	pub SignedPhase: u32 = prod_or_fast!(
+		EPOCH_DURATION_IN_SLOTS / 4,
+		(1 * MINUTES).min(EpochDuration::get().saturated_into::<u32>() / 2),
```
kianenigma, line 418:
```suggestion
		(1 * MINUTES).max(EpochDuration::get().saturated_into::<u32>() / 2),
```

runtime/rococo/Cargo.toml
kianenigma, line 208:
```suggestion
fast-runtime = []

```

runtime/polkadot/src/lib.rs
```diff
@@ -1211,15 +1225,22 @@ impl paras_registrar::Config for Runtime {
 	type WeightInfo = weights::runtime_common_paras_registrar::WeightInfo<Runtime>;
 }
 
+pub const LEASE_PERIOD: BlockNumber = 12 * WEEKS;
+
+#[cfg(feature = "fast-runtime")]
```
kianenigma, line 1230: you can, in principle, also make this be generated by your macro to achieve a cleaner code.

kianenigma (approved):
LGTM

runtime/kusama/src/lib.rs
apopiak: The 1 is intentional. When testing (at least in my tests) you don't want a long queue of proposals. You want to vote as quickly as possible on the proposal you put up.

runtime/kusama/src/lib.rs
apopiak: Yes (at least for the testing scenarios I have in mind)

runtime/common/src/lib.rs
apopiak, line 193:
```suggestion
/// Macro to set a value (e.g. when using the `parameter_types` macro) to either a production value
/// or to an environment variable or testing value (in case the `fast-runtime` feature is selected).
/// Note that the environment variable is evaluated _at compile time_.
```

apopiak:
> I think another alternative implementation here is to just have the macro apply some constant reduction to all the constants than defining both.
> 
> So having `fast-runtime` divides any macro wrapped value by `1_000` or an environment variable amount.
> 
> Not at flexible, but probably just as useful with less code.

This would definitely only work for very few values. E.g. launch period can be reduced to 1 block without issues, but the election workers need enough time to submit their solutions (5 blocks (2min session time / 4) seems to work, didn't figure out the lower bound).
I much prefer this approach.

chevdor (approved):

crystalin:
How about disabling the UnexpectedEpochChange also with it ? It is quite common for test network to get stuck by it

gilescope:
I take your point but talking with a few people it sounds like there should be a separate feature that turns on settings that make things more robust in terms of max finality lag and max backoffs. This is purely timings, but features are designed to be additive so I think they would play well together.

gilescope:
bot merge

PR #4344, opened 2021-11-22
companion for #8783 - jsonrpsee

niklasad1:


jsdw (approved):

node/metrics/src/tests.rs
```diff
@@ -25,7 +25,7 @@ use std::collections::HashMap;
 
 const DEFAULT_PROMETHEUS_PORT: u16 = 9616;
 
-#[substrate_test_utils::test]
+#[substrate_test_utils::test(flavor = "multi_thread")]
```
jsdw, line 28: @niklasad1 is it something like this that maybe leads to the issue with port 9944 being hardcoded in the test code being hit?

niklasad1: most likely, I have pushed the latest stuff that hopefully fixes that.

dvdplm (approved):

PR #4352, opened 2021-11-23
Add Provisioner dispute metrics 

### LINKED ISSUE #4326: Metric on recent disputes
The number of recent disputes (+vote count in each of them) can become large in cases of spam. It would be good to know how many votes/disputes the block producer has to process on each `create_inherent` call.

sandreim:
Fixes #4326

skip check-dependent-cumulus


node/core/parachains-inherent/src/metrics.rs
```diff
@@ -0,0 +1,65 @@
...
+#[derive(Default, Clone)]
+pub struct Metrics(Option<MetricsInner>);
+
+impl Metrics {
+	pub fn inc_votes_by(&self, votes: usize) {
+		if let Some(metrics) = &self.0 {
+			metrics.votes.inc_by(votes.try_into().unwrap_or(0));
+		}
+	}
+	pub fn inc_disputes_by(&self, disputes: usize) {
```
drahnr, line 38: I would prefer to signal the reader that this is the number of disputes sets of the _inherent_. We should also include a count for the number of actual votes included.

drahnr: Could we add a `_inherent_` to the name to make this more explicit?

drahnr (approved):
A small nit about naming, otherwise LGTM

sandreim:
> A small nit about naming, it's a bit awkward to use the `subsystem-utils` outside of a subsystem - I would have imagined this to live in the `ProvisionerSubsystem` - otherwise LGTM

Agreed it is awkward, but this is due only to the Metrics helpers. Should we split it into `susbsytem-utils` and `utils`. I metrics are useful outside of susbsystems. Yeah, you are right, ProvisionerSubsystem` might be a better place but I preferred to increment the metrics as close asap before create_inherent does it's processing.

drahnr:
Pick one (or both) :) - I'd have a slight preference to move it to the `ProvisionerSubsystem`, but this is really up to you. Whatever you see fit.

emostov (approved):

node/core/provisioner/src/lib.rs
```diff
@@ -385,6 +388,10 @@ async fn collect_disputes(
 				.into_iter()
 				.map(|(s, i, sig)| (DisputeStatement::Invalid(s), i, sig));
 
+			metrics.inc_valid_statements_by(valid_statements.len());
```
rphmeier, line 391: It's worth noting that the approach of incrementing metrics at this point will count votes many times. Not necessarily a problem, but worth improving documentation about.

What I mean is that the`collect_disputes` function is supposed to return all 'recent' disputes, so that's absolutely every statement from every dispute from the last several sessions - a session being 10 minutes on Rococo, 1 hour on Kusama, and 4 hours on Polkadot. The runtime filters these.

So `collect_disputes` is called every time any node authors a block (I don't believe it should be called if the node isn't authoring a block, but maybe it will) - so it will depend on the number of validators. The metric is probably only useful from a network-wide viewpoint as opposed to a single-node viewpoint, or at least, the values will vary a lot from node to node based on how often they're selected to author blocks.

eskimor (approved):

Jaysonmaldx (approved):

node/core/provisioner/src/lib.rs
sandreim: Thanks for the insights, I'll add comments/docs.

sandreim:
bot merge

node/core/provisioner/src/metrics.rs
```diff
@@ -22,8 +22,11 @@ struct MetricsInner {
 	inherent_data_requests: prometheus::CounterVec<prometheus::U64>,
 	request_inherent_data: prometheus::Histogram,
 	provisionable_data: prometheus::Histogram,
+
+	/// The dispute_statement.* metrics trak how many disputes/votes the runtime will have to process. It will count
+	/// all recent statements meaning every dispute from last sessions: 10 min on Rococo, 60 min on Kusama and
+	/// 4 hours on Polkadot. The metrics are updated only when the node authors block, so values vary across nodes.
```
drahnr, line 28: `cargo spellcheck fix --checker=hunspell $(git diff --only-names $(git merge-base origin/master)..HEAD)` is a nice helper, (or even a good pre-commit hook when removing the `fix`)

PR #4357, opened 2021-11-24
don't initiate disputes for preparation timeout

rphmeier:


node/core/approval-voting/src/lib.rs
```diff
@@ -2310,17 +2310,25 @@ async fn launch_approval(
...
-							candidate_hash,
-							candidate.clone(),
-							false,
-						)
-						.into(),
-					)
-					.await;
+				match reason {
+					InvalidCandidate::ExecutionError(_) => {
+						// We don't raise disputes on execution errors that are likely
```
drahnr, line 2315: Should we print a `warn!`? This should rarely happen after all right?

pepyakin:
Subsumed by #4364

PR #4358, opened 2021-11-24
Fix use of weight limit errors

gavofyork:
The weight limit error `TooMuchWeightRequired` (used by `Transact` to indicate an invalid upper limit for the dispatchable call) was being incorrectly used in the place of `WeightLimitReached`.

This PR fixes that usage and renames the offending error into something more obvious.

Cumulus companion: https://github.com/paritytech/cumulus/pull/791


KiChjang (approved):
Will need to remember to update the pending XCMv3 PR as well.

PR #4360, opened 2021-11-24
Impose new restrictions on paras init and cleanup

pepyakin:
First PR of the #3211 series.
Related https://github.com/paritytech/polkadot/pull/4322

For upcoming PVF pre-checking feature we will need to impose a couple of
new restrictions for:

- `schedule_para_initialize`.
- `schedule_para_cleanup`.

Specifically, for the former we do not want to allow registration of
wasm blob that is empty, i.e. 0 bytes. While that currently already
does not make a lot of sense, it allows us to simplify the PVF
pre-checking logic: if this PR is deployed before the following changes
for PVF prechecking then we can be sure that no paras onboarding have to
have to go through the PVF pre-checking. In case, we deploy it
altogether this property will allow us to distingush paras that came in
before PVF pre-checking.

For `schedule_para_cleanup` we do not want to allow offboarding of paras
that are undergoing the upgrade process. While this is not a harsh
restriction this change allows us to avoid making the PVF prechecking
more complicated than it has to be.

### DIFF
--- a/runtime/common/src/paras_registrar.rs
+++ b/runtime/common/src/paras_registrar.rs
@@ -157,6 +157,8 @@ pub mod pallet {
 		ParaLocked,
 		/// The ID given for registration has not been reserved.
 		NotReserved,
+		/// Registering parachain with empty code is not allowed.
+		EmptyCode,
 	}
 
 	/// Pending swap operations.
@@ -547,6 +549,7 @@ impl<T: Config> Pallet<T> {
 		parachain: bool,
 	) -> Result<(ParaGenesisArgs, BalanceOf<T>), sp_runtime::DispatchError> {
 		let config = configuration::Pallet::<T>::config();
+		ensure!(validation_code.0.len() > 0, Error::<T>::EmptyCode);
 		ensure!(validation_code.0.len() <= config.max_code_size as usize, Error::<T>::CodeTooLarge);
 		ensure!(
 			genesis_head.0.len() <= config.max_head_data_size as usize,
--- a/runtime/parachains/src/paras.rs
+++ b/runtime/parachains/src/paras.rs
@@ -870,8 +870,10 @@ impl<T: Config> Pallet<T> {
 	pub(crate) fn schedule_para_initialize(id: ParaId, genesis: ParaGenesisArgs) -> DispatchResult {
 		let scheduled_session = Self::scheduled_session();
 
-		// Make sure parachain isn't already in our system.
+		// Make sure parachain isn't already in our system and that the onboarding parameters are
+		// valid.
 		ensure!(Self::can_schedule_para_initialize(&id, &genesis), Error::<T>::CannotOnboard);
+		ensure!(!genesis.validation_code.0.is_empty(), Error::<T>::CannotOnboard);
 
 		ParaLifecycles::<T>::insert(&id, ParaLifecycle::Onboarding);
 		UpcomingParasGenesis::<T>::insert(&id, genesis);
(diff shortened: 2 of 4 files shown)


runtime/parachains/src/paras.rs
```diff
@@ -1817,55 +1833,51 @@ mod tests {
...
-				assert_eq!(Paras::current_code(&para_id), Some(original_code.clone()));
-				check_code_is_stored(&original_code);
-				check_code_is_stored(&new_code);
+			// Enact the upgrade.
+			//
+			// For that run to block #7 and submit a new head.
+			assert_eq!(expected_at, 7);
+			run_to_block(7, None);
+			assert_eq!(<frame_system::Pallet<Test>>::block_number(), 7);
+			Paras::note_new_head(para_id, Default::default(), expected_at);
```
apopiak, line 1845:
```suggestion
			assert_ok!(Paras::note_new_head(para_id, Default::default(), expected_at));
```
assert ok?

runtime/common/src/paras_registrar.rs
```diff
@@ -547,6 +549,7 @@ impl<T: Config> Pallet<T> {
 		parachain: bool,
 	) -> Result<(ParaGenesisArgs, BalanceOf<T>), sp_runtime::DispatchError> {
 		let config = configuration::Pallet::<T>::config();
+		ensure!(validation_code.0.len() > 0, Error::<T>::EmptyCode);
```
apopiak, line 552: Should we add a test?

apopiak (commented):
looks fine to me, not sure I can judge the broader implications (because I'm not deeply familiar with the paras_registrar)

runtime/parachains/src/paras.rs
pepyakin: ah, no, `note_new_head` does not even return a result, it's `Weight`

pepyakin: I yanked that commit

runtime/parachains/src/paras.rs
```diff
@@ -886,10 +888,24 @@ impl<T: Config> Pallet<T> {
 
 	/// Schedule a para to be cleaned up at the start of the next session.
 	///
-	/// Will return error if para is not a stable parachain or parathread.
+	/// Will return error if either is true:
+	///
+	/// - para is not a stable parachain or parathread.
```
bkchr, line 893: What is a stable parachain?

bkchr (approved):

runtime/parachains/src/paras.rs
pepyakin: a stable parachain or parathread is a parachain which `ParaLifecycles::<T>::get(&para)` is either `Parachain` or `Parathread`, and not e.g. `Onboarding` or other transitory state.

Do you think this should be clarified?

bkchr: Ahh okay, never heard this before. 

`para is not an on-boarded parachain or parathread.`

Would maybe have been more clear to me. 

Or you could use something like
`[`ParaLifecycle::is_stable`] returns `false` for para`

This way we would directly link to the particular type and make it more clear?

pepyakin: Sure, good idea

pepyakin:
* master
  * **PR #4371** 
    * **PR #4372** 
      * **PR #4360**  (this PR)

shawntabrizi (approved):
Looks good to me.

Issue #4293, opened 2021-11-15
Investigate avoiding disputes in case of PVF preparation transient errors


pepyakin:
A validator that cannot validate a candidate due to a preparation error should strive to not cast an invalid vote, so that disputes are not initiated due to transient errors.

The caveat here is that we do not always know if an error is transient or it is a deliberate attack. However, after we land https://github.com/paritytech/polkadot/issues/3211 this will become much more easier: if the supermajority voted for a PVF then we can be reasonable sure that the is indeed transient.

burdges:
We'll just compile the PVFs an earlier though, no?  We could finalize that 2/3rd of validators compiled the PVF correctly even.  We'd then have overlapping 2/3rds that do not work out, but ignoring that likely works fine.

PR #4364, opened 2021-11-24
Treat non-deterministic prep errors as internal errors

pepyakin:
Closes https://github.com/paritytech/polkadot/issues/4293

This PR changes the way how we treat a certain subset of PVF preparation
errors. Specifically, now only the deterministic errors are treated as
invalid candidates. That is, the errors that are easily
attributable to either the the PVF contents or the wasmtime code, but
not e.g. I/O errors that could be triggered by the OS (insufficient
memory, disk failure, too much load, etc). The latter are treated as
internal errors and thus do not trigger the disputes.

This behavior will change eventually with #3211: we will treat all preparation errors as internal errors after pre-checking is online.

skip check-dependent-cumulus

### DIFF
--- a/node/core/pvf/src/error.rs
+++ b/node/core/pvf/src/error.rs
@@ -48,9 +48,9 @@ pub enum ValidationError {
 /// of the candidate [`polkadot_parachain::primitives::ValidationParams`] and the PVF.
 #[derive(Debug, Clone)]
 pub enum InvalidCandidate {
-	/// The failure is reported by the worker. The string contains the error message.
-	///
-	/// This also includes the errors reported by the preparation pipeline.
+	/// PVF preparation ended up with a deterministic error.
+	PrepareError(String),
+	/// The failure is reported by the execution worker. The string contains the error message.
 	WorkerReportedError(String),
 	/// The worker has died during validation of a candidate. That may fall in one of the following
 	/// categories, which we cannot distinguish programmatically:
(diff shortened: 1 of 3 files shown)


node/core/pvf/src/error.rs
```diff
@@ -78,13 +78,33 @@ pub enum InvalidCandidate {
...
+		// the sc-executor/wasmtime logic.
+		//
+		// For now, at least until the PVF pre-checking lands, the deterministic errors will be
+		// treated as `InvalidCandidate`. Should those occur they could potentially trigger disputes.
+		//
+		// All non-deterministic errors are qualified as `InternalError`s and will not trigger
+		// disputes.
+		match error {
+			PrepareError::Prevalidation(err) => ValidationError::InvalidCandidate(
+				InvalidCandidate::PrepareError(format!("prevalidation: {}", err)),
```
drahnr, line 97: Can't we include a `PrepareError` instead of stringifing?

drahnr (approved):
small nit, otherwise LGTM

node/core/pvf/src/error.rs
pepyakin: That idea crossed my mind too and I ditched this:

1. We do not need to know (yet) what exactly was this error. Ultimately the string is required for supplying into `ExecutionError`.
2. We actually do not want the full `PrepareError` in `InvalidCandidate::PrepareError` because the full `PrepareError` contains both "deterministic" and "undeterminstic" error, whereas `InvalidCandidate::PrepareError` promises to pass only deterministic variants. While that's not the end of the world I prefer to not include it, because I strive to limit the impossible states on the type level and that also IMO could confuse people thinking that all those errors are possible.

drahnr: I see, I have an idea how to solve this by only allowing the sane, sparse set to occur, but that's for a follow up PR :)

pepyakin:
* master
  * **PR #4364**  (this PR)

node/core/pvf/src/error.rs
```diff
@@ -78,13 +78,33 @@ pub enum InvalidCandidate {
...
+			),
+			PrepareError::Preparation(err) => ValidationError::InvalidCandidate(
+				InvalidCandidate::PrepareError(format!("preparation: {}", err)),
+			),
+			PrepareError::Panic(err) => ValidationError::InvalidCandidate(
+				InvalidCandidate::PrepareError(format!("panic: {}", err)),
+			),
+			PrepareError::TimedOut => ValidationError::InternalError("prepare: timeout".to_owned()),
+			PrepareError::DidNotMakeIt =>
+				ValidationError::InternalError("prepare: did not make it".to_owned()),
```
apopiak, line 107: If the previous error "communication error" was accurate I would add it, maybe as:
```suggestion
				ValidationError::InternalError("prepare: did not make it (communication error)".to_owned()),
```
to distinguish it more from the timeout (the distinction between "did not make it" and "timeout" is not obvious to me on first read.

pepyakin: "Communication error" does not cover every class of error that could potentially occur, which is why I chose such a vague error name.

Issue #4272, opened 2021-11-12
[zombinet] introduce helper crate to support coordinated / relative block malus impls


drahnr:
Problem:

For some test cases we require the possibility to coordinate between malicious actors. For this we need a "backchannel" to the test-runner needs to be implemented to provide key value based information as a bidirectional message stream that can be referenced.

Impl hints: 

For compatibility and ease of use, values should be `scale`-encoded. Every value must have a key. Once a key is sent to the test-runner, it should also be sent to all other malicious entities, so it's a star network with the test-runner in the center.

Impl details:

* Provide a `impl FusedStream` interface to consume the updates.
* Provide a `async send(key: &'static str, val: impl scale_codec::Encode)` to send the value to the runner
* Assure no context is required, but only standalone `fn`s, the context should be global/static.

Required for #4271

PR #4377, opened 2021-11-25
[zombinet] initial implementation of zombienet backchannel

pepoviola:
Hi @drahnr / @Lldenaurois, 
                                                 This is the initial implementation of the backchannel for `zombienet`, targeting https://github.com/paritytech/polkadot/issues/4272. 

The general idea is that there is a server side running in zombienet environment that has an `http` and a `ws` listeners,
  - The `http` interface accept  `post` request to store new items that should be available for use by the `test-runner`.
  - The `ws` interface `broadcasted` those new items to all connected malus actors.

*Notes*:
  - The two `env` vars are defaulted to the ones we need to use in our `CI`, since we can resolve `backchannel` to the correct endpoint.
  - The `connect` fn create the `ws` connection and return the `receiver` part in the `inner` field.
  - I only tested locally (with a small client that I create), so we need to complete the integration to the `malus` crate in order to test the complete flow locally or in the `CI`.
  - I'm not sure if the use of the `scale` encode if correct because in the zombienet side is stored as `string`.
  - A simple `test` mocking the `zombienet` server should be implemented after complete the integration.
  
Thanks!


node/zombienet-backchannel/src/lib.rs
```diff
@@ -0,0 +1,71 @@
...
+
+
+impl ZombienetBackchannel {
+    pub async fn connect() -> Result<ZombienetBackchannel,BackchannelError> {
+        let ws_url = format!("ws://{}:{}/ws", *BACKCHANNEL_HOST, *BACKCHANNEL_PORT);
+        log::debug!("Connecting to : {}",&ws_url);
+        let (ws_stream, _) = connect_async(ws_url).await.map_err(|_| BackchannelError::CantConnectToWS)?;
+        let (_write, read) = ws_stream.split();
+
+        Ok( ZombienetBackchannel {
```
drahnr, line 48: Idea: could we use a slighly different pattern where we use an `fn init` to setup the initial connection - assuming this is at the very beginning of the `malus` invocation. That then stores the `sink` and the `source` part of `ws` so we can use both? The `sink` is easily copyable, the `source` would need some additional thoughts. This would allow to consume messages in multiple places without many extra changes across the code base.

Even if this a bit further out, we should try to make the API compatible in that direction, i.e. move some logic from `lazy_static` to a global `fn initialize` . Happy to discuss this further.

node/zombienet-backchannel/src/lib.rs
```diff
@@ -0,0 +1,71 @@
...
+        })
+    }
+
+    pub async fn send(key: &'static str, val: impl scale_codec::Encode) -> Result<(),BackchannelError> {
+        let client = reqwest::Client::new();
+        let encoded = val.encode();
+        let body = String::from_utf8_lossy(&encoded);
+        let zombienet_endpoint = format!("http://{}:{}/{}", *BACKCHANNEL_HOST, *BACKCHANNEL_PORT, key);
+
+        log::debug!("Sending to : {}",&zombienet_endpoint);
```
drahnr, line 59: We prefer to use the `tracing` crate on the node side, for `target=` in malicious nodes we used the 😈 emoji, but feel free to pick any of the obviously non-bono ones

node/zombienet-backchannel/Cargo.toml
```diff
@@ -0,0 +1,22 @@
...
+description = "Zombienet backchannel to notify test runner and coordinate with malus actors."
+license = "GPL-3.0-only"
+version = "0.9.13"
+authors = ["Parity Technologies <admin@parity.io>"]
+edition = "2018"
+readme = "README.md"
+publish = false
+
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
```
drahnr, line 12: nit: remove default cargo comment

drahnr (changes requested):
See inline comments

node/zombienet-backchannel/src/lib.rs
pepoviola: Hi @drahnr, thanks for the feedback! I like the idea of use a `init` fn to initialize the `ws` connection and store both parts to use later. If we want to follow this path I think we can use a `broadcaster` to connect the `source` part of the `ws` and forward the message to multiple consumers.

Also, a more general question about how the `malus` will consume this information, Does the `malus` will need to `store` the received values to reads when is needed right? I was thinking in a `hashmap` as we store in the `zombienet` part and follow that direction another option is to pass to the `init` fn the hashmap behind an `RwLock` since most of the time the receiving part of the `ws` will `write` new values and the `malus` actor will `read` (since now all the messages received in the zombienet part are forwarded to all connected actors). 
**But** it's just and idea and I'm not sure if this is the direction we want follow.  Happy to discuss this further :)

Thanks!

pepoviola:
Hi @drahnr, I updated the implementation follow the pattern from our last talk. Now you have an `init` fn that connect you to the `ws` server on zombienet side and then two `fn`

- `subscribe` : give you a `receiver` to get the new item from the `ws`
- `send` : Send new items to the `ws`.

Let me know if you want to change/refactor something.

---

Examples of usage could be:

- receive part
```rust
    let backchannel = ZombienetBackchannel::init().await.expect("Error connecting ws");

    let mut rx = backchannel.subscribe();
    while let Ok(backchannel_item) = rx.recv().await {
        println!("{:?}", backchannel_item);
    }
```

- send
```rust
     let backchannel = ZombienetBackchannel::init().await.expect("Error connecting ws");
     backchannel.send(<key>, <value>)
```

---

Ping @Lldenaurois, let me know if you want to chat about test cases 0002/0003 for malus integration.

Thanks!!

node/zombienet-backchannel/src/lib.rs
```diff
@@ -0,0 +1,107 @@
...
+#[derive(Clone, Debug, Serialize, Deserialize)]
+pub struct BackchannelItem {
+	key: String,
+	value: String,
+}
+
+pub const ZOMBIENET: &str = "ZOMBIENET_BACKCHANNEL🧟🧟🧟";
+
+impl ZombienetBackchannel {
+	pub async fn init() -> Result<ZombienetBackchannel, BackchannelError> {
```
drahnr, line 49: I think this should be split into `fn init() -> Result<(), _> {}` and fn sender() -> Result<ZombienetBackchannel, BackchannelError>`

node/zombienet-backchannel/src/lib.rs
```diff
@@ -0,0 +1,107 @@
...
+	ws_tx: SplitSink<WebSocketStream<MaybeTlsStream<TcpStream>>, Message>,
+}
+
+#[derive(Clone, Debug, Serialize, Deserialize)]
+pub struct BackchannelItem {
+	key: String,
+	value: String,
+}
+
+pub const ZOMBIENET: &str = "ZOMBIENET_BACKCHANNEL🧟🧟🧟";
```
drahnr, line 46: Just `🧟🧟🧟` is fine :)

pepoviola: changed :)

node/zombienet-backchannel/src/lib.rs
pepoviola: Hi @drahnr, thanks for the feedback! Just to clarify:
 
- The idea is to call `init` at the start of the `malus`, and that will create the `ws` connection to the `zombienet` test-runner.

```rust
zombienet::init().await?
```

Then in other places we want to have access to either `send` or `subscribe` using the `broadcaster` associated fn and return an instance to allow us to interact with the underline `ws` like
```rust
zombienet::broadcaster().send(Foo(Something))
```
or
```rust
let receiver = zombienet::broadcaster().subscribe();
```

So broadcaster will need access to the initially created instance. I can implement this as a singleton if it's ok with you.

Thanks!

drahnr: Exactly :) Singelton is fine, it's a test helper only.

node/zombienet-backchannel/Cargo.toml
```diff
@@ -0,0 +1,22 @@
+[package]
+name = "zombienet-backchannel"
+description = "Zombienet backchannel to notify test runner and coordinate with malus actors."
+license = "GPL-3.0-only"
+version = "0.9.13"
+authors = ["Parity Technologies <admin@parity.io>"]
+edition = "2018"
```
drahnr, line 7:
```suggestion
edition = "2021"
```

node/zombienet-backchannel/src/lib.rs
```diff
@@ -0,0 +1,154 @@
...
+	key: String,
+	value: String,
+}
+
+pub struct Broadcaster;
+
+pub const ZOMBIENET: &str = "🧟ZOMBIENET🧟";
+
+impl Broadcaster {
+	pub fn subscribe(&self) -> Result<broadcast::Receiver<BackchannelItem>, BackchannelError> {
```
drahnr, line 57: Missing some docs here :)

node/zombienet-backchannel/src/lib.rs
```diff
@@ -0,0 +1,154 @@
...
+				}
+			});
+
+			*zombienet_bkc = Some(ZombienetBackchannel { broadcast_tx: tx, ws_tx: tx_relay });
+			return Ok(());
+		}
+
+		Err(BackchannelError::AlreadyInitialized)
+	}
+
```
drahnr, line 146: CC

drahnr (approved):
LGTM besides the lack of documentation :)

pepoviola:
Thanks for the feedback @drahnr, I added some `docs` to those methods :)
Thx!

drahnr:
Could you merge master, so we can merge this :)

pepoviola:
Hi @drahnr, merged :) but now the `smoke test` is failing for the same reason of https://github.com/paritytech/polkadot/pull/4519#issuecomment-1006666242 (the collator version).
Thanks!

drahnr:
Please merge master again

pepoviola:
> Please merge master again

Done 👍, thanks!

node/zombienet-backchannel/src/lib.rs
```diff
@@ -0,0 +1,155 @@
...
+pub struct Broadcaster;
+
+pub const ZOMBIENET: &str = "🧟ZOMBIENET🧟";
+
+impl Broadcaster {
+	/// Return a subscriber that will receive all message broadcasted by the zombienet backchannel
+	/// websocket server.
+	pub fn subscribe(&self) -> Result<broadcast::Receiver<BackchannelItem>, BackchannelError> {
+		let mut zombienet_bkc = ZOMBIENET_BACKCHANNEL.lock().unwrap();
+		let sender = zombienet_bkc.as_mut().unwrap().broadcast_tx.clone();
```
drahnr, line 58: `expect("Why this cannot error erver. qed")` would be appreciated.

node/zombienet-backchannel/src/lib.rs
```diff
@@ -0,0 +1,155 @@
...
+		if zombienet_bkc.is_none() {
+			return Err(BackchannelError::Uninitialized)
+		}
+		let encoded = val.encode();
+		let backchannel_item = BackchannelItem {
+			key: key.to_string(),
+			value: String::from_utf8_lossy(&encoded).to_string(),
+		};
+
+		let sender = zombienet_bkc.as_mut().unwrap().ws_tx.clone();
```
drahnr, line 82: CC

node/zombienet-backchannel/src/lib.rs
```diff
@@ -0,0 +1,155 @@
...
+}
+
+impl ZombienetBackchannel {
+	pub async fn init() -> Result<(), BackchannelError> {
+		let mut zombienet_bkc = ZOMBIENET_BACKCHANNEL.lock().unwrap();
+		if zombienet_bkc.is_none() {
+			let backchannel_host =
+				env::var("BACKCHANNEL_HOST").unwrap_or_else(|_| "backchannel".to_string());
+			let backchannel_port =
+				env::var("BACKCHANNEL_PORT").unwrap_or_else(|_| "3000".to_string());
```
drahnr, line 99: I'd recommend parsing the content of those first, before assuming their content is sane, and only replace the parsed values by defaults if they didn't error. This way errors are caught early on.

pepoviola: Sounds good, thanks for the feedback. I just pushed the validation. *Note* that the host inside a namespace could be something like `abc` that k8s will resolve, so I only check that the host `is not empty` string.

pepoviola:
Hi @drahnr, thanks for the feedback. I already made those changes 👍 
Thanks!

drahnr (approved):

PR #4389, opened 2021-11-27
rococo-runtime: Switch to latest `construct_runtime!` syntax

bkchr:
Besides that it fixes pallet macro errors in other crates that popped up
because of this switch.

### DIFF
--- a/runtime/rococo/src/validator_manager.rs
+++ b/runtime/rococo/src/validator_manager.rs
@@ -16,84 +16,87 @@
 
 //! A pallet for managing validators on Rococo.
 
-use frame_support::{decl_error, decl_event, decl_module, decl_storage, traits::EnsureOrigin};
 use sp_staking::SessionIndex;
 use sp_std::vec::Vec;
 
+pub use pallet::*;
+
 type Session<T> = pallet_session::Pallet<T>;
 
-/// Configuration for the parachain proposer.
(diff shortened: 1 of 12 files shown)


runtime/rococo/src/validator_manager.rs
```diff
@@ -16,84 +16,87 @@
...
 
-decl_error! {
-	pub enum Error for Module<T: Config> {}
-}
+		/// Privileged origin that can add or remove validators.
+		type PrivilegedOrigin: EnsureOrigin<<Self as frame_system::Config>::Origin>;
+	}
 
-decl_storage! {
-	trait Store for Module<T: Config> as ParachainProposer {
```
thiolliere, line 48: the pallet prefix is changed from ParachainProposer to ValidatorManager, I guess we should migrate the storage

bkchr: Yeah I know, but the storage items are just for storing temporary data. Aka, we don't really need to migrate them.

thiolliere: ok then all good to me

thiolliere (approved):

thiolliere:
I checked the metadata and there is no difference apart from:
* prefix changing for validator manager pallet, but this is known to be fine
* some call with no variant in beefy pallet, which is fine
* more information on storage on some pallets

PR #4401, opened 2021-11-29
add Encointer as TrustedTeleporter to Kusama

brenzi:
This PR prepares for the case where Encointer's proposal to be onboarded as a common good parachain will pass.

### DIFF
--- a/runtime/kusama/src/lib.rs
+++ b/runtime/kusama/src/lib.rs
@@ -1365,8 +1365,10 @@ pub type XcmRouter = (
 parameter_types! {
 	pub const Kusama: MultiAssetFilter = Wild(AllOf { fun: WildFungible, id: Concrete(KsmLocation::get()) });
 	pub const KusamaForStatemine: (MultiAssetFilter, MultiLocation) = (Kusama::get(), Parachain(1000).into());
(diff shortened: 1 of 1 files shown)


brenzi:
@joepetrowski May I request your review? It's copy-paste from our previous PR #4213 for testnets

joepetrowski (approved):
Looks OK to me. Question is whether we:

1. Merge now and put it in the runtime before Encointer would onboard. If the referendum to add Encointer as a common good chain failed, we could remove it from the runtime later.
2. Only merge after Encointer gets accepted by Kusama referendum as a common good chain.

Will let others chime in on the two options.

brenzi:
not sure what that CI error is about. Seems unrelated to this PR

rrtti:
I would suggest to wait until the chain is registered in order to merge this. The team has requested for the common good slot motion to kick off from lease period 18.

brenzi:
Now that Encointer has been ratified to become a common good parachain on Kusama, can we merge this for the next runtime upgrade?

joepetrowski (commented):
LGTM. @KiChjang ?

KiChjang (approved):
Double-checked the ParaId for Encointer just now; everything checks out.

PR #4407, opened 2021-11-29
pvf-precheck: Add `sign` in subsystem-util

pepyakin:
Right now, most of operations that sign stuff in polkadot protocol are
handled by a very convenient tool - `Signed`. However `Signed` assumes
that whatever is signed is anchored to some `parent_hash` which works
for most cases, but does not work for others.

One instance of such a case is pre-checking (#3211). There validators
submit signed votes on-chain. A vote is valid for the entire session. If
we were to use `Signed` we would have to root a vote in some block of
that session and during vote verification check that this block is
indeed within the session. This is especially annoying since we agreed
to use unsigned extrinsics to submit votes and we need to make the
unsigned extrinsic validation as slim as possible.

(FWIW, the definition of a pre-checking vote can be seen in the next
diff in the stack)

That's the reason why we opted-out from using `Signed` for pre-checking
and decided to go with the manual signing approach. Almost every piece
of machinery is in place except for signing which is presented in this
PR.

### DIFF
--- a/node/subsystem-util/src/lib.rs
+++ b/node/subsystem-util/src/lib.rs
@@ -53,7 +53,7 @@ use polkadot_primitives::v1::{
 	AuthorityDiscoveryId, CandidateEvent, CommittedCandidateReceipt, CoreState, EncodeAs,
 	GroupIndex, GroupRotationInfo, Hash, Id as ParaId, OccupiedCoreAssumption,
 	PersistedValidationData, SessionIndex, SessionInfo, Signed, SigningContext, ValidationCode,
-	ValidationCodeHash, ValidatorId, ValidatorIndex,
+	ValidationCodeHash, ValidatorId, ValidatorIndex, ValidatorSignature,
 };
 use sp_application_crypto::AppKey;
 use sp_core::{traits::SpawnNamed, Public};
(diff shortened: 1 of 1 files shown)


pepyakin:
* master
  * **PR #4407**  (this PR)
    * **PR #4406** 

bkchr (approved):

node/subsystem-util/src/lib.rs
```diff
@@ -237,6 +237,27 @@ pub async fn signing_key_and_index(
...
+	data: &[u8],
+) -> Result<Option<ValidatorSignature>, KeystoreError> {
+	use std::convert::TryInto;
+
+	let signature =
+		CryptoStore::sign_with(&**keystore, ValidatorId::ID, &key.into(), &data).await?;
+
+	match signature {
+		Some(sig) =>
+			Ok(Some(sig.try_into().map_err(|_| KeystoreError::KeyNotSupported(ValidatorId::ID))?)),
```
drahnr, line 256: Dropping the conversion error is interesting, we are dropping information here which might be valuable to diagnose what's the error.

drahnr (approved):

node/subsystem-util/src/lib.rs
pepyakin: I should've mentioned that I just repeat the picture I found there.

https://github.com/paritytech/polkadot/blob/8952974037e854fe4de52c211802f23cb9aa0ae3/primitives/src/v1/signed.rs#L238

Lmk, if this is something you want to change

bkchr: The error is `()` anyway, so not really valuable.

PR #4406, opened 2021-11-29
pvf-precheck: Add `PvfCheckStatement` to polkadot-primitives

pepyakin:
This is an insubstantial PR that just unlocks PRs down the line. This PR
is a part of #3211.

Regarding the `PvfCheckStatement` struct itself: this is a structure
that will be used to convert from/into the binary representation and
ultimately will be used to sign and submit votes onto chain.

PR #4408, opened 2021-11-29
pvf-precheck: Strip PastCodeMeta

pepyakin:
This PR is a part of
https://github.com/paritytech/polkadot/issues/3211.

This PR prepares ground for the following runtime changes required for
PVF pre-checking. Specifically, we do several changes here:

1. We remove `validation_code_at` and `validation_code_hash_at`. Those
   functions are not used. They were added in the early days with intent
   to use it later but turned out that we do not need them.
2. We replace `validation_code_hash_at` with just `current_code_hash`
   for the case of inclusion and candidate checking.
3. We also replace `last_code_upgrade` with a direct query into
   `FutureCodeHash` and `UpgradeRestrictionSignal`. Those in conjunction
   should replace the logic that was used for allowing/disallowing
   upgrades. This requires special attention of the reviewers.
4. Then we remove the machinery required to support those queries.
   Specifically the code related to `UseCodeAt`. We do not need it since
   we do not answer the historical queries. However, we still leave all
   the data on-chain. At some point we may clean it up, but that would
   be needed to be done with a dedicated migration which can be done as
   follow-up.
5. Some now irrelevant tests were removed and/or adapted.

### DIFF
--- a/roadmap/implementers-guide/src/runtime/paras.md
+++ b/roadmap/implementers-guide/src/runtime/paras.md
@@ -217,13 +217,6 @@ CodeByHash: map ValidationCodeHash => Option<ValidationCode>
 * `note_new_head(ParaId, HeadData, BlockNumber)`: note that a para has progressed to a new head,
   where the new head was executed in the context of a relay-chain block with given number. This will
   apply pending code upgrades based on the block number provided. If an upgrade took place it will clear the `UpgradeGoAheadSignal`.
-* `validation_code_at(ParaId, at: BlockNumber, assume_intermediate: Option<BlockNumber>)`: Fetches
-  the validation code to be used when validating a block in the context of the given relay-chain
-  height. A second block number parameter may be used to tell the lookup to proceed as if an
-  intermediate parablock has been included at the given relay-chain height. This may return past,
-  current, or (with certain choices of `assume_intermediate`) future code. `assume_intermediate`, if
-  provided, must be before `at`. If the validation code has been pruned, this will return `None`.
-* `validation_code_hash_at(ParaId, at: BlockNumber, assume_intermediate: Option<BlockNumber>)`: Just like `validation_code_at`, but returns the code hash.
 * `lifecycle(ParaId) -> Option<ParaLifecycle>`: Return the `ParaLifecycle` of a para.
 * `is_parachain(ParaId) -> bool`: Returns true if the para ID references any live parachain,
   including those which may be transitioning to a parathread in the future.
@@ -231,9 +224,7 @@ CodeByHash: map ValidationCodeHash => Option<ValidationCode>
   including those which may be transitioning to a parachain in the future.
 * `is_valid_para(ParaId) -> bool`: Returns true if the para ID references either a live parathread
   or live parachain.
-* `last_code_upgrade(id: ParaId, include_future: bool) -> Option<BlockNumber>`: The block number of
-  the last scheduled upgrade of the requested para. Includes future upgrades if the flag is set.
-  This is the `expected_at` number, not the `activated_at` number.
+* `can_upgrade_validation_code(ParaId) -> bool`: Returns true if the given para can signal code upgrade right now.
 
 ## Finalization
 
--- a/runtime/parachains/src/paras.rs
+++ b/runtime/parachains/src/paras.rs
@@ -76,16 +76,6 @@ pub struct ParaPastCodeMeta<N> {
 	last_pruned: Option<N>,
 }
 
-#[cfg_attr(test, derive(Debug, PartialEq))]
-enum UseCodeAt<N> {
-	/// Use the current code.
-	Current,
-	/// Use the code that was replaced at the given block number.
-	/// This is an inclusive endpoint - a parablock in the context of a relay-chain block on this fork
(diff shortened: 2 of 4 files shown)


pepyakin:
* master
  * **PR #4407** 
    * **PR #4406** 
      * **PR #4409** 
        * **PR #4408**  (this PR)

roadmap/implementers-guide/src/runtime/paras.md
```diff
@@ -217,13 +217,6 @@ CodeByHash: map ValidationCodeHash => Option<ValidationCode>
 * `note_new_head(ParaId, HeadData, BlockNumber)`: note that a para has progressed to a new head,
   where the new head was executed in the context of a relay-chain block with given number. This will
   apply pending code upgrades based on the block number provided. If an upgrade took place it will clear the `UpgradeGoAheadSignal`.
-* `validation_code_at(ParaId, at: BlockNumber, assume_intermediate: Option<BlockNumber>)`: Fetches
-  the validation code to be used when validating a block in the context of the given relay-chain
-  height. A second block number parameter may be used to tell the lookup to proceed as if an
-  intermediate parablock has been included at the given relay-chain height. This may return past,
-  current, or (with certain choices of `assume_intermediate`) future code. `assume_intermediate`, if
-  provided, must be before `at`. If the validation code has been pruned, this will return `None`.
-* `validation_code_hash_at(ParaId, at: BlockNumber, assume_intermediate: Option<BlockNumber>)`: Just like `validation_code_at`, but returns the code hash.
```
pepyakin, line 226: We also need to remove `last_code_upgrade` from here and replace it with the new `can_upgrade_validation_code`

runtime/parachains/src/paras.rs
```diff
@@ -1123,16 +1019,10 @@ impl<T: Config> Pallet<T> {
...
-			if let Some(at) = Self::future_code_upgrade_at(id) {
-				return Some(at)
-			}
-		}
-
-		Self::past_code_meta(&id).most_recent_change()
+	/// If a candidate from the specified parachain were submitted at the current block, this
+	/// function returns if that candidate passes the acceptance criteria.
+	pub(crate) fn can_upgrade_validation_code(id: ParaId) -> bool {
+		FutureCodeHash::<T>::get(&id).is_none() && UpgradeRestrictionSignal::<T>::get(&id).is_none()
```
pepyakin, line 1025: This predicate seems to be equivalent to the previous condition, i.e.:

```rust
let valid_upgrade_attempt = <paras::Pallet<T>>::last_code_upgrade(para_id, true)
				.map_or(true, |last| {
					last <= self.relay_parent_number &&
						self.relay_parent_number.saturating_sub(last) >=
							self.config.validation_upgrade_frequency
				});
```

But needs double-checking

rphmeier:
> We remove validation_code_at and validation_code_hash_at. Those functions are not used. They were added in the early days with intent to use it later but turned out that we do not need them.

https://github.com/paritytech/polkadot/blob/0043fe5721047f99277206cb00e61877dde61ef0/runtime/parachains/src/inclusion.rs#L969

We might want to keep it so this line can be altered as a part of asynchronous backing. Otherwise code changes would result in service interruptions for a few blocks. Worth noting.

runtime/parachains/src/inclusion.rs
```diff
@@ -962,7 +958,7 @@ impl<T: Config> CandidateCheckContext<T> {
 			Error::<T>::NotCollatorSigned,
 		);
 
-		let validation_code_hash = <paras::Pallet<T>>::validation_code_hash_at(para_id, now, None)
+		let validation_code_hash = <paras::Pallet<T>>::current_code_hash(para_id)
```
rphmeier, line 961: i.e. this was previously future-proof changes to add asynchronous backing (old relay parent where code hadn't yet changed) but is not future-proof anymore

rphmeier: Probably fine, from a product perspective - if we build parachains 1 block in advance then a code upgrade block would lead to a 6 second service interruption. Really not a problem.

roadmap/implementers-guide/src/runtime/paras.md
rphmeier: why don't we?

pepyakin: Ah, just a reminder for myself

pepyakin:
@rphmeier when I wrote this code that thought crossed my mind. I had to chase it away because I felt it was overthinking: it's not easy to predict the **precise** API that will actually be required by async execution and at the same time there is no benefit for that either: we can easily add/remove parameters and shuffle the code around when the time comes (and when we a better grasp of the actual needs). Then this change also does not alter any on-chain storage, so we can actually restore any code from git history and it will work. Although I had planned to remove the storage eventually as well.

I am also a bit confused since I was under impression that the context should be passed through a relay-chain state merkle proof. That's why here I thought it is acceptable to remove the historical data from the current state.

rphmeier:
Yeah, it's fine as is. The latest version of contextual execution might not need merkle proofs

pepyakin:
bot merge

pepyakin:
(merged assuming this was an implicit approval and due to that this is a small change)

PR #4410, opened 2021-11-29
Fix currently-checking-cache

rphmeier:
We were spawning validation work in approval-voting for every `(candidate, relay-parent)` combination. Since candidate validation is the same regardless of relay-parent of inclusion, the intent of this code was to spawn work only once per candidate. I have fixed the code to do exactly that.


node/core/approval-voting/src/lib.rs
```diff
@@ -529,21 +529,28 @@ impl CurrentlyCheckingSet {
...
-						validator_index,
-						approval_outcome: ApprovalOutcome::TimedOut,
-					},
-					Some(approval_state) => approval_state,
+		match self.candidate_hash_map.entry(candidate_hash) {
+			Entry::Occupied(mut entry) => {
+				// validation already undergoing. just add the relay hash if unknown.
+				if !entry.get().contains(&relay_block) {
+					entry.get_mut().push(relay_block);
 				}
```
bkchr, line 537: If we would directly use a `HashSet`, we would only require a `insert`.

node/core/approval-voting/src/lib.rs
```diff
@@ -529,21 +529,28 @@ impl CurrentlyCheckingSet {
...
-			}));
+			},
+			Entry::Vacant(mut entry) => {
+				// validation not ongoing. launch work and time out the remote handle.
+				let _ = entry.insert(vec![relay_block]);
```
bkchr, line 541:
```suggestion
				entry.insert(vec![relay_block]);
```

Is not required.

bkchr (approved):
A test would be nice, but otherwise looks good.

rphmeier:
@Lldenaurois can you add tests?

drahnr (approved):

node/core/approval-voting/src/tests.rs
```diff
@@ -2259,10 +2272,91 @@ fn subsystem_process_wakeup_trigger_assignment_launch_approval() {
...
+		}
+	);
+
+	recover_available_data(virtual_overseer).await;
+	fetch_validation_code(virtual_overseer).await;
+
+	let first_message = virtual_overseer.recv().await;
+	let second_message = virtual_overseer.recv().await;
+
+	for msg in vec![first_message, second_message].into_iter() {
```
slumber, line 2305: Does it mean the test doesn't care about the order of these two? Otherwise two explicit asserts are better.

Lldenaurois: There is no way to guarantee the ordering as the two blocks are imported asynchronously, so we cannot ensure that one completes before the other starts up.

node/core/approval-voting/src/tests.rs
```diff
@@ -2259,10 +2272,91 @@ fn subsystem_process_wakeup_trigger_assignment_launch_approval() {
...
+			},
+			AllMessages::CandidateValidation(
+				CandidateValidationMessage::ValidateFromExhaustive(_, _, _, _, timeout, tx),
+			) if timeout == APPROVAL_EXECUTION_TIMEOUT => {
+				tx.send(Ok(ValidationResult::Valid(Default::default(), Default::default())))
+					.unwrap();
+			},
+			_ => panic! {},
+		}
+	}
```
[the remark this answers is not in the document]
Lldenaurois: I added a timeout receive at the bottom of the test after the invocation of this function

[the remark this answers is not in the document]
Lldenaurois: That's a good point... Maybe best to include the check in the function rather than below the invocation.

slumber (approved):

PR #4418, opened 2021-11-30
Add doc about runtime version bump

chevdor:
Some doc to avoid releasing a **new** runtime using the same `core_version` as another previously (pre)released one.

skip check-dependent-cumulus


bkchr (approved):

thiolliere (approved):

bkchr:
bot merge

PR #4409, opened 2021-11-29
pvf-precheck: Candidate Validation Changes

pepyakin:
This PR is a part of #3211.

As part of this PR we are integrating the new validation host function, 
that is `precheck_pvf`, into the candidate-validation subsystem, which will 
be the main entrypoint for pre-checking subsystem (To be introduced by 
PRs down the stack). 

skip check-dependent-cumulus

Issue #4248, opened 2021-11-09
Go-ahead signal makes validation upgrade delay of less than 2 invalid


rphmeier:
Changes here are to update the `HostConfiguration` documentation, the setters, and any places where `HostConfiguration` is defined for testnets.

Related to #4246. 

With go-ahead signals, code upgrades are meant to be applied on the first parablock included with a relay-parent >= the first relay chain block where the go-ahead signal was set.

The go-ahead signal is incompatible with delay '0' because that implies instant upgrades. In the old system, delay '0' meant that the code upgrade was applied as soon as the parablock announcing it was included. With go-ahead signals, this is the earliest point where the go-ahead can be set, so the go-ahead logic is incompatible with the old system.

The go-ahead signal is incompatible with delay '1' because a block built on top of relay-parent X would be included in relay-parent X+2, at which point a block with relay-parent X+1 could be backed. The earliest time a go-ahead signal can be set is at relay block X+2, so the relay chain would expect the block with relay-parent X+1 to set the code even though it hadn't seen the go-ahead signal.

Actually, if we extend to contextual execution (#3779), any `validation_upgrade_delay` less than or equal to some parameter `max_relay_parent_age` is disallowed. We can think of the current system as having `max_relay_parent_age = 1`.

PR #4420, opened 2021-11-30
pvf-precheck: update configuration module

### LINKED ISSUE #4010: Add pre-checking into the configuration runtime module

pepyakin:
This PR is a part of #3211.

This PR adds three new fields into the `HostConfiguration` structure.
The fields are going to be used in PRs down the stack.

This change requires migration, so this PR performs runtime storage
migration for configuration module from version 1 to version 2. Migration was performed in very similar manner to https://github.com/paritytech/polkadot/pull/3575.

This PR closes #4010 and subsumes #4177.

There are several TODOs:

- [x] Fill in the PR number for the reference in migration code
- [x] Decide what weight use for `set_pvf_checking_enabled`. It seems
  that u32 will be close enough, but maybe we still want to add weight
  for boolean parameters (prolly a followup)
- [x] Test the upgrade (tested it manually by running a local polkadot-launch node and slapping an upgrade)

skip check-dependent-cumulus

### DIFF
--- a/runtime/parachains/src/configuration.rs
+++ b/runtime/parachains/src/configuration.rs
@@ -177,6 +177,31 @@ pub struct HostConfiguration<BlockNumber> {
 	/// The maximum amount of weight any individual upward message may consume. Messages above this
 	/// weight go into the overweight queue and may only be serviced explicitly.
 	pub ump_max_individual_weight: Weight,
+	/// This flag controls whether PVF pre-checking is enabled.
+	///
+	/// If the flag is false, the behavior should be exactly the same as prior. Specifically, the
+	/// upgrade procedure is time-based and parachains that do not look at the go-ahead signal
+	/// should still work.
+	pub pvf_checking_enabled: bool,
+	/// If an active PVF pre-checking vote observes this many number of sessions it gets automatically
+	/// rejected.
+	///
+	/// 0 means PVF pre-checking will be rejected on the first observed session unless the voting
+	/// gained supermajority before that the session change.
+	pub pvf_voting_ttl: SessionIndex,
+	/// The lower bound number of blocks an upgrade can be scheduled.
+	///
+	/// Typically, upgrade gets scheduled [`validation_upgrade_delay`] relay-chain blocks after
+	/// the relay-parent of the parablock that signalled the validation code upgrade. However,
+	/// in the case a pre-checking voting was concluded in a longer duration the upgrade will be
+	/// scheduled to the next block.
+	///
+	/// That can disrupt parachain inclusion. Specifically, it will make the blocks that were
+	/// already backed invalid.
+	///
+	/// To prevent that, we introduce the minimum number of blocks after which the upgrade can be
+	/// scheduled. This number is controlled by this field.
+	pub minimum_validation_upgrade_delay: BlockNumber,
 }
 
 impl<BlockNumber: Default + From<u32>> Default for HostConfiguration<BlockNumber> {
@@ -222,6 +247,9 @@ impl<BlockNumber: Default + From<u32>> Default for HostConfiguration<BlockNumber
 			hrmp_max_parathread_outbound_channels: Default::default(),
 			hrmp_max_message_num_per_candidate: Default::default(),
 			ump_max_individual_weight: 20 * WEIGHT_PER_MILLIS,
+			pvf_checking_enabled: false,
+			pvf_voting_ttl: 2u32.into(),
+			minimum_validation_upgrade_delay: 0.into(),
 		}
 	}
 }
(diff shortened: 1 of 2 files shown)


pepyakin:
* master
  * **PR #4409** 
    * **PR #4408** 
      * **PR #4420**  (this PR)

runtime/parachains/src/configuration.rs
```diff
@@ -947,6 +975,49 @@ pub mod pallet {
 			});
 			Ok(())
 		}
+
+		/// Enable or disable PVF pre-checking. Consult the field documentation prior executing.
+		#[pallet::weight((
+			T::WeightInfo::set_config_with_u32(), // TODO: should we change that to bool?
```
KiChjang, line 981: This should be fine, the code for setting a u32 vs setting a boolean is pretty much identical, aside from the fact that they're different data types... perhaps we could rename the weighing function into something else to make it more clear? I just don't really have a good name, uhh maybe `set_config_with_primitive_type`?

KiChjang (approved):
Feel like we should have something similar to database migration files a la Ruby on Rails, removing old migration code sounds kind of wrong to me but can understand why it makes sense in the blockchain world. Anyway, it's a larger discussion and shouldn't block this PR from landing.

runtime/parachains/src/configuration.rs
```diff
@@ -222,6 +247,9 @@ impl<BlockNumber: Default + From<u32>> Default for HostConfiguration<BlockNumber
 			hrmp_max_parathread_outbound_channels: Default::default(),
 			hrmp_max_message_num_per_candidate: Default::default(),
 			ump_max_individual_weight: 20 * WEIGHT_PER_MILLIS,
+			pvf_checking_enabled: false,
+			pvf_voting_ttl: 2u32.into(),
+			minimum_validation_upgrade_delay: 0.into(),
```
rphmeier, line 252: https://github.com/paritytech/polkadot/issues/4248

Affected by this?

runtime/parachains/src/configuration.rs
pepyakin: Hm, looking at the `trait WeightInfo` I see that there are several items that could have the name `set_config_with_primitive_type`. E.g. I don't think that there is a really big difference between `with_option_u32` and `with_u32`. I do not see a `weights.rs` file, it would be interesting to see the actual numbers there.

Right now I just removed this todo and added a comment that it should be fine. I think this could be addressed in a follow-up. We can continue the discussion here though.

runtime/parachains/src/configuration.rs
pepyakin: I think it may be related, but not completely: the paras changes (to be submitted in the next diff) will actually behave the same way at it does now with this configuration defaults.

runtime/parachains/src/configuration/migration.rs
```diff
@@ -16,13 +16,304 @@
...
+no_show_slots                            : pre.no_show_slots,
+n_delay_tranches                         : pre.n_delay_tranches,
+zeroth_delay_tranche_width               : pre.zeroth_delay_tranche_width,
+needed_approvals                         : pre.needed_approvals,
+relay_vrf_modulo_samples                 : pre.relay_vrf_modulo_samples,
+ump_max_individual_weight                : pre.ump_max_individual_weight,
+
+pvf_checking_enabled: false,
+pvf_voting_ttl: 2u32.into(),
+minimum_validation_upgrade_delay: pre.chain_availability_period + 10u32.into(),
```
apopiak, line 192: Is this (`minimum_validation_upgrade_delay > chain_availability_period` an invariant that the `set_minimum_validation_upgrade_delay` dispatchable should check?

apopiak (commented):
LGTM
:+1: for the migration tests

runtime/parachains/src/configuration/migration.rs
pepyakin: yeah, that's an invariant. Checking it in the setters makes sense from a glance. Thinking about this, we also need to consider cases when we do need to override the value skipping the invariant protection, it seems. Anyway, that's out of scope for this change, esp. considering that none of the existing setters check the invariants.

pepyakin:
bot merge

PR #4427, opened 2021-12-01
Update bridges subtree r/w

antonio-dropulic:
In preparation for [Rococo-Wococo runtime integration](https://github.com/paritytech/parity-bridges-common/issues/1213).


svyatonik (approved):
Thanks a lot! One last thing - please also note [my last mistake](https://github.com/paritytech/polkadot/pull/3369#issuecomment-868511905) - iirc you should use "Rebase and merge", not "Squash and merge".

tomusdrw (approved):
Also note this only updates the bridge code and does not integrate it with runtimes just yet (that's going to be a follow up PR), so there is not much to review here.

PR #4452, opened 2021-12-02
Restore Rococo <> Wococo bridges (headers + messages)

svyatonik:
May be tested with scripts in https://github.com/svyatonik/rococo-wococo.test and updated relayer from https://github.com/paritytech/parity-bridges-common/pull/1241 . This version:
1) doesn't include any message filters;
2) needs message relayers to be running in altruistic mode, because R/W tokens has no value;
3) requires relayers fund account creation on R/W once this goes live;
4) since we have been running R<>W headers bridge for a while and I believe Rococo hasn't been restarted since that, we don't need to (re)initialize headers bridge. Instead we shall run standalone headers relay for a while && let it sync missing headers.


runtime/rococo/src/lib.rs
```diff
@@ -256,18 +256,18 @@ construct_runtime! {
 		// It might seem strange that we add both sides of the bridge to the same runtime. We do this because this
 		// runtime as shared by both the Rococo and Wococo chains. When running as Rococo we only use
 		// `BridgeWococoGrandpa`, and vice versa.
-		// BridgeRococoGrandpa: pallet_bridge_grandpa::{Pallet, Call, Storage, Config<T>} = 40,
-		// BridgeWococoGrandpa: pallet_bridge_grandpa::<Instance1>::{Pallet, Call, Storage, Config<T>} = 41,
+		BridgeRococoGrandpa: pallet_bridge_grandpa::{Pallet, Call, Storage, Config<T>} = 40,
+		BridgeWococoGrandpa: pallet_bridge_grandpa::<Instance1>::{Pallet, Call, Storage, Config<T>} = 41,
 
 		// Validator Manager pallet.
 		ValidatorManager: validator_manager,
```
tomusdrw, line 263: `ValidatorManager` index is going to change after this goes live - can we move it right after "MmrLeaf" but before Grandpa pallets to prevent that?

tomusdrw (approved):
> since we have been running R<>W headers bridge for a while and I believe Rococo hasn't been restarted since that, we don't need to (re)initialize headers bridge. Instead we shall run standalone headers relay for a while && let it sync missing headers.

Actually it was restarted and the header relayers are not running since then. We need to initialize it from scratch and I think it's best to run the complex relayer right from scratch.

svyatonik:
bot merge

PR #4453, opened 2021-12-02
Enable dispute coordinator by default

eskimor:
@drahnr already has another PR up which enables disputes conditionally on the target chain. Once we have disputes enabled unconditionally, this PR should get merged as well.


rphmeier:
We should not enable for Polkadot, yet. Only Kusama. Once the disputes module is on Polkadot we can enable this for Kusama. So we need to replace the compile-time flag with a runtime check.

drahnr:
Obsolete since #5184 is closed and master already contains the changes.

PR #4457, opened 2021-12-03
pvf-precheck: Integrate PVF pre-checking into paras module

### LINKED ISSUE #4009: Implement pre-checking changes in the paras module
For the rough (and probably slightly dated) description refer to the [design document](https://hackmd.io/@Vq3VW1JpRhOHoVQBe2V6ww/Sk7GjYlVF)

pepyakin:
Closes #4009

This is the most of the runtime-side change needed for #3211.

Here is how it works.

The PVF pre-checking can be triggered either by an upgrade or by
onboarding (i.e. calling `schedule_para_initialize`). The PVF
pre-checking process is identified by the PVF code hash that is being
voted on. If there is already PVF pre-checking process running, then no
new PVF pre-checking process will be started. Instead, we just subscribe
to the existing one.

If there is no PVF pre-checking process running but the PVF code hash
was already saved in the storage, that necessarily means (I invite the
reviewers to double-check this invariant) that the PVF already passed
pre-checking. This is equivalent to instant approving of the PVF.

The pre-checking process can be concluded either by obtaining a
supermajority or if it expires.

Each validator checks the list of PVFs available for voting. The vote is
binary, i.e. accept or reject a given PVF. As soon as the supermajority
of votes are collected for one of the sides of the vote, the voting is
concluded in that direction and the effects of the voting are enacted.

Only validators from the active set can participate in the vote. The set
of active validators can change each session. That's why we reset the
votes each session. A voting that observed a certain number of sessions
will be rejected.

The effects of the PVF accepting depend on the operations requested it:

1. All onboardings subscribed to the approved PVF pre-checking process will
get scheduled and after passing 2 session boundaries they will be onboarded.
2. All upgrades subscribed to the approved PVF pre-checking process will
get scheduled very similarly to the existing process. Upgrades with
pre-checking are really the same process that is just delayed by the
time required for pre-checking voting. In case of instant approval the
mechanism is exactly the same. This is important from parachains
compatibility standpoint since following the delayed upgrade requires
the parachain to implement
https://github.com/paritytech/cumulus/pull/517.

In case, PVF pre-checking process was concluded with rejection, then all
the requesting operations get cancelled. For onboarding it means it gets
without movement: the lifecycle of such parachain is terminated on the
`Onboarding` state and after rejection the lifecycle is none. That in
turn means that the caller can attempt registering the parachain once
more. For upgrading it means that the upgrade process is aborted: that
flashes go-ahead signal with `Abort` flag.

Rejection leads to removing the allegedly bad validation code from the
chain storage. Among other things, this implies that the operation can
be re-requested. That allows for retrying an operation in case there was
some bug. At the same time it does not look as a DoS vector due to the
caching performed by the nodes.

PVF pre-checking can be enabled and disabled. Initially, according to
the changes in #4420, this mechanism is disabled. Triggering the PVF
pre-checking when it is disabled just means that we insta approve the
requesting operation. This should lead to the behavior being unchanged.

Follow-ups:

- expose runtime APIs


pepyakin:
Oh sh~, TIL: "votings" is not a thing.

runtime/parachains/src/paras.rs
```diff
@@ -524,9 +665,174 @@ pub mod pallet {
...
+			let signing_payload = stmt.signing_payload();
+			ensure!(
+				signature.verify(&signing_payload[..], &validator_public),
+				Error::<T>::PvfCheckInvalidSignature,
+			);
+
+			let mut active_vote = PvfActiveVoteMap::<T>::get(&stmt.subject)
+				.ok_or(Error::<T>::PvfCheckSubjectInvalid)?;
+
+			ensure!(
```
drahnr, line 699: :+1: a comment about the importance of this one preventing votes in both directions, would be nice.

runtime/parachains/src/paras.rs
```diff
@@ -524,9 +665,174 @@ pub mod pallet {
...
+				.get(validator_index)
+				.ok_or(Error::<T>::PvfCheckValidatorIndexOutOfBounds)?;
+
+			let signing_payload = stmt.signing_payload();
+			ensure!(
+				signature.verify(&signing_payload[..], &validator_public),
+				Error::<T>::PvfCheckInvalidSignature,
+			);
+
+			let mut active_vote = PvfActiveVoteMap::<T>::get(&stmt.subject)
```
drahnr, line 696: nit: Is there a particular reason we don't `try_mutate` this in place? It feels a bit error prone having the vote tracking the vote at the very end in an `else` case.

runtime/parachains/src/paras.rs
```diff
@@ -524,9 +665,174 @@ pub mod pallet {
...
+				// The supermajority quorum has been achieved.
+				//
+				// Remove the PVF vote from the active map and finalize the PVF checking according
+				// to the outcome.
+				PvfActiveVoteMap::<T>::remove(&stmt.subject);
+				PvfActiveVoteList::<T>::mutate(|l| {
+					if let Ok(i) = l.binary_search(&stmt.subject) {
+						l.remove(i);
+					}
+				});
```
drahnr, line 723: Reason to track it twice in a `Map` and `List` is entirely based on the perf of iterating over a `StorageMap`, correct?

runtime/parachains/src/paras.rs
```diff
@@ -892,55 +1422,156 @@ impl<T: Config> Pallet<T> {
...
-					let insert_idx = upgrade_cooldowns
-						.binary_search_by_key(&next_possible_upgrade_at, |&(_, b)| b)
-						.unwrap_or_else(|idx| idx);
-					upgrade_cooldowns.insert(insert_idx, (id, next_possible_upgrade_at));
-				});
+		// para signals an update to the same code? This does not make a lot of sense, so abort the
+		// process right away.
+		weight += T::DbWeight::get().reads(1);
+		if CurrentCodeHash::<T>::get(&id) == Some(code_hash) {
+			UpgradeGoAheadSignal::<T>::insert(&id, UpgradeGoAhead::Abort);
```
drahnr, line 1469: nit: Adding a log message here would be appreciated.

runtime/parachains/src/paras.rs
```diff
@@ -781,32 +1095,248 @@ impl<T: Config> Pallet<T> {
...
+		//   introduce some unnecessary burden.
+		//
+		// So instead of going through it right now, we will do something sneaky. Specifically:
+		//
+		// - Insert the `CurrentCodeHash` now, instead during the onboarding. That would allow to
+		//   get rid of hashing of the validation code when onboarding.
+		//
+		// - Replace `validation_code` with a sentinel value: an empty vector. This should be fine
+		//   as long we do not allow registering parachains with empty code. At the moment of writing
+		//   this should already be the case.
```
drahnr, line 1321: idea: Could we use a different sentinel value? Does a wasm (and compressed wasm) block have a fixed prefix? We could exploit this to make it explicit.

drahnr (commented):
Generally looks good, I have to review a few sections more carefully:

* [x]  the "hack" 
* [ ] ref counting consistency

All my comments are just nits, code looks pretty nice (minus the existence of "hack")

drahnr:
Uh, also: could you include the ascii graph in the comment :)

pepyakin:
ah, right, I can add it in the next PR in the stack: https://github.com/paritytech/polkadot/pull/4459

runtime/parachains/src/paras.rs
```diff
@@ -203,6 +207,96 @@ pub struct ParaGenesisArgs {
...
+		Some(*accept_vote || *reject_vote)
+	}
+
+	/// Returns `None` if the quorum is not reached, or the direction of the decision.
+	fn quorum(&self, n_validators: usize) -> Option<PvfCheckOutcome> {
+		let q_threshold = primitives::v1::supermajority_threshold(n_validators);
+		if self.votes_accept.count_ones() >= q_threshold {
+			Some(PvfCheckOutcome::Accepted)
+		} else if self.votes_reject.count_ones() >= q_threshold {
+			Some(PvfCheckOutcome::Rejected)
```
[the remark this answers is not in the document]
pepyakin: You mean switch those checks so that we are more defensive here?

runtime/parachains/src/paras.rs
pepyakin: Yeah. I used the same approach when I was in a similar situation for messages. Arguably, there are more messages than PVF votes at the same time. However, I also think that it is simpler this way: it is easier to reason about the `Map` and `List` combination, IMO, than `iter_prefix`. Although I admit that it is a bit more verbose.

runtime/parachains/src/paras.rs
pepyakin: Good call

runtime/parachains/src/paras.rs
pepyakin: I think we could, but TBH I think empty vector is a better fit here (although marginally).

However, that all does not matter since ideally we get rid of that hack by untangling the `ParaGenesis` argument from the storage. I actually did do that but realized that it would require a non-trivial migration and the diff is already hefty.

runtime/parachains/src/paras.rs
pepyakin: will be fixed in a follow-up of a follow-up

runtime/parachains/src/paras.rs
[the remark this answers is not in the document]
drahnr: Exactly

runtime/parachains/src/paras.rs
drahnr: Just making sure :)

runtime/parachains/src/configuration.rs
```diff
@@ -73,6 +73,9 @@ pub struct HostConfiguration<BlockNumber> {
 	/// This parameter affects the upper bound of size of `CandidateCommitments`.
 	pub hrmp_max_message_num_per_candidate: u32,
 	/// The minimum period, in blocks, between which parachains can update their validation code.
+	///
+	/// If PVF pre-checking is enabled this should be greater than the maximum number of blocks
+	/// PVF pre-checking can take.
```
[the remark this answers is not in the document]
pepyakin: I meant to be vague that the pre-checking should finish first and only then a new upgrade will be allowed. It just so happens that the pre-checking timeout is typically takes the most time. However, it is also possible that the last decisive vote comes in on the last block of the session and thus ends the PVF voting, so it is not always timeout.

runtime/parachains/src/paras.rs
pepyakin: Well, I tried to come up with a comment and couldn't. That does not seem to be that important, since everyone sees each other's vote on-chain. Perhaps it just allows for wasting resources.

runtime/parachains/src/paras.rs
pepyakin: yeah it was also bothering me, but I think in this case cure is the same or worse than the disease. 

In the case of achieving the quorum, we should remove the vote from the list and the map. Probably we can switch the entire storage value to be `OptionQuery`, but then we will have to deal with this everywhere as well. Then, I am trying to put mutations of the map closer to the list, but in this case the mutation will happen inside of `try_mutate` but the list mutation will be left alone.

We could perhaps refactor the code so that we return an enum, but that's additional code only to make sure we do not accidentally miss this update, so it feels to me not worth it

runtime/parachains/src/configuration.rs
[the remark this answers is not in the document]
drahnr: Exhaustiveness is imho better than trying to avoid it. Helps reviewing the code and on-boarding new people.

drahnr (approved):

drahnr:
The change-set looks good to me :+1:  Ref-counting logic and usage looks correct. The Hack does not show any logic flaws, the logic matches the comments.

Calling sequence issues that could be caused by `force_*` operations were explicitly _not_ checked.

pepyakin:
Hey, @KiChjang, could you review the benchmarking part of this PR? It seems that you took an initial shot at paras benchmarks.

KiChjang (approved):
Benchmarks LGTM

KiChjang:
Actually, we may also want to update the benchmarks on rococo/westend/any chain that contains the paras pallet.

emostov:

runtime/parachains/src/paras.rs
```diff
@@ -902,55 +1435,163 @@ impl<T: Config> Pallet<T> {
...
+		UpgradeRestrictionSignal::<T>::insert(&id, UpgradeRestriction::Present);
+
+		weight += T::DbWeight::get().reads_writes(1, 1);
+		let next_possible_upgrade_at = relay_parent_number + cfg.validation_upgrade_frequency;
+		<Self as Store>::UpgradeCooldowns::mutate(|upgrade_cooldowns| {
+			let insert_idx = upgrade_cooldowns
+				.binary_search_by_key(&next_possible_upgrade_at, |&(_, b)| b)
+				.unwrap_or_else(|idx| idx);
+			upgrade_cooldowns.insert(insert_idx, (id, next_possible_upgrade_at));
+		});
```
bkchr, line 1505: Why not set this when the upgrade was accepted? Otherwise, if the values are choosen "unlucky", it could happen that we finish an upgrade and the next one is directly doable?

runtime/parachains/src/paras.rs
```diff
@@ -997,10 +1638,44 @@ impl<T: Config> Pallet<T> {
 				T::DbWeight::get().reads_writes(1, 1 + 0)
 			}
 		} else {
+			// This means there is no upgrade scheduled.
+			//
+			// In case the upgrade was aborted by the relay-chain we should reset
+			// the `Abort` signal.
+			UpgradeGoAheadSignal::<T>::remove(&id);
 			T::DbWeight::get().reads_writes(1, 1)
```
bkchr, line 1646:
```suggestion
			T::DbWeight::get().reads_writes(1, 2)
```

bkchr (approved):
Nice pr

runtime/parachains/src/paras.rs
```diff
@@ -791,32 +1107,249 @@ impl<T: Config> Pallet<T> {
...
+	fn enact_pvf_rejected(
+		code_hash: &ValidationCodeHash,
+		causes: Vec<PvfCheckCause<T::BlockNumber>>,
+	) -> Weight {
+		let mut weight = T::DbWeight::get().writes(1);
+
+		for cause in causes {
+			// Whenever PVF pre-checking is started or a new cause is added to it, the RC is bumped.
+			// Now we need to unbump it.
+			weight += Self::decrease_code_ref(code_hash);
```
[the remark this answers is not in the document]
pepyakin: Yes. 

The idea is that paras retains the code whenever it encounters that code. That would be the code added by either onboarding or upgrading. Each of these processes will bump the RC, since they will each be initiated by a different para. If PVF pre-checking failed though we need to get rid of the code, hence we decrease the RC here.

Also there is an invariant: we assume that after iterating over all causes the RC should reach 0. Why? Because when we just started PVF pre-checking, we checked that there is no such code [^1] (If there were, then the pre-checking wouldn't have started and the code was approved immediately), and thus the pre-checking which rejection we are handling here is the sole owner of the `code_hash`

[^1]: The condition is actually more complex: it's no such code or pvf pre-checking disabled. If it is disabled though we would not reject the PVF.

runtime/parachains/src/paras.rs
```diff
@@ -534,9 +676,175 @@ pub mod pallet {
...
+			};
+
+			match active_vote.has_vote(stmt.validator_index.0 as usize) {
+				Some(false) => (),
+				Some(true) => return Err(InvalidTransaction::Custom(INVALID_TX_DOUBLE_VOTE).into()),
+				None => return Err(InvalidTransaction::Custom(INVALID_TX_BAD_VALIDATOR_IDX).into()),
+			}
+
+			Ok(())
+		}
```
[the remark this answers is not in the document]
bkchr: > Note that here we do not check if the signature is correct also

Yes I have seen this, otherwise you could directly have used the default impl of `pre_dispatch` that just calls `validate`.

runtime/parachains/src/paras.rs
pepyakin: 1. In this PR, I mention [this](https://github.com/paritytech/polkadot/pull/4457/files#diff-9d731023bd72eb1b3b196e18a4c373d586ab6379f01a2029799cafed907fc8d5R78) for the documentation. We should enforce the invariant on this in #4248 or similar.
2. It is not possible as of currently, since if the upgrade was not consumed by the para (i.e. it did not produce any new heads, see `note_new_head`) then `can_upgrade_validation_code` will return `false`. This means that such a candidate will be rejected by backing and by the on-chain logic in inclusion module.
3. If we do want to address this, I think it should go in a follow-up.

pepyakin: So I ended up not fixing it, see #4603.

PR #4460, opened 2021-12-03
Update docs on `validation_upgrade_frequency`

joepetrowski:
Frequency is the inverse of time, so this variable is actually the _maximum_ frequency of or _minimum_ time period between upgrades.

Would consider renaming the variable to `validation_upgrade_period`.


pepyakin:
Agreed. FWIW, I was planning to rename it to `validation_upgrade_cooldown`.

pepyakin (approved):

Issue #4139, opened 2021-10-24
SetAppendix xcm message cannot pass Barrier check


GopherJ:
```
2021-10-24 14:39:54.007 TRACE tokio-runtime-worker xcm::should_execute: did not pass barrier:

origin: MultiLocation { parents: 0, interior: X1(Parachain(2085)) }
message: Xcm([
  SetAppendix(Xcm([ReportError { query_id: 1, dest: MultiLocation { parents: 0, interior: X1(Parachain(2085)) }, max_response_weight: 10000 }])),
  WithdrawAsset(MultiAssets([MultiAsset { id: Concrete(MultiLocation { parents: 0, interior: Here }), fun: Fungible(50000000000) }])),
  BuyExecution { fees: MultiAsset { id: Concrete(MultiLocation { parents: 0, interior: Here }), fun: Fungible(50000000000) }, weight_limit: Unlimited },
  Transact { origin_type: SovereignAccount, require_weight_at_most: 3000000000, call: [24, 2, 8, 4, 3, 0, 93, 25, 155, 83, 85, 8, 153, 12, 89, 244, 17, 117, 118, 23, 144, 76, 230, 92, 144, 95, 206, 214, 135, 139, 172, 251, 242, 109, 59, 74, 30, 151, 15, 96, 199, 42, 57, 94, 214, 8, 24, 1, 0, 0, 6, 0, 0, 93, 25, 155, 83, 85, 8, 153, 12, 89, 244, 17, 117, 118, 23, 144, 76, 230, 92, 144, 95, 206, 214, 135, 139, 172, 251, 242, 109, 59, 74, 30, 151, 15, 96, 199, 42, 57, 94, 214, 8, 0] }, RefundSurplus,
  DepositAsset { assets: Definite(MultiAssets([MultiAsset { id: Concrete(MultiLocation { parents: 0, interior: Here }), fun: Fungible(50000000000) }])), max_assets: 1, beneficiary: MultiLocation { parents: 0, interior: X1(AccountId32 { network: Any, id: [112, 97, 114, 97, 37, 8, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0] }) } }
]),
max_weight: 10000000000,
weight_credit: 0
```

GopherJ:
maybe we can have a barrier `AllowTopLevelAppendix`

[...]

PR #4467, opened 2021-12-05
allow SetAppendix(Xcm(vec])) xcm message to pass barrier check

GopherJ:
Signed-off-by: Cheng JIANG <alex_cj96@foxmail.com>

With more and more usage of xcm.Transact, we really need the error reporting feature. However it's not possible with current kusama runtime because this kind of message cannot pass barrier check. This really cause troubles for parachain teams to test and use this feature

close: #4139

### DIFF
--- a/xcm/xcm-builder/src/barriers.rs
+++ b/xcm/xcm-builder/src/barriers.rs
@@ -65,7 +65,13 @@ impl<T: Contains<MultiLocation>> ShouldExecute for AllowTopLevelPaidExecutionFro
 		);
 		ensure!(T::contains(origin), ());
 		let mut iter = message.0.iter_mut();
-		let i = iter.next().ok_or(())?;
+		let mut i = iter.next().ok_or(())?;
+		match i {
+			SetAppendix(Xcm(ref instrs)) if matches!(&instrs[..], &[ReportError { .. }]) => {
+				i = iter.next().ok_or(())?;
+			},
+			_ => (),
+		}
 		match i {
 			ReceiveTeleportedAsset(..) |
 			WithdrawAsset(..) |


GopherJ:
Hi @KiChjang  we kind need this feature to be on live kusama, could you guide me on this?

KiChjang:
Sorry for not replying back, I'll have a look and see whether this makes sense.

xcm/xcm-builder/src/barriers.rs
```diff
@@ -65,7 +65,13 @@ impl<T: Contains<MultiLocation>> ShouldExecute for AllowTopLevelPaidExecutionFro
 		);
 		ensure!(T::contains(origin), ());
 		let mut iter = message.0.iter_mut();
-		let i = iter.next().ok_or(())?;
+		let mut i = iter.next().ok_or(())?;
+		match i {
+			SetAppendix(Xcm(ref instrs)) if matches!(&instrs[..], &[ReportError { .. }]) => {
```
KiChjang, line 70: This here is basically ignoring the `SetAppendix` instruction if it appears before an asset transfer instruction. The one thing that I am concerned about is whether the `SetAppendix` instruction is properly paid for, or whether it's even necessary to pay for it. I will likely need to defer to @gavofyork for this change.

KiChjang: Also, it might also be good to include `SetErrorHandler` as well.

gavofyork: Er... no. This is super dangerous. You would just do a `SetAppendix(<huge amount of code>)` followed by a `BuyExecution` which fails. You'd then get all the huge amount of code executed for free.

KiChjang:
@GopherJ Now that I think about it, what's the reason why you cannot append a `SetAppendix` or `SetErrorHandler` right after `BuyExecution`?

GopherJ:
> @GopherJ Now that I think about it, what's the reason why you cannot append a `SetAppendix` or `SetErrorHandler` right after `BuyExecution`?

I'm using pallet xcm's built in report_outcome_notify so it's inserted st the beginning

KiChjang:
Okay, what happens if you create an empty extrinsic and append your XCMs after `SetAppendix`?

KiChjang:
Actually no, the expectation of this API is that your `message` parameter should contain an XCM that has a list of instructions of what to do once the holding register is already filled with assets, e.g. `[ExchangeAsset, WithdrawAsset]`. And so what happens is you call `report_outcome_notify` with this XCM, which prepends a `SetAppendix` to it, and then afterwards you continue to prepend a `BuyExecution`. I suspect that the `WithdrawAsset`/`ReserveAssetDeposited`/`ReceiveTeleportedAsset`/`ClaimAsset` instructions get prepended by the XCM executor (along with `ClearOrigin`), so you don't need to manually prepend those.

xcm/xcm-builder/src/barriers.rs
GopherJ: `matches!(&instrs[..], &[ReportError { .. }])` actually only allows one instruction which is `ReportError`

GopherJ:
> 

This will be disallowed by `AllowTopLevelPaidExecutionFrom` because it's not in the form:

```
ReceiveTeleportedAsset | WithdrawAsset | ReserveAssetDeposited | ClaimAsset
? ClearOrigin
BuyExecution
```

GopherJ:
> Actually no, the expectation of this API is that your `message` parameter should contain an XCM that has a list of instructions of what to do once the holding register is already filled with assets, e.g. `[ExchangeAsset, DepositAsset]`. And so what happens is you call `report_outcome_notify` with this XCM, which prepends a `SetAppendix` to it, and then afterwards you continue to prepend a `BuyExecution`. I suspect that the `WithdrawAsset`/`ReserveAssetDeposited`/`ReceiveTeleportedAsset`/`ClaimAsset` instructions get prepended by the XCM executor (along with `ClearOrigin`), so you don't need to manually prepend those.

Ok got it so currently I can only construct the xcm message manually

GopherJ:
I can test with this but wondering if the following form is as expected:

```
WithdrawAsset
BuyExecution
SetAppendix(ReportError)
WithdrawAsset
BuyExecution
Transact
RefundPlus
DepositAsset
```

GopherJ:
Or maybe this type will work:

```
WithdrawAsset
BuyExecution
SetAppendix(ReportError)
Transact
RefundPlus
DepositAsset
```

gavofyork:
> Or maybe this type will work:
> 
> ```
> WithdrawAsset
> BuyExecution
> SetAppendix(ReportError)
> Transact
> RefundPlus
> DepositAsset
> ```

Yes, this appears sensible.

gavofyork:
Closing as the solution is not clear and doesn't require alterations to this repo.

PR #4470, opened 2021-12-06
Reanchor should return canonical location

gavofyork:
skip-cumulus-check

TODO:
- [x] `fn reanchored(mut assets: Assets, dest: &MultiLocation) -> Result<MultiAssets, XcmError>` should either fail if an asset cannot  be reanchored or possibly return a pair with `(reanchored, unreanchorable)`, with the latter being placed back into `Holding` by the instruction.


xcm/src/v1/multilocation.rs
```diff
@@ -323,6 +323,60 @@ impl MultiLocation {
...
+
+		// 3. Given that we know some of `target` ancestry, ensure that any parents in `self` are
+		// strictly needed.
+		self.simplify(target.interior());
+
+		Ok(())
+	}
+
+	/// Treating `self` as a context, determine how it would be referenced by a `target` location.
+	pub fn inverted(&self, target: &MultiLocation) -> Result<MultiLocation, ()> {
```
KiChjang, line 349: Wonder if we should call the `target` anchor instead? It's from which the new `MultiLocation` is based off on.

xcm/src/v1/multilocation.rs
```diff
@@ -323,6 +323,60 @@ impl MultiLocation {
...
+		use Junction::OnlyChild;
+		let mut ancestry = self.clone();
+		let mut junctions = Junctions::Here;
+		for _ in 0..target.parent_count() {
+			junctions = junctions
+				.pushed_with(ancestry.interior.take_last().unwrap_or(OnlyChild))
```
KiChjang, line 355: I'm not sure about this one, it's taking the last junction from `ancestry` and pushing it to `junctions`?

So let's assume we're working with a smart contract and it's trying to see how a parachain will reference it -- we'll have `Parachain(1050)/PalletIndex(60)/GeneralIndex(1)` as `ancestry` and `../../../Parachain(2042)` as `target`.

This loop here will loop 3 times, and on the each iteration, it'll take the last junction from `ancestry` and append to `junctions`. When the loop finishes, we'd get `GeneralIndex(1)/PalletIndex(60)/Parachain(1050)` as a result for `junctions`, which doesn't sound right to me, aren't we supposed to keep the order of the junctions intact?

I think this is what's supposed to happen:
```suggestion
				.pushed_front_with(ancestry.interior.take_last().unwrap_or(OnlyChild))
```

KiChjang, line 355: I'm not so sure about this one, we're taking the last junction of `ancestry` and pushing into the end of `junctions`?

Suppose we have a smart contract with an `ancestry` of `Parachain(1050)/PalletIndex(60)/GeneralIndex(1)` and we want to figure out how the `target` referenced by `../../../Parachain(2042)` addresses us. This loop here will iterate 3 times, each time taking the _last_ junction of `ancestry` and pushing it to the _end_ of `junctions`, so at the end of the loop, `junctions` will contain `GeneralIndex(1)/PalletIndex(60)/Parachain(1050)`, which is in a reversed order.

Instead, what I think should happen is that the _last_ junction of `ancestry` gets pushed into the _beginning_ of `junctions`, so the code should look like:

```suggestion
				.pushed_front_with(ancestry.interior.take_last().unwrap_or(OnlyChild))
```

With that, we'd get `Parachain(1050)/PalletIndex(60)/GeneralIndex(1)` as expected.

xcm/src/v1/multilocation.rs
```diff
@@ -323,6 +323,60 @@ impl MultiLocation {
 		}
 		Ok(())
 	}
+
+	/// Mutate `self` so that it represents the same location from the point of view of `target`.
+	/// The context of `self` is provided as `ancestry`.
+	///
+	/// Does not modify `self` in case of overflow.
+	pub fn reanchor(&mut self, target: &MultiLocation, ancestry: &MultiLocation) -> Result<(), ()> {
+		// TODO: Optimize this.
```
KiChjang, line 332: So I realized that what we're essentially doing here is to first find the common ancestor between `self` and `target`, then append the necessary junctions that will allow us to reach `self` from the common ancestor, and then we'd simplify the result, given `ancestry`.

I suspect what we can do here is to write the common ancestor from the perspective of `target`, that way we don't have to simplify at the end of the algorithm.

xcm/src/v1/multilocation.rs
KiChjang: Failing test:

```rust
	#[test]
	fn inverted_does_not_work() {
		let ancestry: MultiLocation = (Parachain(1050), PalletInstance(60), GeneralIndex(1)).into();
		let target = (Parent, Parent, Parent, Parachain(2042)).into();
		let expected = (Parent, Parachain(1050), PalletInstance(60), GeneralIndex(1)).into();
		let inverted = ancestry.inverted(&target).unwrap();
		assert_eq!(inverted, expected);
	}
```

xcm/xcm-executor/src/lib.rs
```diff
@@ -498,9 +497,7 @@ impl<Config: config::Config> XcmExecutor<Config> {
 	}
 
 	fn reanchored(mut assets: Assets, dest: &MultiLocation) -> Result<MultiAssets, XcmError> {
```
KiChjang, line 499: Since `reanchor` is infallible now, there's no need to return a `Result` anymore.

xcm/src/v1/multilocation.rs
gavofyork: Write a (failing) test for it?

xcm/src/v1/multilocation.rs
gavofyork: Right. Hence the TODO :)

xcm/xcm-executor/src/assets.rs
```diff
@@ -203,7 +203,34 @@ impl Assets {
...
+				let _ = id.reanchor(target, ancestry);
+				(id, amount)
+			})
+			.collect();
+		let mut non_fungible = Default::default();
+		mem::swap(&mut self.non_fungible, &mut non_fungible);
+		self.non_fungible = non_fungible
+			.into_iter()
+			.map(|(mut class, inst)| {
+				let _ = class.reanchor(target, ancestry);
```
gavofyork, line 233: Should make this return any assets which cannot be reanchored.

xcm/xcm-executor/src/lib.rs
gavofyork: Yeah, we should replace assets which cannot be reanchored back into Holding to avoid crazy behaviour.

xcm/src/v1/multilocation.rs
```diff
@@ -773,9 +827,76 @@ impl TryFrom<MultiLocation> for Junctions {
...
+	use crate::opaque::v1::{Junction::*, NetworkId::*};
 	use parity_scale_codec::{Decode, Encode};
 
+	#[test]
+	fn inverted_works() {
+		let ancestry: MultiLocation = (Parachain(1000), PalletInstance(42)).into();
+		let target = (Parent, PalletInstance(69)).into();
+		let expected = (Parent, PalletInstance(42)).into();
+		let inverted = ancestry.inverted(&target).unwrap();
+		assert_eq!(inverted, expected);
```
gavofyork, line 839:
```suggestion
		let ancestry: MultiLocation = (Parachain(1000), PalletInstance(42)).into();
		let target = (Parent, PalletInstance(69)).into();
		let expected = (Parent, PalletInstance(42)).into();
		let inverted = ancestry.inverted(&target).unwrap();
		assert_eq!(inverted, expected);
		
		let ancestry: MultiLocation = (Parachain(1000), PalletInstance(42), GeneralIndex(1)).into();
		let target = (Parent, Parent, PalletInstance(69), GeneralIndex(2)).into();
		let expected = (Parent, Parent, PalletInstance(42), GeneralIndex(1)).into();
		let inverted = ancestry.inverted(&target).unwrap();
		assert_eq!(inverted, expected);
```

xcm/src/v1/multilocation.rs
gavofyork: It's not re-anchoring - it's only inverting the perspective.

xcm/src/v1/multilocation.rs
gavofyork: Fixed.

xcm/xcm-executor/src/lib.rs
gavofyork: Done.

xcm/src/v1/multilocation.rs
```diff
@@ -323,6 +323,60 @@ impl MultiLocation {
...
+	}
+
+	/// Treating `self` as a context, determine how it would be referenced by a `target` location.
+	pub fn inverted(&self, target: &MultiLocation) -> Result<MultiLocation, ()> {
+		use Junction::OnlyChild;
+		let mut ancestry = self.clone();
+		let mut junctions = Junctions::Here;
+		for _ in 0..target.parent_count() {
+			junctions = junctions
+				.pushed_front_with(ancestry.interior.take_last().unwrap_or(OnlyChild))
```
KiChjang, line 355: Now I wonder if we can even make a better guarantee and use an `expect` here. Having a parent count that is more than the number of junctions in ancestry means that in order to reach `target`, `self` must traverse all the way up to the relay chain _and above_, but I believe it's impossible for the relay chain to have a parent `MultiLocation`? Does it even mean anything weird such as `target` not being in the local consensus system?

yrong:
@gavofyork yes, I understand (806)[https://github.com/paritytech/cumulus/pull/806] is not the right fix, the reason why I attach the [test case](https://github.com/paritytech/cumulus/pull/806#issuecomment-987802834) is just to show you the current way we're using(wrongly represent native asset with x2 prefix) and we(bifrost and acala and maybe other parachain team) should also do some change to introduce this pr or the xcm transfer feature between us will fail.

xcm/src/v1/multilocation.rs
gavofyork: It will do in XCM v3, yes. In any case a misconfiguration or bad input should not lead to a panic.

xcm/xcm-executor/src/lib.rs
```diff
@@ -418,13 +425,17 @@ impl<Config: config::Config> XcmExecutor<Config> {
 				for asset in assets.assets_iter() {
 					Config::AssetTransactor::check_out(&dest, &asset);
 				}
-				let assets = Self::reanchored(assets, &dest)?;
+				// Note that we pass `None` as `maybe_failed_bin` and drop any assets which cannot
+				// be reanchored  because we have already checked all assets out.
```
KiChjang, line 429: Not 100% sure on this one -- the docs on `TransactAsset::check_out` states the following:

```
	/// Implementation note: In general this will do one of two things: On chains where the asset is native,
	/// it will increase the assets in a special "teleported" account so that a) total-issuance is preserved; and
	/// b) to ensure that no more assets can be teleported in than were teleported out overall (this should not
	/// be needed if the teleporting chains are to be trusted, but better to be safe than sorry). On chains where
	/// the asset is not native then it will generally just be a no-op.
```

So that leads me to conclude that the native assets that are checked out are put into the holding register in order to "preserve the total issuance". If we choose to drop the assets here by passing in `None`, that seems to me that it's equivalent to a burn, i.e. it reduces the total issuance of the native asset. Am I conceptualizing this correctly?

xcm/xcm-executor/src/lib.rs
```diff
@@ -401,13 +400,21 @@ impl<Config: config::Config> XcmExecutor<Config> {
 				for asset in deposited.assets_iter() {
 					Config::AssetTransactor::deposit_asset(&asset, &dest)?;
 				}
-				let assets = Self::reanchored(deposited, &dest)?;
+				// Note that we pass `None` as `maybe_failed_bin` and drop any assets which cannot
+				// be reanchored  because we have already called `deposit_asset` on all assets.
```
KiChjang, line 404: Security-wise, this is fine, because the funds are already in the sovereign account of the destination. It's just a bit unfortunate that the destination will have no notification of the assets that were received in its sovereign account on the reserve, and I don't really have a solution for that.

xcm/xcm-executor/src/lib.rs
gavofyork: Yeah, but it's not a big deal. It's very difficult to burn assets in this way. People are also heavily incentivised not to (accidentally) burn assets (as they lose them). And if they do, then the fact that the total issuance is unaffected is no big issue.

xcm/xcm-executor/src/lib.rs
gavofyork: I considered it and I don't think there is an easy one which doesn't involve introducing substantial code complexity and performance loss.

KiChjang (approved):

xcm/src/v1/multilocation.rs
```diff
@@ -773,9 +827,82 @@ impl TryFrom<MultiLocation> for Junctions {
...
+		let expected = GeneralIndex(69).into();
+		location.simplify(&context);
+		assert_eq!(location, expected);
+
+		let mut location: MultiLocation =
+			(Parent, Parent, Parachain(1000), PalletInstance(42), GeneralIndex(69)).into();
+		let context = X3(OnlyChild, Parachain(1000), PalletInstance(42));
+		let expected = GeneralIndex(69).into();
+		location.simplify(&context);
+		assert_eq!(location, expected);
```
shawntabrizi, line 874: Should this pass? Because it does.

```
		let mut location: MultiLocation = (Parent, PalletInstance(42), GeneralIndex(69)).into();
		let context = X2(PalletInstance(30), PalletInstance(42));
		let expected = GeneralIndex(69).into();
		location.simplify(&context);
		assert_eq!(location, expected);
```

I am having a really hard time understanding this function's behavior.

shawntabrizi:
I'm sorry, but I cannot give an approve at the moment. Maybe it is just late for me and some concepts aren't clicking, but not fully understanding all the behaviors of these functions.

I will start looking again tomorrow morning if you are blocked on merge for me.

xcm/src/v1/multilocation.rs
gavofyork: Yes. All `location`s within a consensus system must be taken to be within some local `context` (aka the consensus system's _ancestry_). The ancestry is essentially just the consensus system's location from the point of view of the distant ancestor consensus system under which the local consensus sits. For the Polkadot Relay, this is conventionally empty (as it is its own universe) but for a parachain which knows that it exists within a broader consensus system it would be the `Parachain` junction by which the parent Relay-chain refers to it.

Local `location`s which begin with `Parent` should be "escaping" outside of the local consensus system (otherwise there is no need to begin with `Parent`), however if after the `Parent` node(s), the rest of the location is such that it descends back into the local consensus then this becomes a non-canonical location. Basically, the `Parent` items and their corresponding junctions are superfluous because they're already implied by the ancestry.

In order to normalise it, we need to detect and remove this case.

In this case the `context` is `(PalletInstance(30), PalletInstance(42)` 
and the `location` is `Parent, PalletInstance(42), GeneralIndex(69)`. The location is therefore non-canonical, because the `Parent` is being used to descend outside of the local consensus and beyond `PalletInstance(42)` (the last item of `context` and then re-descend immediately back into `PalletInstance(42)`. Therefore this location is exactly equivalent to the simpler (and shorter) `GeneralIndex(69)`.

shawntabrizi (approved):

xcm/xcm-executor/src/lib.rs
KiChjang: Oh, I thought the `special "teleported" account` here meant the holding register, because if it did, then the total issuance would have _decreased_ as it was removed from holding, and wasn't ever put back into it after reanchoring had failed.

xcm/src/v1/multilocation.rs
shawntabrizi: Yes, I understand now. Basically, we use context and the front of the multilocation and cancel out all "redundant" data, and normalize it.

> The ancestry is essentially just the consensus system's location from the point of view of the most distant ancestor consensus system under which the local consensus sits.

This was very key. Thanks.

zqhxuyuan: 👍 just some note here
image

dzmitry-lahoda: where I can read about stuff happening here? except code:) I read all papers on xcm I found and ported acala tests. so found reanchor stuff. do not get what it is. actually security of xcm is very unclear.

KiChjang: @dzmitry-lahoda Reanchoring is a piece of logic that simply works behind the scenes. As an end-user of the XCM API, you most likely wouldn't need to touch it. What is the exact problem that led you here to understand reanchoring?

dzmitry-lahoda: I am stealing XCM stuff from Acala.  I am ok with that (0, CurrencyId) => CurrencyId. But here we see that it was sent from Relay. Why it is safe to map (0, CurrencyId) => CurrencyId? May be another buggy para chain may send that and mint infinite amounts? So it is not clear what is security XCM provides to parachains against other parachains.

image

KiChjang: @dzmitry-lahoda Substrate/Polkadot doesn't have a concept of `CurrencyId`, and it looks to me to be an Acala/Karura concept. Would you kindly raise this issue in their repo instead?

dzmitry-lahoda: :) Substrate/Polkadot has the location and general key concept. why would somebody trust these are genuine? Asset with location FooBar. Parachain X xcmed FooBar from X to Z. Why would we trust that FooBar was not just minted?

KiChjang: I think you are not understanding that every consensus system (or blockchain in layman terms) is responsible for interpreting incoming XCM messages correctly. As such, _how_ a chain decides to interpret `GeneralKey` or `GeneralIndex` is not up to us to decide.

Thus, would you please raise this issue to Acala/Karura?

dzmitry-lahoda: Still not clear. Kusama can have KSM. Kusama can send (if we call its XCM pallet) to send KSM to paracahain. Can someone send KSM back to Kusama? Does Kusama tracks all places it send KSM to be equal?  I mean DOT. Polkadot is relay chain. Does DOT sent via XCM has any security/alignment after it was send from Polkadot to Parachain?

So right now I speak about Relay and some abstract Parachain. And some 2 abstract parachains having DOT/KSM.  How these could interpret correctly message with asset (1, Here) origin?

KiChjang: `MultiLocation`s are relative paths, as explained in [Gav's blogpost](https://medium.com/polkadot-network/xcm-the-cross-consensus-message-format-3b77b1373392#1330). A parachain of Kusama/Polkadot will identify KSM/DOT as `(1, Here)`. When they send `(1, Here)` to the relay chain, their local XCM executor will reanchor the location to be `(0, Here)`.

As I mentioned in my first message, reanchoring happens automatically behind the scenes and doesn't require any user input. If you are wondering about how assets are transferred between parachains, I have an [answer on Stack Exchange](https://substrate.stackexchange.com/questions/37/how-can-i-transfer-assets-using-xcm) that describes in detail how it is done, and how the trust issues are resolved.

Issue #4449, opened 2021-12-02
Better reversion loop protection


eskimor:
On chain reversion, we will rip out any backed candidate if are able to import the reversion causing dispute. But in case the block producer is for some reason not aware of that dispute, this won't happen and the candidate can become available again in that very block, which in turn will cause the block to get disputed and reverted again. This will happen in a loop, until it is the turn of some block producer that actually has seen the dispute.

We'd like to ensure the chain to make progress as much as possible, so I would suggest the following safe guards:

1. In availability-distribution - take into account the `LeafStatus` and only fetch chunks if it is not `Stale`.
2.  Make the block producer not provide bitfields to the runtime in case of `LeafStatus` `Stale`.

This way the offended candidate will still stay backed, but won't be included for at least one fresh block and thus the chain will be able to progress.

eskimor:
Step 1 will actually not have any effect, because for a stale leaf availability was already distributed so availability-distribution would most likely be a noop anyway - not trying to recover on stale leaves should still be a good idea, but in addition we should make sure not to send out bitfields.

PR #4472, opened 2021-12-06
Revert loop prevention

eskimor:
Avoid including bitfields on `Stale` leaves.

- Only launch `availability-distribution` on fresh leaves
- Only include bitfields in the `provisioner` on fresh leaves
- Only sign bitfields on fresh leaves in `bitfield-signing`

This should make it impossible for a reverted chain to get reverted again, without any new blocks persisted. 

Not triggering `availability-distribution` on stale leaves is mostly redundant, as the chunk will most likely have been fetched already, thus `availability-distribution` will be a no-op anyway.

Not including bitfields in the provisioner should already be enough to prevent reversion loops, still not distributing/providing them in `bitfield-signing` makes things even more robust and also avoids wasting of resources.

Fixes: https://github.com/paritytech/polkadot/issues/4449

                                                                
skip check-dependent-cumulus


node/core/bitfield-signing/src/lib.rs
```diff
@@ -237,23 +237,31 @@ impl JobTrait for BitfieldSigningJob {
...
-			let span = PerLeafSpan::new(span, "bitfield-signing");
+			if let LeafStatus::Stale = leaf.status {
+				tracing::debug!(
+					target: LOG_TARGET,
+					hash = ?leaf.hash,
+					block_number =  ?leaf.number,
+					"Stale leaf - don't sign bitfields."
+				);
+				return Ok(())
+			}
```
drahnr, line 256: :+1:

drahnr (approved):
LGTM

PR #4480, opened 2021-12-07
introduce Dummy trait for benches

drahnr:
During the benchmarks PR some defaults were made unconditional, which enables them in production and allows for some serious footguns. This has been there for a while and is used in quite a few unit tests, besides the runtime benchmarking.

This is the first step to replace the `Default::default()` instances with `Dummy::dummy()`.

Since this is spread out over the codebase, we might want to introduce `derive(Dummy)`.

@zeke could you double check the failing tests, I am not entirely sure how this change could affect this

Blocks https://github.com/paritytech/substrate/pull/10403

### DIFF
--- a/parachain/src/primitives.rs
+++ b/parachain/src/primitives.rs
@@ -41,19 +41,9 @@ pub use polkadot_core_primitives::BlockNumber as RelayChainBlockNumber;
 
 /// Parachain head data included in the chain.
 #[derive(
-	PartialEq,
-	Eq,
-	Clone,
-	PartialOrd,
-	Ord,
-	Encode,
-	Decode,
-	RuntimeDebug,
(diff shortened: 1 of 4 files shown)


gavofyork:
Having reflected, I don't think the dummy trait is sensible either.

Benchmarks and tests should not be falling back to use a single "privileged" account, since it coerces the developer into using one account when several independent accounts would be more fitting.

In tests, always using the same single account could result in a myopia that would result in a conditions not being checked. In benchmarks it might result in over-optimistic results. Either way it's a bad thing.

zeke:
Not sure why I got pinged on this. 🤔

drahnr:
Mea culpa, please ignore :) meant @emostov

PR #4490, opened 2021-12-08
Fix `pallet_xcm::execute`

athei:
The `max_weight` argument is listed in the weight annotation and therefore the weight is already paid by including the extrinsic due to weight fee collection. Therefore we should credit it here.

We should also use weight correction here. I only use it in case we actually do the xcm execution. One could also use it in case we error out before reaching this point. This would add more complexity to the function. Not sure if worth.

I would need to add some additional tests if we consider merging this. Just wanted to make sure that my assumptions about this functions are correct before investing more work.

### DIFF
--- a/xcm/pallet-xcm/src/lib.rs
+++ b/xcm/pallet-xcm/src/lib.rs
@@ -572,7 +572,7 @@ pub mod pallet {
 			origin: OriginFor<T>,
 			message: Box<VersionedXcm<<T as SysConfig>::Call>>,
 			max_weight: Weight,
-		) -> DispatchResult {
+		) -> DispatchResultWithPostInfo {
 			let origin_location = T::ExecuteXcmOrigin::ensure_origin(origin)?;
 			let message = (*message).try_into().map_err(|()| Error::<T>::BadVersion)?;
 			let value = (origin_location, message);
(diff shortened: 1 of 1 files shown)


shawntabrizi (approved):
Yup, this is a good addition to refund weight.

shawntabrizi:
Actually I think there is a bug here:

```
	/// How much weight was used by the XCM execution attempt.
	pub fn weight_used(&self) -> Weight {
		match self {
			Outcome::Complete(w) => *w,
			Outcome::Incomplete(w, _) => *w,
			Outcome::Error(_) => 0,
		}
	}
```

The `weight_used` when returning `Error` should not be zero, but maybe `Weight::MAX`.

athei:
> We probably should not be doing a weight refund returning an `Error`.

Why? In case of `Error` the `XCM` execution didn't even start according to documentation. Also, the `weight_used()` function would be faulty in this case.

Do you also agree with the other change (crediting the `max_weight`)?

shawntabrizi:
My only thought is that we don't support refunds when extrinsics error, and generally that provides an even larger negative incentive for people to submit any transactions which may potentially error.

You can think of the larger the potential fee, the more someone will be careful about sending a message which succeeds. Ideally, the blockchain is only consistent of successful transactions, but instead, we must store bad ones too since we need to charge a fee for that.

Anyway, practically speaking, you are right though, the weight returned here should be safe.

shawntabrizi (approved):

PR #4491, opened 2021-12-08
New changelog scripts

chevdor:
This is an upgrade to the release notes generation system, aligned with how it works in Cumulus:
- use `changelogerator` to generate the list of changes, including labels and label pre-processing, for both polkadot and substrate repos
- compose srtool digests for each chain and the changelogerator outputs for polkadot and substrate into a main context file: `context.json`
- use `tera-cli` to process the templates with the context to produce the final changelog

A sample run can be seen here: https://github.com/chevdor/polkadot/actions/runs/1560746640

The matching release notes can be seen here: https://github.com/chevdor/polkadot/releases/tag/v0.0.0-rc1

skip check-dependent-cumulus


kianenigma (approved):
You can count this as a symbolic approve, expressing my appreciation of this PR and emphasis on how gravely we need it.

PR #4493, opened 2021-12-08
Add CheckNonZeroSender to runtime configs

### LINKED ISSUE #4488: Introduce CheckNonZeroSender to Relay chains

KiChjang:
skip-cumulus-check

Adding the `CheckNonZeroSender` signed extension a la paritytech/substrate#10413.

Fixes #4488.


shawntabrizi (approved):
need to fix compile / tests

bridges/bin/millau/runtime/src/lib.rs
```diff
@@ -531,6 +531,7 @@ pub type SignedBlock = generic::SignedBlock<Block>;
 pub type BlockId = generic::BlockId<Block>;
 /// The `SignedExtension` to the basic transaction logic.
 pub type SignedExtra = (
+	frame_system::CheckNonZeroSender<Runtime>,
```
xlc, line 534: why is this added the top? is it a requirement?

KiChjang: IIRC `SignedExtension`s get run in the order of declaration, so we'd definitely want to check for the non-zero sender somewhere early. What's the problem that you would run into if it's at the top?

xlc: As you can see in this PR, it changes the index of nonce and tip. Not a major issue, but still some additional changes requires more work.

Will it make any visible difference if it is placed after check weight before charge transaction fee?

shawntabrizi: This shouldn't affect the transaction format if that is what your are implying. This check should not affect the signature payload since the extension itself does not contain any data.

xlc: No. I mean it will require us to change code like https://github.com/AcalaNetwork/Acala/blob/a657f2db621a7d2364c168da666b36d0d1ee0d84/runtime/mandala/src/lib.rs#L1975

gavofyork: We should probably switch to using a struct with named fields rather than a tuple here. This will likely keep happening.

gavofyork (approved):

PR #4509, opened 2021-12-10
Don't cache unavailable results.

eskimor:


node/network/availability-recovery/src/lib.rs
```diff
@@ -686,6 +687,38 @@ impl Future for RecoveryHandle {
...
 }
 
+/// Cached result of an availability recovery operation.
+#[derive(Debug, Clone)]
+enum CachedResult {
+	/// Availability was successfully retrieved before.
+	Success(AvailableData),
+	/// Availability was successfully retrieved before, but was found to be invalid.
+	Invalid,
+}
```
pepyakin, line 697: Maybe we can work with names. Like use `Invalid` and `Valid` for variants. Since that will make it related to validity we can name the enum `CachedValidity` as well? This will solve unfortunate naming of `CachedResult::into_result` which sounds a bit tautology.

node/network/availability-recovery/src/lib.rs
```diff
@@ -972,7 +1007,9 @@ impl AvailabilityRecoverySubsystem {
 				}
 				output = state.ongoing_recoveries.select_next_some() => {
 					if let Some((candidate_hash, result)) = output {
-						state.availability_lru.put(candidate_hash, result);
+						if let Ok(result) = CachedResult::try_from(result) {
```
pepyakin, line 1010: on the similar note, this will become

`if let Ok(validity) = CachedValidity::try_from(result) {`

pepyakin (commented):
Naming nits

node/network/availability-recovery/src/lib.rs
```diff
@@ -972,7 +1007,9 @@ impl AvailabilityRecoverySubsystem {
 				}
 				output = state.ongoing_recoveries.select_next_some() => {
 					if let Some((candidate_hash, result)) = output {
-						state.availability_lru.put(candidate_hash, result);
+						if let Ok(result) = CachedValidity::try_from(result) {
```
drahnr, line 1010: This is a bit of a mind-twister, we take a result, and get a `Ok(result)`, this should be `Ok(cachable_validity)`

drahnr (approved):
Names could still get some :bike: :house: -ing - other than that LGTM. It's very clean, but it also adds quite a bit of complexity.

node/network/availability-recovery/src/lib.rs
```diff
@@ -686,6 +687,38 @@ impl Future for RecoveryHandle {
 	}
 }
 
+/// Cached result of an availability recovery operation.
+#[derive(Debug, Clone)]
+enum CachedValidity {
```
drahnr, line 692: isn't this rather a `AvailabilityRecoveryStatus`? Skip the `Availability` if it's too long, but `CachedValidity` is not indicating what it represents.

eskimor: I ended up with `CachedRecovery` as it is concise and describes what it is the best.

node/network/availability-recovery/src/lib.rs
eskimor: haha,  I haven't noticed - that was really bad.

drahnr:
bot merge

PR #4510, opened 2021-12-10
pvf-precheck: teach runtime-api to work with versions

pepyakin:
As part of #3211 we will need to add a couple of runtime APIs. That
change is coming in following PRs.

The runtime API for pre-checking will be introduced with a version bump
for `ParachainHost`. This commit prepares the ground for that change,
by introducing a error condition that signals that the given API is not
supported.


pepyakin:
* master
  * **PR #4510**  (this PR)

eskimor:
Stupid question, for a simple "supported"/"not supported" - why do we need to check versions - can't we just call and handle the error?

eskimor:
Answering my own question: Because then we would also not report an error in case the API got removed later on, without the node code having been adjusted.

eskimor (approved):
Loving it!

node/core/runtime-api/src/lib.rs
```diff
@@ -335,12 +335,52 @@ where
...
+						tracing::warn!(
+							target: LOG_TARGET,
+							"no runtime version is reported"
+						);
+						WIDELY_DEPLOYED_API_VERSION
+					})
+			} else {
+				// The required version is less or equal to the widely deployed runtime API version.
+				WIDELY_DEPLOYED_API_VERSION
+			};
```
bkchr, line 371: FYI, we are caching the runtime version. So, getting the runtime version does not involve calling into the runtime

pepyakin: I actually considered this. My reasoning was that runtime APIs can be called on different blocks. But writing this I realize that this may be not a good reason. 

- We should not worry about the runtime cache evictions that much: in any case there won't be more than two active runtimes, no matter how many heads there are.
- There is no reason for trying to minimize cases where we query runtime version, because it will still be called every block by some subsystem.

Well, I guess we can simplify that!

bkchr: There is also tons of other stuff that will call into the runtime on every block. So, not that much to worry about :P

PR #4521, opened 2021-12-13
minor improvements in inherent data filtering efficiency

drahnr:
Reduces the number of signature checks done


primitives/src/v1/mod.rs
```diff
@@ -1349,9 +1349,41 @@ pub struct DisputeStatementSet {
...
+
+impl Into<DisputeStatementSet> for CheckedDisputeStatementSet {
+	fn into(self) -> DisputeStatementSet {
+		self.0
+	}
+}
+
+impl CheckedDisputeStatementSet {
+	/// Convert from an unchecked, checking _must_ be done in the context.
+	pub fn from_unchecked(unchecked: DisputeStatementSet) -> Self {
```
sandreim, line 1379: Why doesn't impl `From` work here ?

sandreim: Would it make sense to perform the checks here and return an `Option`?

primitives/src/v1/mod.rs
```diff
@@ -1349,9 +1349,41 @@ pub struct DisputeStatementSet {
...
+		&self
+	}
+}
+
 /// A set of dispute statements.
 pub type MultiDisputeStatementSet = Vec<DisputeStatementSet>;
 
+/// A _checked_ set of dispute statements.
+#[derive(Clone)]
+pub struct CheckedDisputeStatementSet(DisputeStatementSet);
```
sandreim, line 1363: Should we use an enum with variants: Checked/Unchecked ?

runtime/parachains/src/disputes.rs
```diff
@@ -104,20 +105,66 @@ impl PunishValidators for () {
...
+				(None, None) => a.session.cmp(&b.session),
+			}
+		});
+		statement_sets.dedup();
+
+		// if there were any duplicates, indicate that to the caller.
+		if n == statement_sets.len() {
+			Ok(())
+		} else {
+			Err(())
```
sandreim, line 150: Would be worth explaining why duplicates are an error and how it is handled. IMO finding duplicates should not be an error condition, but I might be wrong.

drahnr: The caller should be notified and is the one to decide how the handle it. The information should not be swallowed silently, it removes the option to act upon it.

primitives/src/v1/mod.rs
drahnr: No, that would remove type level guarantees that two disjunctive types allow.

primitives/src/v1/mod.rs
drahnr: Technically yes, but that would require additional context and enforce another change in behavior which I wanted to avoid as much as possible. The changeset is already too large.

runtime/parachains/src/paras_inherent/mod.rs
```diff
@@ -955,33 +999,57 @@ fn compute_entropy<T: Config>(parent_hash: T::Hash) -> [u8; 32] {
...
-				(None, Some(_)) => Ordering::Greater,
-				(Some(_), None) => Ordering::Less,
-				// For local disputes, prioritize those that occur at an earlier height.
-				(Some(a_height), Some(b_height)) => a_height.cmp(&b_height),
-				// Prioritize earlier remote disputes using session as rough proxy.
-				(None, None) => a.session.cmp(&b.session),
-			}
-		});
+) -> (Vec<CheckedDisputeStatementSet>, Weight) {
+	// The total weight iff all disputes would be included
```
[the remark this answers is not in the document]
ordian: https://en.wikipedia.org/wiki/If_and_only_if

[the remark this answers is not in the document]
drahnr: Yeah, that second `f` was a bit ambitious :upside_down_face:

primitives/src/v1/mod.rs
eskimor: I would at least call it `unsafe_from_unchecked` or something. Being able to grep through source code for `unsafe` to manually check things the compiler won't check is a nice property to have.

drahnr: Not really keen on using `unsafe` here, this conveys a different meaning in rust-context. Up for alt-names though :bike: :house:

runtime/parachains/src/disputes.rs
```diff
@@ -609,7 +697,8 @@ impl StatementSetFilter {
 				if statement_set.statements.is_empty() {
 					None
 				} else {
-					Some(statement_set)
+					// we just checked correctness when filtering.
+					Some(CheckedDisputeStatementSet::from_unchecked(statement_set))
```
eskimor, line 701: hmmm ... where exactly?

runtime/parachains/src/disputes.rs
```diff
@@ -717,31 +806,21 @@ impl<T: Config> Pallet<T> {
 	/// Handle sets of dispute statements corresponding to 0 or more candidates.
 	/// Returns a vector of freshly created disputes.
 	///
+	/// Assumes `statement_sets` were already de-duplicated.
```
eskimor, line 809: Do duplicates ever make any sense? And if not, could we make it so that `CheckedMultiDisputeStatementSet` gets a smart constructor that ensures this invariant?

eskimor (commented):
First pass.

runtime/parachains/src/disputes.rs
```diff
@@ -100,24 +101,85 @@ impl PunishValidators for () {
...
-	/// dispute statements.
-	fn filter_multi_dispute_data(statement_sets: &mut MultiDisputeStatementSet);
+	/// Remove dispute statement duplicates and sort the non-duplicates based on
+	/// local (lower indicies) vs remotes (higher indices) and age (older with lower indices).
+	fn deduplicate_and_sort_dispute_data(
+		statement_sets: &mut MultiDisputeStatementSet,
+	) -> Result<(), ()> {
+		use core::cmp::Ordering;
+
+		// TODO: Consider trade-of to avoid `O(n * log(n))` average lookups of `included_state`
```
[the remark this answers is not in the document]
drahnr: It was just a thought and is devoid of much meaning. Will remove

runtime/parachains/src/disputes.rs
drahnr: Where the filter is constructed, in https://github.com/paritytech/polkadot/blob/de62962f7f4e21af8e5d95b7e13fbbe61ec81cf9/runtime/parachains/src/disputes.rs#L911

runtime/parachains/src/disputes.rs
eskimor: Ok, I am guessing to make sure a block author is not trying to spam us with lots of duplicate disputes, where we would have to check signatures on each of those. I can see though, that we are ignoring the error on the calling side ... could we log it?

drahnr: Yeah, that's a good point. Logging is the very least we should do. But this is not even a warning, it's only a `debug!`

runtime/parachains/src/disputes.rs
drahnr: This would require the `CheckedMultiDisputeStatementSet` to become a full type rather than an alias. I think it might make sense to create another alias of `DedupSortedMultiDisputeStatementSets` it could be created as the output of `fn deduplicate_and_sort_..` :+1: and then be required by anything following, besides add a `CheckedMultiDisputeStatementSet` that assures sorting.

runtime/parachains/src/disputes.rs
```diff
@@ -100,24 +101,85 @@ impl PunishValidators for () {
...
+	Skip,
+}
+
 /// Hook into disputes handling.
 ///
 /// Allows decoupling parachains handling from disputes so that it can
 /// potentially be disabled when instantiating a specific runtime.
-pub trait DisputesHandler<BlockNumber> {
+///
+/// Returns `Ok(())` if no duplicates were present, `Err(())` otherwise.
```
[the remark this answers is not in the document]
drahnr: Hmm, this should have remained on the member fn, not sure how this happened.

primitives/src/v1/mod.rs
drahnr: Quite the opposite - I find keyword re-use more dangerous, it conflates memory safety and topological safety under a single name/tag `unsafe`. This was discussed in various groups (rust security SIG iirc besides some design lib PRs) and was agreed to not be a good idea. This is just the same thing, so we should not go down this path imho.

drahnr: We eventually want to follow design of `Signed<_>` where the conversion requires checking and only retain the current `fn from_unchecked` for tests.

runtime/parachains/src/disputes.rs
```diff
@@ -100,24 +101,85 @@ impl PunishValidators for () {
...
+			let a_local_block = Self::included_state(a.session, a.candidate_hash);
+			let b_local_block = Self::included_state(b.session, b.candidate_hash);
+			match (a_local_block, b_local_block) {
+				// Prioritize local disputes over remote disputes.
+				(None, Some(_)) => Ordering::Greater,
+				(Some(_), None) => Ordering::Less,
+				// For local disputes, prioritize those that occur at an earlier height.
+				(Some(a_height), Some(b_height)) => a_height.cmp(&b_height),
+				// Prioritize earlier remote disputes using session as rough proxy.
+				(None, None) => a.session.cmp(&b.session),
```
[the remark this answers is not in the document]
Lldenaurois: I think your last suggestion is the most effective. Instead of a sorted vec, we should simply traverse the statement_sets to create two pools of statements to select from.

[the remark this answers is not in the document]
drahnr: :+1:

runtime/parachains/src/disputes.rs
```diff
@@ -100,24 +101,85 @@ impl PunishValidators for () {
...
+	fn filter_dispute_data(
+		statement_set: DisputeStatementSet,
+		max_spam_slots: u32,
+		post_conclusion_acceptance_period: BlockNumber,
+		verify_sigs: VerifyDisputeSignatures,
+	) -> Option<CheckedDisputeStatementSet>;
 
 	/// Handle sets of dispute statements corresponding to 0 or more candidates.
 	/// Returns a vector of freshly created disputes.
 	fn provide_multi_dispute_data(
```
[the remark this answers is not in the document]
drahnr: Well aware that the entire runtime side of the relay chain needs some more attention, I am reluctant to make all these changes in this PR.

runtime/parachains/src/disputes.rs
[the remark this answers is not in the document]
drahnr: Not sure we want to skip the sorting entirely, it allows for `O(n)` deduplication.

chevdor:
Switching to the 0.9.16 after discussion with @drahnr

drahnr:
Superseeded by https://github.com/paritytech/polkadot/pull/4657

PR #4540, opened 2021-12-16
parachains: Fix configuration module

### LINKED ISSUE #4529: Consecutive host configuration updates in two different sessions leads to overwriting the first update
The affected code:

https://github.com/paritytech/polkadot/blob/cf433f5f8faaffa8e07ef965eeea0ae15aa47b69/runtime/parachains/src/configuration.rs#L1074-L1078

### LINKED ISSUE #4533: Configuration module migration may be broken in presence of PendingConfig
The configuration upgrade to version 2, introduced by https://github.com/paritytech/polkadot/commit/10f3c85ebdeaf422fe151096c173b102297a95d1, may be broken in case there are non empty `PendingConfig` entries. If this is the case, then querying such a pending configuration change will lead to panic in the runtime.

pepyakin:
Closes #4529
Closes #4533

I figured that trying to avoid updates does not really worth it to keep.
This is because we seem to not update the configuration often and when
we do we approach this carefully. Thus possibility of a redundant update
is really negligable. At the same time, if such a redundant update does
happen then the effects of that are really small: just some wasted
storage interactions.

On the other hand, making it work was a little bit annoying. With the
proper fix for the pending updates this would be even more annoying
since now we would have to add combinatorically more cases to test this.

So I figured that I will just scrap that and simplify the code.

# Runtime Migration

This PR doesn't introduce a new runtime migration but rather fixes the migration code introduced in https://github.com/paritytech/polkadot/pull/4420

skip check-dependent-cumulus


pepyakin:
* master
  * **PR #4540**  (this PR)

pepyakin:
/benchmark runtime polkadot runtime_parachains::configuration

pepyakin:
/benchmark runtime custom --chain=rococo-dev --steps=50 --repeat=20 --pallet=runtime_parachains::configuration --extrinsic="*" --execution=wasm --wasm-execution=compiled --heap-pages=4096 --header=./file_header.txt --output=./runtime/rococo/src/weights/runtime_parachains_configuration.rs

runtime/parachains/src/configuration.rs
```diff
@@ -361,9 +360,22 @@ pub mod pallet {
...
+	/// Pending configuration changes.
+	///
+	/// This is a list of configuration changes, each with a session index at which it should
+	/// be applied.
+	///
+	/// The list is sorted ascending by session index. Also, this list can only contain at most
+	/// 2 items: for the next session and for the `scheduled_session`.
+	#[pallet::storage]
+	pub(crate) type PendingConfigs<T: Config> =
+		StorageValue<_, Vec<(SessionIndex, HostConfiguration<T::BlockNumber>)>, ValueQuery>;
```
shawntabrizi, line 378: if it can only contain 2 items, why not use a fixed size array?

pepyakin: What would that give though? At the same time, it would be more annoying to change if we were to change the SESSION_DELAY.

shawntabrizi: No specific reason, just thinking that if you are going to have some specific constraint about this storage item, that we can just program that in.

runtime/parachains/src/configuration.rs
```diff
@@ -1054,9 +1065,28 @@ impl<T: Config> Pallet<T> {
...
 	/// Called by the initializer to note that a new session has started.
 	pub(crate) fn initializer_on_new_session(session_index: &SessionIndex) {
-		if let Some(pending) = <Self as Store>::PendingConfig::take(session_index) {
-			<Self as Store>::ActiveConfig::set(pending);
+		let pending_configs = <PendingConfigs<T>>::get();
+		if pending_configs.is_empty() {
+			return
 		}
+
+		let (past_and_now, future) = pending_configs
```
shawntabrizi, line 1073:
```suggestion
		let (past_and_present, future) = pending_configs
```

Extreme nit pick, but this the associated word with "past" and "future" it "present"

runtime/parachains/src/configuration.rs
```diff
@@ -1104,21 +1192,146 @@ mod tests {
...
+			on_new_session(2);
 
 			assert_eq!(Configuration::config(), config);
-			assert_eq!(<Configuration as Store>::PendingConfig::get(3), None);
+			assert_eq!(<Configuration as Store>::PendingConfigs::get(), vec![]);
 		})
 	}
 
+	#[test]
+	fn consequtive_changes_within_one_session() {
```
shawntabrizi, line 1214:
```suggestion
	fn consecutive_changes_within_one_session() {
```

shawntabrizi (approved):
looking good to me

runtime/parachains/src/configuration.rs
pepyakin: Yeah, I am generally pro making it more typesafe, but I believe here it does not provide much benefit while can be potential problem in the future (although I admit that it is also not super likely we will change SESSION_DELAY)

runtime/parachains/src/configuration.rs
```diff
@@ -1104,21 +1192,146 @@ mod tests {
...
+
+			on_new_session(3);
+
+			assert_eq!(Configuration::config(), final_config);
+			assert_eq!(<Configuration as Store>::PendingConfigs::get(), vec![]);
+		});
+	}
+
+	#[test]
+	fn test() {
```
eskimor, line 1287: That test name is not super descriptive.

runtime/parachains/src/configuration.rs
```diff
@@ -1104,21 +1192,146 @@ mod tests {
...
+		new_test_ext(Default::default()).execute_with(|| {
```
eskimor, line 1288: Default::default ... is also not helping.

eskimor (approved):
LGTM

runtime/parachains/src/configuration.rs
pepyakin: Oops, forgot to rename!

runtime/parachains/src/configuration.rs
pepyakin: You know that I hate `Default::default`. Here, however, I stuck to the code around. I can change it, but then we will end up with inconsistency.

eskimor: Yeah, let's fix that in some cleanup PR for all then.

pepyakin:
This PR doesn't introduce a new runtime migration but rather fixes the code in the previously introduced one.

eskimor:
bot merge

PR #4542, opened 2021-12-16
pvf-precheck: hook up runtime API

pepyakin:
This commit hooks up the API provided by #4457 to the runtime API
subsystem. In a following PR this API will be consumed by the PVF
pre-checking subsystem.

skip check-dependent-cumulus

### DIFF
--- a/primitives/src/v1/mod.rs
+++ b/primitives/src/v1/mod.rs
@@ -986,6 +986,7 @@ impl ApprovalVote {
 
 sp_api::decl_runtime_apis! {
 	/// The API for querying the state of parachains on-chain.
+	#[api_version(2)]
 	pub trait ParachainHost<H: Encode + Decode = Hash, N: Encode + Decode = BlockNumber> {
 		/// Get the current validators.
 		fn validators() -> Vec<ValidatorId>;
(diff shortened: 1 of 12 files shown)


primitives/src/v1/mod.rs
```diff
@@ -1054,6 +1055,12 @@ sp_api::decl_runtime_apis! {
 
 		/// Scrape dispute relevant from on-chain, backing votes and resolved disputes.
 		fn on_chain_votes() -> Option<ScrapedOnChainVotes<H>>;
+
+		/// Submits a PVF pre-checking statement into the transaction pool.
+		fn submit_pvf_check_statement(stmt: PvfCheckStatement, signature: ValidatorSignature);
```
eskimor, line 1060: Other kind of statements we imported via inherents. How come we are using the transaction pool here?

eskimor (approved):

primitives/src/v1/mod.rs
pepyakin: It was suggested by Rob [here](https://github.com/paritytech/polkadot/issues/3211). We never really discussed exactly this part, mostly because I agree with this. On one hand, we can afford routing those through the chain: we do not expect a lot of those and they are not too heavy. On the other hand, it greatly simplifies the whole construction. The next PR is for subsystem and you'll see how simple it is.

That being said, in future we can reconsider and implement it via a network protocol if such need arises.

eskimor: Not sure, I fully understand this - maybe a quick call tomorrow?

pepyakin: sure, thing.

eskimor:
bot merge

PR #4549, opened 2021-12-17
Implement transfer_asset on CurrencyAdapter

### LINKED ISSUE #4548: Default beam_asset implementation is not compatible with tuple asset transactor
The test: https://github.com/AcalaNetwork/Acala/commit/f69e348d67fd6eac494730fb9557173667d4add3#diff-505267759159dc9fcbb79d7e6758c08a2b1b5c3402a4a307321153312a31f4de

```
$ RUST_LOG=xcm=trace cargo test -p runtime-integration-tests --features=with-karura-runtime -- --nocapture statemine
   Compiling runtime-integration-tests v2.1.0 (/Users/xiliangchen/projects/acala/Acala/runtime/integration-tests)
    Finished test [unoptimized + debuginfo] target(s) in 14.40s
     Running unittests (target/debug/deps/runtime_integration_tests-1f175f768eff02f5)

running 1 test
[2021-12-17T00:05:51Z TRACE xcm::weight] FixedWeightBounds message: Xcm([ReserveAssetDeposited(MultiAssets([MultiAsset { id: Concrete(MultiLocation { parents: 1, interior: Here }), fun: Fungible(1000000000000) }])), ClearOrigin, BuyExecution { fees: MultiAsset { id: Concrete(MultiLocation { parents: 2, interior: Here }), fun: Fungible(1000000000000) }, weight_limit: Limited(0) }, DepositAsset { assets: Wild(All), max_assets: 1, beneficiary: MultiLocation { parents: 0, interior: X1(AccountId32 { network: Any, id: [5, 5, 5, 5, 5, 5, 5, 5, 5, 5, 5, 5, 5, 5, 5, 5, 5, 5, 5, 5, 5, 5, 5, 5, 5, 5, 5, 5, 5, 5, 5, 5] }) } }])
[2021-12-17T00:05:51Z TRACE xcm::weight] FixedWeightBounds message: Xcm([TransferReserveAsset { assets: MultiAssets([MultiAsset { id: Concrete(MultiLocation { parents: 1, interior: Here }), fun: Fungible(1000000000000) }]), dest: MultiLocation { parents: 0, interior: X1(Parachain(1000)) }, xcm: Xcm([BuyExecution { fees: MultiAsset { id: Concrete(MultiLocation { parents: 2, interior: Here }), fun: Fungible(1000000000000) }, weight_limit: Limited(4000000000) }, DepositAsset { assets: Wild(All), max_assets: 1, beneficiary: MultiLocation { parents: 0, interior: X1(AccountId32 { network: Any, id: [5, 5, 5, 5, 5, 5, 5, 5, 5, 5, 5, 5, 5, 5, 5, 5, 5, 5, 5, 5, 5, 5, 5, 5, 5, 5, 5, 5, 5, 5, 5, 5] }) } }]) }])
[2021-12-17T00:05:51Z TRACE xcm::execute_xcm_in_credit] origin: MultiLocation { parents: 0, interior: X1(AccountId32 { network: Named([87, 101, 115, 116, 101, 110, 100]), id: [4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4] }) }, message: Xcm([TransferReserveAsset { assets: MultiAssets([MultiAsset { id: Concrete(MultiLocation { parents: 1, interior: Here }), fun: Fungible(1000000000000) }]), dest: MultiLocation { parents: 0, interior: X1(Parachain(1000)) }, xcm: Xcm([BuyExecution { fees: MultiAsset { id: Concrete(MultiLocation { parents: 2, interior: Here }), fun: Fungible(1000000000000) }, weight_limit: Limited(4000000000) }, DepositAsset { assets: Wild(All), max_assets: 1, beneficiary: MultiLocation { parents: 0, interior: X1(AccountId32 { network: Any, id: [5, 5, 5, 5, 5, 5, 5, 5, 5, 5, 5, 5, 5, 5, 5, 5, 5, 5, 5, 5, 5, 5, 5, 5, 5, 5, 5, 5, 5, 5, 5, 5] }) } }]) }]), weight_limit: 1000000000, weight_credit: 1000000000
[2021-12-17T00:05:51Z TRACE xcm::weight] FixedWeightBounds message: Xcm([TransferReserveAsset { assets: MultiAssets([MultiAsset { id: Concrete(MultiLocation { parents: 1, interior: Here }), fun: Fungible(1000000000000) }]), dest: MultiLocation { parents: 0, interior: X1(Parachain(1000)) }, xcm: Xcm([BuyExecution { fees: MultiAsset { id: Concrete(MultiLocation { parents: 2, interior: Here }), fun: Fungible(1000000000000) }, weight_limit: Limited(4000000000) }, DepositAsset { assets: Wild(All), max_assets: 1, beneficiary: MultiLocation { parents: 0, interior: X1(AccountId32 { network: Any, id: [5, 5, 5, 5, 5, 5, 5, 5, 5, 5, 5, 5, 5, 5, 5, 5, 5, 5, 5, 5, 5, 5, 5, 5, 5, 5, 5, 5, 5, 5, 5, 5] }) } }]) }])
[2021-12-17T00:05:51Z TRACE xcm::barriers] TakeWeightCredit origin: MultiLocation { parents: 0, interior: X1(AccountId32 { network: Named([87, 101, 115, 116, 101, 110, 100]), id: [4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4] }) }, message: Xcm([TransferReserveAsset { assets: MultiAssets([MultiAsset { id: Concrete(MultiLocation { parents: 1, interior: Here }), fun: Fungible(1000000000000) }]), dest: MultiLocation { parents: 0, interior: X1(Parachain(1000)) }, xcm: Xcm([BuyExecution { fees: MultiAsset { id: Concrete(MultiLocation { parents: 2, interior: Here }), fun: Fungible(1000000000000) }, weight_limit: Limited(4000000000) }, DepositAsset { assets: Wild(All), max_assets: 1, beneficiary: MultiLocation { parents: 0, interior: X1(AccountId32 { network: Any, id: [5, 5, 5, 5, 5, 5, 5, 5, 5, 5, 5, 5, 5, 5, 5, 5, 5, 5, 5, 5, 5, 5, 5, 5, 5, 5, 5, 5, 5, 5, 5, 5] }) } }]) }]), max_weight: 1000000000, weight_credit: 1000000000
[2021-12-17T00:05:51Z TRACE xcm::execute] origin: Some(MultiLocation { parents: 0, interior: X1(AccountId32 { network: Named([87, 101, 115, 116, 101, 110, 100]), id: [4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4] }) }), total_surplus/refunded: 0/0, error_handler_weight: 0
[2021-12-17T00:05:51Z TRACE xcm::fungibles_adapter] transfer_asset what: MultiAsset { id: Concrete(MultiLocation { parents: 1, interior: Here }), fun: Fungible(1000000000000) }, from: MultiLocation { parents: 0, interior: X1(AccountId32 { network: Named([87, 101, 115, 116, 101, 110, 100]), id: [4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4] }) }, to: MultiLocation { parents: 0, interior: X1(Parachain(1000)) }
[2021-12-17T00:05:51Z TRACE xcm::execute_xcm_in_credit] result: Err(ExecutorError { index: 0, xcm_error: FailedToTransactAsset("AssetIdConversionFailed"), weight: 0 })
[2021-12-17T00:05:51Z DEBUG xcm::execute_xcm_in_credit] Execution errored at 0: FailedToTransactAsset("AssetIdConversionFailed") (original_origin: MultiLocation { parents: 0, interior: X1(AccountId32 { network: Named([87, 101, 115, 116, 101, 110, 100]), id: [4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4] }) })
[EventRecord { phase: Phase::Initialization, event: Event::PolkadotXcm(Event::Attempted(Incomplete(1000000000, FailedToTransactAsset("")))), topics: [] }]
test relaychain::statemine::statemine ... ok
```

Note the XCM execution failed with `FailedToTransactAsset("AssetIdConversionFailed")`

This is because 

https://github.com/paritytech/polkadot/blob/bff00af654dba2941af108c7816d45bda7c97cd8/xcm/xcm-executor/src/traits/transact_asset.rs#L92-L106

Is used to do the transfer, which calls

https://github.com/paritytech/polkadot/blob/bff00af654dba2941af108c7816d45bda7c97cd8/xcm/xcm-executor/src/traits/transact_asset.rs#L171-L190

on

https://github.com/paritytech/cumulus/blob/05cc5f0e2acacc18796f45ffa3c7b4626fd1046d/polkadot-parachains/statemine/src/lib.rs#L486

`pub type AssetTransactors = (CurrencyTransactor, FungiblesTransactor);`

It firstly tired the `transfer_asset` on `CurrencyAdapter` and goes to the default implementation

https://github.com/paritytech/polkadot/blob/bff00af654dba2941af108c7816d45bda7c97cd8/xcm/xcm-executor/src/traits/transact_asset.rs#L74-L76

and then try `transfer_asset` on `FungiblesTransactor` again and returns `MatchError::AccountIdConversionFailed`

https://github.com/paritytech/polkadot/blob/bff00af654dba2941af108c7816d45bda7c97cd8/xcm/xcm-builder/src/fungibles_adapter.rs#L121-L140

Now back to the `transfer_asset` of the tuple implementation, because the error is not `Err(XcmError::Unimplemented)` so it did not fallback to the withdraw / deposit and just failed.

One of the fix is implement `transfer_asset` on `CurrencyAdapter` (why it is not implemented on the first place?)

KiChjang:
Fixes #4548.

This makes it so that a tuple of `(CurrencyAdapter, FungiblesAdapter)` will attempt to call the `transfer_asset` method on `CurrencyAdapter` first and see if it fails, before moving onto `FungiblesAdapter`.

Without this change, the `transfer_asset` method on `CurrencyAdapter` uses the default implementation, which returns an `XcmError::Unimplemented` and is thus skipped, which may not be the desired behaviour.


xlc:
Should have some test to improve the poor coverage bit more.

xcm/xcm-builder/src/currency_adapter.rs
```diff
@@ -170,4 +170,15 @@ impl<
...
 		Ok(what.clone().into())
 	}
+
+	fn transfer_asset(
+		asset: &MultiAsset,
+		from: &MultiLocation,
+		to: &MultiLocation,
+	) -> result::Result<Assets, XcmError> {
+		log::trace!(target: "xcm::currency_adapter", "transfer_asset asset: {:?}, from: {:?}, to: {:?}", asset, from, to);
+		let assets = Self::withdraw_asset(asset, from)?;
```
xlc, line 180: I think it will be better to do `Currency::transfer`. It is slightly faster, and emit better event.

shawntabrizi (approved):
yeah, looks totally sensible

xcm/xcm-builder/src/currency_adapter.rs
```diff
@@ -170,4 +170,23 @@ impl<
...
 	}
+
+	fn transfer_asset(
+		asset: &MultiAsset,
+		from: &MultiLocation,
+		to: &MultiLocation,
+	) -> result::Result<Assets, XcmError> {
+		log::trace!(target: "xcm::currency_adapter", "transfer_asset asset: {:?}, from: {:?}, to: {:?}", asset, from, to);
+		let amount: u128 =
+			Matcher::matches_fungible(asset).ok_or(Error::AssetNotFound)?.saturated_into();
```
shawntabrizi, line 181: I might suggest we do a `try_into` rather than saturated into here since failing to convert a value is probably better than saturating here.

KiChjang: Huh interesting, that means I don't even need this intermediate `amount` variable at all, since it looks like the only purpose that it's there is to attempt and convert it to a `u128` and saturate it, before it gets converted back to `Balance` as seen in `balance_amount`.

shawntabrizi: Is there some other code which is making you think this might be the better way to go?

Do you generally agree with my assessment that an error here is nicer than some saturating behavior?

KiChjang: Yeah, the code in `deposit_asset` and `withdraw_asset` both use `saturated_into`, and if we are going to change here, then it makes sense to change it on both of those functions.

I don't really have an opinion on how it should work, but I'm assuming that there's a reason why the existing code chooses to saturate into a `u128` first and converting it back to a `Balance`.

zqhxuyuan:
what's the progress of this PR? I think this issue: https://github.com/paritytech/cumulus/issues/892 also has some related to this PR.

xcm/xcm-builder/src/currency_adapter.rs
gavofyork: I think you're right @KiChjang --- there's really no reason to convert from `Currency::Balance` to `u128` and then back again. It should just be:

```
let balance_amount = Matcher::matches_fungible(asset).ok_or(Error::AssetNotFound)?;
```

gavofyork (approved):

gavofyork (changes requested):
So this "issue" comes about due to the `transfer_asset`/`beam_asset` duality. Basically, `transfer_asset` wasn't really meant to be used directly - it's meant to implement a "correct" transfer - i.e. one which the backend actually implements as a transfer internally.

Instead, the API to be used was really `beam_asset`: this did a transfer by any means necessary - meaning that it would try to use `transfer_asset` (the "correct" way of doing it) and that was unimplemented then fall back on a `withdraw_asset` and `deposit_asset` combination. If `withdraw_asset` (or `transfer_asset`) returned an `AssetNotFound` error, then it would propagate that to the tuple instance which would allow it to move on to the following impls.

I suspect if we renamed `beam_asset` to `transfer_asset` and then `transfer_asset` to `native_transfer_asset`, we might end up fixing the problem without introducing unneeded implementations.

xlc:
maybe we want the rename, but I think we still want the native transfer asset implemented when it is possible as an optimization. it also generates a nicer event

KiChjang:
I agree that we should implement `native_transfer_asset` for the currency adapter -- using `Currency::transfer` instead of a `Currency::withdraw` and a `Currency::deposit` sounds like it is exactly a "correct" internal transfer method for the backend. I might even call it `internal_transfer_asset` instead of `native_transfer_asset` as I think it may reflect the intention better.

apopiak:
Just a quick note that my impression is that this blocks reserve transfers of KSM out of Statemint/e so we might want to resolve it.

gilescope:
Keith says this is ready for review again. It would be great to have this land, so that people could do limited teleport assets from one parachain to another.

gavofyork (approved):

gavofyork:
I don't know why I didn't implement the `transfer_native` function myself for `Currency` impls. That bothers me a little. But I don't see anything wrong with the logic.

KiChjang:
bot merge

PR #4561, opened 2021-12-18
Update Polkadot

Doordashcon:
This PR adds the constant value `EXISTENTIAL_DEPOSIT` to polkadot, kusama & westend runtime constants as needed to update cumulus.


runtime/westend/constants/src/lib.rs
```diff
@@ -20,6 +20,9 @@
 pub mod currency {
 	use primitives::v0::Balance;
 
+	/// The existential deposit. Set to 1/10 of its parent Relay Chain.
```
bkchr, line 23:
```suggestion
	/// The existential deposit. Set to 1/10 of its parent Relay Chain.
```

bkchr (approved):

bkchr (changes requested):
This value should also be used in the runtimes directly.

runtime/polkadot/constants/src/lib.rs
```diff
@@ -22,6 +22,9 @@ pub use self::currency::DOLLARS;
 pub mod currency {
 	use primitives::v0::Balance;
 
+	/// The existential deposit. Set to 1/10 of its parent Relay Chain (v9010).
+	pub const EXISTENTIAL_DEPOSIT: Balance = 10 * CENTS;
```
bkchr, line 26: These values are also not correct and the docs are also not correct.

Doordashcon:
> This value should also be used in the runtimes directly.

should have noticed that earlier, TY

runtime/polkadot/constants/src/lib.rs
Doordashcon: used the values from cumulus, updated now

bkchr (approved):

PR #4564, opened 2021-12-19
Create benchmarks for XCM instructions introduced in v3

KiChjang:


shawntabrizi:
@KiChjang sorry, the bot merged master here. It is not well suited to do benchmarking on a branch which targets something which is not master.

I think if you merge master on gavs branch, the diff will be fixed.

KiChjang:
We gotta redo the benchmark anyway since it was failing benchmark tests.

KiChjang:
> Starting benchmark for branch: kckyeung/xcm-v3-weights (vs master)

No wonder it's merging master... is there a command to tell it to run benchmarks against a different branch?

shawntabrizi:
> > Starting benchmark for branch: kckyeung/xcm-v3-weights (vs master)
> 
> No wonder it's merging master... is there a command to tell it to run benchmarks against a different branch?

unfortunately no, but a good backlog item i will add

xcm/src/v3/mod.rs
```diff
@@ -662,6 +660,7 @@ pub enum Instruction<Call> {
 	/// Kind: *Instruction*
 	///
 	/// Errors: *Fallible*.
+	#[weight_args()]
```
shawntabrizi, line 663: why do you want to bypass the weight args here?

xcm/src/v3/mod.rs
```diff
@@ -678,6 +677,7 @@ pub enum Instruction<Call> {
 	///
 	/// Errors:
 	/// - `ExpectationFalse`: In case any of the expectations are broken.
+	#[weight_args(index: &u32)]
```
shawntabrizi, line 680: And why modify them here?

xcm/src/v3/mod.rs
KiChjang: Because the weight function for this instruction doesn't require any arguments, so it's left empty.

xcm/src/v3/mod.rs
KiChjang: The weight function for this instruction only takes in an `index: &u32` and not all of the fields in the variant, so it's modified.

xcm/pallet-xcm-benchmarks/src/generic/benchmarking.rs
```diff
@@ -306,13 +305,141 @@ benchmarks! {
...
+
+		let instruction = Instruction::ExpectAsset(assets.into());
+		let xcm = Xcm(vec![instruction]);
+	}: {
+		executor.execute(xcm)?;
+	} verify {
+		// `execute` completing successfully is as good as we can check.
+	}
+
+	expect_origin {
```
shawntabrizi, line 345: A simple comment here that worst case scenario is when the expected origin call fails.

xcm/pallet-xcm-benchmarks/src/generic/benchmarking.rs
```diff
@@ -306,13 +305,141 @@ benchmarks! {
...
+
+	expect_pallet {
+		let mut executor = new_executor::<T>(Default::default());
+
+		let instruction = Instruction::ExpectPallet {
+			index: 0,
+			name: b"System".to_vec(),
+			module_name: b"frame_system".to_vec(),
+			crate_major: 4,
+			min_crate_minor: 0,
```
shawntabrizi, line 402: is there a way to make this test more adaptable?

For example, you will easily fail this benchmark when the crate version increments. Better to query some value you know will succeed here.

xcm/pallet-xcm-benchmarks/src/generic/benchmarking.rs
```diff
@@ -306,13 +305,141 @@ benchmarks! {
...
+	clear_transact_status {
+		let mut executor = new_executor::<T>(Default::default());
+		executor.transact_status = MaybeErrorCode::Error(b"MyError".to_vec());
+
+		let instruction = Instruction::ClearTransactStatus;
+		let xcm = Xcm(vec![instruction]);
+	}: {
+		executor.execute(xcm)?;
+	} verify {
+		assert_eq!(executor.transact_status, MaybeErrorCode::Success);
```
shawntabrizi, line 440: Not a note for this PR, but it is strange to me the transaction status defaults to success versus some kind of "unreported'

shawntabrizi (approved):
some comments, but should be good

xcm/pallet-xcm-benchmarks/src/generic/benchmarking.rs
```diff
@@ -306,13 +305,141 @@ benchmarks! {
...
+	}: {
+		_result = executor.execute(xcm);
+	} verify {
+		assert!(matches!(_result, Err(ExecutorError {
+			xcm_error: XcmError::ExpectationFalse,
+			..
+		})));
+	}
+
+	expect_error {
```
kianenigma, line 361: why is the failure case more expensive than the success?

xcm/pallet-xcm-benchmarks/src/generic/benchmarking.rs
```diff
@@ -306,13 +305,141 @@ benchmarks! {
...
+	}
+
+	query_pallet {
+		let query_id = Default::default();
+		let destination = T::valid_destination().map_err(|_| BenchmarkError::Skip)?;
+		let max_weight = Default::default();
+		let mut executor = new_executor::<T>(Default::default());
+
+		let instruction = Instruction::QueryPallet {
+			module_name: b"frame_system".to_vec(),
```
kianenigma, line 384:
```suggestion
			pallet_name: b"frame_system".to_vec(),
```

?

kianenigma (commented):
Not possible for me to be 100% sure if every benchmark is the absolute worse case, everything else looks in order, just small questions/comments. Can treat as an approve since the base branch is not master.

xcm/pallet-xcm-benchmarks/src/generic/benchmarking.rs
KiChjang: It actually isn't; I'm using the failure case here so that I could verify the result, because the success case looks too much like any other successes and you can't really be sure if it worked correctly.

xcm/pallet-xcm-benchmarks/src/generic/benchmarking.rs
KiChjang: Ack, yeah this is an unfortunate naming convention that I settled to -- "module" here means "Rust module" and not "pallet", since it's directly taken from the path to the pallet in `construct_runtime`.

xcm/pallet-xcm-benchmarks/src/generic/benchmarking.rs
KiChjang: I don't think it actually is -- I'm using the failure case here because it's easier to verify that it worked than the success case, since the success case looks practically identical to any other success case.

xcm/pallet-xcm-benchmarks/src/generic/benchmarking.rs
KiChjang: Well hmm, I guess I could create a dummy pallet, import it into the runtime and give it a fixed index location.

xcm/src/v3/mod.rs
shawntabrizi: I don't think this is the right pattern. The weight function should have all parameters matching the instruction. If they are not used, they are just not used, but the instruction logic could potentially change in the future, using one of these parameters, and we dont want to change the function signature.

xcm/src/v3/mod.rs
shawntabrizi: same as above

xcm/pallet-xcm-benchmarks/src/generic/benchmarking.rs
shawntabrizi: Isn't success here means that the expected origin matches? so literally only one input could give a success output?

xcm/pallet-xcm-benchmarks/src/generic/benchmarking.rs
shawntabrizi: would also need to be generic for any runtime we benchmark this on too, so making a custom pallet is not a good choice.

xcm/pallet-xcm-benchmarks/src/generic/benchmarking.rs
KiChjang: Yes, but I can't do a match against the result. If we are to treat the entire XCM executor as a blackbox, then the success case is not distinguishable from the "do nothing" case, since both of them emit no observable result, and hence I've chosen the failure case.

xcm/pallet-xcm-benchmarks/src/generic/benchmarking.rs
kianenigma: hmm but if the success case is is heavier, this leads to inaccurate weight (I guess this is not the case here).

xcm/pallet-xcm-benchmarks/src/generic/benchmarking.rs
kianenigma: fair enough

PR #4575, opened 2021-12-21
paras: add governance control dispatchables

pepyakin:
Part of #3211 

Adds a couple of functions for governance control for the paras module
in the anticipation of PVF pre-checking enabling.

Specifically, this commit adds a function for pre-registering a PVF that
governance trusts enough. This function will come in handy in case there
is a parachain that does not follow the GoAhead signal. That is, does
not include https://github.com/paritytech/cumulus/pull/517.

This may be not an exhaustive list of the functions that may come in
handy. Any suggestions to add more are welcome.

cc @bkchr @shawntabrizi 

TODO:

- [x] kusama benchmarks
- [x] polkadot benchmarks
- [x] rococo benchmarks
- [x] westend


pepyakin:
* master
  * **PR #4575**  (this PR)

runtime/parachains/src/paras.rs
```diff
@@ -1726,7 +1801,10 @@ impl<T: Config> Pallet<T> {
 	fn decrease_code_ref(code_hash: &ValidationCodeHash) -> Weight {
 		let mut weight = T::DbWeight::get().reads(1);
 		let refs = <Self as Store>::CodeByHashRefs::get(code_hash);
-		debug_assert!(refs != 0);
```
pepyakin, line 1729: This is removed just to pass an extra paranoid test: `increase_code_ref_doesnt_have_allergy_on_add_trusted_validation_code`

runtime/parachains/src/paras.rs
```diff
@@ -677,6 +677,79 @@ pub mod pallet {
...
+		///
+		/// Otherwise, the code will be added into the storage. Note that the code will be added
+		/// into storage with reference count 0. This is to account the fact that there are no users
+		/// for this code yet. The caller will have to make sure that this code eventually gets
+		/// used by some parachain or removed from the storage to avoid storage leaks. For the latter
+		/// prefer to use the `poke_unused_validation_code` dispatchable to raw storage manipulation.
+		///
+		/// This function is mainly meant to be used for upgrading parachains that do not follow
+		/// the go-ahead signal while the PVF pre-checking feature is enabled.
+		#[pallet::weight(0)] // TODO:
```
pepyakin, line 693: We still need to run benchmarks for this.

runtime/parachains/src/paras.rs
```diff
@@ -763,6 +771,79 @@ pub mod pallet {
...
+				// There is no vote, but the code exists. Nothing to do here.
+				return Ok(())
+			}
+
+			// At this point the code is unknown and there is no PVF pre-checking vote for it, so we
+			// can just add the code into the storage.
+			//
+			// NOTE That we do not use `increase_code_ref` here, because the code is not yet used
+			// by any parachain.
+			<Self as Store>::CodeByHash::insert(code_hash, &validation_code);
```
eskimor, line 825: So in this case, we completely ignore the onboarding/upgrade machinery? I guess that's fine/intended, but I would like to understand it better: Like how does the actual upgrade work then for example? Code is already on the relay chain, so what steps must a parachain do now to upgrade in this state?

pepyakin: A parachain would attempt an upgrade. When that candidate is enacted, the inclusion module would call [`schedule_code_upgrade`](https://github.com/paritytech/polkadot/blob/3599d0a3cdd5755b0c9d4750060679e75ff30aa7/runtime/parachains/src/paras.rs#L1523) and then control will fall into `kick_off_pvf_check` and this [branch](https://github.com/paritytech/polkadot/blob/3599d0a3cdd5755b0c9d4750060679e75ff30aa7/runtime/parachains/src/paras.rs#L1615) will be taken.

From the parachain's standpoint, it will look like instant PVF approval or equally as good old pre-go-ahead signal upgrade.

Does this make sense?

eskimor: yep it does - thanks! So a parachain would still need to get it's full runtime upgrade through the backing phase, although the code is already on chain. We just bypass the voting, because we trust governance to have the code checked by some other means. :bulb:

eskimor (approved):

pepyakin:

PR #4580, opened 2021-12-22
configuration: Consistency checks for PVF pre-checking

pepyakin:
As was suggested by Alexander Popiak [here][comment], this commit
checks the consistency of the configuration.

[comment]:
https://github.com/paritytech/polkadot/pull/4420#discussion_r764796519

PR #4583, opened 2021-12-22
configuration: validation_upgrade_delay consistency

### LINKED ISSUE #4248: Go-ahead signal makes validation upgrade delay of less than 2 invalid

pepyakin:
Closes #4248

Impose additional constraint on configuration consistency:
`validation_upgrade_delay` should not be less than or equal to 1.

See the original issue for more details.

TODO:

- [x] fix the default chainspec

### DIFF
--- a/runtime/parachains/src/configuration.rs
+++ b/runtime/parachains/src/configuration.rs
@@ -219,7 +219,7 @@ impl<BlockNumber: Default + From<u32>> Default for HostConfiguration<BlockNumber
 			thread_availability_period: 1u32.into(),
 			no_show_slots: 1u32.into(),
 			validation_upgrade_frequency: Default::default(),
-			validation_upgrade_delay: Default::default(),
+			validation_upgrade_delay: 2u32.into(),
 			code_retention_period: Default::default(),
 			max_code_size: Default::default(),
 			max_pov_size: Default::default(),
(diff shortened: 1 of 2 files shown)


pepyakin:
* master
  * **PR #4569** 
    * **PR #4581** 
      * **PR #4580** 
        * **PR #4583**  (this PR)

ordian (approved):

bkchr:
Did you tested this with a lag in approval voting? The delay is calculated from the block the pov was build on and not the one it was enacted?

pepyakin:
For the record, @bkchr meant problem described in this PR https://github.com/paritytech/polkadot/pull/4246. That problem is solved by a previous PR in the stack: https://github.com/paritytech/polkadot/pull/4580. 

It puts a constraint that `minimum_validation_upgrade_delay` [^1] must be greater than both `chain_availability_period` and `thread_availability_period`.

[^1]: the number of blocks that must pass from enacting a candidate that signals an upgrade to the relay chain block number which will be the first relay-parent when the code upgrade is applied.

PR #4596, opened 2021-12-23
availability-distribution: look for leaf ancestors within the same session

### LINKED ISSUE #2513: Availability Distribution: Consider occupied cores from last k blocks within same session

slumber:
Restores the behavior that was removed in https://github.com/paritytech/polkadot/pull/2423
Resolves #2513 

- [x] tests

### DIFF
--- a/node/network/availability-distribution/src/requester/mod.rs
+++ b/node/network/availability-distribution/src/requester/mod.rs
@@ -27,18 +27,23 @@ use std::{
 };
 
 use futures::{
-	channel::mpsc,
+	channel::{mpsc, oneshot},
 	task::{Context, Poll},
 	Stream,
 };
 
-use polkadot_node_subsystem_util::runtime::{get_occupied_cores, RuntimeInfo};
-use polkadot_primitives::v1::{CandidateHash, Hash, OccupiedCore};
+use polkadot_node_subsystem_util::{
+	request_session_index_for_child,
+	runtime::{self, get_occupied_cores, RuntimeInfo},
+};
+use polkadot_primitives::v1::{CandidateHash, Hash, OccupiedCore, SessionIndex};
 use polkadot_subsystem::{
-	messages::AllMessages, ActivatedLeaf, ActiveLeavesUpdate, LeafStatus, SubsystemContext,
+	messages::{AllMessages, ChainApiMessage},
+	ActivatedLeaf, ActiveLeavesUpdate, LeafStatus, SubsystemContext,
 };
 
-use super::{Metrics, LOG_TARGET};
+use super::{Metrics, Result, LOG_TARGET};
+use crate::error::Fatal;
 
 /// Cache for session information.
 mod session_cache;
@@ -64,6 +69,13 @@ pub struct Requester {
 	/// Localized information about sessions we are currently interested in.
 	session_cache: SessionCache,
 
+	/// Chain of up to [`LEAF_ANCESTRY_LEN_WITHIN_SESSION`] ancestors of activated leaves.
+	/// leaf -> [leaf parent, leaf grandparent, ...]
+	/// Fetch tasks also get launched for array `relay_ancestors[leaf]`, this helps with
+	/// possibly missing activated leaves updates, as well as with "slow" validators which do not
+	/// manage to fetch their chunks in time.
+	relay_ancestors: HashMap<Hash, Vec<Hash>>,
+
 	/// Sender to be cloned for `FetchTask`s.
 	tx: mpsc::Sender<FromFetchTask>,
 
@@ -75,14 +87,25 @@ pub struct Requester {
 }
 
 impl Requester {
+	/// How many ancestors of the leaf should we consider along with it.
+	pub(crate) const LEAF_ANCESTRY_LEN_WITHIN_SESSION: usize = 3;
+
 	/// Create a new `Requester`.
 	///
 	/// You must feed it with `ActiveLeavesUpdate` via `update_fetching_heads` and make it progress
 	/// by advancing the stream.
 	pub fn new(metrics: Metrics) -> Self {
 		let (tx, rx) = mpsc::channel(1);
-		Requester { fetches: HashMap::new(), session_cache: SessionCache::new(), tx, rx, metrics }
+		Requester {
+			fetches: HashMap::new(),
+			session_cache: SessionCache::new(),
+			relay_ancestors: HashMap::new(),
+			tx,
+			rx,
+			metrics,
+		}
 	}
+
 	/// Update heads that need availability distribution.
 	///
 	/// For all active heads we will be fetching our chunks for availability distribution.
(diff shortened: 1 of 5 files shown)


node/network/availability-distribution/src/requester/mod.rs
```diff
@@ -215,3 +263,83 @@ impl Stream for Requester {
...
+		response_channel: tx,
+	})
+	.await;
+
+	let ancestors = rx.await.map_err(Fatal::ChainApiSenderDropped)?.map_err(Fatal::ChainApi)?;
+	Ok(ancestors)
+}
+
+/// Request session index for the child of the relay parent from the Runtime API.
+async fn get_session_index_for_child<Context>(
```
eskimor, line 331: See https://github.com/paritytech/polkadot/blob/468a75ed56852afd7d5c161407b3b70d342da64d/node/subsystem-util/src/runtime/mod.rs#L118

node/network/availability-distribution/src/requester/mod.rs
```diff
@@ -91,50 +114,75 @@ impl Requester {
...
+		self.relay_ancestors.insert(leaf, ancestors_in_session);
+
 		Ok(())
 	}
 
 	/// Stop requesting chunks for obsolete heads.
 	///
 	fn stop_requesting_chunks(&mut self, obsolete_leaves: impl Iterator<Item = Hash>) {
-		let obsolete_leaves: HashSet<_> = obsolete_leaves.collect();
+		let obsolete_leaves: HashSet<_> = obsolete_leaves
```
eskimor, line 180: I don't see how this can work. If we always remove the ancestors as well, then this will have the same effect as not taking the ancestors into account at all. For each obsolete leaf we also remove all its ancestors - why add them in the first place?

slumber: That's true, thanks
Should we instead schedule it? Like keep an LRU of fetch tasks for ancestors?

eskimor: If we only add the actually activated head to [`FetchTask::live_in`](https://github.com/paritytech/polkadot/blob/a6e1f37694b2b46b7132216181af4568017a23ca/node/network/availability-distribution/src/requester/fetch_task/mod.rs#L67), even if it was actually spawned for an ancestor - then we should not need to keep track of ancestors separately. (We should change docs accordingly)

For lengthening the fetch time a bit, the cleanup could become a two phase process:

- Before updating `FetchTask::live_in` for the tasks, we find those that already have an empty `live_in` set. We remove it if the block number of the currently deactivated leaf differed from the block number of the last deactivated leaf.*)
- Then we remove any deactivated heads in `FetchTask::live_in`, but don't yet remove the task.


We can keep track of block numbers for all active leaves and always store the block number of the last deactivated.

This should lengthen the fetching phase for another slot (most of the time) and should be robust even in the event of chain reversions.

Note: While we will only use activated heads in `live_in` we do need to use the correct relay parent  for [`relay_parent`](https://github.com/paritytech/polkadot/blob/a6e1f37694b2b46b7132216181af4568017a23ca/node/network/availability-distribution/src/requester/fetch_task/mod.rs#L121)

*) That's a simple heuristic to make sure we are really lengthening the time - it is not 100% but very close and keeps the code a little simpler (especially with chain reversions).

eskimor (commented):
Looks good overall! 

But I noticed a bug, that is not related to your changes but would be nice to be fixed in the course of this PR:

[Here](https://github.com/paritytech/polkadot/blob/b01fc015b2dae448d8b7ed8d7a928d0bf393297e/node/network/availability-distribution/src/requester/mod.rs#L216) we are not actually using the `SessionInfo` for the leaf, but the `SessionInfo` for the leaf's child. Therefore, at session boundaries we will look up the backing group in the wrong session.

I'll tell @sandreim  - would be good to have this fixed in the course of his PR.

slumber:
> Looks good overall!
> 
> But I noticed a bug, that is not related to your changes but would be nice to be fixed in the course of this PR:
> 
> [Here](https://github.com/paritytech/polkadot/blob/b01fc015b2dae448d8b7ed8d7a928d0bf393297e/node/network/availability-distribution/src/requester/mod.rs#L216) we are not actually using the `SessionInfo` for the leaf, but the `SessionInfo` for the leaf's child. Therefore, at session boundaries we will look up the backing group in the wrong session.
> 
> I'll tell @sandreim - would be good to have this fixed in the course of his PR.

Sure, will take care of it

eskimor:
Some tests would be nice - especially for behavior at session boundaries, other than that - good to go.

Lldenaurois (commented):
I still need to review the tests, but the code itself lgtm

node/network/availability-distribution/src/requester/mod.rs
```diff
@@ -91,43 +100,72 @@ impl Requester {
...
-		let activated = activated.and_then(|h| match h.status {
-			LeafStatus::Stale => None,
-			LeafStatus::Fresh => Some(h),
-		});
 		// Order important! We need to handle activated, prior to deactivated, otherwise we might
 		// cancel still needed jobs.
-		self.start_requesting_chunks(ctx, runtime, activated.into_iter()).await?;
+		if let Some(activated) = activated {
+			// Stale leaves happen after a reversion - we don't want to re-run availability there.
+			if let LeafStatus::Fresh = activated.status {
```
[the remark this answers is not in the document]
slumber: You can also express it with
```rust
if matches!(activated, Some(activated) if activated.status == LeafStatus::Fresh) {
```
(provided one derives the PartialEq for status)
But to be honest nested `if let` is the most readable option to me, 2 more lines is not a big deal

node/network/availability-distribution/src/requester/mod.rs
```diff
@@ -215,3 +253,69 @@ impl Stream for Requester {
 		}
 	}
 }
+
+/// Requests up to `limit` ancestor hashes of relay parent in the same session.
+async fn get_block_ancestors_in_same_session<Context>(
```
rphmeier, line 258: Logic here seems accurate accounting for the fact that `get_session_index` is actually returning the session index of the child. Nice

node/network/availability-distribution/src/requester/mod.rs
```diff
@@ -91,43 +100,72 @@ impl Requester {
...
+		// Also spawn or bump tasks for candidates in ancestry in the same session.
+		for hash in std::iter::once(leaf).chain(ancestors_in_session) {
+			let cores = get_occupied_cores(ctx, hash).await?;
 			tracing::trace!(
 				target: LOG_TARGET,
 				occupied_cores = ?cores,
 				"Query occupied core"
 			);
+			// Important:
+			// We mark the whole ancestry as live in the **leaf** hash, so we don't need to track
```
rphmeier, line 160: This comment is great and seems to be correct behavior. I had noticed that `add_cores` might race with `remove_leaf` if you used `hash`. 

It would be even more clear if you wrote "We invoke add_cores with the `leaf` and not with the ancestor `hash` to avoid a race condition" or something like that.

rphmeier (approved):
Looks good. Very much up to standard of neatness, correctness, and testing. Thank you.

Issue #4544, opened 2021-12-16
Runtime metrics boilerplate


sandreim:
This work was deferred from https://github.com/paritytech/polkadot/pull/4483. The main issue is that the runtime metrics registration happens on each runtime call. On top of that, incrementing metrics is not ergonomic because it requires either recreating the runtime metric object every time or passing it around. 

One solution is to move metric registration logic to node side and register all runtime metrics early. This also reduces the amount of data the runtime will send via tracing for updating metrics.

PR #4602, opened 2021-12-24
Parachain runtime metrics followup

sandreim:
fixes https://github.com/paritytech/polkadot/issues/4544

<s>WIP</s>


runtime/parachains/src/metrics.rs
```diff
@@ -0,0 +1,80 @@
...
+use polkadot_runtime_metrics::{Counter, CounterVec};
+
+pub use pallet::*;
+
+#[frame_support::pallet]
+pub mod pallet {
+	use super::*;
+
+	#[pallet::pallet]
+	pub struct Pallet<T>(_);
```
bkchr, line 28: Why is this a pallet?

sandreim: At the time, it made sense because I wanted to follow the pattern in the parachains runtime (runtime lego pieces). Now that you asked this question, I think it actually doesn't fit the pallet definition since this one provides no runtime logic at all. I'll get rid of the pallet.

node/metrics/src/runtime/parachain.rs
```diff
@@ -0,0 +1,57 @@
...
+
+//! Client side declaration and registration of the parachain Prometheus metrics.
+//! All of the metrics have a correspondent runtime metric definition.
+
+use crate::runtime::RuntimeMetricsProvider;
+
+/// Register the parachain runtime metrics.
+pub fn register_metrics(runtime_metrics_provider: &RuntimeMetricsProvider) {
+	runtime_metrics_provider.register_countervec(
+		"polkadot_parachain_inherent_data_weight",
```
bkchr, line 25: The names should be constant in runtime/metrics.rs file. Then you can share them and also not silently break them :D

sandreim: Thanks! Good point.

drahnr:
Generally looks good, I think we'd still want another level of abstraction, where we have specific `fn on_invalid_candidate` / `metrics.on_invalid_candidate` as we have on the node side. This would considerably improve readability since a 3 line inc could be replaced by a single line and the metrics become a bit less interleaved with the actual code

primitives/src/v1/metrics.rs
```diff
@@ -137,3 +106,74 @@ impl From<&'static str> for RuntimeMetricLabel {
...
+	/// Counts the number of dispute statements sets included in a block in `enter_inner`.
+	pub const PARACHAIN_INHERENT_DATA_DISPUTE_SETS_INCLUDED: CounterDefinition =
+		CounterDefinition {
+			name: "polkadot_parachain_inherent_data_dispute_sets_included",
+			description:
+				"Counts the number of dispute statements sets included in a block in `enter_inner`.",
+		};
+
+	/// Counts the number of `valid` and `invalid` bitfields signature checked in `enter_inner`.
+	pub const PARACHAIN_CREATE_INHERENT_BITFIELDS_SIGNATURE_CHECKS: CounterVecDefinition =
```
drahnr, line 173: :+1: I like this

runtime/parachains/src/metrics.rs
```diff
@@ -0,0 +1,60 @@
...
+/// `enter_inner`.
+pub fn disputes_included() -> Counter {
+	Counter::new(PARACHAIN_INHERENT_DATA_DISPUTE_SETS_INCLUDED)
+}
+
+/// Returns a runtime metric instance for counting bitfield signature checks in
+/// `enter_inner`.
+pub fn bitfields_signature_checks() -> CounterVec {
+	CounterVec::new(PARACHAIN_CREATE_INHERENT_BITFIELDS_SIGNATURE_CHECKS)
+}
```
drahnr, line 60: Allover I think I'd prefer a similar pattern to the node side, with `metrics.on_bitfield_signature_verified()`  where the logic (counter inc, dec, the label) is hidden from the user. We could create a dummy `Metrics` type. But I am open to suggestions.

primitives/src/v1/metrics.rs
bkchr: Yeah looks really good!

runtime/parachains/src/metrics.rs
sandreim: Yup, that makes a lot of sense. I am concerned that we need to always construct the metrics struct  to just update one metric which  is a lot of overhead. I am not sure if the compiler will optimize to prevent this.

node/metrics/src/runtime/mod.rs
```diff
@@ -20,11 +20,12 @@
 //! tracing support. This requires that the custom profiler (`TraceHandler`) to be
 //! registered in substrate via a `logger_hook()`. Events emitted from runtime are
 //! then captured/processed by the `TraceHandler` implementation.
+
 #![cfg(feature = "runtime-metrics")]
 
 use codec::Decode;
 use primitives::v1::{
-	RuntimeMetricLabelValues, RuntimeMetricOp, RuntimeMetricRegisterParams, RuntimeMetricUpdate,
+	metric_definitions::*, RuntimeMetricLabelValues, RuntimeMetricOp, RuntimeMetricUpdate,
```
drahnr, line 28: We try to avoid `*` as much as possible.

runtime/parachains/src/metrics.rs
drahnr: Technically, the `struct Metrics;` can be a marker type, and then it should end up in the same instructions :crossed_fingers:

drahnr (commented):
one nit

runtime/parachains/src/metrics.rs
sandreim: I've take a different approach and made the runtime Metrics object holding all counters `const`.

drahnr (approved):
Very nice!

runtime/metrics/src/with_runtime_metrics.rs
```diff
@@ -49,68 +52,57 @@ trait MetricEmitter {
...
-			metric_name: Vec::from(name),
-			op: RuntimeMetricOp::Register(RuntimeMetricRegisterParams::new(
-				Vec::from(description),
-				None,
-			)),
-		};
-
-		Self::emit(&metric_update);
-		Counter { name }
+	/// Create a new counter metric with specified `name`. This metric needs to be registered
```
sandreim, line 98: Needs update.

runtime/metrics/src/with_runtime_metrics.rs
```diff
@@ -49,68 +52,57 @@ trait MetricEmitter {
...
-				op: RuntimeMetricOp::IncrementCounterVec(value, label_values),
-			};
-
-			Self::emit(&metric_update);
-		});
+impl MetricEmitter for LabeledMetric {}
+impl MetricEmitter for Counter {}
+
+impl CounterVec {
+	/// Create a new counter metric with specified `name`. This metric needs to be registered
```
sandreim, line 82: needs update

primitives/src/v1/metrics.rs
```diff
@@ -137,3 +106,74 @@ impl From<&'static str> for RuntimeMetricLabel {
...
+		pub description: &'static str,
+	}
+
+	/// `CounterVec` metric definition.
+	pub struct CounterVecDefinition<'a> {
+		/// The name of the metric.
+		pub name: &'static str,
+		/// The description of the metric.
+		pub description: &'static str,
+		/// The labels names of the metric.
```
bkchr, line 126:
```suggestion
		/// The label names of the metric.
```

bkchr (approved):

PR #4603, opened 2021-12-24
paras: fix upgrade restriction signal

### LINKED ISSUE #3971: `UpgradeRestrictionSignal` is one block off
Assume `UpgradeRestrictionSignal` is set. If `UpgradeRestrictionSignal` is unset at the relay-block `B`, then a parablock that signals an upgrade built with the relay-parent `B-1` can be accepted. However, the post-state at `B-1` forbids an upgrade since `UpgradeRestrictionSignal` is set. That means a non-🤠 parachain will not risk upgrading until it builds a parablock at `B` where the upgrade restriction is lifted.

A potential solution is to unset `UpgradeRestrictionSignal` at `on_finalize` not in `on_initialize`.

pepyakin:
Closes #3971

Read the linked issue.

Apart from that, this addresses the concern raised in this [comment] by
just adding a test. I couldn't find a clean way to reconcile a block
number delay with a PVF voting TTL, so I just resorted to rely on the
test. Should be fine for now.

[comment]: https://github.com/paritytech/polkadot/pull/4457#discussion_r770517562

### DIFF
--- a/runtime/parachains/src/paras.rs
+++ b/runtime/parachains/src/paras.rs
@@ -1004,7 +1004,9 @@ impl<T: Config> Pallet<T> {
 	}
 
 	/// Called by the initializer to finalize the configuration pallet.
-	pub(crate) fn initializer_finalize() {}
+	pub(crate) fn initializer_finalize(now: T::BlockNumber) {
+		Self::process_scheduled_upgrade_cooldowns(now);
+	}
 
 	/// Called by the initializer to note that a new session has started.
 	///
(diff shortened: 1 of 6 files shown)


pepyakin:
* master
  * **PR #4581** 
    * **PR #4580** 
      * **PR #4583** 
        * **PR #4603**  (this PR)

eskimor (approved):
Honestly I don't see an issue with https://github.com/paritytech/polkadot/issues/3971 ... but I also don't know how we handle such things in general. I assume this is to make things consistent with other code?

runtime/parachains/src/configuration.rs
```diff
@@ -73,12 +73,26 @@ pub struct HostConfiguration<BlockNumber> {
...
+	///
 	/// If PVF pre-checking is enabled this should be greater than the maximum number of blocks
 	/// PVF pre-checking can take. Intuitively, this number should be greater than the duration
 	/// specified by [`pvf_voting_ttl`]. Unlike, [`pvf_voting_ttl`], this parameter uses blocks
 	/// as a unit.
 	pub validation_upgrade_frequency: BlockNumber,
 	/// The delay, in blocks, before a validation upgrade is applied.
+	///
+	/// The delay is counted from the relay-parent of the candidate that signalled the upgrade. The
+	/// sum of that block and the this delay is dubbed `expected_at`. The first parachain block with
```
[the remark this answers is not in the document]
pepyakin: Oops. I removed this comment and the comments will be updated with another PR )

runtime/parachains/src/paras.rs
```diff
@@ -1004,7 +1004,9 @@ impl<T: Config> Pallet<T> {
 	}
 
 	/// Called by the initializer to finalize the configuration pallet.
```
bkchr, line 1006: Not your fault, but the comment is c&p?

bkchr (approved):

runtime/parachains/src/paras.rs
pepyakin: yeah, but it does look correct, or?

bkchr: Is this the configuration pallet?

pepyakin: ah gotcha xD will fix

pepyakin: https://github.com/paritytech/polkadot/pull/4646

pepyakin:
It is not critical by any means, but it is an issue, I believe. Fixing it will actually simplify the behavior from my perspective (at the cost of small complication because of having to do logic in on_finalize) and that will less likely shoot us in the foot down the road

pepyakin:
bot merge

Issue #4426, opened 2021-12-01
XCM Generics Benchmarks Leftovers (TODO)


shawntabrizi:
- [ ] Is there a better way to verify that `tranasct` XCM completed successfully? Potentially set some unique storage via the extrinsic and check that storage? (without root?)

```
} verify {
	// TODO make better assertion?
	let num_events2 = frame_system::Pallet::<T>::events().len();
	assert_eq!(num_events + 1, num_events2);
}
```

- [ ]

PR #4605, opened 2021-12-27
Allow verification of sent XCMs in benchmarks

KiChjang:
Partial fix for #4426.

Creates an additional `TestSendXcm` struct in mocks, which is essentially a global queue of XCMs that are meant to be sent by the `XcmSender` as configured in the XCM executor. This will allow us to inspect the outgoing messages by reading off of the queue, and thereby verifying the integrity of our XCM benchmarks.


xcm/pallet-xcm-benchmarks/src/lib.rs
```diff
@@ -64,7 +62,7 @@ pub fn mock_worst_case_holding() -> MultiAssets {
 	const HOLDING_FUNGIBLES: u32 = 99;
 	const HOLDING_NON_FUNGIBLES: u32 = 99;
 	let fungibles_amount: u128 = 100;
-	(0..HOLDING_FUNGIBLES)
+	(1..=HOLDING_FUNGIBLES)
```
KiChjang, line 65: This is done here because the XCM executor automatically removes any fungible assets that have an amount of 0, which would cause some tests to fail as the worst case holding expects a non-existent `GeneralIndex(0)` asset to be there.

shawntabrizi:
This unfortunately will not work right?

Because when running verification on an actual production chain, this mock thing wont exist. So this only works for tests, but verification in benchmarking is quite a bit more than that.

We need some general trait / implementation for both tests and the real queue system.

KiChjang:
Yeah, this is definitely not going to work, closing for now.

PR #4613, opened 2021-12-27
Fix flaky test in polkadot-node-metrics 

### LINKED ISSUE #4609: tests::runtime_can_publish_metrics in polkadot-node-metrics is flakey
I've noticed a couple of times already that the test mentioned in subject fails randomly in PRs with unrelated changes. Here is the last instance of it

https://gitlab.parity.io/parity/polkadot/-/jobs/1308146#L1044

cc @sandreim

sandreim:
Fixes #4609

### DIFF
--- a/node/metrics/src/tests.rs
+++ b/node/metrics/src/tests.rs
@@ -43,7 +43,7 @@ fn runtime_can_publish_metrics() {
 		.expect("failed to start the node process");
 
 	// Enough time to author one block.
-	thread::sleep(Duration::from_secs(10));
+	thread::sleep(Duration::from_secs(30));
 
 	let runtime = tokio::runtime::Runtime::new().expect("failed to create tokio runtime");
 


pepyakin (approved):
Let's see if it works better!

bkchr:
Sleeps in tests are a really really bad idea. You should only use sleep if there is no other way. Here is the proper solution to check that the node has produced a block, this can be done using jsonrpc. You could also use the test service to spawn you a node that makes it easier to check if a block was build.

sandreim:
> Sleeps in tests are a really really bad idea. You should only use sleep if there is no other way. Here is the proper solution to check that the node has produced a block, this can be done using jsonrpc. You could also use the test service to spawn you a node that makes it easier to check if a block was build.

I agree this is bad practice in general.  Thanks for pointing this out.  I'll look into the proper alternatives you mentioned and fix this.

sandreim:
https://github.com/paritytech/polkadot/issues/4622

bkchr:
Ty 👍

PR #4618, opened 2021-12-28
Use derive macro to generate XcmWeightInfo trait

KiChjang:
Per title.

One thing to note is that the derive macro is named `XcmWeightInfoTrait`, as it generates code for a *trait*, and not the *implementation of a trait* as is common with most derive macros. Such a naming scheme would allow us to then create a `XcmWeightInfo` derive macro in the future that would properly generate the implementation of `XcmWeightInfo`, if the need arises.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -11854,6 +11854,7 @@ dependencies = [
 name = "xcm-procedural"
 version = "0.1.0"
 dependencies = [
+ "Inflector",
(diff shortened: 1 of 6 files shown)


bkchr (approved):

shawntabrizi (approved):
thanks a bunch!

KiChjang:
bot merge

PR #4625, opened 2021-12-28
paras: fix placeholder weight

pepyakin:
Setting zero as weight may be a source of problems.

The problem is, a rogue validator can shove a lot of duplicated votes
and thus fill a block with work that may well exceed the weight limit.

### DIFF
--- a/runtime/parachains/src/paras.rs
+++ b/runtime/parachains/src/paras.rs
@@ -765,7 +765,7 @@ pub mod pallet {
 
 		/// Includes a statement for a PVF pre-checking vote. Potentially, finalizes the vote and
 		/// enacts the results if that was the last vote before achieving the supermajority.
-		#[pallet::weight(0)]
+		#[pallet::weight(Weight::MAX)]
 		pub fn include_pvf_check_statement(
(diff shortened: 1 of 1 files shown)


pepyakin:
* master
  * **PR #4625**  (this PR)

runtime/parachains/src/paras.rs
```diff
@@ -765,7 +765,7 @@ pub mod pallet {
 
 		/// Includes a statement for a PVF pre-checking vote. Potentially, finalizes the vote and
 		/// enacts the results if that was the last vote before achieving the supermajority.
-		#[pallet::weight(0)]
+		#[pallet::weight(Weight::MAX)]
```
sandreim, line 768: Should we also create an issue to benchmark and replace with proper value, later ?

sandreim (approved):

runtime/parachains/src/paras.rs
pepyakin: Although it doesn't have a separate issue (yet) it is being tracked by https://github.com/paritytech/polkadot/issues/3211

PR #4644, opened 2021-12-31
Move XCM runtime configurations into their own files

KiChjang:
This is purely a refactoring of XCM config code.

Since misconfigured XCM settings easily causes attacks, I thought that it may be useful to move its configuration into its own file for easier auditing and maintainability.

### DIFF
--- a/runtime/kusama/src/xcm_config.rs
+++ b/runtime/kusama/src/xcm_config.rs
@@ -1,4 +1,4 @@
-// Copyright 2017-2020 Parity Technologies (UK) Ltd.
+// Copyright 2022 Parity Technologies (UK) Ltd.
 // This file is part of Polkadot.
 
 // Polkadot is free software: you can redistribute it and/or modify
--- a/runtime/kusama/src/xcm_config.rs
+++ b/runtime/kusama/src/xcm_config.rs
@@ -1,0 +1,187 @@
+// Copyright 2017-2020 Parity Technologies (UK) Ltd.
+// This file is part of Polkadot.
+
+// Polkadot is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+
+// Polkadot is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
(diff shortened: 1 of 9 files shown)


shawntabrizi:
We should do this for the runtime in general, across all the pallets.

shawntabrizi (approved):

runtime/kusama/src/xcm_config.rs
```diff
@@ -0,0 +1,187 @@
+// Copyright 2017-2020 Parity Technologies (UK) Ltd.
```
kianenigma, line 1:
```suggestion
// Copyright 2022 Parity Technologies (UK) Ltd.
```

runtime/polkadot/src/xcm_config.rs
```diff
@@ -0,0 +1,184 @@
+// Copyright 2021 Parity Technologies (UK) Ltd.
```
kianenigma, line 1:
```suggestion
// Copyright 2022 Parity Technologies (UK) Ltd.
```

runtime/polkadot/src/xcm_config.rs
```diff
@@ -0,0 +1,184 @@
...
+	BackingToPlurality<
+		Origin,
+		pallet_collective::Origin<Runtime, CouncilCollective>,
+		CouncilBodyId,
+	>,
+	// And a usual Signed origin to be used in XCM as a corresponding AccountId32
+	SignedToAccountId32<Origin, AccountId, PolkadotNetwork>,
+);
+
+impl pallet_xcm::Config for Runtime {
```
kianenigma, line 167: I would personally rather have this block in the main runtime file, and all the other noisy `parameter_types!` in a separate file. This allows you to quickly skim the runtime configs with 1 layer of observability and no further. For example, you will know the _name_ of the type parameters, but not their value, and their name could already say a lot of about them.

But this format is also fine.

runtime/rococo/src/xcm_config.rs
```diff
@@ -0,0 +1,170 @@
+// Copyright 2021 Parity Technologies (UK) Ltd.
```
kianenigma, line 1:
```suggestion
// Copyright 2022 Parity Technologies (UK) Ltd.
```

kianenigma: and the rest please.

kianenigma (approved):

runtime/polkadot/src/xcm_config.rs
shawntabrizi: i actually think this content in a separate file is the more natural thing to do. I think that we should make files where configurations with one another are relevant, so all democracy / governance configs in one file. But keep separate things which are not related

KiChjang:
bot merge

Issue #4651, opened 2022-01-03
Runtime debugging not possible due to missing RuntimeDebug implementations


hirschenberger:
I added `sp-debug-derive` with the `force-debug` feature as @al3mart suggested [here](https://matrix.to/#/!HzySYSaIhtyWrwiwEV:matrix.org/$16412030812550AMtvS:matrix.parity.io?via=matrix.parity.io&via=corepaper.org&via=matrix.org)

and removed the debug-when-std conditionals when I got an error of missing Debug implementations as @bkchr suggested.

I worked on a PR by removing some more conditionals but got stuck when structs from `finality-grandpa` got involved as it does not have a dependency on `sp_core` or `sp_debug_derive`

I'm not sure how to continue here, should I add that dependency to `finality-grandpa` also?

bkchr:
Not sure, but I didn't run into any problems: https://github.com/paritytech/polkadot/pull/4652

hirschenberger:
Maybe I did sth. wrong I added:

```
sp-debug-derive = { git = "https://github.com/paritytech/substrate", branch = "master", features = ["force-debug"] }
```
to `runtime/parachains/Cargo.toml` to enable using the log-macros for debugging an issue.

bkchr:
Ahh I did not compile with the runtime benchmarks feature.

bkchr:
My branch works for me with `runtime-benchmarks`

hirschenberger:
Can you check: https://github.com/hirschenberger/polkadot/tree/issue-4651?

It's your branch with my changes on top.

It fails with:

```
    Compiling bp-header-chain v0.1.0 (/home/hirschen/devel/git/polkadot/bridges/primitives/header-chain)
  error[E0277]: `finality_grandpa::Commit<<Header as sp_runtime::traits::Header>::Hash, <Header as sp_runtime::traits::Header>::Number, sp_finality_grandpa::app::Signature, sp_finality_grandpa::app::Public>` doesn't implement `core::fmt::Debug`
    --> /home/hirschen/devel/git/polkadot/bridges/primitives/header-chain/src/justification.rs:38:26
     |
  38 | #[derive(Encode, Decode, RuntimeDebug, Clone, PartialEq, Eq, TypeInfo)]
[...]

PR #4652, opened 2022-01-03
Switch types to use `RuntimeDebug` instead of gated `Debug`

bkchr:
This is useful for when you want to debug in wasm and enable the
`force-debug` feature of `sp-debug-derive`.

Fixes: https://github.com/paritytech/polkadot/issues/4651

### DIFF
--- a/parachain/src/primitives.rs
+++ b/parachain/src/primitives.rs
@@ -113,8 +113,8 @@ impl sp_std::fmt::LowerHex for ValidationCodeHash {
 /// Parachain block data.
 ///
 /// Contains everything required to validate para-block, may contain block and witness data.
-#[derive(PartialEq, Eq, Clone, Encode, Decode, derive_more::From, TypeInfo)]
-#[cfg_attr(feature = "std", derive(Serialize, Deserialize, Debug, MallocSizeOf))]
+#[derive(PartialEq, Eq, Clone, Encode, Decode, derive_more::From, TypeInfo, RuntimeDebug)]
+#[cfg_attr(feature = "std", derive(Serialize, Deserialize, MallocSizeOf))]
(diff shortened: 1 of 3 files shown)


kianenigma (approved):
This should have a small, but non-zero impact on the code size if I am not mistaking. The `RuntimeDebug` will generate a `impl` block for this type in wasm that's not really printing anything, whereas the std-gated `Debug` is literally non-existent in wasm. Can you confirm which one is the case here? Maybe the `RuntimeDebug`'s implementation is also stripped away by the compiler? 

Either way, PR is fine to me.

bkchr:
> The `RuntimeDebug` will generate a `impl` block for this type in wasm that's not really printing anything, whereas the std-gated `Debug` is literally non-existent in wasm.

This is correct. `RuntimeDebug` will print `<wasm:stripped>`

hirschenberger:
Can we please land this quickly, as I need it for debugging my code? Merci.

bkchr:
> I'm running a benchmark with `--execution=native`, so it should not be stripped, right?

Benchmarks are only running in wasm.


> I want to debug a `frame_system::pallet::Config::Event`, which is still `Debug`, `RuntimeDebug` needs to be introduced to substrate as well?

If it is just `Debug`, there is no problem.

hirschenberger:
> > I'm running a benchmark with `--execution=native`, so it should not be stripped, right?
> 
> Benchmarks are only running in wasm.
> 
> > I want to debug a `frame_system::pallet::Config::Event`, which is still `Debug`, `RuntimeDebug` needs to be introduced to substrate as well?
> 
> If it is just `Debug`, there is no problem.

So how can I debug my event then without being stripped away?

bkchr:
This is using this code: https://github.com/paritytech/substrate/blob/master/frame/support/procedural/src/lib.rs#L448-L476

Issue #3243, opened 2021-06-14
Reserved KSM by `nicks` pallet can't be released


michalisFr:
A user has a reserved amount of 10 KSM due to using the `nicks.set_name` extrinsic a long time ago. Now they don't have a way to remove that reserve because the `nicks` pallet is no longer part of the runtime.

Is there a workaround they can use to release that reserve? And if there isn't, how can this be resolved for them and other users who may face this issue?

bkchr:
CC @shawntabrizi

shawntabrizi:
@michalisFr I dont see any on-chain data which has the information about the users who used nicks. Do you have some kind of transaction data or block explorer where we can get the list?

Ultimately if we can collect an accurate list of users who this affects, we can make a simple governance proposal.

michalisFr:
I don't have any @shawntabrizi. We only know of this one user that wrote in support with this issue. I'll ping Jake who's working on a script to finds other users in the same position.

So, I gather there's no way for them to release these funds themselves, right?

xlc:
Here are all the Nicks events https://kusama.subscan.io/event?address=&module=nicks&event=all&startDate=&endDate=
Latest one is a NameSet so this guy likely have some extra reserved balance if the upgrade that removes nicks pallet didn't correctly clean up the state https://kusama.subscan.io/extrinsic/0x121d10050dc3a5d73014432c0f6014d4c5bef79bbe0ec8d1e0ff32a402d51ba2?event=563794-4

jakehemmerle:
fwiw @michalisFr finished a script finding accounts that:

- Have issued at least one successful set_name extrinsic
- Haven't issued a successful clear_name extrinsic after that
- Have at least 10 KSM reserved balance

corrosponding accounts:
[accounts.json.txt](https://github.com/paritytech/polkadot/files/6658750/accounts.json.txt)

Maybe include the script here for ref @michalisFr ?

shawntabrizi:
Yeah, and I would be interested to know how many accounts which satisfy the first two points dont satisfy the third point

michalisFr:
@shawntabrizi I did a little refactoring to get you the output you wanted:
```
Accounts with `set_name`: 153
Accounts that didn't issue a `clear_name`: 115
Accounts without a `clear_name` but less than 10 KSM reserved: 84
Accounts with reserved KSM: 31
```

Here's the script (it's in python): 
[...]

PR #4656, opened 2022-01-04
Fix locked deposits of Nicks pallet 

kianenigma:
The nicks pallet was removed at block 569325 from the runtime, without any consideration of the
fact that numerous accounts had reserved funds in this pallet. This migration is the outcome of
an investigation that tries to refund all of the accounts that had a nick set for them prior to
removal, if they still have the amount in their reserved balance. Otherwise, we ignore the
refund.

closes https://github.com/paritytech/polkadot/issues/3243

For the sake of posterity, the code that's used to generate this fix is here:  https://gist.github.com/kianenigma/a34bcc2efcf753d986c199da62adffbd

--- 

## Release Note Addition

Please make sure this is added to the release note of 9150 runtime for transparency:

> This runtime upgrade contains a custom migration that fixes the locked deposit participants of the historical `Nicks` pallet, which was only used temporarily in the Kusama runtime. See the list of affected accounts and more description [here](https://github.com/paritytech/polkadot/pull/4656).


michalisFr:
I have found a subset of @kianenigma's accounts, the ones that have a 10 KSM reserve and some that have 0.01 KSM (I wasn't looking for 0.01 KSM reserves, as I wasn't aware the reserved amount had changed). My results for the accounts with 10 KSM reserve agree with Kian's results, as do the accounts with 0.01 KSM that I found.

My approach can be found here:
https://github.com/paritytech/polkadot/issues/3243
but Kian's methodology is more solid.

The only account that potentially has a reserve from `nicks` but we couldn't determine it with certainty is this one:
CmD9vaMYoiKe7HiFnfkftwvhKbxN9bhyjcDrfFRGbifJEG8
and so it isn't included in this PR.

runtime/kusama/src/lib.rs
```diff
@@ -1640,6 +1640,1251 @@ pub type Executive = frame_executive::Executive<
...
+pub struct RefundNickPalletDeposit;
+
+impl RefundNickPalletDeposit {
+	fn execute(check: bool) -> Weight {
+		let accounts_and_deposits = vec![
+			(
+				[
+					116, 157, 220, 147, 166, 93, 254, 195, 175, 39, 204, 116, 120, 33, 44, 183,
+					212, 176, 192, 53, 127, 239, 53, 160, 22, 57, 102, 171, 83, 51, 183, 87,
+				],
```
shawntabrizi, line 1657: would like to have the account id written as a comment above these different hexs for easier audit

runtime/kusama/src/lib.rs
```diff
@@ -1640,6 +1640,1251 @@ pub type Executive = frame_executive::Executive<
...
+					126, 224, 152, 119, 99, 164, 118, 38, 81, 33, 204, 1, 140, 4, 39,
+				],
+				10000000000,
+			),
+			(
+				[
+					218, 1, 7, 123, 220, 2, 95, 215, 121, 204, 33, 201, 118, 7, 39, 236, 7, 229,
+					42, 161, 50, 65, 11, 130, 229, 250, 186, 203, 111, 69, 176, 85,
+				],
+				10000000000000,
```
shawntabrizi, line 1819: this has more zeros than the others

runtime/kusama/src/lib.rs
kianenigma: that's a good idea actually. I will add them, please manually check a few of them? they are all auto generated, so if one is correct, they all should be fine.

runtime/kusama/src/lib.rs
```diff
@@ -1640,6 +1640,1407 @@ pub type Executive = frame_executive::Executive<
...
+				],
+				10000000000,
+			),
+			(
+				// FagAVsTYT8QghxypUtLcfnmnnhPhPpf854UNuptpQKuNndK
+				[
+					132, 250, 52, 33, 111, 141, 173, 223, 29, 75, 206, 46, 56, 72, 71, 124, 222,
+					140, 211, 139, 226, 179, 165, 161, 38, 254, 36, 111, 96, 86, 124, 60,
+				],
+				10000000000000,
```
kianenigma, line 2995: note that some accounts reserved 10KSM and some 10mKSM because of how the logic of the pallet changed, so variability here is expected.

shawntabrizi (approved):

kianenigma:
bot merge

PR #4663, opened 2022-01-05
[ci] .gitlab-ci.yml chores

alvicsam:
CI team has a docker image for pipelines `paritytech/ci-linux:production` which has previous stable and nightly versions of rust. We want to run Polkadot pipeline with `paritytech/ci-linux:staging` image which has the latest rust stable and nightly versions (the same way as we do for substrate).
So the initial idea of this PR was to add exception in rules for downstream pipeline in order not to run them all for testing the image.

What is done: 
- renamed rules to refs in the same way as in other projects
- moved some rules and added anchors
- renamed `malus.Dockerfile` so it complies [Naming policy](https://github.com/paritytech/ci_cd/wiki/Policies-and-Processes:-Dockerfiles-naming-convention)

https://github.com/paritytech/ci_cd/issues/249


pepoviola (approved):

TriplEight (approved):
LGTM. This will also come in handy when we will evolve companion PRs into running not just one job but trigger a part of the pipeline.

PR #4681, opened 2022-01-10
XCM v3: Bridge infrastructure

gavofyork:
Introduces XCM protocol elements needed for bridging and removes XCM v0.

Introduces a conceptual _Universal Context_ under which all global consensus systems exist. The context (what used to be called "ancestry") of a consensus system (chain) should now be in terms of this Universal Context. A new junction type is introduced meant for global consensus systems which sit directly under the Universal Context.

`NetworkId` is reworked to remove the `Any` which allows it to be used within the new `Junction::GlobalConsensus`.

_Local Consensus_ is defined as all locations under the same overall Consensus system as the local location. _Non-Local Consensus_ (may sometimes be called Remote Consensus) is defined as all other locations. Bridges exist as a means of communication between systems which view each other as being under Non-Local Consensus.

`ExportXcm` is a new trait added similar to `SendXcm`, except that it works for sending messages to a Non-Local Consensus. `SendXcm` has been reworked to be two-stage, with error reporting and pricing happening in the first stage and then actual delivery happening in the second.

Three new `SendXcm` impls have been added to manage routing over a bridge (`ExportXcm`) impl. One is for when the bridge is local, one for when the bridge is in _Local Consensus_ but is not on the local chain and when it does not expect any fees to be paid. And the final one when the bridge is not local and does expect fees to be paid.

All instructions which themselves initiate a transport are now 

## SendXcm/ExportXcm and Weights

Weight for SendXcm/ExportXcm will be a constant/fixed value for a runtime. The reasons for this are:
1. They will tend to be similar for all modes of sending since send/export operations tend to  just place the message in a storage queue for later delivery.
2. Payment for operations using SendXcm in XCM would generally need to over-estimate without forcing the user to provide a hint in the XCM program since the destination cannot generally be known prior to instruction execution, when weight is initially measured.
3. Measuring weights for sending operations and using these values dynamically is a non-trivial code addition adding to overall complexity and attack surface. 
4. Though over-estimation of weight eats into performance, it is safe.

However, though *execution weights* are fixed, *pricing* happens independently and can be arbitrary.

## Usage

There should be a separate local router `LocalRouter` for routing within Local Consensus. For general routing (which may itself need to use the `LocalRouter`), one or more Bridging Routers may need to be amalgamated into a second general `Router` type. 

A bridging Router will be configured with the XCM location of the Bridge it routes over as well as the prices for usage. This router checks for non-local consensus destinations (i.e. those which ascend to the Universal Ancestor and descend into a consensus which is not local). If a message is sent to such a destination and it is supported by a bridge, it will lookup the correct XCM location for the bridge then encompass the message in a `ExportMessage` instruction, providing payment with a prefixed `WithdrawAsset`/`BuyExecution` if needed.

When a message is routed into a bridge, the bridge checks the origin and if a descendent of the overall consensus (which it should be representing), then a `DescendOrigin` will be prefixed. This message will then be queued for sending over the bridge together with the destination location relative to the destination consensus which the remote side of the bridge represents.

Bridge hubs will be configured as `UniversalAliases` for all networks to which they bridge over. When a message comes over a bridge they will send the message on to the destination consensus chain using intra-consensus transport, prefixed with a `UniversalOrigin` instruction with the source consensus system.

- Rework `NetworkId` into `Option<NetworkId>` and remove `NetworkId::Any`.
- New junction: `Junction::GlobalConsensus(NetworkId)`
- New instruction: `Instruction::UniversalOrigin`
- New `xcm_executor::Config` item: `type UniversalAliases`

CC @svyatonik 

## TODO 
- [x] New instruction: `ExportMessage` + `xcm_executor::Config` type item to configure it.
- [x] New `SendXcm` impl which detects destinations beyond our consensus and extracts the remote network and location within it. If a known (configurable by trait parameters) bridge exists for the remote network it prepends and packs the message and sends to the bridge, e.g.: `[WithdrawAsset, BuyExecution, ExportMessage { network: remote_network, destination: remote_destination, xcm: Xcm(vec![UniversalOrigin(local_network), DescendOrigin(local_origin), { msg }]) }]`. The type which configures this should be the same type which configures the `UniversalAliases` type.
- [x] Slim API ready for bridge
- [x] Cost/price impls
  - [x] `LocalUnpaidExporter::send_cost`
  - [x] `UnpaidRemoteExporter::send_cost`
  - [x] `SovereignPaidRemoteExporter::send_cost`
  - [x] `HaulBlobExporter::export_price`
- [x] Tests
  - [x] New barriers
  - [x] Costs for `send_xcm`
  - [x] Prices for `export_price`
  - [x] `ExportMessage` charging
- [x] Further consideration of bridge usage costs: will be fixed fee per message with network-incentivised relayers. Weight for `ExportMessage` should be adjusted accordingly.
- [x] Sanity check Polkadot/Kusama changes

## Next steps (further PRs)

- Introduce new config into Rococo
- Local dispatcher for a bridge endpoint
  - Queued execution for messages bound locally
  - SendXcm for messages bound elsewhere


runtime/common/src/xcm_sender.rs
```diff
@@ -20,30 +20,33 @@ use parity_scale_codec::Encode;
 use runtime_parachains::{configuration, dmp};
 use sp_std::marker::PhantomData;
 use xcm::latest::prelude::*;
+use SendError::*;
```
gdeek, line 23: whats this for?

xcm/src/v3/junction.rs
```diff
@@ -0,0 +1,205 @@
...
+	/// Convert `self` into a `MultiLocation` containing 0 parents.
+	///
+	/// Similar to `Into::into`, except that this method can be used in a const evaluation context.
+	pub const fn into_location(self) -> MultiLocation {
+		MultiLocation { parents: 0, interior: Junctions::X1(self) }
+	}
+
+	/// Convert `self` into a `MultiLocation` containing `n` parents.
+	///
+	/// Similar to `Self::into`, with the added ability to specify the number of parent junctions.
```
KiChjang, line 194:
```suggestion
	/// Similar to `Self::into_location`, with the added ability to specify the number of parent junctions.
```

xcm/src/v3/junction.rs
```diff
@@ -0,0 +1,205 @@
...
+	pub const fn into_exterior(self, n: u8) -> MultiLocation {
+		MultiLocation { parents: n, interior: Junctions::X1(self) }
+	}
+
+	/// Convert `self` into a `MultiLocation` containing 0 parents.
```
KiChjang, line 199:
```suggestion
	/// Convert `self` into a `VersionedMultiLocation` containing 0 parents.
```

xcm/src/v3/junctions.rs
```diff
@@ -0,0 +1,591 @@
...
+	/// Convert `self` into a `MultiLocation` containing 0 parents.
+	///
+	/// Similar to `Into::into`, except that this method can be used in a const evaluation context.
+	pub const fn into_location(self) -> MultiLocation {
+		MultiLocation { parents: 0, interior: self }
+	}
+
+	/// Convert `self` into a `MultiLocation` containing `n` parents.
+	///
+	/// Similar to `Self::into`, with the added ability to specify the number of parent junctions.
```
KiChjang, line 127:
```suggestion
	/// Similar to `Self::into_location`, with the added ability to specify the number of parent junctions.
```

xcm/src/v3/junctions.rs
```diff
@@ -0,0 +1,591 @@
...
+	}
+
+	/// Returns a reference iterator over the junctions in reverse.
+	#[deprecated(note = "Please use iter().rev()")]
+	pub fn iter_rev(&self) -> impl Iterator + '_ {
+		self.iter().rev()
+	}
+
+	/// Consumes `self` and returns an iterator over the junctions in reverse.
+	#[deprecated(note = "Please use into_iter().rev()")]
```
KiChjang, line 450: Do we still want to keep deprecated code around?

xcm/src/v3/junction.rs
```diff
@@ -0,0 +1,205 @@
...
+	}
+}
+
+impl From<Vec<u8>> for Junction {
+	fn from(id: Vec<u8>) -> Self {
+		Self::GeneralKey(id)
+	}
+}
+
+impl TryFrom<OldJunction> for Junction {
```
KiChjang, line 165: Does this need to be a `TryFrom` when the function never returns an Err variant?

xcm/src/v3/multilocation.rs
```diff
@@ -0,0 +1,727 @@
...
+	use parity_scale_codec::{Decode, Encode};
+
+	#[test]
+	fn conversion_works() {
+		let x: MultiLocation = Parent.into();
+		assert_eq!(x, MultiLocation { parents: 1, interior: Here });
+		//		let x: MultiLocation = (Parent,).into();
+		//		assert_eq!(x, MultiLocation { parents: 1, interior: Here });
+		//		let x: MultiLocation = (Parent, Parent).into();
+		//		assert_eq!(x, MultiLocation { parents: 2, interior: Here });
```
KiChjang, line 503: What's with the commented out code here?

xcm/src/v3/junction.rs
gavofyork: Earlier versions of the function were not infallible, and over time as more changes creep in, these converters have a habit of becoming fallible. So it's easiest to keep it as a `TryFrom` for now.

PR #4683, opened 2022-01-10
paras: Add runtime events for PVF pre-checking

pepyakin:
In this PR, paras module emit runtime events on certain PVF pre-checking
related conditions.

Specifically, there are 3 new events in the paras module:

1. PvfCheckStarted
2. PvfCheckAccepted
3. PvfCheckRejected

All of those have identifiers for the parachain that triggered the PVF
pre-checking and the validation code that goes through the pre-checking.

The mechanics of those are as follows. Each time a new PVF is added, be
it due to onboarding or upgrading, the `PvfCheckStarted` will be
triggered. If another parachain triggers a pre-checking process for the
validation code which is already being pre-checked, another
`PvfCheckStarted` event will be triggered with the corresponding para
id.

When the PVF pre-checking voting for a PVF was finished, several
`PvfCheckAccepted/Rejected` events will be triggered: one for each para id that
was subscribed to this check (i.e. was a "cause" for it).

If the PVF pre-checking is disabled, then one can still expect these
events to be fired. Since insta PVF approval is syncronous, the
`PvfCheckStarted` will be followed by the `PvfCheckAccepted` with the
same validation code and para id.

If somebody is interested in following validation code changes for a PVF
of a parachain, they would need to subscribe to those events. I did not
supply the topics for the events, since I am not sure if that's needed
or will be used, but they can be added later if needed.

### DIFF
--- a/runtime/parachains/src/paras/mod.rs
+++ b/runtime/parachains/src/paras/mod.rs
@@ -314,6 +314,16 @@ enum PvfCheckCause<BlockNumber> {
 	},
 }
 
+impl<BlockNumber> PvfCheckCause<BlockNumber> {
+	/// Returns the ID of the para that initiated or subscribed to the pre-checking vote.
+	fn para_id(&self) -> ParaId {
+		match *self {
+			PvfCheckCause::Onboarding(id) => id,
+			PvfCheckCause::Upgrade { id, .. } => id,
+		}
+	}
+}
+
 /// Specifies what was the outcome of a PVF pre-checking vote.
 #[derive(Copy, Clone, Encode, Decode, RuntimeDebug, TypeInfo)]
 enum PvfCheckOutcome {
@@ -467,6 +477,15 @@ pub mod pallet {
 		NewHeadNoted(ParaId),
 		/// A para has been queued to execute pending actions. `para_id`
 		ActionQueued(ParaId, SessionIndex),
+		/// The given para either initiated or subscribed to a PVF check for the given validation
+		/// code. `code_hash` `para_id`
+		PvfCheckStarted(ValidationCodeHash, ParaId),
+		/// The given validation code was rejected by the PVF pre-checking vote.
+		/// `code_hash` `para_id`
+		PvfCheckAccepted(ValidationCodeHash, ParaId),
+		/// The given validation code was accepted by the PVF pre-checking vote.
+		/// `code_hash` `para_id`
+		PvfCheckRejected(ValidationCodeHash, ParaId),
 	}
 
 	#[pallet::error]
@@ -1345,6 +1364,9 @@ impl<T: Config> Pallet<T> {
 	) -> Weight {
 		let mut weight = 0;
 		for cause in causes {
+			weight += T::DbWeight::get().reads_writes(3, 2);
+			Self::deposit_event(Event::PvfCheckAccepted(*code_hash, cause.para_id()));
+
 			match cause {
 				PvfCheckCause::Onboarding(id) => {
 					weight += Self::proceed_with_onboarding(*id, sessions_observed);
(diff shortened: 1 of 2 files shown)


pepyakin:
* master
  * **PR #4683**  (this PR)
    * **PR #4684** 

runtime/parachains/src/paras/mod.rs
```diff
@@ -467,6 +477,12 @@ pub mod pallet {
 		NewHeadNoted(ParaId),
 		/// A para has been queued to execute pending actions. `para_id`
 		ActionQueued(ParaId, SessionIndex),
+		/// The given para either initiated or subscribed to a PVF check for the given validation
+		/// code. `code_hash` `para_id`
+		PvfCheckStarted(ValidationCodeHash, ParaId),
```
bkchr, line 482: FYI: We started now slowly migrating to named event arguments. Which doesn't require these comments on what is what .

runtime/parachains/src/paras/mod.rs
```diff
@@ -1345,6 +1361,9 @@ impl<T: Config> Pallet<T> {
 	) -> Weight {
 		let mut weight = 0;
 		for cause in causes {
+			weight += T::DbWeight::get().reads_writes(3, 2);
+			Self::deposit_event(Event::PvfCheckConcluded(*code_hash, cause.para_id(), true));
```
bkchr, line 1365: Looks like this should be fixed in Substrate? So that `deposit_event` returns the weight instead of you are required to calculate this here manually?

runtime/parachains/src/paras/mod.rs
```diff
@@ -467,6 +477,12 @@ pub mod pallet {
 		NewHeadNoted(ParaId),
 		/// A para has been queued to execute pending actions. `para_id`
 		ActionQueued(ParaId, SessionIndex),
+		/// The given para either initiated or subscribed to a PVF check for the given validation
+		/// code. `code_hash` `para_id`
+		PvfCheckStarted(ValidationCodeHash, ParaId),
+		/// The PVF pre-checking for the given validation code was concluded with the given outcome
+		/// (either accepted or rejected). `code_hash` `para_id` `accepted`
+		PvfCheckConcluded(ValidationCodeHash, ParaId, bool),
```
bkchr, line 485: IMO we could also just split this up into two variants.

bkchr (approved):

runtime/parachains/src/paras/mod.rs
pepyakin: What's your reasoning here?

runtime/parachains/src/paras/mod.rs
pepyakin: Yeah, I know. I did not want to open up this can of worms though. There are plenty examples of this, even in the polkadot repo, e.g.: 

https://github.com/paritytech/polkadot/blob/7311a3aa9f83730a079b0899c34bb813dd96c4e1/runtime/parachains/src/inclusion/mod.rs#L780

And it is not limited to `deposit_event` though. One of the worst cases IMO is the `Currency` trait where you can't even assume what the concrete implementation would do. 

IMO this problem should be addressed as a whole, holistically.

Anyway, I do not think that problem is within the scope of this particular PR (unless, you propose to remove the weight accounting for this case to make it inline with other cases found elsewhere). I see a solution should be first discussed in an issue in Substrate (cc @shawntabrizi), not in this thread. As soon as we have a solution though, I'd be down to migrate this code (and the rest of polkadot's for that matter) to the new mechanism.

runtime/parachains/src/paras/mod.rs
pepyakin: Thanks for the heads up!

runtime/parachains/src/paras/mod.rs
bkchr: Yeah, I also don't wanted you to fix this as part of this pr.

runtime/parachains/src/paras/mod.rs
bkchr: You directly see the output based on the event name.

pepyakin: Ok, don't feel about this strongly. [Here](https://github.com/paritytech/polkadot/compare/9537c19cb666468688f7e4bba3c81877ffd773cf..bec1cc031c7b0f8942ca8c3d006952f1a092b974) is the fix

PR #4684, opened 2022-01-10
paras: do not allow PVF vote submission if disabled

pepyakin:
if the PVF pre-checking is disabled the runtime dispatchable will reject
any attempts of submission. This is also concern the unsigned tx
validation.

Right now, the `include_pvf_check_statement` dispatchable is effectively
uncallable because of the weight set to the maximum value. If we were to
benchmark it, it would become includable in a block, but since there
will be no active votes, the dispatchable won't do anything.

However, it will execute some code, like signature validation and
querying some storage entries. To be completely safe, we can bail out
early if the `pvf_checking_enabled` config is disabled. That's what this
PR does.


pepyakin:
* master
  * **PR #4683** 
    * **PR #4684**  (this PR)

bkchr (approved):

Issue #4610, opened 2021-12-27
pvf-precheck: the subsystem


pepyakin:
Implement the PVF pre-checking subsystem as described in https://github.com/paritytech/polkadot/issues/3211

Issue #2867, opened 2021-04-09
PVF validation host: hook up heads up signal


pepyakin:
The new PVF validation host allows its client, candidate validation, to inform about soon needed PVFs via a so called "heads up" signal. 

Candidate validation subsystem should periodically collect all active PVFs' code hashes and pass it to the PVF validation host. The interval doesn't matter that much from standpoint of PVF validation host, but from a glance it makes sense to do this once every session.

It would be good to have https://github.com/paritytech/polkadot/issues/2866 first, so that we don't have to shoot validation code preimages. Also we must have a way to query "relevant" PVFs from the paras module in the runtime.

burdges:
I mentioned availability reconstruction of code in #2866 but again like with blocks do feel free to optimize by fetching the full code fro anyone who possesses it.  If we construct that system before the availability reconstruction part then we need to be able to adequately test the availability reconstruction part whenever we eventually write it.

PR #4685, opened 2022-01-10
impl guide: Describe the PVF check subsystem

### LINKED ISSUE #4611: implementer's guide: fill out the description for PVF pre-checker subsystem

pepyakin:
closes #4611

A follow-up for the recently merged PVF pre-checking subsystem (aka pvf-checker) https://github.com/paritytech/polkadot/issues/4610

### DIFF
--- a/roadmap/implementers-guide/src/node/utility/pvf-prechecker.md
+++ b/roadmap/implementers-guide/src/node/utility/pvf-prechecker.md
@@ -10,8 +10,22 @@ This subsytem does not produce any output messages either. The subsystem will, h
 
 ## Functionality
 
-TODO: Write up the description of the functionality of the PVF pre-checker. https://github.com/paritytech/polkadot/issues/4611
+If the node is running in a collator mode, this subsystem will be disabled. The PVF pre-checker subsystem keeps track of the PVFs that are relevant for the subsystem. 
+
+To be relevant for the subsystem, a PVF must be returned by `pvfs_require_precheck` [`pvfs_require_precheck` runtime API][PVF pre-checking runtime API] in any of the active leaves. If the PVF is not present in any of the active leaves, it ceases to be relevant.
+
+When a PVF just becomes relevant, the subsystem will send a message to the [Candidate Validation] subsystem asking for the pre-check.
+
+Upon receving a message from the candidate-validation subsystem, the pre-checker will note down that the PVF has its judgement and will also sign and submit a [`PvfCheckStatement`] via the [`submit_pvf_check_statement` runtime API][PVF pre-checking runtime API]. In case, a judgement was received for a PVF that is no longer in view it is ignored. It is possible that the candidate validation was not able to check the PVF. In that case, the PVF pre-checker will abstain and won't submit any check statements.
+
+Since a vote only is valid during [one session][overview], the subsystem will have to resign and submit the statements for for the new session. The new session is assumed to be started if at least one of the leaves has a greater session index that was previously observed in any of the leaves.
+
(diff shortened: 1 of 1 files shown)


pepyakin:
* master
  * **PR #4685**  (this PR)

roadmap/implementers-guide/src/node/utility/pvf-prechecker.md
```diff
@@ -10,8 +10,22 @@ This subsytem does not produce any output messages either. The subsystem will, h
...
+
+Since a vote only is valid during [one session][overview], the subsystem will have to resign and submit the statements for for the new session. The new session is assumed to be started if at least one of the leaves has a greater session index that was previously observed in any of the leaves.
+
+The subsystem tracks all the statement that it submitted within a session. If for some reason a PVF became irrelevant and then becomes relevant again, the subsystem will not submit a new statement for that PVF within the same session.
+
+If the node is not in the active validator set, it will still perform all the checks. However, it will only submit the check statements when the node is in the active validator set.
```
eskimor, line 25: Hmm, why actually?

pepyakin: I do not think this is something that would be attackable and the validators in the active set should be ready for this so we cannot really use the freed up performance as an advantage on the validators not in the active set (and they won't do any work either). On the other hand, if we do this, then pre-check will actually prepare the PVF and save it in the cache and as soon as the validator gets into the active set, even though it did not participate in the pre-checking vote, it will be ready to execute candidates from the get go.

In other words, we are being optimistic here, because I assume/hope 99.9% of pre-checks will end up with accepting the PVF. At the same time, if the PVF does not get accepted, then it is just a little bit of wasted effort for an inactive validator that does not do much anyway.

pepyakin: That's, by the way, is a poor man's heads up signal https://github.com/paritytech/polkadot/issues/2867

eskimor: Got it - I mixed that up, with whether the node is a validator at all. You are talking about the set that is restricted by `max_validators...`. In that case I think the code is wrong though. You are using the `validators` runtime call, which ends up reading `ActiveValidatorKeys`, which in turn gets set to the [`active set`](https://github.com/paritytech/polkadot/blob/4165c7f83a85e12c7ec56366a9559bf0eba25f8d/runtime/parachains/src/shared.rs#L83), instead of all potential validators.

pepyakin: When I say active set / activate validator set (both in my message and in the code) I refer to the whole set of validators returned by `validator`

When I say when the node is not in active set, I mean the set of all nodes that do run the overseer but are not collator and which do not have keys for the validator ids in the `validators` list. That definition, hopefully, should only include validators but not full nodes (they do not run overseer) and collators (they have IsCollator::Yes).

eskimor: :bulb:  uh right - you do the validation regardless whether or not you find your key in `validators`, but are you sure full nodes don't run the overseer?

I believe checking for  `role.is_authority()` in [this line](https://github.com/paritytech/polkadot/blob/22e5c04008dd85ddd36c4af74e3bdcdb20bb0645/node/service/src/lib.rs#L735) would make sense - I don't see any code that disables the overseer if `is_authority()` is false, only [chain selection](https://github.com/paritytech/polkadot/blob/22e5c04008dd85ddd36c4af74e3bdcdb20bb0645/node/service/src/lib.rs#L737) is different. :thinking:

pepyakin: I moved this discussion into a separate issue.

eskimor (approved):

Issue #4691, opened 2022-01-11
Inaccessible data stored at session pallet


kianenigma:
https://github.com/paritytech/substrate/pull/9878/ was added to both chains seemingly around 9.12 release, yet we somehow missed its companion. Luckily, this is now all good since after 28 days, the pallet is fully functioning in its new storage prefix (`Historical`). 

The problem is that the two storage items of this pallet still live under the old prefix as well, with their data being stale and not really useful anymore. 

We need a migration that goes and cleans everything under `Session::HistoricalSessions` and `Session::StoredRanges`.

PR #4692, opened 2022-01-11
Fix Migrations for the upcoming release

kianenigma:
- Removes the buggy migration that is no longer needed. 
- I actually forgot to include the migration for the nicks pallet fix into `Executive` 🤦 
- made this to remember a cleanup: https://github.com/paritytech/polkadot/issues/4691

### DIFF
--- a/runtime/kusama/src/lib.rs
+++ b/runtime/kusama/src/lib.rs
@@ -1634,7 +1634,7 @@ pub type Executive = frame_executive::Executive<
 	frame_system::ChainContext<Runtime>,
 	Runtime,
 	AllPalletsWithSystem,
-	(SessionHistoricalPalletPrefixMigration, SchedulerMigrationV3),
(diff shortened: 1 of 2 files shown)


runtime/kusama/src/lib.rs
```diff
@@ -3041,8 +3041,9 @@ impl OnRuntimeUpgrade for RefundNickPalletDeposit {
 	}
 
 	#[cfg(feature = "try-runtime")]
-	fn pre_migrate() -> Result<(), &'static str> {
-		Self::execute(true)
+	fn pre_upgrade() -> Result<(), &'static str> {
+		let _ = Self::execute(true);
+		Ok(())
```
bkchr, line 3046: Why not return the result?

bkchr (approved):

andresilva (approved):

kianenigma:
bot merge

Issue #4489, opened 2021-12-08
Optimize reanchor


gavofyork:
So I realized that what we're essentially doing here is to first find the common ancestor between `self` and `target`, then append the necessary junctions that will allow us to reach `self` from the common ancestor, and then we'd simplify the result, given `target`.

I suspect what we can do here is to write the common ancestor from the perspective of `target`, that way we don't have to simplify at the end of the algorithm.

_Originally posted by @KiChjang in https://github.com/paritytech/polkadot/pull/4470#discussion_r763422057_

shawntabrizi:
We will have a hiring candidate working on this one.

emmaling27:
> We will have a hiring candidate working on this one.

Hey! Just started looking at this. I was messing with different inputs in the tests to make sure I understand the behavior of the current implementation of `reanchor` correctly, and the following gave me an unexpected result
```
let mut id: MultiLocation = (Parent, Parachain(2000), Parachain(4000), GeneralIndex(42)).into();
let ancestry = Parachain(2000).into();
let target = (Parent, Parachain(1000), Parachain(3000)).into();
let expected = (Parent, Parent, Parachain(4000), GeneralIndex(42)).into();
id.reanchor(&target, &ancestry).unwrap();
assert_eq!(id, expected);
```
I got 
```
left: `MultiLocation { parents: 2, interior: X3(Parachain(2000), Parachain(4000), GeneralIndex(42)) }`,
 right: `MultiLocation { parents: 2, interior: X2(Parachain(4000), GeneralIndex(42)) }`', xcm/src/v1/multilocation.rs:952:9
```
Do we want the `Parachain(2000)` to be repeated (it is in the ancestor and the multilocation of the reanchored id now)? I assumed not, since visiting it twice would be redundant, but maybe I'm missing some context. It also seems like there's not validation on the `ancestry` we pass into `reanchor`. How do we know that this `ancestry` is valid?

KiChjang:
It looks to me that the `id` is not simplified, and here's why: `ancestry` is the path that the relay chain must take in order to get to our current consensus system, and so we can see that we're on parachain 2000. The ancestry should therefore not contain any parent junctions. We don't have any validation functions to check for a properly formed `ancestry`, but perhaps we can do so by making sure that the `ancestry` only has interior junctions and no parents, but it's not really the scope of this PR however.

[...]

PR #4697, opened 2022-01-12
Emma/optimize reanchor

emmaling27:
This PR improves the algorithm for `reanchor` - see https://github.com/paritytech/polkadot/issues/4489 for more details.

Instead of finding the common ancestor, appending junctions, and simplifying, we instead handle the cases where `target` and `self` are not on the same path directly by adding parents and `ancestry` as a junction if needed, and then in the case of `target` and `self` on the same branch, walk down both paths until they split, return the number of parents as the number of remaining junctions to reach `target` and the remaining `Junctions` to reach `self` as the new path. I increased test coverage, too.

### DIFF
--- a/xcm/src/v1/multilocation.rs
+++ b/xcm/src/v1/multilocation.rs
@@ -335,7 +335,7 @@ impl MultiLocation {
 			for i in 0..target.interior.len() {
 				if target.interior.at(i) != self.interior.at(i) {
 					first_different = Some(i);
-					break;
+					break
 				}
 			}
 
@@ -356,7 +356,9 @@ impl MultiLocation {
 		} else if target.parents > self.parents {
 			// Handle when the target has more parents so we have to go through ancestry
 			self.parents = target.len() as u8 - 1;
-			self.interior.push_front(ancestry.interior.first().unwrap_or(&Junction::OnlyChild).clone()).map_err(|_| ())?;
+			self.interior
+				.push_front(ancestry.interior.first().unwrap_or(&Junction::OnlyChild).clone())
+				.map_err(|_| ())?;
 		} else if target.parents < self.parents {
 			// Handle when the id has more parents so we must go all the way up the target path
 			self.parents = self.parents + target.interior().len() as u8;
--- a/xcm/src/v1/multilocation.rs
+++ b/xcm/src/v1/multilocation.rs
@@ -17,7 +17,11 @@
 //! Cross-Consensus Message format data structures.
 
 use super::Junction;
-use core::{convert::TryFrom, mem, result};
+use core::{
+	cmp::max,
+	convert::TryFrom,
+	mem, result,
+};
 use parity_scale_codec::{Decode, Encode};
 use scale_info::TypeInfo;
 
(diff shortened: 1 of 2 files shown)


xcm/src/v1/multilocation.rs
```diff
@@ -329,18 +329,40 @@ impl MultiLocation {
...
+				// In the case where the paths are the same all the way to the target,
+				// remove all the junctions on the shared path
+				for _ in 0..target.interior.len() {
+					self.parents = 0;
+					self.take_first_interior();
+				}
+			}
+		} else if target.parents > self.parents {
+			// Handle when the target has more parents so we have to go through ancestry
+			self.parents = target.len() as u8 - 1;
```
KiChjang, line 358: This is assuming that the `target` only has 1 more parent than `self`, which isn't always the case. I expect to see a loop here in fact.

xcm/src/v1/multilocation.rs
```diff
@@ -329,18 +329,40 @@ impl MultiLocation {
...
+			self.interior
+				.push_front(ancestry.interior.first().unwrap_or(&Junction::OnlyChild).clone())
```
KiChjang, line 360: This is only grabbing the first junction off of the ancestry? That doesn't seem correct to me. Could you try creating test cases where the `ancestry` isn't just `Parachain(2000)`? For example, imagine we're on a smart contract, and so the `ancestry` could look something like `Parachain(2022)/PalletInstance(50)/GeneralIndex(32)`.

xcm/src/v1/multilocation.rs
```diff
@@ -329,18 +329,40 @@ impl MultiLocation {
...
+				.map_err(|_| ())?;
+		} else if target.parents < self.parents {
+			// Handle when the id has more parents so we must go all the way up the target path
+			self.parents = self.parents + target.interior().len() as u8;
```
KiChjang, line 364: If `self`/`id` has more parents, then it doesn't sound like it's simply an addition here, I'd expect to see something like `(self.parents - target.parents) + target.interior().len() as u8`. The `self.parents - target.parents` part is the path used by the `target` to reach the part of `ancestry` in which `self`/`id` actually branches off from and starts concatenating interior junctions.

xcm/src/v1/multilocation.rs
emmaling27: I was actually wondering about this but then I forgot to leave a comment! Thanks!

KiChjang:
@emmaling27 I appreciate the effort in trying to understand how `MultiLocation`s and the reanchoring algorithm works! This PR however still needs more work.

On a top-level view, I'd like to keep things a bit more declarative as possible, i.e. giving labels and names to algorithms and procedures, given the difficultly of understanding `MultiLocation` operations. What it still looks like to me, regardless of how `self.parents` compares with `target.parents`, is that the code is still trying to look for the common ancestor between `self` and `parent`, so there's no reason not to create a new internal function called `mutate_to_common_ancestor` or something similar just for this specific step in the algorithm. From there on, you can then append the remaining junctions to reach the final `MultiLocation`.

xcm/src/v1/multilocation.rs
emmaling27: Good point, I made a false assumption that the ancestry only had once junction, which made it hard to reason about the multiple parents as well. I'm confused at what the multilocation ancestry is relative to - will ancestry ever have parents > 0? Are `self` and `target` MultiLocations relative to the first junction in `ancestry` or the last one?

emmaling27: Also, I'm not sure what we would want to default to if there isn't an anything in `ancestry.interior`. I feel like we should error here, but the way I'm familiar with doing error handling in rust would require me to change the fn signature to be `Result<(), Error>` and I haven't seen that elsewhere in this file. How does Result<(), ()> work? From my perspective, this gives you no information on how something failed, right?

KiChjang: > will ancestry ever have parents

No, as I have mentioned in the issue, `ancestry` is not supposed to have parents, and in fact in an upcoming XCMv3 PR, we have specifically typed it as `InteriorMultiLocation`, a change from just `MultiLocation`.

The default of using `OnlyChild` looks fine to me.

Issue #4675, opened 2022-01-08
Add explicit logs for node authority status changes.


sandreim:
We've discovered that we don't have any logs to indicate if a node changes it's `authority` status. This information would have saved us a lot of debugging time.

bkchr:
Why isn't that being tracked by some external tool? A node doesn't really know if it is an authority, because it doesn't know its stash account.

PR #4699, opened 2022-01-12
Add node authority status metric

sandreim:
Fixes https://github.com/paritytech/polkadot/issues/4675

### DIFF
--- a/node/core/chain-selection/src/lib.rs
+++ b/node/core/chain-selection/src/lib.rs
@@ -22,7 +22,9 @@ use polkadot_node_subsystem::{
 	messages::{ChainApiMessage, ChainSelectionMessage},
 	overseer, FromOverseer, OverseerSignal, SpawnedSubsystem, SubsystemContext, SubsystemError,
 };
+use polkadot_node_subsystem_util::Validator;
 use polkadot_primitives::v1::{BlockNumber, ConsensusLog, Hash, Header};
+use sp_keystore::SyncCryptoStorePtr;
 
 use futures::{channel::oneshot, future::Either, prelude::*};
 use kvdb::KeyValueDB;
@@ -307,13 +309,23 @@ pub struct Config {
 pub struct ChainSelectionSubsystem {
 	config: Config,
 	db: Arc<dyn KeyValueDB>,
+	keystore: SyncCryptoStorePtr,
+}
+
+/// A structure to pass arguments to the subsystem main loop.
+struct ChainSelectionSubsystemArgs<'a, Context, B> {
+	ctx: &'a mut Context,
+	backend: &'a mut B,
+	stagnant_check_interval: &'a StagnantCheckInterval,
+	clock: &'a (dyn Clock + Sync),
+	keystore: SyncCryptoStorePtr,
 }
 
 impl ChainSelectionSubsystem {
 	/// Create a new instance of the subsystem with the given config
 	/// and key-value store.
-	pub fn new(config: Config, db: Arc<dyn KeyValueDB>) -> Self {
-		ChainSelectionSubsystem { config, db }
+	pub fn new(config: Config, db: Arc<dyn KeyValueDB>, keystore: SyncCryptoStorePtr) -> Self {
+		ChainSelectionSubsystem { config, db, keystore }
 	}
 }
 
@@ -329,9 +341,15 @@ where
 		);
 
 		SpawnedSubsystem {
-			future: run(ctx, backend, self.config.stagnant_check_interval, Box::new(SystemClock))
-				.map(Ok)
-				.boxed(),
+			future: run(
+				ctx,
+				backend,
+				self.config.stagnant_check_interval,
+				Box::new(SystemClock),
+				self.keystore,
+			)
+			.map(Ok)
+			.boxed(),
 			name: "chain-selection-subsystem",
 		}
 	}
@@ -342,13 +360,21 @@ async fn run<Context, B>(
 	mut backend: B,
 	stagnant_check_interval: StagnantCheckInterval,
 	clock: Box<dyn Clock + Send + Sync>,
+	keystore: SyncCryptoStorePtr,
 ) where
 	Context: SubsystemContext<Message = ChainSelectionMessage>,
 	Context: overseer::SubsystemContext<Message = ChainSelectionMessage>,
 	B: Backend,
 {
 	loop {
-		let res = run_until_error(&mut ctx, &mut backend, &stagnant_check_interval, &*clock).await;
+		let res = run_until_error(ChainSelectionSubsystemArgs {
+			ctx: &mut ctx,
+			backend: &mut backend,
+			stagnant_check_interval: &stagnant_check_interval,
+			clock: &*clock,
+			keystore,
+		})
+		.await;
 		match res {
 			Err(e) => {
 				e.trace();
@@ -368,18 +394,20 @@ async fn run<Context, B>(
 //
 // A return value of `Ok` indicates that an exit should be made, while non-fatal errors
 // lead to another call to this function.
-async fn run_until_error<Context, B>(
-	ctx: &mut Context,
-	backend: &mut B,
-	stagnant_check_interval: &StagnantCheckInterval,
-	clock: &(dyn Clock + Sync),
+async fn run_until_error<'a, Context, B>(
+	args: ChainSelectionSubsystemArgs<'a, Context, B>,
 ) -> Result<(), Error>
 where
 	Context: SubsystemContext<Message = ChainSelectionMessage>,
 	Context: overseer::SubsystemContext<Message = ChainSelectionMessage>,
 	B: Backend,
 {
+	let ChainSelectionSubsystemArgs { ctx, backend, stagnant_check_interval, clock, keystore } =
+		args;
+
 	let mut stagnant_check_stream = stagnant_check_interval.timeout_stream();
+	let mut is_validator = false;
+	let mut sender = ctx.sender().clone();
 	loop {
 		futures::select! {
 			msg = ctx.recv().fuse() => {
@@ -398,6 +426,23 @@ where
 							).await?;
 
 							backend.write(write_ops)?;
+
+							if is_validator != Validator::new(leaf.hash, keystore.clone(), &mut sender).await.is_ok() {
+								is_validator = !is_validator;
+								tracing::info!(
+									"👮 Authority status changed to `{}` at block #{}({})",
+									is_validator,
+									leaf.number,
+									leaf.hash
+								);
+							} else {
+								tracing::info!(
+									"🚔 Authority status is `{}` at block #{}({})",
+									is_validator,
+									leaf.number,
+									leaf.hash
+								);
+							}
 						}
 					}
 					FromOverseer::Signal(OverseerSignal::BlockFinalized(h, n)) => {
--- a/node/core/chain-selection/src/lib.rs
+++ b/node/core/chain-selection/src/lib.rs
@@ -22,9 +22,7 @@ use polkadot_node_subsystem::{
 	messages::{ChainApiMessage, ChainSelectionMessage},
 	overseer, FromOverseer, OverseerSignal, SpawnedSubsystem, SubsystemContext, SubsystemError,
 };
-use polkadot_node_subsystem_util::Validator;
 use polkadot_primitives::v1::{BlockNumber, ConsensusLog, Hash, Header};
-use sp_keystore::SyncCryptoStorePtr;
 
 use futures::{channel::oneshot, future::Either, prelude::*};
 use kvdb::KeyValueDB;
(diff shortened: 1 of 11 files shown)


eskimor (commented):
While there is no "perfect" subsystem to add these generic logs, I would prefer to use one where we are already doing the check. gossip-support would make the "most" sense to me. Any reason why you added the logs to chain selection in particular?

sandreim:
> While there is no "perfect" subsystem to add these generic logs, I would prefer to use one where we are already doing the check. gossip-support would make the "most" sense to me. Any reason why you added the logs to chain selection in particular?

I would say it was path of least resistance given the fact that I tried overseer, runtime-api and a new tiny subsystem just for this. 

That being said, you are right, gossip support seems the way to go here as the check is already done in there.

node/core/chain-selection/src/lib.rs
```diff
@@ -398,6 +426,23 @@ where
...
+									leaf.number,
+									leaf.hash
+								);
+							} else {
+								tracing::info!(
+									"🚔 Authority status is `{}` at block #{}({})",
+									is_validator,
+									leaf.number,
+									leaf.hash
+								);
```
bkchr, line 444: Spamming the node output on every block isn't a nice idea.

node/core/chain-selection/src/lib.rs
```diff
@@ -398,6 +426,23 @@ where
 							).await?;
 
 							backend.write(write_ops)?;
+
+							if is_validator != Validator::new(leaf.hash, keystore.clone(), &mut sender).await.is_ok() {
+								is_validator = !is_validator;
+								tracing::info!(
+									"👮 Authority status changed to `{}` at block #{}({})",
```
bkchr, line 433: This is also wrong on node startup, because the authority status didn't changed. Just your boolean changes here.

bkchr (commented):
I'm still not convinced that we need this. However, if you really insist on this, why not add a prometheus metric for this?

node/core/chain-selection/src/lib.rs
sandreim: Ah, this should have been a trace.

node/core/chain-selection/src/lib.rs
sandreim: This is as intended - it confirms the node is in the authority set. The semantics are off indeed when the node starts, but for debugging purposes this still has value.

ordian:
> That being said, you are right and I totally missed this - gossip support seems the way to go here as the check is already done in there.

Note that in gossip-support we're querying `discovery_keys`, which also past session validators. To check only the current session, we can also query the current `n_validators` (e.g. by querying `util::request_validators`) and check if `our_index is < n_validators`.

ordian:
Would it be good to add a prometheus metric for this as well?

sandreim:
> Would it be good to add a prometheus metric for this as well?

IMO the log is enough, but if folks things this is useful I'll add it.

node/core/chain-selection/src/lib.rs
eskimor: Uh, i somehow thought this is only triggered when the status changes - then `info` should be fine as well.

sandreim: This is triggered when it doesn't change so it's spamming as Basti pointed out. If you think it's not useful I'll remove it.

eskimor: got it - else should be `trace`, or removed -  yep.

drahnr: Remove please, it's another trace bloater. The issue we had is really a blue-moon one.

node/core/chain-selection/src/lib.rs
bkchr: Nodes don't restart that often normally (assuming they are not crashing because of a bug), so you will need to search in loki for this piece of log line. 

I still don't see the advantage. With a prometheus metric you could at least see the value directly.

sandreim: IMO it's reasonable to search the logs for this piece of information, but I agree that metrics would convey the same information more efficiently. I am a bit on the fence, but it seems like consensus on this PR is that this should be a metric.

node/core/chain-selection/src/lib.rs
sandreim: done!

node/network/gossip-support/src/metrics.rs
```diff
@@ -33,6 +33,12 @@ struct MetricsInner {
 }
 
 impl Metrics {
+	#[cfg(test)]
+	/// Dummy constructor for testing.
```
drahnr, line 37: usually we reverse the order
```
/// Foo
#[cfg(test)]
`

File: node/network/gossip-support/src/lib.rs [hunk 4]
```diff
@@ -212,7 +223,14 @@ where
 				}
 
 				let all_authorities = determine_relevant_authorities(ctx, relay_parent).await?;
-				let our_index = ensure_i_am_an_authority(&self.keystore, &all_authorities).await?;
+				let our_index = ensure_i_am_an_authority(&self.keystore, &all_authorities)
+					.await
+					.map_err(|e| {
+						self.metrics.on_is_not_authority();
```
eskimor, line 229: I find side effects in `map` functions rather ugly - I think an explicit pattern match would be better and more readable, also hardly more code.

The other thing is, as @ordian mentioned - this won't do exactly what you want as `ensure_i_am_an_authority` will check whether we are an authority in a couple of sessions, not just the current one.

sandreim: > I find side effects in `map` functions rather ugly - I think an explicit pattern match would be better and more readable, also hardly more code.
I agree in the general case, but in this particular one there are no side effects, it's a metric increment which is on the same level with logging wrt side effects (state). Initially I went for the match, but I like it better this way. If you have a hard preference here I'll switch to match.

> The other thing is, as @ordian mentioned - this won't do exactly what you want as `ensure_i_am_an_authority` will check whether we are an authority in a couple of sessions, not just the current one.

Yeah, that makes sense, but my thinking was this doesn't need to be very accurate so, assuming couple of sessions is not hours it might be OK. @ordian @eskimor how many blocks does this translate to currently ?

ordian (approved):
LGTM

node/network/gossip-support/src/lib.rs
```diff
@@ -209,6 +220,12 @@ where
 						"New session detected",
 					);
 					self.last_session_index = Some(session_index);
+
+					// Update the authority status metric on every new session.
+					match util::Validator::new(leaf, self.keystore.clone(), ctx.sender()).await {
+						Ok(_) => self.metrics.on_is_authority(),
+						Err(_) => self.metrics.on_is_not_authority(),
```
ordian, line 227: we should probably ignore runtime errors

sandreim: definitely

node/network/gossip-support/src/metrics.rs
```diff
@@ -0,0 +1,67 @@
...
+
+	/// Set the authority flag.
+	pub fn on_is_authority(&self) {
+		if let Some(metrics) = &self.0 {
+			metrics.is_authority.set(1);
+		}
+	}
+
+	/// Unset the authority flag.
+	pub fn on_is_not_authority(&self) {
```
drahnr, line 50:
```suggestion
	pub fn on_role_is_not_authority(&self) {
```

node/network/gossip-support/src/lib.rs
```diff
@@ -209,6 +220,14 @@ where
 						"New session detected",
 					);
 					self.last_session_index = Some(session_index);
+
+					// Update the authority status metric on every new session.
+					match util::Validator::new(leaf, self.keystore.clone(), ctx.sender()).await {
+						Ok(_) => self.metrics.on_is_authority(),
```
drahnr, line 226:
```suggestion
						Ok(_) => self.metrics.on_role_is_authority(),
```

node/network/gossip-support/src/lib.rs
```diff
@@ -209,6 +220,14 @@ where
...
+						Err(util::Error::NotAValidator) => self.metrics.on_is_not_authority(),
```
drahnr, line 227:
```suggestion
						Err(util::Error::NotAValidator) => self.metrics.on_role_is_not_authority(),
```

drahnr (approved):
Minor bike shed nit, other than that, LGTM :+1:

node/network/gossip-support/src/metrics.rs
```diff
@@ -0,0 +1,67 @@
...
+
+impl Metrics {
+	/// Dummy constructor for testing.
+	#[cfg(test)]
+	pub fn new_dummy() -> Self {
+		Self(None)
+	}
+
+	/// Set the authority flag.
+	pub fn on_is_authority(&self) {
```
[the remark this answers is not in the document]
ordian: I think authority is too ambiguous, what we are checking is "parachain validator".
Authority also includes block producers outside of parachain validators.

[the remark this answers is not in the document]
sandreim: Yes, but "parachain validator" doesn't sound accurate to me, since the validator(authority) in our context is a block producer which includes validating the parachain candidates.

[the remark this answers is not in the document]
ordian: Glossary[^1] actually says that "Parachain Validators" are the backing groups 
> A subset of validators assigned during a period of time to back candidates for a specific parachain

However, note that of all the relay chain validators (1000 on Kusama), we select maxValidators (200 on Kusama) and only these do approval-checking of parachains. The relay chain block produces may or may not be in this set at any given time. We may eventually set maxValidators to None, but I wanted to point out `authorities` is the superset of what we're checking.

[^1]: https://w3f.github.io/parachain-implementers-guide/glossary.html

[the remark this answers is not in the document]
sandreim: Thanks for posting this. That means I have to change the code so that we're checking the superset or update the metric definition and naming so it makes sense.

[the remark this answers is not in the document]
eskimor: I don't wan to bikeshed even more, but I find that `role_authority` irritating as in [other places](https://github.com/paritytech/polkadot/blob/22e5c04008dd85ddd36c4af74e3bdcdb20bb0645/node/service/src/lib.rs#L726) "authority role" just means the node has been started with `--validator`, regardless on whether the node has actual keys and in fact is a validator.

:grimacing:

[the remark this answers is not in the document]
eskimor: In my understanding we have been quite consistent in the code base to talk about authorities, meaning any potential block producer (has `AuthorityDiscoveryKey`) and (parachain-)validators having a key in `SessionInfo::validators`.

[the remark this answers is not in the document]
sandreim: IMO this metric should not actually be tracking the cli arguments. For this to fully make sense I'll have to include 2 metrics - one will say if the node is currently a relay chain validator while the other will say if the node is in the approval checking group. I'll fix the naming along the way to make it consistent with what I said above.

primitives/src/v2/mod.rs
```diff
@@ -28,7 +28,8 @@ use parity_util_mem::MallocSizeOf;
 
 /// Information about validator sets of a session.
 #[derive(Clone, Encode, Decode, RuntimeDebug, TypeInfo)]
-#[cfg_attr(feature = "std", derive(PartialEq, MallocSizeOf))]
+#[cfg_attr(feature = "std", derive(PartialEq, MallocSizeOf, Default))]
+#[cfg_attr(all(feature = "std", test), derive(Default))]
```
drahnr, line 32: No default for test, add an explicit `dummy_session_info` to the `primitives-test` crate

node/network/gossip-support/src/metrics.rs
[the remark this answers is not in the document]
sandreim: done, please take another look

primitives/src/v2/mod.rs
```diff
@@ -28,7 +28,8 @@ use parity_util_mem::MallocSizeOf;
 
 /// Information about validator sets of a session.
 #[derive(Clone, Encode, Decode, RuntimeDebug, TypeInfo)]
-#[cfg_attr(feature = "std", derive(PartialEq, MallocSizeOf))]
+#[cfg_attr(feature = "std", derive(PartialEq, MallocSizeOf, Default))]
```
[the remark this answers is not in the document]
drahnr: See here: https://github.com/paritytech/polkadot/pull/4699#discussion_r784090200

primitives/src/v2/mod.rs
sandreim: Forgot to fix this one before pushing ...

node/network/gossip-support/src/tests.rs
```diff
@@ -254,6 +275,17 @@ fn issues_a_connection_request_on_new_session() {
...
 		test_neighbors(overseer).await;
 
+		assert_matches!(
+			overseer_recv(overseer).await,
+			AllMessages::RuntimeApi(RuntimeApiMessage::Request(
+				relay_parent,
+				RuntimeApiRequest::SessionInfo(1, sender),
+			)) => {
+				assert_eq!(relay_parent, hash);
+				sender.send(Ok(Some(SESSION_INFO.clone()))).unwrap();
```
ordian, line 285: alternatively, you can simply send `None` here, doesn't seem to be used anyway

node/network/gossip-support/src/metrics.rs
```diff
@@ -0,0 +1,90 @@
...
+/// Dispute Distribution metrics.
+#[derive(Clone, Default)]
+pub struct Metrics(Option<MetricsInner>);
+
+#[derive(Clone)]
+struct MetricsInner {
+	/// Tracks authority status for producing relay chain blocks.
+	is_authority: Gauge<U64>,
+	/// Tracks authority status for parachain approval checking.
+	is_parachain_validator: Gauge<U64>,
```
ordian, line 34: It's possible to have an inconsistent state with `is_authority: 0, is_parachain_validator: 1`. But I don't know a better way of expressing an enum with prometheus (could be one metrics with 3 values, but less readable).

ordian (approved):

node/network/gossip-support/src/metrics.rs
sandreim: The flow prevents that, because if we do not set` is_authority` to 1 we no longer check if it's a parachain validator.

eskimor (approved):

node/network/gossip-support/src/metrics.rs
sandreim: the impl was wrong, as if we are no longer authority it wouldn't uncheck the parachain validator status. Fixed it.

node/network/gossip-support/src/tests.rs
sandreim: done

ordian:
bot merge

ordian:
bot merge

PR #4710, opened 2022-01-13
Migration for phragmen election locks migration

emostov:
Applies the migration from https://github.com/paritytech/substrate/pull/10649 to polkadot and kusama. They current plan is to apply this migration in the runtime after https://github.com/paritytech/substrate/pull/10646 is included. This way we can be sure the list of accounts to migrate will not grow because the bug allowing it will have been removed.

References the bug fixed in https://github.com/paritytech/substrate/pull/10646

### DIFF
--- a/runtime/polkadot/src/lib.rs
+++ b/runtime/polkadot/src/lib.rs
@@ -1627,7 +1627,7 @@ impl OnRuntimeUpgrade for FixCouncilDepositMigration {
 
 	#[cfg(feature = "try-runtime")]
 	fn pre_upgrade() -> Result<(), &'static str> {
-		Self::execute(true);
+		let _ = Self::execute(true);
 		Ok(())
 	}
 }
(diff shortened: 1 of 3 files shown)


runtime/polkadot/src/lib.rs
```diff
@@ -1626,7 +1627,8 @@ impl OnRuntimeUpgrade for FixCouncilDepositMigration {
 
 	#[cfg(feature = "try-runtime")]
 	fn pre_upgrade() -> Result<(), &'static str> {
-		Self::execute(true)
+		Self::execute(true);
+		Ok(())
```
emostov, line 1631: The changes here and the parallel code in the kusama runtime should be superseded by whatever gets merged by https://github.com/paritytech/polkadot/pull/4692

kianenigma (approved):

emostov:
Closing until appropriate merge

PR #4712, opened 2022-01-13
Remove `Default` bound on `AccountId` types under the xcm directory

### LINKED ISSUE #4687: Implications of Using `AccountId::default()` as parent `AccountId` for Location-Conversion
Given a note on the default behaviour of `AccountId32`

Quote:
> We recently discovered that anyone can sign transactions in Substrate with a zero account (AccountId::default()). The reason behind this is that cryptographically a private key of zero makes to an actual complete curve in sr25519 and ed255519. This is coming directly from schnorrkel.

Resolved for substrate via https://github.com/paritytech/substrate/pull/10403

Aren't the following lines a bit dangerous, given that many parachains will most likely plug in the normal `AccountId32` type there?

https://github.com/paritytech/polkadot/blob/5a7004f22b43e46e73ed36c87c9159283e025a56/xcm/xcm-builder/src/location_conversion.rs#L40-L59

KiChjang:
Fixes #4687.

cumulus companion: paritytech/cumulus#901

### DIFF
--- a/xcm/xcm-builder/src/currency_adapter.rs
+++ b/xcm/xcm-builder/src/currency_adapter.rs
@@ -54,7 +54,7 @@ impl From<Error> for XcmError {
 /// ```
 /// use frame_support::parameter_types;
 /// use xcm::latest::prelude::*;
-/// use xcm_builder::{ParentIsDefault, CurrencyAdapter, IsConcrete};
+/// use xcm_builder::{ParentIsAllZeroes, CurrencyAdapter, IsConcrete};
 ///
 /// /// Our chain's account id.
 /// type AccountId = sp_runtime::AccountId32;
@@ -67,7 +67,7 @@ impl From<Error> for XcmError {
 ///
 /// /// Some items that implement `Convert<MultiLocation, AccountId>`. Can be more, but for now we just assume we accept
 /// /// messages from the parent (relay chain).
-/// pub type LocationConvertor = (ParentIsDefault<RelayChain>);
+/// pub type LocationConvertor = (ParentIsAllZeroes<RelayChain>);
 ///
 /// /// Final currency adapter. This can be used in `xcm::Config` to specify how asset related transactions happen.
 /// pub type AssetTransactor = CurrencyAdapter<
--- a/xcm/xcm-builder/src/currency_adapter.rs
+++ b/xcm/xcm-builder/src/currency_adapter.rs
@@ -52,7 +52,9 @@ impl From<Error> for XcmError {
 ///
 /// # Example
 /// ```
+/// use parity_scale_codec::Decode;
 /// use frame_support::parameter_types;
+/// use sp_runtime::traits::TrailingZeroInput;
 /// use xcm::latest::prelude::*;
 /// use xcm_builder::{ParentIsAllZeroes, CurrencyAdapter, IsConcrete};
 ///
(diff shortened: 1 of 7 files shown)


xcm/xcm-builder/src/location_conversion.rs
```diff
@@ -35,22 +35,26 @@ impl<Network: Get<NetworkId>, AccountId: From<[u8; 32]> + Into<[u8; 32]> + Clone
...
+/// for `AccountId`.
+pub struct ParentIsAllZeroes<AccountId>(PhantomData<AccountId>);
+impl<AccountId: Decode + Eq + Clone> Convert<MultiLocation, AccountId>
+	for ParentIsAllZeroes<AccountId>
 {
 	fn convert_ref(location: impl Borrow<MultiLocation>) -> Result<AccountId, ()> {
 		if location.borrow().contains_parents_only(1) {
-			Ok(AccountId::default())
+			Ok(AccountId::decode(&mut TrailingZeroInput::zeroes())
+				.expect("infinite length input; no invalid inputs for type; qed"))
```
bkchr, line 47: For what is this account being used?

KiChjang: It looks like it is only used in testing, in cases where the test wants to convert a message originating from a parent location to just the zero account, i.e. a "don't care but signed" account.

As such, there's not really a huge issue leaving it as `ParentIsDefault`, other than to keep in line with the `Default` bound removal on `AccountId`.

shawntabrizi: I think we should remove this trait from the public api and make it test only.

xlc: I think default account is always zero account.

bkchr: Yeah `Default == Zero`.

I'm also for removing this or doing it like @xlc proposed.

bkchr:
@KiChjang what is the status on this? I need this for my Substrate pr that removes the `Default` for `AccountId32`

KiChjang:
@bkchr I'll need to figure out what the parent accounts are for Statemint/e, currently in the test networks I still have them set to the zero account, and it is definitely not going to be the zero account on production networks.

shawntabrizi:
@KiChjang you could use something like `hash("parent_parachain")` or `hash("polkadot/kusama")` or something like that. Otherwise speak to Gav to finalize an answer.

xcm/xcm-builder/src/currency_adapter.rs
```diff
@@ -52,22 +52,26 @@ impl From<Error> for XcmError {
...
 ///
 /// /// Our chain's account id.
 /// type AccountId = sp_runtime::AccountId32;
 ///
 /// /// Our relay chain's location.
 /// parameter_types! {
 ///     pub RelayChain: MultiLocation = Parent.into();
-///     pub CheckingAccount: AccountId = Default::default();
+///     pub CheckingAccount: AccountId =
+///         AccountId::decode(&mut TrailingZeroInput::zeroes()).unwrap();
```
bkchr, line 68: Do we really want to have such "hacks" in the docs?

bkchr (approved):

xcm/xcm-builder/src/currency_adapter.rs
KiChjang: I don't know what else you can put into here, you want me to do the same thing with `PalletId` instead?

KiChjang:
```suggestion
///     pub CheckingAccount: AccountId = PalletId(*b"checking").into_account();
```

shawntabrizi (approved):

bkchr (approved):

KiChjang:
bot merge

PR #4716, opened 2022-01-13
staking miner: spawn separate task for each block

niklasad1:
This PR makes a couple of optimizations:
- move the extrinsic construction and remote-externalities (view) per block into a separate tokio task.
- handle `JsonRpseeError::RestartNeeded` explicitly if that occurs the connection is closed and the client is useless after that, except some buffered responses on subscriptions but it doesn't help because we need the account state (via RPC calls) which will fail.
- update jsonrpsee

However, I haven't been able to quantify that the optimizations compared to current one so you know, so it's just my hunch that it and  `headers` should be never be missed because of `buffer subscriptions` gets full and dropped.


utils/staking-miner/src/monitor.rs
```diff
@@ -64,120 +68,245 @@ async fn ensure_no_previous_solution<
...
-				if ensure_signed_phase::<Runtime, Block>(client, hash).await.is_err() {
-					log::debug!(target: LOG_TARGET, "phase closed, not interested in this block at all.");
-					continue;
-				};
+			let at = tokio::select! {
+				maybe_rp = subscription.next() => {
+					match maybe_rp {
+						Some(Ok(r)) => r,
+						// Custom `jsonrpsee` message; should not occur.
+						Some(Err(RpcError::SubscriptionClosed(reason))) => {
```
emostov, line 99: For the `Some(Err(_))` cases why not an error but still retry with `subscription = client.subscribe(&sub, None, &unsub).await?;`?

emostov: (Not saying we should change this, but just want to understand the rational)

utils/staking-miner/src/monitor.rs
```diff
@@ -64,120 +68,245 @@ async fn ensure_no_previous_solution<
...
+			shared: SharedConfig,
+			config: MonitorConfig,
+		) {
+
+			let hash = at.hash();
+			log::trace!(target: LOG_TARGET, "new event at #{:?} ({:?})", at.number, hash);
+
+			// if the runtime version has changed, terminate.
+			if let Err(err) = crate::check_versions::<Runtime>(&*client).await {
+				let _ = tx.send(err.into());
```
emostov, line 150: Just to confirm my understanding, this will send an error back to the thread that spawned this and that will be caught by the above line `maybe_err = rx.recv()`?

utils/staking-miner/src/monitor.rs
```diff
@@ -64,120 +68,245 @@ async fn ensure_no_previous_solution<
...
+			}
 
-				// grab an externalities without staking, just the election snapshot.
-				let mut ext = crate::create_election_ext::<Runtime, Block>(
-					shared.uri.clone(),
-					Some(hash),
-					vec![],
-				).await?;
+			// grab an externalities without staking, just the election snapshot.
+			let mut ext = match crate::create_election_ext::<Runtime, Block>(
```
emostov, line 161: Beyond the scope of this PR, but wondering if it makes sense to parallize `check_versions` & `ensure_signed_phase` & `create_election_ext` - so we can start creating the election_ext immediatley and if its not the correct version or the signed phase we can send a message to the thread of `create_election_ext` to stop?

utils/staking-miner/src/monitor.rs
niklasad1: good catch, yupp I agree that we should create a new subscription if this occurs.

currently this is unreachable anyways because it's a custom message we have introduced in jsonrpsee once the server closes down subscription. I don't think jsonrpc has this FWIW.

utils/staking-miner/src/monitor.rs
niklasad1: yeah, exactly.

utils/staking-miner/src/monitor.rs
niklasad1: Should be possible but to lowest hanging fruit is to re-use the existing `ws connection` and to pass `Arc<WsClient>` into `remote ext` instead of creating a new connection for each `create_election_ext` I think.

I think it takes a couple of seconds to establish a new connection.

emostov: Yeah makes sense - what you are saying might be a good issue for interviewing as well cc @kianenigma

emostov (approved):
lgtm

utils/staking-miner/src/monitor.rs
```diff
@@ -97,20 +97,20 @@ macro_rules! monitor_cmd_for { ($runtime:tt) => { paste::paste! {
 						Some(Ok(r)) => r,
 						// Custom `jsonrpsee` message; should not occur.
 						Some(Err(RpcError::SubscriptionClosed(reason))) => {
-							log::debug!("[rpc]: subscription closed by the server: {:?}, starting a new one", reason);
+							log::warn!(target: LOG_TARGET, "subscription to {} terminated: {:?}. Retrying..", reason, sub);
```
niklasad1, line 100:
```suggestion
							log::warn!(target: LOG_TARGET, "subscription to {} terminated: {:?}. Retrying..", sub, reason);
```

utils/staking-miner/README.md
```diff
@@ -59,3 +59,10 @@ docker run --rm -it \
     -e URI=wss://your-node:9944 \
     staking-miner dry-run
 ```
+
+### Test locally
```
kianenigma, line 63: 👍

utils/staking-miner/src/monitor.rs
```diff
@@ -64,120 +68,245 @@ async fn ensure_no_previous_solution<
...
+				},
+			};
+
+			while let Some(rp) = tx_subscription.next().await {
+				let status_update = match rp {
+					Ok(r) => r,
+					// Custom `jsonrpsee` message; should not occur.
+					Err(RpcError::SubscriptionClosed(reason)) => {
+						log::warn!(
+							"[rpc]: subscription closed by the server: {:?}; continuing...",
```
kianenigma, line 249:
```suggestion
						log::warn!(
							target: LOG_TARGET,
							"[rpc]: subscription closed by the server: {:?}; continuing...",
```

kianenigma: and in other places, otherwise we might miss them in log filtering.

kianenigma (approved):
Looks broadly good to me. If tested and works in westend, then can merge ahead. Thanks!

kianenigma:
> However, I haven't been able to quantify that the optimizations compared to current one so you know, so it's just my hunch that it and headers should be never be missed because of buffer subscriptions gets full and dropped.

All in all, I would argue that for now we don't necessarily need to care too much about optimisations etc. At this point, my main concern would be for the code to be maintainable. The two main steps to achieve this are: 

1. Make it independent of the runtime (which is our next objective) 
2. Keep the code relatively simple and clean (which I can say IMO is and this PR also helps).

niklasad1:
Yeah, you made that clear. Yes tested both on the local dev chain and westend but not remember which commit of this I tested.

I realized that the logs from `jsonrpsee` doesn't work anymore here because `jsonrpsee` migrated to `tracing` should fine to skip for now?

PR #4719, opened 2022-01-14
Refactor beefy-mmr implementation

Wizdave97:
Companion PR to https://github.com/paritytech/substrate/pull/10664

### DIFF
--- a/bridges/bin/millau/runtime/src/lib.rs
+++ b/bridges/bin/millau/runtime/src/lib.rs
@@ -32,7 +32,7 @@ pub mod rialto_messages;
 
 use crate::rialto_messages::{ToRialtoMessagePayload, WithRialtoMessageBridge};
 
-use beefy_primitives::{crypto::AuthorityId as BeefyId, mmr::{MmrLeafVersion, BeefyDataProvider}, ValidatorSet};
+use beefy_primitives::{crypto::AuthorityId as BeefyId, mmr::{MmrLeafVersion}, ValidatorSet};
 use bridge_runtime_common::messages::{
 	source::estimate_message_dispatch_and_delivery_fee, MessageBridge,
 };
@@ -278,19 +278,10 @@ parameter_types! {
 	pub LeafVersion: MmrLeafVersion = MmrLeafVersion::new(0, 0);
 }
 
-pub struct ParachainHeadsProvider;
-
-impl BeefyDataProvider for ParachainHeadsProvider {
-	fn extra_data() -> Vec<u8> {
-		Vec::new()
-	}
-}
-
-
 impl pallet_beefy_mmr::Config for Runtime {
 	type LeafVersion = LeafVersion;
 	type BeefyAuthorityToMerkleLeaf = pallet_beefy_mmr::BeefyEcdsaToEthereum;
-	type BeefyDataProvider = ParachainHeadsProvider;
+	type BeefyDataProvider = ();
 }
 
 parameter_types! {
--- a/bridges/bin/millau/runtime/src/lib.rs
+++ b/bridges/bin/millau/runtime/src/lib.rs
@@ -32,7 +32,7 @@ pub mod rialto_messages;
 
 use crate::rialto_messages::{ToRialtoMessagePayload, WithRialtoMessageBridge};
 
-use beefy_primitives::{crypto::AuthorityId as BeefyId, mmr::MmrLeafVersion, ValidatorSet};
+use beefy_primitives::{crypto::AuthorityId as BeefyId, mmr::{MmrLeafVersion, BeefyDataProvider}, ValidatorSet};
 use bridge_runtime_common::messages::{
 	source::estimate_message_dispatch_and_delivery_fee, MessageBridge,
 };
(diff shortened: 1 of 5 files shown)


bridges/bin/millau/runtime/src/lib.rs
```diff
@@ -279,10 +279,19 @@ parameter_types! {
...
+		Vec::new()
+	}
+}
+
+
 impl pallet_beefy_mmr::Config for Runtime {
 	type LeafVersion = LeafVersion;
 	type BeefyAuthorityToMerkleLeaf = pallet_beefy_mmr::BeefyEcdsaToEthereum;
-	type ParachainHeads = ();
+	type BeefyDataProvider = ParachainHeadsProvider;
```
seunlanlege, line 294: actually provide a `BeefyDataProvider` implementation for `()` in the substrate repo

acatangiu (approved):
Looks good!

runtime/rococo/src/lib.rs
```diff
@@ -741,10 +736,25 @@ parameter_types! {
 	pub LeafVersion: MmrLeafVersion = MmrLeafVersion::new(0, 0);
 }
 
+pub struct ParasProvider;
+impl BeefyDataProvider for ParasProvider {
+	fn extra_data() -> Vec<u8> {
+		let mut para_heads = Paras::parachains()
```
acatangiu, line 742: To fix build:

```suggestion
		let mut para_heads: Vec<(u32, Vec<u8>)> = Paras::parachains()
```

runtime/rococo/src/lib.rs
```diff
@@ -741,10 +736,25 @@ parameter_types! {
 	pub LeafVersion: MmrLeafVersion = MmrLeafVersion::new(0, 0);
 }
 
+pub struct ParasProvider;
```
svyatonik, line 739: @acatangiu This is a breaking change. Do we need a runtime storage upgrade here? What are possible outcomes of not having an upgrade?

What I mean: previously MMR leaf was explicitly storing parachain heads merkle root (`pub parachain_heads: MerkleRoot`), which is a `H256` iiuc. Scale codes is encoding `H256` without any prefix - so `H256([0u8; 32])` will be encoded as `[0u8; 32]`. Now leaf has some extra data (`pub leaf_extra: Vec<u8>`), which is encoded with prefix - i.e. `vec![0u8; 32]` will be stored as `[32, 0, 0, ..., 0, 0]`. So iiuc all existing MMR hashes (roots, peaks) will break if this will be changed without upgrade.

There are several solutions imo:
1) ignore and let it break - I don't know about consequences, though. Can Rococo/BEEFY stall because of that? Do we care?
2) make a runtime upgrade. This would require update of all stored leafs. IIUC leafs are stored in offchain storage and hashes are stored in runtime storage. So it'll be painful to do an upgrade here;
3) somehow change leaf encoding to match previous. I have no idea now how to do that. Probably instead of storing `Vec<u8>`, we need to add generic parameter to leaf? I.e. change `MmrLeaf<BlockNumber, Hash, MerkleRoot>` to `MmrLeaf<BlockNumber, Hash, MerkleRoot, ExtraData>` and set this `ExtraData` to `H256` for Rococo

acatangiu: Good catch!

BEEFY is not running on Rococo afaik and we want to bring it back with Lean BEEFY (https://github.com/paritytech/grandpa-bridge-gadget/issues/350).

Even if it is running and we just make this change, it should only break BEEFY (and nothing else) for a while (until all nodes are updated), right? Since BEEFY is not used for anything right now and we don't even have alarms or anything I guess that's the easiest way.. wdyt?

svyatonik: > Even if it is running and we just make this change, it should only break BEEFY (and nothing else) for a while (until all nodes are updated), right? 

I don't know to be honest :) If it breaks BEEFY, then are you sure that BEEFY will be able to recover itself from this? I.e. it won't use any obsolete MMR leafs for anything + it won't use updated hashes to access obsolete leafs? If you're sure that it won't, then we're fine.

The fact that mmr leaf has version field in it, makes me think that the only possible problem is that we won't be able to verify or generate proofs for things that has happened before this upgrade (without some special treatment).

> Since BEEFY is not used for anything right now and we don't even have alarms or anything I guess that's the easiest way.. wdyt?

Honestly I'd keep previous encoding (so option 3 from above) :) It also has additional advantages imo - not storing that additional compact int (vector length), applying version field of the leaf structure to all fields (see https://github.com/paritytech/substrate/pull/10664#discussion_r838487623), no need for double encoding/decoding (I mean - now you need to decode/encode the leaf itself first and then to decode/encode extra data). But I'm not a pro here, so I'd prefer to leave this decision to you :)

acatangiu: > Honestly I'd keep previous encoding (so option 3 from above) :) It also has additional advantages imo - not storing that additional compact int (vector length), applying version field of the leaf structure to all fields (see https://github.com/paritytech/substrate/pull/10664#discussion_r838487623), no need for double encoding/decoding (I mean - now you need to decode/encode the leaf itself first and then to decode/encode extra data).

I agree.

acatangiu (approved):

svyatonik (approved):

PR #4732, opened 2022-01-17
XCM v3: Introduce querier field into `QueryReponse`

gavofyork:
Introduce an additional field into the `QueryResponse` XCM instruction `querier` indicating the original initiator which led to this response being sent (from the point of view of the receiver). Typically this will simply be `Here`, since the receiver of the response will be the one who originally made the query.

The XCM pallet's query response handler system now allows expected query responses to have have an expected initiator of the response. Generally this is set to `Here` when new queries are initiated.

Before any query response is executed using the XCM pallet's response-dispatch system, the `querier` field is matched against what is expected. If it differs, then the `ExpectedResponse` barrier doesn't pass and the response is ignored.

### DIFF
--- a/xcm/pallet-xcm/src/lib.rs
+++ b/xcm/pallet-xcm/src/lib.rs
@@ -1413,13 +1413,13 @@ pub mod pallet {
 	}
 
 	impl<T: Config> OnResponse for Pallet<T> {
-		fn expecting_response(origin: &MultiLocation, query_id: QueryId, querier: &Option<MultiLocation>) -> bool {
+		fn expecting_response(origin: &MultiLocation, query_id: QueryId, querier: Option<&MultiLocation>) -> bool {
 			match Queries::<T>::get(query_id) {
 				Some(QueryStatus::Pending { responder, maybe_match_querier, .. }) =>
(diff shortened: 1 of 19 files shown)


xcm/pallet-xcm/src/lib.rs
```diff
@@ -1375,7 +1494,33 @@ pub mod pallet {
...
+							Ok(mq) => mq,
+							Err(_) => {
+								Self::deposit_event(Event::InvalidQuerierVersion(
+									origin.clone(),
+									query_id,
+								));
+								return 0
+							},
+						};
+						if querier.map_or(false, |q| q != &match_querier) {
```
KiChjang, line 1514: Hmm, this is somewhat surprising... in this code branch, I already have a `match_querier`, meaning that I do expect that there should be a querier location, but the `map_or(false, ..)` here means that I can refuse to supply a querier location, and it would still pass through, instead of being rejected.

@viniul Could you verify that my logic above is correct?

gavofyork, line 1514:
```suggestion
						if querier.map_or(true, |q| q != &match_querier) {
```

KiChjang (approved):

PR #4736, opened 2022-01-18
Check that `try-runtime` compiles in CI

bkchr:
It also moves some simple scripts directly into the .gitlab.yaml file.


kianenigma (approved):
danke

PR #4745, opened 2022-01-19
Introduce ParentChainId

KiChjang:
This PR creates a new struct called ParentChainId, and it will be primarily used by parachains to create an AccountId for their parent chain, so e.g. Statemine will have a `ParentChainId(b"kusama".try_into().unwrap()).into_account()` to refer to the sovereign account of Kusama locally.

Same as paritytech/substrate#10695, but moved over here since `ParentChainId` fits the purpose of Polkadot much better.

### DIFF
--- a/runtime/common/src/lib.rs
+++ b/runtime/common/src/lib.rs
@@ -37,17 +37,20 @@ mod integration_tests;
 #[cfg(test)]
 mod mock;
 
+use parity_scale_codec::{Decode, Encode};
 pub use frame_support::weights::constants::{
 	BlockExecutionWeight, ExtrinsicBaseWeight, RocksDbWeight,
 };
 use frame_support::{
 	parameter_types,
+	storage::bounded_vec::BoundedVec,
 	traits::{ConstU32, Currency, OneSessionHandler},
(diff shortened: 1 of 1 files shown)


runtime/common/src/lib.rs
```diff
@@ -59,6 +62,15 @@ pub use pallet_timestamp::Call as TimestampCall;
 #[cfg(any(feature = "std", test))]
 pub use sp_runtime::BuildStorage;
 
+/// A parent blockchain identifier. Used primarily when parachains need to refer to a parent chain's
+/// `AccountId`.
+#[derive(Clone, Eq, PartialEq, Encode, Decode, scale_info::TypeInfo)]
+pub struct ParentChainId(pub BoundedVec<u8, ConstU32<20>>);
```
kianenigma, line 68: unclear to me why this is always a 20 byte identifier?

bkchr: `PalletId` also supports some fixed number of bytes.

bkchr: But I don't like that we need to have some unchecked call or similar here. This also doesn't really make any sense to use a bounded vec here?

KiChjang:
Doesn't seem required now.

PR #4753, opened 2022-01-20
Create a more rigid overseer builder pattern that fails at compile time

vstakhov:
This pull requests addresses #3772 

The general idea is to generalise all fields arguments and fix them in two states:

* `Uninit<T>` for uninitialised fields
* `Init<T>` for initialised fields

With this approach, each setter function accepts builder template with `Uninit<Field>` and returns a builder where this field is changed to `Init<Field>`. Subsequently `build` function accepts all templated fields to be `Init<T>` and fails to compile if any of the required fields is missing.


node/overseer/overseer-gen/proc-macro/src/impl_builder.rs
```diff
@@ -107,46 +61,379 @@ pub(crate) fn impl_builder(info: &OverseerInfo) -> proc_macro2::TokenStream {
...
+	// A helper structure to carry fields of a builder to produce setters
+	struct BuilderFieldHelper {
+		// Used to distinguish baggage and subsystems
+		is_subsystem: bool,
+		// Valid for baggage as all subsystems are generics
+		is_generic: bool,
+		// Field name
+		name: Ident,
+		// Field type (either generic or not)
+		field_ty: TokenStream,
```
drahnr, line 73: Please _never_ use `TokenStream` in a struct, when there is a better way of representing that as a type.

node/overseer/overseer-gen/proc-macro/src/impl_builder.rs
```diff
@@ -107,46 +61,379 @@ pub(crate) fn impl_builder(info: &OverseerInfo) -> proc_macro2::TokenStream {
 		})
 		.collect::<Vec<_>>();
 
-	let generics = quote! {
-		< S, #( #baggage_generic_ty, )* >
-	};
-	let where_clause = quote! {
+	// A helper structure to carry fields of a builder to produce setters
+	struct BuilderFieldHelper {
```
drahnr, line 65: This tries to unify across baggage fields (which might or might not have relevant generic) and subsystems (which have generic) which I don't think is a good idea. After all they have different trait bounds etc which causes the whole issue you mitigated by using `TokenStream`s.

node/overseer/overseer-gen/proc-macro/src/impl_builder.rs
```diff
@@ -107,46 +61,379 @@ pub(crate) fn impl_builder(info: &OverseerInfo) -> proc_macro2::TokenStream {
...
+		order: usize,
+	}
+
+	let mut field_helpers = info.subsystems()
+		.iter()
+		.filter(|ssf| !ssf.wip)
+		.enumerate()
+		.map(|(idx, ssf)| {
+			let prefixed_ident = format_ident!("__{}",
+				ssf.name.to_string().to_class_case());
```
drahnr, line 92: This is only necessary due to the abstraction chosen. If we can avoid this call, we can avoid an entire dependency package.

node/overseer/overseer-gen/proc-macro/src/impl_builder.rs
```diff
@@ -107,46 +61,379 @@ pub(crate) fn impl_builder(info: &OverseerInfo) -> proc_macro2::TokenStream {
...
+			let consumes = &ssf.consumes;
+			let generic_ty = &ssf.generic;
+			BuilderFieldHelper {
+				is_subsystem: true,
+				is_generic: true,
+				name: ssf.name.clone(),
+				field_ty: quote!(#generic_ty),
+				placeholder_type: quote! { #prefixed_ident },
+				where_clause: Some(quote! {
```
drahnr, line 101: you can still use the approach, but use https://docs.rs/syn/latest/syn/macro.parse_quote.html instead of `quote!`

node/overseer/overseer-gen/proc-macro/src/impl_builder.rs
```diff
@@ -107,46 +61,379 @@ pub(crate) fn impl_builder(info: &OverseerInfo) -> proc_macro2::TokenStream {
...
+		.collect::<Vec<_>>();
+	let where_clause_elts = field_helpers
+		.iter()
+		.filter_map(|fh| fh.where_clause.as_ref())
+		.collect::<Vec<_>>();
+
+	let builder_where_clause = quote! {
 		where
-			S: #support_crate ::SpawnNamed,
+			#( #where_clause_elts, )*
```
drahnr, line 157: what's elts?

node/overseer/overseer-gen/proc-macro/src/impl_builder.rs
```diff
@@ -107,46 +61,379 @@ pub(crate) fn impl_builder(info: &OverseerInfo) -> proc_macro2::TokenStream {
...
+						}
+						else {
+							quote! { #fname: OverseerFieldInit::<#var_type>(var) }
+						}
+					}
+					else {
+						quote!{ #fname: self.#fname }
+					}
+				})
+				.collect::<Vec<_>>();
```
drahnr, line 237: Lacking some abstraction here, this is quite some visual load. Impl `trait quote::ToTokens` would solve this and encapsulate this at the very least https://docs.rs/quote/latest/quote/trait.ToTokens.html

node/overseer/overseer-gen/proc-macro/src/impl_builder.rs
```diff
@@ -107,46 +61,379 @@ pub(crate) fn impl_builder(info: &OverseerInfo) -> proc_macro2::TokenStream {
...
+		.iter()
+		.map(|fh| &fh.inited_ty)
+		.collect::<Vec<_>>();
+	// Defines all fields to be set to uninit<T>
+	let uninit_builder_field_def = field_helpers
+		.iter()
+		.map(|fh| {
+			let (field_name, field_ty) = (&fh.name, &fh.field_ty);
+			if fh.is_subsystem {
+				quote! { #field_name: OverseerFieldUninit::<SubsystemFieldInitMethod<#field_ty>>::default() }
```
drahnr, line 362: Why is there a nesting here? Could we not make this one level and make `OverseerFieldInit` an `enum`?

node/overseer/overseer-gen/proc-macro/src/parse_struct.rs
```diff
@@ -280,15 +281,6 @@ impl OverseerInfo {
...
 	}
 
-	/// Generic types per subsystem, as defined by the user.
-	pub(crate) fn builder_generic_types(&self) -> Vec<Ident> {
-		self.subsystems
-			.iter()
-			.filter(|ssf| !ssf.wip)
-			.map(|sff| sff.generic.clone())
-			.collect::<Vec<_>>()
-	}
```
drahnr, line 290: I actually think that this level does not necessarily have to be changed. Happy to be convinced otherwise.

node/overseer/overseer-gen/tests/ui/err-05-missing-field.rs
```diff
@@ -0,0 +1,61 @@
...
+        unimplemented!("spawn {} {}", task_name, subsystem_name.unwrap_or("default"))
+    }
+}
+
+struct DummyCtx;
+
+fn main() {
+    let _ = Overseer::builder()
+        .sub0(AwesomeSubSys::default())
+        //.i_like_pie(std::f64::consts::PI) // The filed is not initialised
```
drahnr, line 57: :+1:

drahnr: Since we also wanted to cover double-init, adding an ui test for that as well would be much appreciated.

node/overseer/overseer-gen/proc-macro/src/impl_builder.rs
```diff
@@ -16,66 +16,23 @@
...
 ///
 /// Elements tagged with `wip` are not covered here.
 pub(crate) fn impl_builder(info: &OverseerInfo) -> proc_macro2::TokenStream {
 	let overseer_name = info.overseer_name.clone();
-	let builder = Ident::new(&(overseer_name.to_string() + "Builder"), overseer_name.span());
-	let handle = Ident::new(&(overseer_name.to_string() + "Handle"), overseer_name.span());
-	let connector = Ident::new(&(overseer_name.to_string() + "Connector"), overseer_name.span());
+	let builder = format_ident!("{}Builder", overseer_name);
+	let handle = format_ident!("{}Handle", overseer_name);
+	let connector = format_ident!("{}Connector", overseer_name);
```
drahnr, line 31: :+1:

drahnr (changes requested):
Good start! It requires some additional attention on the way things are represented, and - as you already said - a round of cleanups.

node/overseer/overseer-gen/proc-macro/src/impl_builder.rs
vstakhov: Yes, that was caused just because I wanted to unify baggage and subsystems. I will think about a better approach.

node/overseer/overseer-gen/proc-macro/src/impl_builder.rs
vstakhov: This is added to improve error messages (so they include the violated field name) and avoiding snake_case types.

node/overseer/overseer-gen/proc-macro/src/impl_builder.rs
vstakhov: Elements, meaning it is a vector of clauses.

node/overseer/overseer-gen/proc-macro/src/parse_struct.rs
vstakhov: It should actually be `subsystem_generic_types` then, not `builder_generic_types` as it covers merely subsystem and missing potential generics from baggage (and we have `baggage_generic_types` as well).

node/overseer/overseer-gen/proc-macro/Cargo.toml
```diff
@@ -16,6 +16,7 @@ syn = { version = "1.0.83", features = ["full", "extra-traits"] }
 quote = "1.0.14"
 proc-macro2 = "1.0.36"
 proc-macro-crate = "1.1.0"
+Inflector = "0.11.4"
```
drahnr, line 19:
```suggestion
```

node/overseer/overseer-gen/proc-macro/src/parse_struct.rs
vstakhov: Done in 5d0503c07fa70d6f19e8d754a5e2f17b165b0150

node/overseer/overseer-gen/proc-macro/src/impl_builder.rs
```diff
@@ -210,56 +203,53 @@ pub(crate) fn impl_builder(info: &OverseerInfo) -> proc_macro2::TokenStream {
...
+	let baggage_specific_setters = info.baggage().iter().enumerate().map(|(idx, bag_field)| {
+		// Baggage fields follow subsystems
+		let baggage_idx = idx + subsystem_name.len();
+		let fname = &bag_field.field_name;
+		let ftype = &bag_field.field_ty;
+		let impl_generics = placeholder_type[..baggage_idx]
+			.iter()
+			.chain(placeholder_type[baggage_idx + 1..].iter());
+		let other_field_name =
+			field_name[..baggage_idx].iter().chain(field_name[baggage_idx + 1..].iter());
```
drahnr, line 224:
```suggestion
			field_name.iter().enumerate().filter(|(idx,_)| { baggage_idx != idx });
```

node/overseer/overseer-gen/proc-macro/src/impl_builder.rs
```diff
@@ -210,56 +203,53 @@ pub(crate) fn impl_builder(info: &OverseerInfo) -> proc_macro2::TokenStream {
...
+		let return_type_generics = produce_setter_generic_permutation(
+			ftype,
+			baggage_idx,
+			placeholder_type.as_slice(),
+			&Ident::new("OverseerFieldInit", Span::call_site()),
+		);
+		let additional_baggage_generic = if bag_field.generic {
+			quote! {#ftype,}
+		} else {
+			quote!()
```
drahnr, line 241:
```suggestion
			TokenStream::new()
```

node/overseer/overseer-gen/proc-macro/src/impl_builder.rs
```diff
@@ -601,43 +591,50 @@ pub(crate) fn impl_task_kind(info: &OverseerInfo) -> proc_macro2::TokenStream {
...
+			.map(|field_name| format_ident!("__{}", field_name.to_string().to_class_case())),
 	);
-	result.extend(baggage_names
-		.iter()
-		.map(|field_name| format_ident!("__{}",
-			field_name.to_string().to_class_case()))
+	result.extend(
+		baggage_names
+			.iter()
+			.map(|field_name| format_ident!("__{}", field_name.to_string().to_class_case())),
```
drahnr:
```suggestion
			.map(|field_name| format_ident!("__{}", field_name.to_string().to_class_case()))
```

node/overseer/overseer-gen/tests/ui/err-08-duplicate-subsystem.stderr
```diff
@@ -0,0 +1,10 @@
+error[E0599]: no method named `sub0` found for struct `OverseerBuilder<_, OverseerFieldInit<AwesomeSubSys>, OverseerFieldUninit<f64>>` in the current scope
+  --> tests/ui/err-08-duplicate-subsystem.rs:57:10
+   |
+22 | #[overlord(signal=SigSigSig, error=OverseerError, event=Event, gen=AllMessages)]
+   | -------------------------------------------------------------------------------- method `sub0` not found for this
+...
+57 |         .sub0(AwesomeSubSys::default()) // Duplicate subsystem
+   |          ^^^^-------------------------- help: remove the arguments
```
drahnr, line 8: This error message is very unfortunate. it leaks some generated inner type info.

node/overseer/overseer-gen/proc-macro/Cargo.toml
vstakhov: It is done now.

node/overseer/overseer-gen/proc-macro/src/impl_builder.rs
```diff
@@ -210,56 +203,53 @@ pub(crate) fn impl_builder(info: &OverseerInfo) -> proc_macro2::TokenStream {
...
-								#other_field_name: self.#other_field_name,
-							)*
-							spawner: self.spawner,
-						}
+	let baggage_specific_setters = info.baggage().iter().enumerate().map(|(idx, bag_field)| {
+		// Baggage fields follow subsystems
+		let baggage_idx = idx + subsystem_name.len();
+		let fname = &bag_field.field_name;
+		let ftype = &bag_field.field_ty;
+		let impl_generics = placeholder_type[..baggage_idx]
```
drahnr, line 220: Can we split the placeholder types into baggage and non baggage and avoid the `panic`-y `[]`-operators

drahnr (changes requested):
Superficial pass

node/overseer/overseer-gen/proc-macro/src/impl_builder.rs
vstakhov: Well, it would be `field_name.iter().enumerate().filter(|(idx,_)| { baggage_idx != *idx }).map(|(_, field)| field)` which looks much cryptic to me. Alternative is `let mut other_field_name = field_name.clone()` followed by `other_field_name.remove(idx)` which is simple but not very smart in general.

node/overseer/overseer-gen/proc-macro/src/impl_builder.rs
```diff
@@ -19,63 +19,20 @@ use syn::Ident;
...
-			builder_generic_ty[idx] = format_ident!("NEW");
-
-			let generics_ts = quote! {
-				<S, #( #baggage_generic_ty, )* #( #builder_generic_ty, )* >
-			};
-
-			(generics_ts, to_be_replaced_ty.clone(), to_be_replaced_name, to_keep_name)
-		})
-		.collect::<Vec<(_, _, _, _)>>()
-}
```
drahnr, line 53: :+1: this was far from perfect to begin with, but I think it could have been reworked directly to return a `syn::Generics` which is what it generates after all

node/overseer/overseer-gen/proc-macro/src/impl_builder.rs
```diff
@@ -19,63 +19,20 @@ use syn::Ident;
...
-	let subsystem_name_replace_with = &info
-		.subsystem_names_without_wip()
-		.iter()
-		.map(|subsystem_name| format_ident!("replace_{}", subsystem_name))
-		.collect::<Vec<_>>();
-
-	let builder_generic_ty = &info.builder_generic_types();
+	let subsystem_type = &info.subsystem_generic_types();
 
+	let consumes = &info.consumes_without_wip();
```
drahnr, line 35: The `wip` denotes that the subsystem is stilll work in progress, and hess no channels will be hooked up. So the name conveys a bit of a skewed meaning.

node/overseer/overseer-gen/proc-macro/src/impl_builder.rs
```diff
@@ -107,46 +62,243 @@ pub(crate) fn impl_builder(info: &OverseerInfo) -> proc_macro2::TokenStream {
...
+	// Helpers to use within quote! macros
+	let subsystem_where_clause = &info
+		.subsystems()
+		.iter()
+		.map(|subsys| {
+			let generic_ty = &subsys.generic;
+			let consumes = &subsys.consumes;
+			quote! {
+				#generic_ty : #support_crate ::Subsystem<#subsystem_ctx_name< #consumes >, #error_ty>
+			}
```
drahnr, line 74: could be a `WherePredicate` collected into a `Punctuated<WherePredicate, Comma>,`

node/overseer/overseer-gen/proc-macro/src/impl_builder.rs
```diff
@@ -107,46 +62,243 @@ pub(crate) fn impl_builder(info: &OverseerInfo) -> proc_macro2::TokenStream {
...
+		})
+		.collect::<Vec<_>>();
+	let builder_where_clause = quote! {
 		where
-			S: #support_crate ::SpawnNamed,
+			#( #subsystem_where_clause, )*
+			S: #support_crate ::SpawnNamed + Send
```
drahnr, line 80: nit: is there a reason we require send here?

node/overseer/overseer-gen/proc-macro/src/impl_builder.rs
```diff
@@ -107,46 +62,243 @@ pub(crate) fn impl_builder(info: &OverseerInfo) -> proc_macro2::TokenStream {
...
+				&Ident::new("OverseerFieldInit", Span::call_site()),
+			);
+			let other_field_name =
+				field_name[..idx].iter().chain(field_name[idx + 1..].iter()).collect::<Vec<_>>();
+			let replace_modified_generics = return_type_generics
+				.iter()
+				.enumerate()
+				.map(|(other_idx, generic)| {
+					if other_idx == idx {
+						quote!(OverseerFieldInit<NEW>)
```
drahnr, line 135: nit: 
```suggestion
						quote! { OverseerFieldInit<NEW> }
```

drahnr (changes requested):
Looks much much better :+1:

node/overseer/overseer-gen/proc-macro/src/impl_builder.rs
drahnr: Good point, looking `derive_replacable_generic_lists` this must be easier to achieve, I find this code rather convoluted and there must be an easier to comprehend structure.

node/overseer/overseer-gen/proc-macro/src/impl_builder.rs
```diff
@@ -578,3 +603,51 @@ pub(crate) fn impl_task_kind(info: &OverseerInfo) -> proc_macro2::TokenStream {
...
+/// This helper function is used to create a replacement for a placeholder types
+/// replacing a field matches specific index with a replacement ident
+/// For example, we have `<A, B, C>` and we want to produce `<A, Init<T>, C>`,
+/// so we call this function for `idx=1` and replacement ident as `Init<T>`
+fn produce_setter_generic_permutation<T>(
+	field_type: T,
+	field_idx: usize,
+	placeholder_types: &[Ident],
+	replacement_type: &Ident,
+) -> Vec<TokenStream>
```
drahnr, line 628: nit: should returen `syn::Generics`

node/overseer/overseer-gen/proc-macro/src/impl_builder.rs
```diff
@@ -578,3 +594,51 @@ pub(crate) fn impl_task_kind(info: &OverseerInfo) -> proc_macro2::TokenStream {
...
+}
+
+/// This helper function is used to create a replacement for a placeholder types
+/// replacing a field matches specific index with a replacement ident
+/// For example, we have `<A, B, C>` and we want to produce `<A, Init<T>, C>`,
+/// so we call this function for `idx=1` and replacement ident as `Init<T>`
+fn produce_setter_generic_permutation<T>(
+	field_type: T,
+	field_idx: usize,
+	placeholder_types: &[Ident],
```
drahnr, line 626: I don't quite get why we need both `field_type` and `field_idx`? Couldn't you just drop `field_type`, avoid all the preprocessing of `placeholder_types` (skipping the `field_idx` actually) and do smth along the lines of `builder_generic_ty[idx] = format_ident!("NEW");`?

drahnr (changes requested):

node/overseer/overseer-gen/proc-macro/src/impl_builder.rs
vstakhov: It is also required in the setter: https://github.com/paritytech/polkadot/blob/master/node/overseer/overseer-gen/proc-macro/src/impl_builder.rs#L252

Please correct me if I'm wrong but it seems that if `Send` is required in a setter signature it should also be required in the common `where` clause. Or `Send` could be removed from the both places...

node/overseer/overseer-gen/proc-macro/src/impl_builder.rs
```diff
@@ -107,46 +62,234 @@ pub(crate) fn impl_builder(info: &OverseerInfo) -> proc_macro2::TokenStream {
...
+					if other_idx == idx {
+						quote!(OverseerFieldInit<NEW>)
+					} else {
+						generic.clone()
+					}
+				})
+				.collect::<Vec<_>>();
+
+			quote! {
+				#[allow(non_camel_case_types)]
```
drahnr, line 134: We should be able to manage without this, right?

vstakhov: No, removing this attribute produces a warning like the following:

```
warning: type parameter `__plinkos` should have an upper camel case name
  --> node/overseer/overseer-gen/examples/dummy.rs:96:2
   |
96 |     plinkos: GoblinTower,
   |     ^^^^^^^ help: convert the identifier to upper camel case: `Plinkos`
   |
   = note: `#[warn(non_camel_case_types)]` on by default
```

In fact, whenever we use any combination of the placeholder types, except full initialized or full uninitialized, we need to add this override.

This particular place is transformed to the following output in `cargo expand --example dummy`:

```
impl<S, AwesomeSubSys, __plinkos, __i_like_pi, __i_like_generic>
    XxxBuilder<S, OverseerFieldUninit<AwesomeSubSys>, __plinkos, __i_like_pi, __i_like_generic>
where
    AwesomeSubSys: Subsystem<XxxSubsystemContext<MsgStrukt>, Yikes>,
    GoblinTower: Subsystem<XxxSubsystemContext<Plinko>, Yikes>,
    S: crate::SpawnNamed + Send,
```

drahnr: Because the type parameter should be `GoblinTower` (the generic).

drahnr: `__i_like_pi` should also not become a generic in the builder, the type is fixed in the declaration. That's the difference between baggage and subsystem. Baggage items must be explicitly declared as generic or as a type, any `#[subsystem]` annotation will make it an _implicit_ (= generated) generic where the user just gives the name (for orientation, this usually matches the subsystem implementation struct name), which we should then continue using as generic.

vstakhov: The idea was to create a generic for each field, so a builder has a generic parameter (placeholder) for each field. Then we specialize builder impl for each field being set or unset and we use generics for this type of specializations where needed (e.g. for subsystems or generic baggage). The builder struct itself looks like the following one:

```
pub struct XxxBuilder<S, __sub0, __plinkos, __i_like_pi, __i_like_generic> {
    sub0: __sub0,
    plinkos: __plinkos,
    i_like_pi: __i_like_pi,
    i_like_generic: __i_like_generic,
    spawner: ::std::option::Option<S>,
}
```

So no original generics are preserved in the initial structure.

node/overseer/overseer-gen/proc-macro/src/impl_builder.rs
```diff
@@ -85,11 +50,15 @@ pub(crate) fn impl_builder(info: &OverseerInfo) -> proc_macro2::TokenStream {
 	let channel_name_rx = &info.channel_names_without_wip("_rx");
 	let channel_name_unbounded_rx = &info.channel_names_without_wip("_unbounded_rx");
 
-	let baggage_generic_ty = &info.baggage_generic_types();
 	let baggage_name = &info.baggage_names();
-	let baggage_ty = &info.baggage_types();
-
-	let subsystem_ctx_name = format_ident!("{}SubsystemContext", overseer_name);
+	let baggage_generic_ty = &info.baggage_generic_types();
+	let baggage_placeholder_ty = baggage_name
```
drahnr, line 55: nit: I think palceholder is slight misnomer, it's really the init state as a generic iiuc, no idea to express this concisely

node/overseer/overseer-gen/proc-macro/src/impl_builder.rs
drahnr: I don't quite remember why this was added. We do spawn the subsystems on a threadpool, but at that point the generics are already specialized/known :monocle_face: I'd say get the tests to pass and then let's try to remove it.

node/overseer/overseer-gen/proc-macro/src/impl_builder.rs
```diff
@@ -107,46 +76,235 @@ pub(crate) fn impl_builder(info: &OverseerInfo) -> proc_macro2::TokenStream {
...
 
-	// all subsystems must have the same context
-	// even if the overseer does not impose such a limit.
-	let builder_additional_generics = quote! {
-		<#( #builder_generic_ty, )* >
-	};
+	let field_name = subsystem_name.iter().chain(baggage_name.iter()).collect::<Vec<_>>();
+	let field_type = subsystem_type
+		.iter()
+		.chain(info.baggage().iter().filter_map(|bag| bag.field_ty.get_ident()))
```
drahnr, line 94: This should not be a `filter_map`, `x.field_ty.get_ident()` being `None` is a bug, that should not cause broken code to be generated. We only support named fields for the overseer pattern. If it's handled earlier, it should be a `expect`.

node/overseer/overseer-gen/proc-macro/src/impl_builder.rs
```diff
@@ -107,46 +76,235 @@ pub(crate) fn impl_builder(info: &OverseerInfo) -> proc_macro2::TokenStream {
...
+		.enumerate()
+		.map(|(idx, ssf)| {
+			let fname = &ssf.name;
+			let ftype = &ssf.generic;
+			let subsystem_consumes = &ssf.consumes;
+			// Remove placeholder at specific field position
+			let impl_generics = &(subsystem_placeholder_ty[..idx]
+				.iter()
+				.chain(subsystem_placeholder_ty[idx + 1..].iter())
+				.collect::<Vec<_>>());
```
drahnr, line 115: nit: I do not like this. It looks / feels very unidiomatic. It also changes the order of generics. `let mut subsystem_placeholder_ty = subsystem_placeholder_ty.clone(); subsystem_placeholder_ty[idx] = foo;` reads significantly better.

node/overseer/overseer-gen/proc-macro/src/impl_builder.rs
```diff
@@ -107,46 +76,235 @@ pub(crate) fn impl_builder(info: &OverseerInfo) -> proc_macro2::TokenStream {
...
+				.map(|gen_ty| {
+					let ret : syn::GenericArgument = parse_quote!(#gen_ty);
+					ret
+				})
+				.collect::<Vec<_>>();
+			uninit_generics[idx] = parse_quote!{ FieldMissing<#ftype> };
+			let mut return_type_generics = uninit_generics.clone();
+			return_type_generics[idx] = parse_quote!{ FieldInit<#ftype> };
+			let mut replace_modified_generics = uninit_generics.clone();
+			replace_modified_generics[idx] = parse_quote!{ FieldInit<NEW> };
```
drahnr, line 132: :+1:

node/overseer/overseer-gen/proc-macro/src/impl_builder.rs
```diff
@@ -107,46 +76,235 @@ pub(crate) fn impl_builder(info: &OverseerInfo) -> proc_macro2::TokenStream {
...
+			.chain(baggage_placeholder_ty[idx + 1..].iter());
+		let other_baggage_name =
+			baggage_name[..idx].iter().chain(baggage_name[idx + 1..].iter());
+
+		let mut uninit_generics = baggage_placeholder_ty
+			.iter()
+			.map(|gen_ty| {
+				let ret : syn::GenericArgument = parse_quote!(#gen_ty);
+				ret
+			})
```
drahnr, line 235:
```suggestion
			.map(|gen_ty| -> syn::GenericArgument {
				parse_quote!(#gen_ty)
			})
```

node/overseer/overseer-gen/proc-macro/src/impl_builder.rs
```diff
@@ -107,46 +76,235 @@ pub(crate) fn impl_builder(info: &OverseerInfo) -> proc_macro2::TokenStream {
...
+		/// Type for the initialised field of the overseer builder
+		pub enum FieldInit<T> {
+			/// Defer initialization to a point where the `handle` is available.
+			Fn(SubsystemInitFn<T>),
+			/// Directly initialize the subsystem with the given subsystem type `T`.
+			/// Also used for baggage fields
+			Value(T),
+		}
+		/// Type marker for the uninitialised field of the overseer builder
+		/// Phantom data is used for type hinting when creating uninitialized
```
drahnr, line 284: I'd recommend to use `cargo-spellcheck`, it suggest you to enclose unknown things in `-ticks which is a nice convenience.

drahnr: Having the `PhantomData<T>` really helps to avoid specifying the generics when instantiating the `FooBuilder` when calling `Foo::builder()`, which is very convenient in our use case, imho the effect of something as subtle as this should be named explicitly in the code.

node/overseer/overseer-gen/proc-macro/src/impl_builder.rs
```diff
@@ -107,46 +76,235 @@ pub(crate) fn impl_builder(info: &OverseerInfo) -> proc_macro2::TokenStream {
...
+		/// builder.
+		#[derive(Debug)]
+		pub struct FieldMissing<T>(::core::marker::PhantomData<T>);
+
+		/// Trait used to mark fields status in a builder
+		trait OverseerFieldSt<T> {}
```
drahnr, line 290: nit: I buy an `a`, a `t` and an `e` - jokes aside, avoid abbreviations like these, it makes the code harder to read and we optimize for readability (and where it's not the case yet, will eventually)

node/overseer/overseer-gen/proc-macro/src/impl_builder.rs
```diff
@@ -107,46 +76,235 @@ pub(crate) fn impl_builder(info: &OverseerInfo) -> proc_macro2::TokenStream {
...
+
+		impl<T> OverseerFieldSt<T> for FieldInit<T> {}
+		impl<T> OverseerFieldSt<T> for FieldMissing<T> {}
+
+		impl<T> std::default::Default for FieldMissing<T> {
```
drahnr, line 295:
```suggestion
		impl<T> ::std::default::Default for FieldMissing<T> {
```

node/overseer/overseer-gen/proc-macro/src/impl_builder.rs
```diff
@@ -193,103 +351,61 @@ pub(crate) fn impl_builder(info: &OverseerInfo) -> proc_macro2::TokenStream {
...
-				#(
-					#subsystem_name: Default::default(),
-				)*
-				#(
-					#baggage_name: None,
-				)*
+					#(
+						#field_name: FieldMissing::<#field_type>::default(),
+					)*
 					spawner: None,
```
drahnr, line 380: nit: I think this should also become a proper `Init`, `Missing`?

node/overseer/overseer-gen/proc-macro/src/impl_builder.rs
```diff
@@ -334,13 +450,9 @@ pub(crate) fn impl_builder(info: &OverseerInfo) -> proc_macro2::TokenStream {
...
 				#(
-					// TODO generate a builder pattern that ensures this
-					// TODO https://github.com/paritytech/polkadot/issues/3427
 					let #subsystem_name = match self. #subsystem_name {
-						FieldInitMethod::Fn(func) => func(handle.clone())?,
-						FieldInitMethod::Value(val) => val,
-						FieldInitMethod::Uninitialized =>
-							panic!("All subsystems must exist with the builder pattern."),
+						FieldInit::Fn(func) => func(handle.clone())?,
+						FieldInit::Value(val) => val,
```
drahnr, line 455: :tada:

drahnr (commented):
Almost there! Only a few nits left and idiomatic things :+1:

node/overseer/overseer-gen/proc-macro/src/impl_builder.rs
vstakhov: I have changed it in https://github.com/paritytech/polkadot/pull/4753/commits/0af20ce752fc228314f5ec37a44e9943dcd78ab0

With `filter_map` approach we miss all complex fields (like `Vec<T>`)

node/overseer/overseer-gen/proc-macro/src/impl_builder.rs
vstakhov: This takes all placeholders but one. It could be done via `clone+remove`, of course.

node/overseer/overseer-gen/proc-macro/src/impl_builder.rs
vstakhov: Done, thanks!

node/overseer/overseer-gen/proc-macro/src/impl_builder.rs
drahnr: Why does it take all but one? Why not all and replace the one in question?

node/overseer/overseer-gen/proc-macro/src/impl_builder.rs
vstakhov: Do you mean without `Field` part? I was under impression that it could clash with some other types in the code.

drahnr: Yep

bkchr:
A different approach for this could have been what I have done for the wasm-builder builder: https://github.com/paritytech/substrate/blob/master/utils/wasm-builder/src/builder.rs#L100

Aka have different types per "stage". Would require always a fixed order of calling the individual methods, but would maybe give some better error messages?

vstakhov:
The error message is not so bad in fact: there is a mention of the type of the missing field, e.g. `Missing<Foo>` and it's relative order in the original Overseer structure.

node/overseer/overseer-gen/tests/ui/err-08-duplicate-subsystem.stderr
vstakhov: It seems to be better now, isn't it?

vstakhov:
```
error[E0599]: no method named `sub0` found for struct `OverseerBuilder<_, Init<AwesomeSubSys>, Missing<f64>>` in the current scope
```
When we are trying to set another `AwesomeSubSys` over.

node/overseer/overseer-gen/proc-macro/src/impl_builder.rs
drahnr: Leave as is for now.

node/overseer/overseer-gen/proc-macro/src/impl_builder.rs
```diff
@@ -193,38 +371,20 @@ pub(crate) fn impl_builder(info: &OverseerInfo) -> proc_macro2::TokenStream {
...
+		pub struct #builder <S, #( #subsystem_passthrough_generics, )* #( #baggage_passthrough_generics, )*> {
 			#(
-				#subsystem_name : FieldInitMethod< #builder_generic_ty >,
+				#subsystem_name: #subsystem_passthrough_generics,
 			)*
 			#(
-				#baggage_name : ::std::option::Option< #baggage_ty >,
+				#baggage_name: #baggage_passthrough_generics,
 			)*
 			spawner: ::std::option::Option< S >,
```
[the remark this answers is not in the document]
vstakhov: Yes, I have this in my todo as well. Similarly, we might need to remove `+Send` trait on Spawner.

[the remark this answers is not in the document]
drahnr: I already implemented this now, and will push in a few minutes.

node/overseer/overseer-gen/proc-macro/src/impl_builder.rs
```diff
@@ -215,10 +218,10 @@ pub(crate) fn impl_builder(info: &OverseerInfo) -> proc_macro2::TokenStream {
 		// Baggage fields follow subsystems
 		let fname = &bag_field.field_name;
 		let field_type = &bag_field.field_ty;
-		let impl_generics = recollect_without_idx(&baggage_passthrough_generics[..], idx);
+		let impl_generics = recollect_without_idx(&baggage_state_generics[..], idx);
```
drahnr, line 221: Actually they are both, `state` and `passthrough`? How about `baggage_passthrough_state_generics`

node/overseer/overseer-gen/tests/ui/err-08-duplicate-subsystem.stderr
drahnr: :+1:

node/overseer/overseer-gen/proc-macro/src/impl_builder.rs
vstakhov: It is a bit too verbose, but more descriptive indeed.

node/overseer/overseer-gen/proc-macro/src/impl_builder.rs
```diff
@@ -107,49 +81,256 @@ pub(crate) fn impl_builder(info: &OverseerInfo) -> proc_macro2::TokenStream {
...
+			.map(|gen_ty| parse_quote!(#gen_ty))
+			.collect::<Vec<syn::GenericArgument>>();
+		pre_setter_generics[idx] = parse_quote! { Missing<#field_type> };
+
+		let mut post_setter_generics = pre_setter_generics.clone();
+		post_setter_generics[idx] = parse_quote! { Init<#field_type> };
+
+		// Baggage can also be generic, so we need to include that to a signature
+		let additional_baggage_generic = if bag_field.generic {
+			quote! {#field_type,}
```
drahnr, line 235: nit: remove the `,` here..

node/overseer/overseer-gen/proc-macro/src/impl_builder.rs
```diff
@@ -107,49 +81,256 @@ pub(crate) fn impl_builder(info: &OverseerInfo) -> proc_macro2::TokenStream {
...
 
-	let builder_where_clause = quote! {
-		where
-			S: #support_crate ::SpawnNamed,
-		#(
-			#builder_generic_ty : Subsystem<#subsyste_ctx_name< #consumes >, #error_ty>,
-		)*
-	};
+		quote! {
+			impl <InitStateSpawner, #additional_baggage_generic #( #subsystem_passthrough_state_generics, )* #( #impl_baggage_state_generics, )* >
```
drahnr, line 241: ...and add it here instead

node/overseer/overseer-gen/proc-macro/src/impl_builder.rs
drahnr: nit: also a misnomer, it's not additional, it has been there from the get go :)

node/overseer/overseer-gen/proc-macro/src/impl_builder.rs
```diff
@@ -107,49 +81,256 @@ pub(crate) fn impl_builder(info: &OverseerInfo) -> proc_macro2::TokenStream {
...
+			/// Defer initialization to a point where the `handle` is available.
+			Fn(SubsystemInitFn<T>),
+			/// Directly initialize the subsystem with the given subsystem type `T`.
+			/// Also used for baggage fields
+			Value(T),
+		}
+		/// Type marker for the uninitialized field of the overseer builder.
+		/// `PhantomData` is used for type hinting when creating uninitialized
+		/// builder, e.g. to avoid specifying the generics when instantiating
+		/// the `FooBuilder` when calling `Foo::builder()`
```
drahnr, line 303: :+1:

node/overseer/overseer-gen/proc-macro/src/impl_builder.rs
```diff
@@ -192,39 +373,26 @@ pub(crate) fn impl_builder(info: &OverseerInfo) -> proc_macro2::TokenStream {
...
+	ts.extend(quote!{
+		/// Builder pattern to create compile time safe construction path.
+		pub struct #builder <InitStateSpawner, #( #subsystem_passthrough_state_generics, )* #( #baggage_passthrough_state_generics, )*> {
 			#(
-				#subsystem_name : FieldInitMethod< #builder_generic_ty >,
+				#subsystem_name: #subsystem_passthrough_state_generics,
 			)*
 			#(
-				#baggage_name : ::std::option::Option< #baggage_ty >,
+				#baggage_name: #baggage_passthrough_state_generics,
```
drahnr, line 385: :+1:

node/overseer/overseer-gen/proc-macro/src/impl_builder.rs
```diff
@@ -351,11 +513,11 @@ pub(crate) fn impl_builder(info: &OverseerInfo) -> proc_macro2::TokenStream {
 					let (signal_tx, signal_rx) = #support_crate ::metered::channel(SIGNAL_CHANNEL_CAPACITY);
 
 					// Generate subsystem name based on overseer field name.
-					let mut subsystem_string = String::from(stringify!(#subsystem_name));
+					let subsystem_string = String::from(stringify!(#subsystem_name));
 					// Convert owned `snake case` string to a `kebab case` static str.
 					let subsystem_static_str = Box::leak(subsystem_string.replace("_", "-").into_boxed_str());
```
drahnr, line 518: note to myself: this could be extracted as generated `&'static str` rather than the adhoc string generation.

node/overseer/overseer-gen/proc-macro/src/impl_builder.rs
vstakhov: Unfortunately, `,` can not be removed from here, as the `else` branch might have an empty token leading to double `,` syntax error. Renaming the variable is a good idea anyway.

drahnr: It could, but there is not much gain in this context, after all - discard my comment

PR #4756, opened 2022-01-20
XCM v3: Context & ID hash

gavofyork:
Simple refactor ensuring that `send_xcm` returns the hash of the message as sent.

This should be combined with a new context system to allow the functions in traits used by `XcmExecutor` to have knowledge of the context of their instruction. The context passed into the trait functions should include at least the `origin: Option<MultiLocation>` and `message_hash: XcmHash`. E.g.:

```
pub struct XcmContext {
  origin: Option<MultiLocation>,
  message_hash: XcmHash,
}
```

Traits designed only for use with XCM can have this passed in to their relevant functions. Traits not designed for use only in XCM can have converter structs which discard the context before forwarding.

## Topic

There should be a new item in the XCVM: `topic: Option<[u8; 32]>`. Context should be extended to include a `topic: Option<[u8; 32]>` field, representing this value. There should be two new instructions:

```
- `SetTopic([u8; 32])`: Sets the XCVM's topic to `Some`.
- `ResetTopic`: Sets the XCVM's topic to `None`.
```

### `XcmContext`

Several APIs now accept an additional `context: &XcmContext` parameter (except in the case of `withdraw_asset`, which accepts a `maybe_context: Option<&XcmContext>`).

This parameter must be added, but its value may be ignored.

### `XcmHash`

`SendXcm` and `ExportXcm` must both provide an `XcmHash` value back from `deliver` function if it returns `Ok`. It is up to the trait implementation exactly how this hash is determined, however in general it should identify the message which was sent and should be identifiable *on both sides of the transport*. Generally this will mean hashing the data which is encoded and set on the wire (or the common part of it, if it changes over the transport).

### `XcmExecutor`

`XcmExecutor::execute` has now been renamed to `XcmExecutor::bench_process` and is only enabled for benchmarks. Prior to this it has collided with the trait function `ExecuteXcm::execute`.


xcm/src/v3/traits.rs
```diff
@@ -354,15 +365,23 @@ pub trait SendXcm {
 	/// If it is not a destination which can be reached with this type but possibly could by others, then it *MUST*
 	/// return `CannotReachDestination`. Any other error will cause the tuple implementation to exit early without
 	/// trying other type fields.
-	fn send_xcm(destination: impl Into<MultiLocation>, message: Xcm<()>) -> SendResult;
+	fn send_xcm(
+		destination: impl Into<MultiLocation>,
+		message: Xcm<()>,
+		context: XcmContext,
```
gavofyork, line 371: `send_xcm` does not need a context.

KiChjang:
Oh shoot, since the bridge PR replaces the `send_xcm` method with `validate` and `deliver`, we now can't hash the incoming `Xcm` and then return it in `deliver`, because the XCM is now processed in `validate`! We'll either need to return the hash in validate or store it somewhere when we're in `validate`, then retrieve the hash in `deliver` and return it.

xcm/xcm-executor/src/lib.rs
```diff
@@ -128,8 +124,12 @@ impl<Config: config::Config> ExecuteXcm<Config::Call> for XcmExecutor<Config> {
 
 		let mut vm = Self::new(origin);
 
+		// We hash the message here instead of inside the loop, because `message` may have been
+		// reassigned due to the existence of the error handler or the appendix.
+		let message_hash = message.using_encoded(sp_io::hashing::blake2_256);
```
gavofyork, line 129: The XCM version on the receiving side may not be the same as it was on the sending side (owing to automated translation between versions).

Since all XCMs are always sent as a `VersionedXcm` while on the wire, it's better to hash the encoded `VersionedXcm`. This also has the advantage that it avoids a needless `.encode()`. This ensure that it's definitely the same hash as on the sending side.

xcm/xcm-executor/src/lib.rs
```diff
@@ -420,26 +471,32 @@ impl<Config: config::Config> XcmExecutor<Config> {
...
+						XcmContext { origin: self.origin.clone(), message_hash, topic: self.topic };
+					Config::AssetTransactor::deposit_asset(&asset, &dest, context)?;
 				}
 				// Note that we pass `None` as `maybe_failed_bin` and drop any assets which cannot
 				// be reanchored  because we have already called `deposit_asset` on all assets.
 				let assets = Self::reanchored(deposited, &dest, None);
 				let mut message = vec![ReserveAssetDeposited(assets), ClearOrigin];
 				message.extend(xcm.0.into_iter());
-				Config::XcmSender::send_xcm(dest, Xcm(message)).map_err(Into::into)
+				Config::XcmSender::send_xcm(dest, Xcm(message))?;
```
gavofyork, line 498: Missing the context?

xcm/xcm-executor/src/lib.rs
```diff
@@ -451,27 +508,32 @@ impl<Config: config::Config> XcmExecutor<Config> {
...
+				Ok(())
 			},
 			ReportHolding { response_info, assets } => {
 				// Note that we pass `None` as `maybe_failed_bin` since no assets were ever removed
 				// from Holding.
 				let assets =
 					Self::reanchored(self.holding.min(&assets), &response_info.destination, None);
-				Self::respond(self.origin.clone(), Response::Assets(assets), response_info)
+				Self::respond(self.origin.clone(), Response::Assets(assets), response_info)?;
```
gavofyork, line 535: missing the context?

xcm/xcm-executor/src/traits/filter_asset_location.rs
```diff
@@ -14,27 +14,36 @@
...
 /// Filters assets/location pairs.
 ///
 /// Can be amalgamated into tuples. If any item returns `true`, it short-circuits, else `false` is returned.
 pub trait FilterAssetLocation {
 	/// A filter to distinguish between asset/location pairs.
-	fn filter_asset_location(asset: &MultiAsset, origin: &MultiLocation) -> bool;
+	fn filter_asset_location(
+		asset: &MultiAsset,
+		origin: &MultiLocation,
+		context: XcmContext,
```
gavofyork, line 27: Don't think this is needed as it's not a state-changing action.

gavofyork (commented):
Generally looking like it's going in the right direction, aside from the remarks.

runtime/westend/src/weights/xcm/pallet_xcm_benchmarks_generic.rs
```diff
@@ -51,57 +51,57 @@ impl<T: frame_system::Config> WeightInfo<T> {
...
-		(5_238_000 as Weight)
+		(3_485_000 as Weight)
 	}
 	pub(crate) fn set_error_handler() -> Weight {
-		(5_104_000 as Weight)
+		(3_314_000 as Weight)
 	}
 	pub(crate) fn set_appendix() -> Weight {
-		(5_095_000 as Weight)
+		(3_334_000 as Weight)
```
gavofyork, line 76: 33% reduction in weights?

KiChjang: Huh, could it be related to the increased DB cache size? I can bench again to see if we get anything different.

xcm/pallet-xcm/src/lib.rs
```diff
@@ -1306,22 +1310,24 @@ impl<T: Config> Pallet<T> {
...
-		let maybe_fee_payer = if interior != Junctions::Here {
+		let maybe_fee_payer: Option<MultiLocation> = if interior != Junctions::Here {
 			message.0.insert(0, DescendOrigin(interior.clone()));
 			Some(interior.into())
 		} else {
 			None
 		};
 		log::trace!(target: "xcm::send_xcm", "dest: {:?}, message: {:?}", &dest, &message);
+		let message_hash =
+			VersionedXcm::from(message.clone()).using_encoded(sp_io::hashing::blake2_256);
```
gavofyork, line 1324: This is a heavy operation. If `charge_fees` needs to be passed in the context, then refactor it so that `validate_send` returns the message hash rather than `deliver`.

runtime/westend/src/weights/xcm/pallet_xcm_benchmarks_generic.rs
ggwpez: Weights in the Polkadot repo were not updated after adding the Production optimizations in https://github.com/paritytech/substrate/pull/10692.  
So this should happen for other runtimes and pallets as well.

xcm/src/v3/traits.rs
```diff
@@ -246,15 +246,17 @@ pub trait ExecuteXcm<Call> {
...
 		origin: impl Into<MultiLocation>,
 		pre: Self::Prepared,
+		hash: XcmHash,
 		weight_credit: Weight,
 	) -> Outcome;
 
-	/// Execute some XCM `message` from `origin` using no more than `weight_limit` weight. The weight limit is
-	/// a basic hard-limit and the implementation may place further restrictions or requirements on weight and
-	/// other aspects.
+	/// Execute some XCM `message` with the versioned `hash` from `origin` using no more than `weight_limit` weight.
```
gavofyork, line 253:
```suggestion
	/// Execute some XCM `message` with the message `hash` from `origin` using no more than `weight_limit` weight.
```

xcm/src/v3/traits.rs
```diff
@@ -265,16 +267,17 @@ pub trait ExecuteXcm<Call> {
 			message,
 			weight_limit,
 		);
-		Self::execute_xcm_in_credit(origin, message, weight_limit, 0)
+		Self::execute_xcm_in_credit(origin, message, hash, weight_limit, 0)
 	}
 
-	/// Execute some XCM `message` from `origin` using no more than `weight_limit` weight.
+	/// Execute some XCM `message` with the versioned `hash` from `origin` using no more than `weight_limit` weight.
```
gavofyork, line 273:
```suggestion
	/// Execute some XCM `message` with the message `hash` from `origin` using no more than `weight_limit` weight.
```

xcm/xcm-executor/src/lib.rs
```diff
@@ -271,13 +279,24 @@ impl<Config: config::Config> XcmExecutor<Config> {
...
+			topic: None,
 			_config: PhantomData,
 		}
 	}
 
 	/// Execute the XCM program fragment and report back the error and which instruction caused it,
 	/// or `Ok` if there was no error.
 	pub fn execute(&mut self, xcm: Xcm<Config::Call>) -> Result<(), ExecutorError> {
+		let message_hash =
+			VersionedXcm::from(xcm.clone()).using_encoded(sp_io::hashing::blake2_256);
```
gavofyork, line 291: This won't really work since the hash is not of the message which was used for transport. Also the message hash will never be reported back to the caller.

What is the use-case of this function? If just tests, then it should be marked as such.

xcm/xcm-executor/src/lib.rs
```diff
@@ -304,7 +323,7 @@ impl<Config: config::Config> XcmExecutor<Config> {
 
 	/// Execute any final operations after having executed the XCM message.
 	/// This includes refunding surplus weight, trapping extra holding funds, and returning any errors during execution.
-	pub fn post_execute(mut self, xcm_weight: Weight) -> Outcome {
+	pub fn post_execute(mut self, xcm_weight: Weight, message_hash: [u8; 32]) -> Outcome {
```
gavofyork, line 326: `message_hash` should probably be stored in `self` rather than having the caller pass it in twice since it was already passed by the caller in `execute`/`execute_with_hash`.

xcm/xcm-executor/src/lib.rs
```diff
@@ -399,13 +424,22 @@ impl<Config: config::Config> XcmExecutor<Config> {
 	}
 
 	/// Process a single XCM instruction, mutating the state of the XCM virtual machine.
-	fn process_instruction(&mut self, instr: Instruction<Config::Call>) -> Result<(), XcmError> {
+	fn process_instruction(
+		&mut self,
+		instr: Instruction<Config::Call>,
+		message_hash: XcmHash,
```
gavofyork, line 430: Just store in `self`.

xcm/xcm-executor/src/lib.rs
```diff
@@ -527,19 +576,24 @@ impl<Config: config::Config> XcmExecutor<Config> {
...
 					FeeReason::Report,
-				)
+				)?;
+				Ok(())
 			},
 			DepositAsset { assets, beneficiary } => {
 				let deposited = self.holding.saturating_take(assets);
 				for asset in deposited.into_assets_iter() {
-					Config::AssetTransactor::deposit_asset(&asset, &beneficiary)?;
+					let context =
```
gavofyork, line 585: Store context in `self` and just pass in `&self.context` to avoid cloning.

xcm/xcm-simulator/example/src/parachain.rs
```diff
@@ -275,19 +276,22 @@ pub mod mock_msg_queue {
...
 					},
-					Ok(Err(())) => {
-						Self::deposit_event(Event::UnsupportedVersion(id));
-					},
-					Ok(Ok(x)) => {
-						let outcome = T::XcmExecutor::execute_xcm(Parent, x.clone(), limit);
-						<ReceivedDmp<T>>::append(x);
-						Self::deposit_event(Event::ExecutedDownward(id, outcome));
+					Ok(versioned) => {
+						let hash = versioned.using_encoded(sp_io::hashing::blake2_256);
```
gavofyork, line 285: Hash this first to avoid needless re-encoding.

gavofyork (commented):
Still some changes needed yet.

xcm/xcm-simulator/example/src/parachain.rs
gavofyork: (Just use `id`, even)

gavofyork:
Should all be good now.

KiChjang (approved):

Polkadot-Forum:
This pull request has been mentioned on **Polkadot Forum**. There might be relevant details there:

https://forum.polkadot.network/t/substrate-etl-polkadot-xcmtransfers-dataset-xcmv3-context-matching/1732/1

PR #4759, opened 2022-01-21
Update XCM AlwaysLatest/Release to V2

### LINKED ISSUE #4734: Update `AlwaysRelease` and `AlwaysLatest`
https://github.com/paritytech/polkadot/blob/dee0681cccc650d7b2ba5face7f15eb49bdc9b98/xcm/src/lib.rs#L438-L442

Now that v2 is stable and audited, I would assume that it should be the default for both `AlwaysRelease` and `AlwaysLatest`.

KiChjang:
Fixes #4734.

### DIFF
--- a/xcm/src/lib.rs
+++ b/xcm/src/lib.rs
@@ -436,10 +436,10 @@ impl WrapVersion for AlwaysV2 {
 }
 
 /// `WrapVersion` implementation which attempts to always convert the XCM to the latest version before wrapping it.
-pub type AlwaysLatest = AlwaysV1;
+pub type AlwaysLatest = AlwaysV2;
 
 /// `WrapVersion` implementation which attempts to always convert the XCM to the release version before wrapping it.
-pub type AlwaysRelease = AlwaysV0;
+pub type AlwaysRelease = AlwaysV2;
 
 pub mod prelude {
 	pub use super::{


xcm/src/lib.rs
```diff
@@ -436,10 +436,10 @@ impl WrapVersion for AlwaysV2 {
 }
 
 /// `WrapVersion` implementation which attempts to always convert the XCM to the latest version before wrapping it.
-pub type AlwaysLatest = AlwaysV1;
+pub type AlwaysLatest = AlwaysV2;
 
 /// `WrapVersion` implementation which attempts to always convert the XCM to the release version before wrapping it.
-pub type AlwaysRelease = AlwaysV0;
+pub type AlwaysRelease = AlwaysV2;
```
kianenigma, line 442: it seems to me like we already moved to v2 somewhere, but forgot to update these? If so, can we have a test that would prevent this again?

KiChjang: I think you are talking about the safe XCM version or the `CurrentXcmVersion` in the XCM pallet, which is different from the `AlwaysLatest` and `AlwaysRelease` type aliases here. This type here is primarily used in the configs of the `ChildParachainRouter` and the XCMP queue pallet for the purpose of version negotiation. I believe nowadays chains use the XCM pallet as a default, but people can still freely choose to substitute the XCM pallet with these two types as a quick way to simply mandate all incoming and outgoing XCM versions to be of a particular version, instead of discovering and negotiating with the destination chain.

PR #4760, opened 2022-01-21
Use `define_benchmarks!`

ggwpez:
https://github.com/paritytech/substrate/pull/10592 made it possible to reduce the combination of `add_benchmark`+`list_benchmark` to just one `define_benchmarks`.  
This MR refactors the code to use it.  

Same as in Cumulus https://github.com/paritytech/cumulus/pull/926

## @ reviewers
- [ ] The `rialto-parachain` runtime config somehow has no `list_benchmark` (nor `benchmark_metadata`) call, only `add`.
- [ ] The correctness of this change can be verified by listing all benchmarks: `cargo run --features=runtime-benchmarks -- benchmark --chain=westend-dev --pallet="*" --extrinsic="*" --list` for the chain specs: `westend`, `rococo`, `polkadot`, `kusama` but I could not test it for `rialto`, `rialto-parachains` and `millau` .


shawntabrizi (approved):

kianenigma (approved):
>  The rialto-parachain runtime config somehow has no list_benchmark (nor benchmark_metadata) call, only add.


I think it is fine to just ping someone from the bridge team and let them update to the new syntax if/when they wish.

ggwpez:
bot merge

PR #4771, opened 2022-01-23
Use BlockImport and not BlockImportNotification to determine active leaves.

rphmeier:
Block import notifications were too flaky, esp. when concerning small forks of the type that occurs when reverting. With this PR we now aggressively consider every single block the node imports successfully (even the millions during initial sync) an active leaf, at least temporarily.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6677,6 +6677,7 @@ name = "polkadot-overseer"
 version = "0.9.13"
 dependencies = [
  "assert_matches",
+ "async-trait",
  "femme",
  "futures 0.3.19",
  "futures-timer",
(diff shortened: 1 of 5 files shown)


rphmeier:
Approval voting and the network bridge already have logic to limit functionality while doing the initial sync. We may want to introduce that functionality into more subsystems if they become spammy, annoying, or heavy during initial sync. The main thing I worry about with this change is that it would lead to sync being slow or unpredictable things happening during sync.

We should come up with some clever heuristic for figuring out which blocks to ignore for import notifications at some point.

drahnr (commented):
LGTM, but the side effects this most likely will have need to be mitigated before merging, at the very least verified.

bkchr:
> small forks of the type that occurs when reverting

Where do we filter them in the current code? I would say that you already get these notifications as the moment.

rphmeier:
Iced this because the bug this attempts to fix turned out to be far more insidious.

rphmeier:
Closing. We can revisit this if necessary but it is probably not necessary.

PR #4782, opened 2022-01-25
Update Versi chainspec to make use of most recent Rococo runtime

Lldenaurois:


node/service/src/chain_spec.rs
```diff
@@ -157,7 +157,24 @@ pub fn rococo_config() -> Result<RococoChainSpec, String> {
...
+	Ok(RococoChainSpec::from_genesis(
+		"Versi",
+		"versi_v1_3",
+		ChainType::Live,
+		move || RococoGenesisExt {
+			runtime_genesis_config: rococo_development_config_genesis(wasm_binary),
+			// Use 1 minute session length.
+			session_length_in_blocks: Some(100),
+		},
+		vec![],
```
Lldenaurois, line 171: Bootnodes are being set to null here, which I think is alright since we have our own spec for deployments in ansible:

https://github.com/paritytech/substrate/blob/master/client/chain-spec/src/chain_spec.rs#L263

node/service/res/versi.json
```diff
@@ -6,11 +6,7 @@
     "/ip4/34.90.151.124/tcp/30333/p2p/12D3KooWF7BUbG5ErMZ47ZdarRwtpZamgcZqxwpnFzkhjc1spHnP",
     "/ip4/34.90.137.14/tcp/30333/p2p/12D3KooWLcpkpvjr5ccgtUdTSYtNDjEdsDcPNrt2Rb7yXuAf7bUE",
     "/ip4/35.204.67.254/tcp/30333/p2p/12D3KooWGjEEDmNbBkXLM1uKMseK9iYD3osKA4JGdGKMZDCusjd6",
-    "/ip4/34.90.121.39/tcp/30333/p2p/12D3KooWBhkZQydNHDR3XSehnrfj1KNFCdpwgDrYpX54FrUR1FRS",
-    "/ip4/34.91.145.35/tcp/30333/p2p/12D3KooWBuLAMevZexnFKCgTyoz3AnHQn98D9cfe1Mg3kPoCjkwf",
-    "/ip4/34.91.77.80/tcp/30333/p2p/12D3KooWA5BAM71y9NtV5NH6EjANgYKRZ8jNLJ5z8GJ5RPdjt63n",
-    "/ip4/34.91.84.25/tcp/30333/p2p/12D3KooWSV4VqhBHZKKBsZKmVU462qRW9PmXTSuYvuajt1P93djA",
-    "/ip4/34.91.97.19/tcp/30333/p2p/12D3KooWD6wC88atMMyVeP6ZKg9sK7QmUL8x8m1RxMW8rhv2vWyg"
+    "/ip4/34.90.121.39/tcp/30333/p2p/12D3KooWBhkZQydNHDR3XSehnrfj1KNFCdpwgDrYpX54FrUR1FRS"
```
eskimor: Are those bootnodes actually correct?

node/service/src/chain_spec.rs
eskimor: Let's just kick them above then.

node/service/src/chain_spec.rs
```diff
@@ -157,7 +157,24 @@ pub fn rococo_config() -> Result<RococoChainSpec, String> {
...
-	VersiChainSpec::from_json_bytes(&include_bytes!("../res/versi.json")[..])
+	let wasm_binary = rococo::WASM_BINARY.ok_or("Versi development wasm not available")?;
+
+	Ok(RococoChainSpec::from_genesis(
+		"Versi",
+		"versi_v1_3",
+		ChainType::Live,
+		move || RococoGenesisExt {
+			runtime_genesis_config: rococo_development_config_genesis(wasm_binary),
+			// Use 1 minute session length.
```
eskimor, line 168: Really one minute?

eskimor (commented):
Looks good.

node/service/src/chain_spec.rs
Lldenaurois: Ah, damn I need to update the comment. 100 blocks is 10 minutes

node/service/src/chain_spec.rs
Lldenaurois: I would go even further to suggest that we should remove the versi.json entirely.

It would be our responsibility to set all the bootnodes and telemetry values in our ansible scripts.

eskimor (approved):

PR #4783, opened 2022-01-25
secondary filtering

drahnr:
Adds some extra checks.

TODO:

* [x] test case

### DIFF
--- a/runtime/parachains/src/disputes.rs
+++ b/runtime/parachains/src/disputes.rs
@@ -516,14 +516,19 @@ pub mod pallet {
 bitflags::bitflags! {
 	#[derive(Default)]
 	struct DisputeStateFlags: u8 {
+		/// The byzantine threshold of `f + 1` votes (and hence participating validators) was reached.
 		const CONFIRMED = 0b0001;
+		/// Is the supermajority for validity of the dispute reached.
 		const FOR_SUPERMAJORITY = 0b0010;
+		/// Is the supermajority against the validity of the block reached.
 		const AGAINST_SUPERMAJORITY = 0b0100;
 	}
 }
 
 impl DisputeStateFlags {
 	fn from_state<BlockNumber>(state: &DisputeState<BlockNumber>) -> Self {
+		// Only correct since `DisputeState` is _always_ initialized
+		// with the validator set based on the session.
 		let n = state.validators_for.len();
 
 		let byzantine_threshold = byzantine_threshold(n);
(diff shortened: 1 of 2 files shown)


runtime/parachains/src/disputes.rs
```diff
@@ -1001,6 +1027,10 @@ impl<T: Config> Pallet<T> {
 						// by the duplicate checks above. It's only the first one which
 						// may not already have been filtered out.
 						filter.remove_index(first_index_in_set);
+
+						// Removing individual statments can cause the dispute to become onesided.
+						// Checking that (again) is done after the loop. Remove the bit indices.
+						summary.new_flags.set(validator_index.0 as _, false);
```
eskimor, line 1033: Why is this suddenly necessary?

runtime/parachains/src/disputes.rs
```diff
@@ -1026,6 +1056,37 @@ impl<T: Config> Pallet<T> {
...
+						// Note that this does not distinguish between pro or cons votes,
+						// since allowing both of them, even if the spam threshold would be reached
+						// is a good thing.
+						// Overflow is no concern, disputes are limited by weight.
+						DisputeStatement::Valid(_) => {
+							if Some(true) == summary.new_flags.get(v_i) {
+							vote_for_count += 1;
+						}}
+						DisputeStatement::Invalid(_) => {
+							if Some(true) == summary.new_flags.get(v_i) {
```
drahnr, line 1080: @eskimor it's used here (and nowhere else) to make sure we only count newly added votes, which wasn't done before

runtime/parachains/src/disputes.rs
drahnr: See my next comment, it's not used outside the scope of this `fn` so it's alright to modify it at this point.

runtime/parachains/src/disputes.rs
```diff
@@ -652,8 +670,12 @@ impl<BlockNumber: Clone> DisputeStateImporter<BlockNumber> {
 					.map(|i| (ValidatorIndex(i as _), SpamSlotChange::Dec))
 					.collect()
 			},
-			(true, true) | (true, false) => {
-				// nothing to do. (true, false) is also impossible.
+			(true, false) => {
+				log::error!("Dispute statments are never removed. qed");
```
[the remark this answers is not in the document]
drahnr: It's both, but it's also the runtime so we don't want to panic if some unrelated code change _might_ make this possible in the future.

runtime/parachains/src/disputes.rs
eskimor: Now, that I check this more closely, it makes even less sense. `new_flags` has type: [DisputeStateFlags](https://github.com/paritytech/polkadot/blob/31a254b9174388422f9fe94a831c90a64f4c172c/runtime/parachains/src/disputes.rs#L498), I don't understand why we would use `ValidatorIndex` as index here. :confused:

drahnr: Yeah, you're correct. That makes no sense at all.

dvdplm:
A better title&description would be good here.

tdimitrov (approved):
I had a code walkthrough with @drahnr. The change is reasonable, tests are passing => should be good to go.

eskimor (approved):

Issue #4576, opened 2021-12-21
Drop candidates with insufficient backing in runtime + drop node side check


eskimor:
Currently we rely on node side code to not provide candidates in the paras inherent which have insufficient backing and simply err out on block production if that did not work out. This is not ideal, at the very least because this way we have to make damn sure to always have compatible limits on the node side and in the runtime.

Instead of mechanisms like querying the runtime for the correct value, we should simply drop the node side check and properly filter out candidates in `create_inherent` that lack required backing.

So this issue should be implemented in two phases:

- [ ] Provide a runtime upgrade which does proper filtering instead of bailing
- [ ] Remove the node side check and just provide candidates to the runtime as soon as we have them

PR #4784, opened 2022-01-25
runtime: skip insufficiently backed candidates instead of bailing

slumber:
Part of #4576

### DIFF
--- a/runtime/parachains/src/inclusion/mod.rs
+++ b/runtime/parachains/src/inclusion/mod.rs
@@ -492,9 +492,10 @@ impl<T: Config> Pallet<T> {
 			Vec::with_capacity(candidates.len());
 
 		// Do all checks before writing storage.
-		let core_indices_and_backers = {
+		// This is an `Vec` of (candidate, core index, backers bits, group index).
+		let candidates_with_cores_and_backers = {
 			let mut skip = 0;
-			let mut core_indices_and_backers = Vec::with_capacity(candidates.len());
+			let mut candidates_with_cores_and_backers = Vec::with_capacity(candidates.len());
 			let mut last_core = None;
 
 			let mut check_assignment_in_order = |assignment: &CoreAssignment| -> DispatchResult {
@@ -521,13 +522,13 @@ impl<T: Config> Pallet<T> {
 			// In the meantime, we do certain sanity checks on the candidates and on the scheduled
 			// list.
 			'next_backed_candidate: for (candidate_idx, backed_candidate) in
-				candidates.iter().enumerate()
+				candidates.into_iter().enumerate()
 			{
 				if let FullCheck::Yes = full_check {
 					check_ctx.verify_backed_candidate(
 						parent_hash,
 						candidate_idx,
-						backed_candidate,
+						&backed_candidate,
 					)?;
 				}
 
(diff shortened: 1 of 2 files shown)


rphmeier:
I don't like this change.

Previously, it was always the case that you could scrape the block bodies and know that parachain candidates were actually backed within a block.

Lldenaurois:
> I understand it, is about having good runtime-level filtering at the `create_inherent` phase, not during block execution itself. Even that, IMO, should be up for discussion because node-side logic will always be faster than runtime-side, although runtime-side is without a doubt easier.

I am also of the opinion that we should always bail if block execution contains anything that is not sufficient to act upon, i.e. insufficiently backed candidate, unsigned bitfields, disputes with no vote for/against etc

eskimor:
I am not really arguing against that. I did suggest that change because:

1. This way we are not introducing two code paths, but keep the one we have right now.
2. It should not really make much of a difference - assuming the block producer is honest, the filtering will never happen. If it is dishonest the candidate will be dropped, instead of the whole inherent - which should be fine as well, as everything else gets checked too. Although it will increase potential attack vectors in case of subtle bugs.

Not saying I particularly like the solution, but given the current code structure it seemed like the least intrusive path. In general we should revisit the create_inherent/block import duality - @drahnr  is already doing some writeup on that - really looking forward to it. Given our discussions we had with @jakoblell, I think the two code paths are unavoidable, but I think we need a more structured approach for making sure they are consistent. E.g. have something like having a `check_....` function for each `create_....` function, where each of those pairs can be reviewed & tested individually with a strong guarantee that if each of those pairs is correct, the whole stack will be correct.

drahnr (changes requested):
I don't think we should be more lenient on-chain.

rphmeier:
>  In general we should revisit the create_inherent/block import duality - @drahnr is already doing some writeup on that - really looking forward to it.

Me too. Refactoring is needed, clearly, and I think a 'check' / 'operate' duality is very reasonable. 

> We will only provide the very next candidate of any prospective chain to create_inherent, which will accept it when sufficiently backed, if not it will be accepted (likely) at the next block. 

Yes

> Relying purely on the node side for e.g. signature checking at block production

I think our ending point should be to allow the node-side to specify in the `InherentData` which checks it would like the runtime to do. As you mention, working on the node-side will simply be more efficient and but working in the runtime is more obviously correct. So we might start with node-side code that does very little by way of checking, but end with node-side code that does every possible check and the runtime only needs to do pass-through.

> I am also of the opinion that we should always bail if block execution contains anything that is not sufficient to act upon, i.e. insufficiently backed candidate, unsigned bitfields, disputes with no vote for/against etc

Anyway, I'm not sure what this has to do with making the on-chain logic more lenient but it seems we have a rough consensus that it's a bad idea. That is, it seems that myself, @drahnr, @Lldenaurois are against it and @eskimor, you seem to be neutral? If I understand you correctly, it seems that you've only suggested this change as a route to refactoring and you don't consider it an end-goal unto itself. Please correct me if that is an inaccurate portrayal. If that sounds right, then I move to close this PR.

eskimor:
Neutral ... I guess, less opposed. I did not like it, but it seemed ok until we have that refactor. Anyhow, this is not urgent as we are well aware that node side checks and runtime checks have to stay in sync for the time being, so I think we can manage to not break chains for a while, even if we have to adjust the backing threshold again.

rphmeier:
FWIW #4790 might help.

Issue #3779, opened 2021-09-03
Asynchronous Backing Spec & Tracking Issue


rphmeier:
### Tracking

- [x] Backwards compatible runtime changes: accept older relay-parents (#4786)
  - [ ] Deploy to Kusama
  - [ ] Deploy to Polkadot
- [x] Implementation of 'Context' logic on the node-side (#4790)
- [x] Implementation of a 'Prospective Parachains' subsystem that tracks backed blocks (#4913 )
  - [ ] Deploy to Kusama parameterized without 'look-forward'
  - [ ] Deploy to Polkadot parameterized without 'look-forward'
- [x] Update backing & collation subsystems to take impulses from prospective parachains instead of active-leaves directly
  - [x] collation-generation (https://github.com/paritytech/polkadot/pull/7405)
  - [x] collator-protocol (#5740)
  - [x] candidate-backing (#5557)
  - [x] statement-distribution (#5999)
  - [x] provisioner: choose backed candidates from Prospective Parachains instead of Backing. Find ones that the new block will accept (#5711)
- [ ] Final deployment
  - [ ] Test on Versi, Rococo with 'look-forward' parameter 
  - [ ] Collect & publish throughput improvements, fiddle with parameters
  - [ ] Deploy with farther look-forward to Kusama
  - [ ] Deploy with farther look-forward to Polkadot

Board: https://github.com/orgs/paritytech/projects/37

-----

This is a long-standing idea which has been pending a writeup for some time.

I previously referred to this as "Contextual Execution", but I'd prefer to rename it to "Asynchronous Backing", which is the goal, rather than "contextual execution" which doesn't describe the purpose very well and is fact an implementation detail.

## Terminology

* The **relay parent** is the relay-chain block that a parachain block anchors to.
* A parachain block is **backed** when it is first appears in the relay chain but is not necessarily available.
* A parachain block is **included** when it has appeared in the relay chain and has been signified as available by a majority of validators. Note that this happens at the earliest one relay-chain block after backing.
* The **base parent** of a parachain block is the parachain block of the same parachain which is _included_ as-of the relay-parent. This could be described as a function `base_parent(relay_parent) -> Parachain Block Hash`
* The **context** of a parachain block P is a set of constraints on execution for a parachain block derived from the relay-parent R of P and all the parachain blocks in-between `P` and `base_parent(R)`. This could be described as a function `context(R, [intermediate_blocks]) -> Constraints`.

## Motivation

[...]

PR #4786, opened 2022-01-26
Runtime changes for Asynchronous Backing

rphmeier:
cc #3779

This changes the runtime to accept parachain candidates which are based on relay-parents older than the most recent relay-parent.
  * Paras module tracks a few recent relay-parents
  * Inclusion module checks that relay-parent is one of those and is advancing
  * DMP module advancement criterion has been changed slightly

TODOs:
  - [x] Get all tests compiling
  - [x] Add tests for the test cases outlined in TODOs and even more if possible
  - [x] Update or add a new runtime API to make `check_validation_outputs` work based on relay-parent (postponed)
  - [x] Implementer's guide changes including a few paragraphs explaining why we do things this way.
  - [ ] Ensure that this code works with the existing node-side code:
      - [x] in local testnets 
      - [ ] on Versi/Rococo


slumber:
> /benchmark runtime kusama runtime_parachains::paras

https://github.com/paritytech/polkadot/issues/4789

rphmeier:
Changes look great. Thanks @slumber . I'd like to see more tests for the inclusion module as a follow-up, to make sure this actually works with older relay-parents.

runtime/parachains/src/paras_inherent/mod.rs
```diff
@@ -97,6 +98,81 @@ impl DisputedBitfield {
 	}
 }
 
+/// The maximum amount of relay-parent lookback.
+// TODO [now]: put this in the configuration module or file an issue to discuss
```
rphmeier, line 102: this needs to link to the github issue.

rphmeier:
I'm thinking of making a few changes to this PR:
  1. Track/compute the `GroupRotationInfo` as well at each relay-parent
  2. Evaluate backing signatures & group bitfield based on this group rotation info
  3. Clear `AllowedRelayParents` on session change

This is to avoid "re-backing" being necessary at all on the node-side and makes handling of group rotations much less complex overall.

slumber:
> I'm thinking of making a few changes to this PR:
> 
> 1. Track/compute the `GroupRotationInfo` as well at each relay-parent
> 2. Evaluate backing signatures & group bitfield based on this group rotation info
> 3. Clear `AllowedRelayParents` on session change
> 
> This is to avoid "re-backing" being necessary at all on the node-side and makes handling of group rotations much less complex overall.

For 1&2: when processing backed candidates we read cores assignments from storage, thus having both core index and group index for each candidate:
https://github.com/paritytech/polkadot/blob/0a1456dcd69c0dfe33db12440352d043b09e9d79/runtime/parachains/src/paras_inherent/mod.rs#L597
So while trying to understand a rationale, I assumed that we want to drop it and use the `GroupRotationInfo` instead, easily obtained from scheduler:
https://github.com/paritytech/polkadot/blob/0a1456dcd69c0dfe33db12440352d043b09e9d79/runtime/parachains/src/scheduler.rs#L675-L682
This info then should be used for getting the group index. But we still need a core index for this:
https://github.com/paritytech/polkadot/blob/0a1456dcd69c0dfe33db12440352d043b09e9d79/primitives/src/v1/mod.rs#L718-L722

And I don't think there's an interface in the runtime for this, thus the question is: why are we not happy with `scheduler::scheduled`? This will also hint on how obtaining core indices should be implemented for our case.

rphmeier:
The motivation is that we want to allow candidates which were backed by a previously assigned group to be accepted by the relay chain.

> thus the question is: why are we not happy with scheduler::scheduled

`Scheduler::scheduled` currently does 2 things that, based on the new motivation, probably shouldn't be mixed.
1. It determines the cores which are unoccupied
2. It determines which validator group is assigned to the core _right now_.

We don't need (1) to change but we want to make inclusion more flexible on (2) so we can also accept blocks from the previous groups

runtime/parachains/src/inclusion/mod.rs
```diff
@@ -571,9 +571,12 @@ impl<T: Config> Pallet<T> {
 						// account for already skipped, and then skip this one.
 						skip = i + skip + 1;
 
+						// The candidate based upon relay parent `N` should be backed by a group
+						// assigned to core at block `N + 1`. Thus, `relay_parent_number + 1`
+						// will always land in the current session.
 						let group_idx = <scheduler::Pallet<T>>::group_assigned_to_core(
 							assignment.core,
-							relay_parent_number,
+							relay_parent_number + One::one(),
```
rphmeier, line 580: 👍

runtime/parachains/src/shared.rs
```diff
@@ -38,6 +39,77 @@ pub(crate) const SESSION_DELAY: SessionIndex = 2;
 #[cfg(test)]
 mod tests;
 
+/// The maximum amount of relay-parent lookback.
+// TODO [now]: put this in the configuration module (https://github.com/paritytech/polkadot/issues/4841).
```
rphmeier, line 43:
```suggestion
// TODO: put this in the configuration module (https://github.com/paritytech/polkadot/issues/4841).
```

I use TODO [now] for stuff that should be done in the current PR. If there's an open issue, then the [now] should be removed.

runtime/parachains/src/shared.rs
```diff
@@ -90,6 +168,9 @@ impl<T: Config> Pallet<T> {
 		new_config: &HostConfiguration<T::BlockNumber>,
 		all_validators: Vec<ValidatorId>,
 	) -> Vec<ValidatorId> {
+		// Drop allowed relay parents buffer on a session change.
+		AllowedRelayParents::<T>::mutate(|tracker| tracker.buffer.clear());
```
rphmeier, line 172: This seems correct. The behavior is somewhat nuanced.

`on_new_session` is invoked in the `on_finalize` logic of the block that changes the session. So when we call `on_initialize` in the next block, it'll add a relay-parent to the tracker which is the last block of the prior session. This is fine, because we say that a candidate with a relay-parent R should be backed with the group assigned in number(R) + 1, which is guaranteed to be in the current session.

Sounds right?

slumber: > 

Yes,
To put it simple, we always allow building on top of the previous block
With async backing it will be some lookback which is kind of disabled on session boundaries

Probably worth leaving a note about it in comments?

rphmeier: Yeah, comments are always helpful.

PR #4790, opened 2022-01-27
inclusion emulator logic for asynchronous backing

rphmeier:
Part of #3779 

Note that this PR isn't being made against master but rather against the async-backing feature branch.

This PR introduces primitives for the creation of 'fragment-trees' which are trees of prospective parachain blocks that are later evaluated against the state of the relay-chain in the future. In short, this lays the groundwork for the introduction of an off-chain "prediction layer" that enables validators and collators to mostly decouple the act of extending parachains from the growth of the relay chain.

There are 3 main primitives exposed:
  - `Constraints`: Constraints on inputs and outputs of a parachain block.
  - `ConstraintModifications`: further modifications to constraints as a result of a prospective parachain block
  - `Fragment`: a single link of a prospective parachain, representing a set of operating constraints that the parachain block was built under as well as the modifications produced by that parachain block.

The module docs go into much more detail on fragments, fragment-trees, and constraints.

This also introduces skeletal 'VStaging' primitives including a basic 'VStaging' version of the ParachainHost runtime API which will grow as we continue to commit to this feature branch.

TODO:
 - [x] Finish adding tests
 - [x] Fix warnings


node/subsystem-util/src/inclusion_emulator/staging.rs
```diff
@@ -0,0 +1,1226 @@
...
+	BlockNumber, CandidateCommitments, CollatorId, CollatorSignature, Hash, HeadData, Id as ParaId,
+	PersistedValidationData, UpgradeRestriction, ValidationCodeHash,
+};
+use std::collections::HashMap;
+
+/// Constraints on inbound HRMP channels.
+#[derive(Debug, Clone, PartialEq)]
+pub struct InboundHrmpLimitations {
+	/// An exhaustive set of all valid watermarks, sorted ascending
+	pub valid_watermarks: Vec<BlockNumber>,
```
eskimor, line 126: We should really strive for making sorted assumptions more explict. Would a `BTreeSet` work here?

PR #4796, opened 2022-01-27
Fix sorting for 0.9.16

eskimor:


chevdor (approved):

eskimor:
> could you backport the rest of the commits of #4795?

I really don't want to. I checked, functionality wise it is the same. This waiting for CI to finally carry on with tasks is driving me crazy - I want that to finally stop.

eskimor:
I would prefer a workflow next time, where fixes just go to the release branch and once the release is done, we merge the release branch to master. - Would be way less tedious and error prone.

drahnr:
If we only have one release at a time, this might work sufficiently well, iff and only iff master doesn't move too fast. As soon as one has more than one release branch at the same time or master picking up with our team growth, this will become a mess.

eskimor:
> If we only have one release at a time, this might work sufficiently well, iff and only iff master doesn't move too fast. As soon as one has more than one release branch at the same time or master picking up with our team growth, this will become a mess.

Once the release is out, this is a different thing- then I am actually fine with back-porting. I am only concerned with the release branch prior to release.

eskimor:
> It seems the real problem here is long CI times. My main concern if we're to backport some urgent changes for 0.9.17 to 0.9.16 branch we might end up fixing some merge conflicts, which is unnecessary.

Yes, long CI times are a big source of delays. Amplified by unnecessary/unknown changes, which cause PRs to fail, although they worked fine when based on master. So you end up not only waiting once for CI, but several times for a single stupid PR, that actually was already fine on master.

chevdor:
Could you please change the base so we merge this PR into `wk-prep-v0.9.16-rc7` ?

PR #4809, opened 2022-01-30
Update docs and enable DOT-over-XCM

gavofyork:
A few docs were outdated - they have been brought up to date.

This basically just enables the four transactions for doing a reserve asset transfer or teleportation of DOT to a parachain.

It disables sending XCM for now since it's useless without the destination chain supporting `DescendOrigin`. Expect for it to be re-enabled along with XCM v3 which will enable support for receiving messages beginning with `DescendOrigin`.


runtime/kusama/src/xcm_config.rs
```diff
@@ -169,14 +165,20 @@ pub type LocalOriginToLocation = (
...
+	// the DOT to send from the Relay-chain). But it's useless until we bring in XCM v3 which will
+	// make `DescendOrigin` a bit more useful.
+	type SendXcmOrigin = xcm_builder::EnsureXcmOrigin<Origin, ()>;
 	type XcmRouter = XcmRouter;
-	// Anyone can execute XCM messages locally...
+	// Anyone can execute XCM messages locally.
 	type ExecuteXcmOrigin = xcm_builder::EnsureXcmOrigin<Origin, LocalOriginToLocation>;
-	// ...but they must match our filter, which rejects all.
-	type XcmExecuteFilter = Nothing;
+	type XcmExecuteFilter = Everything;
```
KiChjang, line 176: Setting this to `Everything` means that anyone can call `execute` on the XCM pallet and provide an arbitrary XCM as an extrinsic to be executed. If we're 100% confident that XCM is good to go at its current form, then there shouldn't be any issues setting it to `Everything`, but just making sure we're on the same page.

KiChjang: Ah, I see, this is for Kusama, definitely not a huge problem then.

KiChjang (approved):

rdovgan (approved):

joepetrowski:
This should also activate the `UmpSink` to handle messages? Otherwise assets on paras will be dropped when they think they've sent them back.
https://github.com/paritytech/polkadot/blob/467544f5db344f546090259dabd0b6a8e6688046/runtime/polkadot/src/lib.rs#L1213

shawntabrizi (approved):

joepetrowski (approved):

Issue #3740, opened 2021-08-30
Staking miner: Check the queue one last time before submission


kianenigma:
Such submissions are failing because the miner checks the queue once (which is empty), and then computes a solution for 20 seconds, and then submits it, assuming that the queue is still empty. It should check one last time, after the mining is finish, if the anything exists in the queue or not. 

The root cause of this is that the current code requires a full `ext` to check the queue: https://github.com/paritytech/polkadot/blob/0a86997fba66211fe75d53146a67b58cb399901b/utils/staking-miner/src/monitor.rs#L49

And the `ext` that we are accessing is simply outdated, because it was fetched some 20 seconds ago.

The way to solve this is to: 
- rewrite `ensure_no_previous_solution` such that it directly query the chain state, not requiring an ext, so that we can be sure that no other solution from us lives in the queue.
- add `ensure_no_better_solution` that works in the same manner, but checks that no better solution exists in the queue.
- add a configuration enum to configure what we should do about the second check if it does NOT pass, because we might still want to be greedy and submit anyways, even though a better solution is queued. The most simple reason for this could be because we believe the the claimed better solution is fake. 

```rust 
enum SubmissionStrategy {
    /// Only submit if at the time, we are the best.
    OnlySubmitIfLeading,
    // Always submit. 
    AlwaysSubmit,
    /// Submit if we are leading, or if the solution that's leading is more that the given `Perbill`
    /// better than us. This helps detect obviously fake solutions and still combat them.
    SubmitIfClaimBetterThan(Perbill)
}
```

niklasad1:
Closed by #4819

PR #4819, opened 2022-01-31
staking miner: Check the queue one last time before submission 

niklasad1:
Builds on https://github.com/paritytech/polkadot/pull/4716

Closing #3740


utils/staking-miner/src/monitor.rs
```diff
@@ -50,138 +63,337 @@ async fn ensure_no_previous_solution<
...
-			Ok(())
+	let indices_key = storage_value(EPM_MODULE_PREFIX, b"SignedSubmissionIndices");
+
+	let indices: SubmissionIndicesOf<T> = get_storage(client, rpc_params! {indices_key, at})
+		.await
+		.map_err::<Error<T>, _>(Into::into)?
+		.unwrap_or_default();
+
+	for (_score, id) in indices {
+		let key = storage_value_by_key::<Twox64Concat>(
```
niklasad1, line 78: we could make a batch request for this instead, not sure what's best trade-off TBH.

utils/staking-miner/src/main.rs
```diff
@@ -294,6 +297,32 @@ enum Solvers {
...
+	// Submit if we are leading, or if the solution that's leading is more that the given `Perbill`
+	// better than us. This helps detect obviously fake solutions and still combat them.
+	SubmitIfClaimBetterThan(Perbill),
+}
+
+impl FromStr for SubmissionStrategy {
+	type Err = std::num::ParseFloatError;
+
+	fn from_str(s: &str) -> Result<Self, Self::Err> {
+		let res = if s == "only-submit-if-leading" {
```
kianenigma, line 315:
```suggestion
		let res = if s == "--if-leading" {
```

utils/staking-miner/src/main.rs
```diff
@@ -294,6 +297,32 @@ enum Solvers {
...
+			Self::OnlySubmitIfLeading
+		} else if s == "always-submit" {
```
kianenigma, line 317:
```suggestion
		} else if s == "--always" {
```

utils/staking-miner/src/main.rs
```diff
@@ -294,6 +297,32 @@ enum Solvers {
...
+			Self::AlwaysSubmit
+		} else {
+			Self::SubmitIfClaimBetterThan(Perbill::from_float(s.parse()?))
+		};
```
kianenigma, line 321: this should parse something like `--if-better-than 10` (10 would be passed to `Perbill::from_precent`)

utils/staking-miner/src/monitor.rs
```diff
@@ -54,17 +63,63 @@ async fn ensure_no_previous_solution<
...
+	at: B::Hash,
 	us: &AccountId,
 ) -> Result<(), Error<T>> {
-	use EPM::signed::SignedSubmissions;
-	ext.execute_with(|| {
-		if <SignedSubmissions<T>>::get().iter().any(|ss| &ss.who == us) {
-			Err(Error::AlreadySubmitted)
-		} else {
-			Ok(())
+	let indices_key = storage_value(EPM_MODULE_PREFIX, b"SignedSubmissionIndices");
```
kianenigma, line 70: hard-coding `EPM_MODULE_PREFIX` is wrong, since it is coming from the name of the pallet in `construct_runtime`, and it could be different. 

You can however get this from `T`: https://paritytech.github.io/substrate/master/frame_support/traits/trait.PalletInfo.html#tymethod.name

kianenigma, line 70: Also, as discussed, you can make `SignedSubmissionIndices` public.

utils/staking-miner/src/monitor.rs
```diff
@@ -54,17 +63,63 @@ async fn ensure_no_previous_solution<
...
+		.unwrap_or_default();
+
+	for (_score, id) in indices {
+		let key = storage_value_by_key::<Twox64Concat>(
+			EPM_MODULE_PREFIX,
+			b"SignedSubmissionsMap",
+			&id.encode(),
+		);
+
+		if let Some(submission) =
```
kianenigma, line 84: I would wait until your other PR is merged (which you can do ASAP since it is all approved now) to rewrite this stuff with the new PRC helpers? I think the code would be cleaner.

utils/staking-miner/src/monitor.rs
```diff
@@ -54,17 +63,63 @@ async fn ensure_no_previous_solution<
...
 		}
-	})
+	}
+
+	Ok(())
+}
+
+/// Queries the chain for the best solution and checks whether the computed score
+/// is better than best known.
+async fn ensure_no_better_solution<T: EPM::Config, B: BlockT>(
```
kianenigma, line 100: This one needs a rewrite, there must have been a misunderstanding about it. 

We want to compare ourselves with two queues.

1. The stuff stored in `SignedSubmissionsMap`. These are unverified solutions, and are sorted based on score. We want to scan these solutions, and then react to them based on `SubmissionStrategy`. Recall that the reason that we might want to do something like `AlwaysSubmit` is that these solutions are unverified, and they might have a false high claimed score. 

2. Then you have `QueuedSolution`. This is a single verified solution, that is the current best solution. In practice, we don't need to check this, since it is always empty during the signed phase (it is only set at the end of the signed phase). But since you already have the code, we should just basically treat it always as `OnlySubmitIfLeading`. As in, we check this, if it exists and if it has a better score than us, then we don't submit. Otherwise, we do.

kianenigma (changes requested):
- Some misunderstanding about the queues that need to be checked.
- Can use `PalletInfo`.
- Can use the nice RPC helpers from the other PR.

utils/staking-miner/src/monitor.rs
niklasad1: I had some troubles to actually use the `StorageValue trait` here because of some trait bounds that are missing from the `Runtime`.

```bash
let xx = StorageValue::<EPM::SignedSubmissions::<Runtime>>::hashed_key();

    = note: required because of the requirements on the impl of `FullEncode` for `SignedSubmissions<kusama_runtime::Runtime>`
    = note: required because of the requirements on the impl of `FullCodec` for `SignedSubmissions<kusama_runtime::Runtime>`
note: required by a bound in `hashed_key`
```

utils/staking-miner/src/main.rs
```diff
@@ -298,6 +301,42 @@ enum Solvers {
...
+///
+impl FromStr for SubmissionStrategy {
+	type Err = String;
+
+	fn from_str(s: &str) -> Result<Self, Self::Err> {
+		let res = if s == "if-leading" {
+			Self::IfLeading
+		} else if s == "always" {
+			Self::Always
+		} else if s.starts_with("percent-better ") {
```
niklasad1, line 330: this is a little but weird/controversial one need to passing this a string such as:

```$ target/release/staking-miner --uri ws://localhost:9999 --seed //Alice monitor --submission-strategy "percent-better 1" phrag-mms
```

we could just take a number but :point_up: is a bit more readable

utils/staking-miner/src/monitor.rs
```diff
@@ -131,39 +205,57 @@ macro_rules! monitor_cmd_for { ($runtime:tt) => { paste::paste! {
...
+			let check_version_fut = tokio::spawn(async move {
+				crate::check_versions::<Runtime>(&rpc3).await
+			});
+
+			let ext_fut = tokio::spawn(async move {
+				crate::create_election_ext::<Runtime, Block>(rpc4, Some(hash), vec![]).await
+			});
+
+			// Run the calls concurrently and return once all has completed or any failed.
+			let res = tokio::try_join!(
```
niklasad1, line 244: //cc @emostov now this these calls are performed in parallel however the code is a bit nasty with the clones

utils/staking-miner/src/monitor.rs
```diff
@@ -131,39 +205,57 @@ macro_rules! monitor_cmd_for { ($runtime:tt) => { paste::paste! {
...
+
+			let check_version_fut = tokio::spawn(async move {
+				crate::check_versions::<Runtime>(&rpc3).await
+			});
+
+			let ext_fut = tokio::spawn(async move {
+				crate::create_election_ext::<Runtime, Block>(rpc4, Some(hash), vec![]).await
+			});
+
+			// Run the calls concurrently and return once all has completed or any failed.
```
niklasad1, line 243:
```suggestion
			// Run the calls in parallel and return once all has completed or any failed.
```

utils/staking-miner/src/monitor.rs
```diff
@@ -200,6 +293,25 @@ macro_rules! monitor_cmd_for { ($runtime:tt) => { paste::paste! {
...
+
+			let ensure_no_better_fut = tokio::spawn(async move {
+				ensure_no_better_solution::<Runtime, Block>(&rpc1, hash, score, config.submission_strategy).await
+			});
+
+			let ensure_no_signed_fut = tokio::spawn(async move {
+				ensure_no_ready_solution::<Runtime, Block>(&rpc2, hash, score).await
+			});
+
+			// Run the calls concurrently and return once all has completed or any failed.
```
niklasad1, line 307:
```suggestion
			// Run the calls in parallel and return once all has completed or any failed.
```

niklasad1:
@kianenigma can you take another look at this? I tried to address your grumbles and I hope I got it right this time

utils/staking-miner/src/main.rs
```diff
@@ -318,6 +357,10 @@ struct MonitorConfig {
 	/// The solver algorithm to use.
 	#[clap(subcommand)]
 	solver: Solvers,
+
+	/// Submission strategy to use.
```
emostov, line 361:
```suggestion
	/// Submission strategy to use. Ex: `--submission-strategy "percent-better 1"`.
```
Not sure what the convention is for clap CLIs, but it would be nice if the user could get an example arg for percent-better since its a bit tricky

utils/staking-miner/src/monitor.rs
```diff
@@ -43,21 +47,95 @@ async fn ensure_signed_phase<T: EPM::Config, B: BlockT<Hash = Hash>>(
...
+	let indices_key = StorageKey(EPM::SignedSubmissionIndices::<T>::hashed_key().to_vec());
+
+	let indices: SubmissionIndicesOf<T> = rpc
+		.get_storage_and_decode(&indices_key, Some(at))
+		.await
+		.map_err::<Error<T>, _>(Into::into)?
+		.unwrap_or_default();
+
+	// BTreeMap is ordered, take last to get the max score.
+	if let Some(curr_max_score) = indices.into_iter().last().map(|(s, _)| s) {
```
emostov, line 107: Not sure how all the underlying code works, but this looks like it will linearly traverse all the indices to get to the last one. `BTreeMap::last_key_value` presumably can do this with just logarithmic complexity. So this could be changed to `if let Some(curr_max_score) = indices.into_iter().last_key_value().map(|(s, _)| s)`

utils/staking-miner/src/monitor.rs
```diff
@@ -43,21 +47,95 @@ async fn ensure_signed_phase<T: EPM::Config, B: BlockT<Hash = Hash>>(
...
+	}
+
+	Ok(())
+}
+
+/// Checks whether the score is better than checked solution on chain.
+/// The actual execution strategy is ignored because this is checked.
+///
+/// The checked score is set in the end of the signed phase.
+async fn ensure_better_than_ready_solution<T: EPM::Config, B: BlockT>(
```
emostov, line 120: I don't think this is neccesary at all. The first time a `ReadySolution` is available is when we finalize the signed phase, which is at the very end. We should already check that the we are in the signed phase, which should implicitly ensure the `ReadySolution` does not exist. So I think you can remove this function

utils/staking-miner/src/monitor.rs
```diff
@@ -131,39 +209,58 @@ macro_rules! monitor_cmd_for { ($runtime:tt) => { paste::paste! {
 			config: MonitorConfig,
 		) {
 
+			async fn flatten<T>(
```
emostov, line 212: Just out of curiosity, why a function instead of a closue? Is it a limitation of the `async` keyword?

utils/staking-miner/src/monitor.rs
```diff
@@ -131,39 +209,58 @@ macro_rules! monitor_cmd_for { ($runtime:tt) => { paste::paste! {
...
+				}
+			}
+
 			let hash = at.hash();
 			log::trace!(target: LOG_TARGET, "new event at #{:?} ({:?})", at.number, hash);
 
-			// if the runtime version has changed, terminate.
+			// block on this because if this fails there is no way to recover from
+			// that error i.e, upgrade/downgrade required.
 			if let Err(err) = crate::check_versions::<Runtime>(&rpc).await {
```
emostov, line 227: imo, if possible we should try and put this in the try join and then have a special match case for the `check_versions` error where we do `tx.send(err.into()); return;`

utils/staking-miner/src/monitor.rs
emostov: > But since you already have the code, we should just basically treat it always as OnlySubmitIfLeading. As in, we check this, if it exists and if it has a better score than us, then we don't submit. Otherwise, we do.

Not sure if I am reading this correctly, but I think there is no need to check the queued solution - we could instead check the phase. AFAICT in the current code its an invariant that if we set a QueuedSolution, then by the end of the block the phase will not be Signed

utils/staking-miner/src/monitor.rs
```diff
@@ -200,6 +298,25 @@ macro_rules! monitor_cmd_for { ($runtime:tt) => { paste::paste! {
...
+			let rpc1 = rpc.clone();
+			let rpc2 = rpc.clone();
+
+			let ensure_no_better_fut = tokio::spawn(async move {
+				ensure_no_better_solution::<Runtime, Block>(&rpc1, hash, score, config.submission_strategy).await
+			});
+
+			let ensure_no_signed_fut = tokio::spawn(async move {
+				ensure_better_than_ready_solution::<Runtime, Block>(&rpc2, hash, score).await
+			});
```
emostov, line 310:
```suggestion
```

utils/staking-miner/src/monitor.rs
```diff
@@ -200,6 +298,25 @@ macro_rules! monitor_cmd_for { ($runtime:tt) => { paste::paste! {
...
+
+			// Run the calls in parallel and return once all has completed or any failed.
+			if tokio::try_join!(
+				flatten(ensure_no_better_fut),
+				flatten(ensure_no_signed_fut),
```
emostov, line 315:
```suggestion
```

utils/staking-miner/src/monitor.rs
emostov: If we need to check the phase we can just query `EPM::CurrentPhase::<Runtime>`, which should be a lot cheaper then trying to query for a full solution

utils/staking-miner/src/monitor.rs
```diff
@@ -43,21 +47,95 @@ async fn ensure_signed_phase<T: EPM::Config, B: BlockT<Hash = Hash>>(
...
+	score: sp_npos_elections::ElectionScore,
+	strategy: SubmissionStrategy,
 ) -> Result<(), Error<T>> {
-	use EPM::signed::SignedSubmissions;
-	ext.execute_with(|| {
-		if <SignedSubmissions<T>>::get().iter().any(|ss| &ss.who == us) {
-			Err(Error::AlreadySubmitted)
-		} else {
-			Ok(())
+	let epsilon = match strategy {
```
emostov, line 91: 👍

utils/staking-miner/src/monitor.rs
niklasad1: The `last_first APIs` are not stabilized yet so I can't use them yet but those seems to be optimized [see this](https://github.com/rust-lang/rust/issues/62924)

Since `BTreeMap::iter` implements `DoubleEndedIterator` this should be faster than linear AFAIU.

utils/staking-miner/src/monitor.rs
niklasad1: the issue is that  `try_join` will return as soon one future fails i.e, returns `Err()`

so this might fail but something else always get first and the main task will never get notified that 
the runtime version is outdated and it will try to spawn new tasks that always will fail..

that's the reason why it did it like this but I guess I could do `join(check_version, try_join(a, b, c))` instead

utils/staking-miner/src/main.rs
niklasad1: yeah, this is a bit weird maybe using an integer will work better with the CLI help menu.

utils/staking-miner/src/monitor.rs
niklasad1: async closures are not supported in rust yet, so that's why it's a separate function.

This could be a helper method because `StakingMinerError` needs `T` from the `Runtime` I made it an internal helper.. :(

utils/staking-miner/src/main.rs
kianenigma: I think this is fine 👍

utils/staking-miner/src/main.rs
```diff
@@ -586,7 +629,7 @@ async fn main() {
 	log::info!(target: LOG_TARGET, "connected to chain {:?}", chain);
 
 	any_runtime_unit! {
-		check_versions::<Runtime>(&rpc).await
+		check_versions::<Runtime>(&rpc).await.expect("Runtime version is invalid, terminating")
```
kianenigma, line 632: no -- we just spoke of this, we only need to check this strictly in `monitor` in other commands it can be a mismatch.

utils/staking-miner/src/monitor.rs
```diff
@@ -43,21 +47,95 @@ async fn ensure_signed_phase<T: EPM::Config, B: BlockT<Hash = Hash>>(
...
+{
+	let indices_key = StorageKey(EPM::SignedSubmissionIndices::<T>::hashed_key().to_vec());
+
+	let indices: SubmissionIndicesOf<T> = rpc
+		.get_storage_and_decode(&indices_key, Some(at))
+		.await
+		.map_err::<Error<T>, _>(Into::into)?
+		.unwrap_or_default();
+
+	for (_score, idx) in indices {
```
kianenigma, line 67: I kinda wanted to rewrite this with `.any()` but I realized async code in closures it still a PITA right?

utils/staking-miner/src/monitor.rs
```diff
@@ -43,21 +47,95 @@ async fn ensure_signed_phase<T: EPM::Config, B: BlockT<Hash = Hash>>(
...
+	let indices: SubmissionIndicesOf<T> = rpc
+		.get_storage_and_decode(&indices_key, Some(at))
+		.await
+		.map_err::<Error<T>, _>(Into::into)?
+		.unwrap_or_default();
+
+	// BTreeMap is ordered, take last to get the max score.
+	if let Some(curr_max_score) = indices.into_iter().last().map(|(s, _)| s) {
+		if !score.strict_threshold_better(curr_max_score, epsilon) {
+			return Err(Error::BetterScoreExist)
```
kianenigma, line 109:
```suggestion
			return Err(Error::StrategyNotSatisfied)
```

utils/staking-miner/src/monitor.rs
kianenigma: yeah, in the previous round of review I opted to keep this part, even though it should never be needed. Let's do what @emostov suggests for now: remove this check for simplicity.

utils/staking-miner/src/monitor.rs
kianenigma: AFAIK `fn` and `const` inside a body of the function are identical to being places in global scope. So, unlike closures, they do not capture anything from their scope.

utils/staking-miner/src/monitor.rs
niklasad1: I realized it would be slightly better to have a runtime version subscription here instead of polling `runtime_version` all over the place, when it received a new version -> quit all

utils/staking-miner/src/monitor.rs
niklasad1: it's possible but then we need convert the futures to a stream or batch them together.

Something like [FuturesOrdered](https://docs.rs/futures/latest/futures/stream/struct.FuturesUnordered.html#) or JSON-RPC batch request might make this is little bit nicer to read.

niklasad1:
> LGTM, as a bonus test, you can try and run this against wesntend or kusama (you probably need to "fake" the linked runtime spec version).

Running it right now, I'll keep you posted.

utils/staking-miner/src/monitor.rs
niklasad1: I'm going to revert the `create_election_ext` because it increases network usage significantly even if the we drop the requests the server will answer it anyway.

In addition we only need to fetch these in signed phase anyway still no performance issues what I'm aware of anyway.

utils/staking-miner/src/monitor.rs
niklasad1: unfortunately I ran into https://github.com/paritytech/substrate/issues/10910 occasionally when I tested it

niklasad1:
@emostov can you take another look at this as well?

utils/staking-miner/src/monitor.rs
```diff
@@ -131,39 +184,52 @@ macro_rules! monitor_cmd_for { ($runtime:tt) => { paste::paste! {
...
 			}
 
-			// we prefer doing this check before fetching anything into a remote-ext.
-			if ensure_signed_phase::<Runtime, Block>(&rpc, hash).await.is_err() {
-				log::debug!(target: LOG_TARGET, "phase closed, not interested in this block at all.");
+			let rpc1 = rpc.clone();
+			let rpc2 = rpc.clone();
+			let account = signer.account.clone();
+
+			let signed_phase_fut = tokio::spawn(async move {
```
emostov, line 211: I assume we do this here to prevent heavy network usage from `create_election_ext` and then again right before submitting to make sure its not pointless to submit? 

I think its fine for now, but in the future I think we should look into only doing it right before submitting. Even if it ends up with a lot of wasted network usage it should be fine as long as we are not executing this function righter _before_ the signed phase, since that is when we want to be really fast to get the solution in. But if there fear is that this will be executing at the end of the sign phase I think its fine because we will have awhile until the next signed phase, so the extra network usage won't slow itself down. (Assuming its just traffic to a node on the local network)

emostov (approved):

niklasad1:
bot merge

PR #4824, opened 2022-02-01
Update Crowdloan Account to FundIndex

shawntabrizi:


runtime/common/src/crowdloan.rs
```diff
@@ -171,7 +171,7 @@ pub struct FundInfo<AccountId, Balance, BlockNumber, LeasePeriod> {
 	/// as `BlockNumber`.
 	pub last_period: LeasePeriod,
 	/// Index used for the child trie of this fund
```
shawntabrizi, line 173:
```suggestion
	/// Unique index used to represent this fund.
```

runtime/common/src/crowdloan.rs
```diff
@@ -658,42 +658,42 @@ impl<T: Config> Pallet<T> {
 	///
 	/// This actually does computation. If you need to keep using it, then make sure you cache the
 	/// value and only call this once.
-	pub fn fund_account_id(index: ParaId) -> T::AccountId {
+	pub fn fund_account_id(index: FundIndex) -> T::AccountId {
 		T::PalletId::get().into_sub_account(index)
```
shawntabrizi, line 662: this is the main change, and UIs will need to update logic

runtime/common/src/crowdloan.rs
```diff
@@ -246,8 +246,8 @@ pub mod pallet {
 
 	/// Tracker for the next available trie index
 	#[pallet::storage]
-	#[pallet::getter(fn next_trie_index)]
-	pub(super) type NextTrieIndex<T> = StorageValue<_, u32, ValueQuery>;
+	#[pallet::getter(fn next_fund_index)]
+	pub(super) type NextFundIndex<T> = StorageValue<_, u32, ValueQuery>;
```
shawntabrizi, line 250: this is also a breaking change that needs to be migrated

PR #4825, opened 2022-02-01
collation-generation: Support compressed PoVs directly

bkchr:
This adds support for returning a compressed PoV directly to the
collation generation. This is mainly to not requiring to compress a PoV
twice.

cumulus companion: https://github.com/paritytech/cumulus/pull/942


ordian (approved):

bkchr:
We did not really test it. I would say it depends on the data that is part of the state proof.

xlc:
While compact encoding is used for extrinsic, they are not (generally) used for storages. So that means balances will have a lot of empty bytes which are potentially compressible.

PR #4827, opened 2022-02-01
Add `matches_prefix` function to `MultiLocation` and `Junctions`

apopiak:
What it says on the tin :wink: 

Motivated by https://github.com/paritytech/cumulus/pull/936#discussion_r796234771

Key motivation: `match_and_split` only allows a single `Junction` after the prefix but it can be useful to be able to just check for a prefix and ignore what comes after.

Possible extension: We could introduce a corresponding `match_and_split_tail` (or similar, just a naming suggestion) to allow splitting off arbitrary suffixes.

### DIFF
--- a/xcm/src/v0/multi_location.rs
+++ b/xcm/src/v0/multi_location.rs
@@ -367,6 +367,31 @@ impl MultiLocation {
 		return self.at(prefix.len())
 	}
 
+	/// Returns whether `self` begins with or is equal to `prefix`.
+	///
+	/// # Example
+	/// ```rust
+	/// # use xcm::v0::{Junction::*, MultiLocation::*};
+	/// # fn main() {
+	/// let m = X4(Parent, PalletInstance(3), OnlyChild, OnlyChild);
+	/// assert!(m.matches_prefix(&X2(Parent, PalletInstance(3))));
+	/// assert!(m.matches_prefix(&m));
+	/// assert!(!m.matches_prefix(&X2(Parent, GeneralIndex(99))));
+	/// assert!(!m.matches_prefix(&X1(PalletInstance(3))));
+	/// # }
+	/// ```
+	pub fn matches_prefix(&self, prefix: &MultiLocation) -> bool {
+		if self.len() < prefix.len() {
+			return false
+		}
+		for i in 0..prefix.len() {
+			if prefix.at(i) != self.at(i) {
+				return false
+			}
+		}
+		true
+	}
+
 	/// Mutates `self`, suffixing it with `new`. Returns `Err` in case of overflow.
 	pub fn push(&mut self, new: Junction) -> result::Result<(), ()> {
 		let mut n = MultiLocation::Null;
--- a/xcm/src/v1/multilocation.rs
+++ b/xcm/src/v1/multilocation.rs
@@ -253,6 +253,26 @@ impl MultiLocation {
 		self.interior.match_and_split(&prefix.interior)
 	}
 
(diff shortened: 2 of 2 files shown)


shawntabrizi (approved):

xcm/src/v0/multi_location.rs
```diff
@@ -367,6 +367,31 @@ impl MultiLocation {
...
+	/// # use xcm::v0::{Junction::*, MultiLocation::*};
+	/// # fn main() {
+	/// let m = X4(Parent, PalletInstance(3), OnlyChild, OnlyChild);
+	/// assert!(m.matches_prefix(&X2(Parent, PalletInstance(3))));
+	/// assert!(m.matches_prefix(&m));
+	/// assert!(!m.matches_prefix(&X2(Parent, GeneralIndex(99))));
+	/// assert!(!m.matches_prefix(&X1(PalletInstance(3))));
+	/// # }
+	/// ```
+	pub fn matches_prefix(&self, prefix: &MultiLocation) -> bool {
```
bkchr, line 383:
```suggestion
	pub fn starts_with(&self, prefix: &MultiLocation) -> bool {
```

Sounds more like the other methods available in rust

xcm/src/v0/multi_location.rs
```diff
@@ -367,6 +367,31 @@ impl MultiLocation {
...
+		if self.len() < prefix.len() {
+			return false
+		}
+		for i in 0..prefix.len() {
```
bkchr, line 387:
```suggestion
		prefix.iter().zip(self.iter()).all(|l, r| l == r)
```

This could be used to replace the rest of the function

xcm/src/v0/multi_location.rs
```diff
@@ -367,6 +367,31 @@ impl MultiLocation {
 		return self.at(prefix.len())
 	}
 
+	/// Returns whether `self` begins with or is equal to `prefix`.
+	///
+	/// # Example
+	/// ```rust
+	/// # use xcm::v0::{Junction::*, MultiLocation::*};
+	/// # fn main() {
```
bkchr, line 375:
```suggestion
```

Rust does this automatically

xcm/src/v0/multi_location.rs
```diff
@@ -367,6 +367,31 @@ impl MultiLocation {
...
+	/// let m = X4(Parent, PalletInstance(3), OnlyChild, OnlyChild);
+	/// assert!(m.matches_prefix(&X2(Parent, PalletInstance(3))));
+	/// assert!(m.matches_prefix(&m));
+	/// assert!(!m.matches_prefix(&X2(Parent, GeneralIndex(99))));
+	/// assert!(!m.matches_prefix(&X1(PalletInstance(3))));
+	/// # }
```
bkchr, line 381:
```suggestion
```

KiChjang (commented):
Code looks good, but was wondering if `match_and_split` can use `match_prefix`/`starts_with` underneath as well.

KiChjang:
Code should also be made against the XCMv3 branch instead of master, i.e. against `gav-xcm-v3`. cc @gavofyork

gavofyork (changes requested):
Should be against XCM v3 branch. Also the various suggested by @bkchr

xcm/src/v0/multi_location.rs
apopiak: That's how I implemented it in Cumulus, but here I figured I'd keep to the imperative style of `match_and_split` :smile: 
Can definitely do.

apopiak:
suderseded by #4835

PR #4835, opened 2022-02-02
Add `starts_with` function to `MultiLocation` and `Junctions`

apopiak:
What it says on the tin :wink: 

Motivated by https://github.com/paritytech/cumulus/pull/936#discussion_r796234771

Key motivation: `match_and_split` only allows a single `Junction` after the prefix but it can be useful to be able to just check for a prefix and ignore what comes after.

Possible extension: We could introduce a corresponding `match_and_split_tail` (or similar, just a naming suggestion) to allow splitting off arbitrary suffixes.

supersedes #4827

### DIFF
--- a/xcm/src/v1/multilocation.rs
+++ b/xcm/src/v1/multilocation.rs
@@ -834,13 +834,8 @@ impl Junctions {
 	/// # }
 	/// ```
 	pub fn match_and_split(&self, prefix: &Junctions) -> Option<&Junction> {
-		if prefix.len() + 1 != self.len() {
+		if prefix.len() + 1 != self.len() || !self.starts_with(prefix) {
 			return None
-		}
-		for i in 0..prefix.len() {
-			if prefix.at(i) != self.at(i) {
-				return None
-			}
 		}
 		self.at(prefix.len())
 	}
--- a/xcm/src/v0/multi_location.rs
+++ b/xcm/src/v0/multi_location.rs
@@ -356,13 +356,8 @@ impl MultiLocation {
 	/// # }
 	/// ```
 	pub fn match_and_split(&self, prefix: &MultiLocation) -> Option<&Junction> {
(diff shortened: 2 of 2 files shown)


xcm/src/v1/multilocation.rs
```diff
@@ -253,6 +253,26 @@ impl MultiLocation {
...
+	///
+	/// # Example
+	/// ```rust
+	/// # use xcm::v1::{Junctions::*, Junction::*, MultiLocation};
+	/// # fn main() {
+	/// let m = MultiLocation::new(1, X3(PalletInstance(3), OnlyChild, OnlyChild));
+	/// assert!(m.starts_with(&MultiLocation::new(1, X1(PalletInstance(3)))));
+	/// assert!(!m.starts_with(&MultiLocation::new(1, X1(GeneralIndex(99)))));
+	/// assert!(!m.starts_with(&MultiLocation::new(0, X1(PalletInstance(3)))));
+	/// # }
```
bkchr, line 267:
```suggestion
	/// let m = MultiLocation::new(1, X3(PalletInstance(3), OnlyChild, OnlyChild));
	/// assert!(m.starts_with(&MultiLocation::new(1, X1(PalletInstance(3)))));
	/// assert!(!m.starts_with(&MultiLocation::new(1, X1(GeneralIndex(99)))));
	/// assert!(!m.starts_with(&MultiLocation::new(0, X1(PalletInstance(3)))));
```

bkchr (approved):

KiChjang (approved):
Should we also use `starts_with` in the implementation of `match_and_split` as well?

apopiak:
> Should we also use `starts_with` in the implementation of `match_and_split` as well?

Could do, iteration in the middle could be replaced by `if !starts_with(prefix) { return None; }` I guess.

gavofyork (approved):

Issue #4763, opened 2022-01-21
Support running full nodes with overseer enabled


bkchr:
For parachains it would be nice to run full nodes with overseer enabled. We don't require a full overseer running, some stripped down version would be okay (mainly with networking and what is required for this). With https://github.com/paritytech/cumulus/pull/927 we will require the overseer for full nodes to be able to recover povs. Currently that works because we accidentally have polkadot nodes always run as collators when run inside a parachain node. However, we would like to solve this properly.

CC @drahnr

rphmeier:
We could start to make subsystems aware of the type of node they're embedded in (validator, collator, regular full-node) so that they can internally disable code-paths that aren't relevant or are too heavy for the context. It might also let us finally remove the `collator_side` / `validator_side` distinction in collator-protocol.

drahnr:
I assume we would separate the subset functionality based on the _role_ a node has.

If so, this would be rather simply to integrate and actually simplify the overseer initialization in the  `polkadot` package, since we could pass the role to the `impl OverseerGen` to which is then responsible to setup a concrete overseer instance. 

One caveat: Something needs to drain the channels or they will eventually overflow, depending on which subsystems are being removed / left in place. If a separate overseer struct is intended for different roles, this could also work, yet increase work to integrate with the malus/nemesis implementations.

bkchr:
> Something needs to drain the channels or they will eventually overflow,

Can you not just have some dummy subsystem that drains the channels?

drahnr:
You can and that'd be the easiest, except for the case where bsckchannels are used. The it depends how the cancellation errors are handled, but this has to be investigated if there are any such channels.

rphmeier:
> If so, this would be rather simply to integrate and actually simplify the overseer initialization in the polkadot package, since we could pass the role to the impl OverseerGen to which is then responsible to setup a concrete overseer instance.

The fact that we have to drain channels with 'dummy' subsystems in certain places is a byproduct of trying to do runtime polymorphism this way. i.e. role at runtime determining the set of subsystems at runtime. I think it might still be easier to have all subsystems decide internally how to behave depending on role.

bkchr:
FYI:

[...]

PR #4840, opened 2022-02-03
allow overseer to be enabled anyways

drahnr:
Ref #4763

CC @bkchr 

If this deviates from your expectation of behavior, then let's schedule a talk.

Assumption here was, we are not running disputes coordinator and as such we want to follow the longest chain as relay chain selection algorithm would do. A keystore most be present for the overseer to launch.


bkchr (approved):
Sorry, I had already seen it, but had forgotten to review it :D

bkchr:
Ahh and thanks

PR #4848, opened 2022-02-04
Remove stale migrations post 9.16 release

kianenigma:
All of the removed migrations are already applied. I think `SchedulerMigrationV3` is also removed but have to double check.


runtime/polkadot/src/lib.rs
```diff
@@ -1446,205 +1446,11 @@ pub type Executive = frame_executive::Executive<
 	frame_system::ChainContext<Runtime>,
 	Runtime,
 	AllPalletsWithSystem,
-	(SchedulerMigrationV3, FixCouncilDepositMigration),
+	SchedulerMigrationV3,
```
bkchr, line 1449: They should also be removed?

bkchr (changes requested):

runtime/polkadot/src/lib.rs
kianenigma: Done.

bkchr (approved):

bkchr:
Still has some unused imports :P

kianenigma:
Screen Shot 2022-02-08 at 17 52 42
 broken namely due to

runtime/polkadot/src/lib.rs
```diff
@@ -1630,7 +1629,7 @@ impl FixCouncilDepositMigration {
 
 impl OnRuntimeUpgrade for FixCouncilDepositMigration {
 	fn on_runtime_upgrade() -> frame_support::weights::Weight {
-		if VERSION.spec_version == 9150 {
+		if VERSION.spec_version == 9180 {
```
bkchr, line 1632: I think I already said this once, but do we really want this? I mean this now already failed once because of this :P

bkchr (approved):

runtime/polkadot/src/lib.rs
kianenigma: better not execute something that execute it twice. 

This was originally added in 9150, and no one removed it (the blame for which is on me as well) until now. If we didn't have had this `if`, we would have executed this 3 times already in 9.15, 9.16, and 9.17 :D

michalisFr (approved):

bkchr:
@kianenigma there is a warning while compiling.

PR #4849, opened 2022-02-04
staking miner: reuse ws conn for remote-ext

niklasad1:
Builds on top of https://github.com/paritytech/polkadot/pull/4716

It passes the already established WebSocket connection down to `remote externalities` to re-use the already established connection for that part.


emostov:
From what I understand the changes lgtm. I looked at the diff via https://github.com/paritytech/polkadot/compare/na-staking-miner-playground..na-staking-miner-reuse-ws-conn#

To make sure I understand please let me know if the following is correct:

`&*client` => deref's to the inner value and then we get a simple pointer it. This is used when the compiler knows `client` won't get dropped before the function it is getting passed too executes.

`client.clone()` => gives us a new Arc, increasing the reference counter. We need to use this when we spawn a new thread that is not guaranteed to finish before `client` goes out of scope in the current function

emostov (approved):

niklasad1:
You got everything correct Zeke again :)

Yeah, we could wrap it in a new type and `impl Deref trait` to get rid of that horrible syntax

utils/staking-miner/src/rpc.rs
```diff
@@ -0,0 +1,147 @@
...
+pub trait RpcApi {
+	/// Fetch system name
+	#[method(name = "system_chain")]
+	async fn system_chain(&self) -> RpcResult<String>;
+
+	/// Fetch a storage key
+	#[method(name = "state_getStorage")]
+	async fn storage(&self, key: &StorageKey, hash: Option<&Hash>) -> RpcResult<Option<Bytes>>;
+
+	/// Fetch the runtime version
```
kianenigma, line 52:
```suggestion
	/// Fetch the runtime version.
```

nit: punctuation in docs are not consistent.

utils/staking-miner/src/rpc.rs
```diff
@@ -0,0 +1,147 @@
...
+	}
+}
+
+#[rpc(client)]
+pub trait RpcApi {
+	/// Fetch system name
+	#[method(name = "system_chain")]
+	async fn system_chain(&self) -> RpcResult<String>;
+
+	/// Fetch a storage key
```
kianenigma, line 48:
```suggestion
	/// Fetch a storage key.
```

utils/staking-miner/src/rpc.rs
```diff
@@ -0,0 +1,147 @@
...
+
+impl std::fmt::Display for RpcHelperError {
+	fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+		<RpcHelperError as std::fmt::Debug>::fmt(self, f)
+	}
+}
+
+#[rpc(client)]
+pub trait RpcApi {
+	/// Fetch system name
```
kianenigma, line 44:
```suggestion
	/// Fetch system name.
```

utils/staking-miner/src/rpc.rs
```diff
@@ -0,0 +1,147 @@
...
+	#[method(name = "payment_queryInfo")]
+	async fn payment_query_info(
+		&self,
+		encoded_xt: &Bytes,
+		at: Option<&Hash>,
+	) -> RpcResult<RuntimeDispatchInfo<Balance>>;
+
+	/// Dry run an extrinsic at a given block. Return SCALE encoded ApplyExtrinsicResult.
+	#[method(name = "system_dryRun", aliases = ["system_dryRunAt"])]
+	async fn dry_run(&self, extrinsic: &Bytes, at: Option<&Hash>) -> RpcResult<Bytes>;
```
kianenigma, line 65: I think `Hash` is `Copy` and making it a ref doesn't make sense

utils/staking-miner/src/rpc.rs
```diff
@@ -0,0 +1,147 @@
...
+		item = Header
+	)]
+	fn subscribe_finalized_heads(&self) -> RpcResult<()>;
+}
+
+/// Wraps a shared websocket JSON-RPC client that can be cloned.
+#[derive(Clone, Debug)]
+pub(crate) struct SharedRpcClient(Arc<WsClient>);
+
+impl Deref for SharedRpcClient {
```
kianenigma, line 99: noice

utils/staking-miner/src/rpc.rs
```diff
@@ -0,0 +1,147 @@
...
+impl SharedRpcClient {
+	/// Consume and extract the inner client.
+	pub fn into_inner(self) -> Arc<WsClient> {
+		self.0
+	}
+
+	/// Create a new shared RPC client.
+	pub(crate) async fn new(uri: &str) -> Result<Self, RpcError> {
+		let client = WsClientBuilder::default()
+			.connection_timeout(Duration::from_secs(20))
```
kianenigma, line 116: coudl turn 20 and 60 to constants.

utils/staking-miner/src/rpc.rs
```diff
@@ -0,0 +1,147 @@
...
+			.max_request_body_size(u32::MAX)
+			.request_timeout(Duration::from_secs(60))
+			.build(uri)
+			.await?;
+		Ok(Self(Arc::new(client)))
+	}
+
+	/// Get the storage item.
```
kianenigma, line 124: I would expect this doc to explain how this is different than `self.storage`. In the same spirit, I would name it `decode_storage` or `get_storage_decoded`.

utils/staking-miner/src/rpc.rs
```diff
@@ -0,0 +1,147 @@
...
+			let decoded = <T as codec::Decode>::decode(&mut &*bytes.0)
+				.map_err::<RpcHelperError, _>(Into::into)?;
+			Ok(Some(decoded))
+		} else {
+			Ok(None)
+		}
+	}
+}
+
+/// Takes a future that returns `Bytes` and tries the decode those bytes into the type `Dec`
```
kianenigma, line 140: missing punctuation.

utils/staking-miner/src/monitor.rs
```diff
@@ -71,23 +64,21 @@ macro_rules! monitor_cmd_for { ($runtime:tt) => { paste::paste! {
...
 		type StakingMinerError = Error<$crate::[<$runtime _runtime_exports>]::Runtime>;
 
-		let (sub, unsub) = if config.listen == "head" {
-			("chain_subscribeNewHeads", "chain_unsubscribeNewHeads")
-		} else {
-			("chain_subscribeFinalizedHeads", "chain_unsubscribeFinalizedHeads")
-		};
-
-		let mut subscription: Subscription<Header> = client.subscribe(&sub, None, &unsub).await?;
+		let heads_subscription = ||
```
kianenigma, line 74: bad choice of name -- should not use `heads` in it, right?

kianenigma (approved):
very nice refactors, appreciated!

utils/staking-miner/src/rpc.rs
niklasad1: fair enough, it was more to have a uniform API since all other methods takes that params `by ref` but sure I can change that.

niklasad1:
bot merge

PR #4852, opened 2022-02-06
Use Length Summation for Slot Winners

shawntabrizi:
Rather than calculating slot winners with simply the sum of bids for each lease period, we use a weighted sum, where each bid counts for the lease periods before it too, taking into account the total time tokens are locked for, rather than just how much are locked.

So...

Before:

```
1 + 2 + 3 + 4 = 10
```

After:

```
1 * 1 + 2 * 2 + 3 * 3 + 4 * 4 = 30
```

Thus a single bid for lease period 8 for 10 UNITS will be worth 80.

A bid for lease periods 7 and 8 for 10 UNITS will be worth 150.

etc...

This may have impact in some fringe scenarios as shown in the change in test in this PR.

### DIFF
--- a/runtime/common/slot_range_helper/src/lib.rs
+++ b/runtime/common/slot_range_helper/src/lib.rs
@@ -147,11 +147,22 @@ macro_rules! generate_slot_range_len {
 	) => {
 		/// Return the length of occupying a `SlotRange`.
 		///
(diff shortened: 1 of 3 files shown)


shawntabrizi:
doesnt actually solve what we want

PR #4859, opened 2022-02-07
Zombienet: add disputes test

sandreim:


ordian (approved):
Does it include functionality of [0001 test](https://github.com/paritytech/polkadot/blob/master/node/malus/integrationtests/0001-dispute-valid-block.feature)? Should we replace it if so?

drahnr (approved):

sandreim:
> Does it include functionality of [0001 test](https://github.com/paritytech/polkadot/blob/master/node/malus/integrationtests/0001-dispute-valid-block.feature)? Should we replace it if so?

Yes it does. That test is simpler and smaller scale focused on testing the malus functionality. I wouldn't remove it or change, as it can help reason about failures of this new test.

zombienet_tests/functional/0002-parachains-disputes.feature
```diff
@@ -0,0 +1,69 @@
+Description: Disputes initiation, conclusion and lag
+Network: ./0002.parachains-disputes-local.toml
```
sandreim, line 2:
```suggestion
Network: ./0002.parachains-disputes.toml
```

zombienet_tests/functional/0002-parachains-disputes.feature
```diff
@@ -0,0 +1,69 @@
+Description: Disputes initiation, conclusion and lag
+Network: ./0002.parachains-disputes.toml
```
sandreim, line 2:
```suggestion
Network: ./0002-parachains-disputes.toml
```

PR #4896, opened 2022-02-10
Baseline weights for `force_apply_min_commission`

emostov:


runtime/polkadot/src/weights/pallet_staking.rs
```diff
@@ -60,7 +60,7 @@ impl<T: frame_system::Config> pallet_staking::WeightInfo for WeightInfo<T> {
 	// Storage: BagsList ListNodes (r:3 w:3)
 	// Storage: BagsList ListBags (r:2 w:2)
 	fn bond_extra() -> Weight {
-		(93_200_000 as Weight)
+		(56_216_000 as Weight)
```
emostov, line 63: @ggwpez or @shawntabrizi is there any reason we would expect an across the board reduction in weights?

ggwpez: Yep since we did not update the Polkadot et.al. weights after [#10692](https://github.com/paritytech/substrate/pull/10692) and https://github.com/paritytech/bench-bot/pull/79 which added some optimization flags.  
Re-running Substrate was good enough for now since Polkadot has that in the release pipeline.

shawntabrizi (approved):

PR #4923, opened 2022-02-15
Add missing XCM benchmarks for ReserveAssetDeposited

### LINKED ISSUE #4915: reserveTransferAssets does not work in Westend 0.9.16
the `reserveTransferAssets` in `xcmPallet` does not seem to work in Westend 0.9.16, failing with `UnweighableMessage` error. I tried `limitedReserveTransferAssets` and is working fine, so the reason seems to be the way the destination chain message is weighed locally (https://github.com/paritytech/polkadot/blob/41ab002d7451766324a9f314fee11c9c53314350/xcm/pallet-xcm/src/lib.rs#L801).

I think the issue is the `ReserveAssetDeposited`, which is set to be the maximum weight in Westend (https://github.com/paritytech/polkadot/blob/41ab002d7451766324a9f314fee11c9c53314350/runtime/westend/src/weights/xcm/mod.rs#L74), and makes the message non-weighable.

Is this intended behavior? I know we can use `limitedReserveTransferAssets` but I was just wondering whether this is something that was foreseen before

KiChjang:
This PR adds the missing weights and benchmark code for the `ReserveAssetDeposited` XCM instruction.

Aims to fix #4915.


KiChjang:
Shucks, the problem here is obvious -- Westend doesn't regard Westmint as a proper reserve location for WND!

runtime/westend/src/lib.rs
```diff
@@ -1547,13 +1547,18 @@ sp_api::impl_runtime_apis! {
 					Westmint::get(),
 					MultiAsset { fun: Fungible(1 * UNITS), id: Concrete(WndLocation::get()) },
 				));
+				pub const TrustedReserve: Option<(MultiLocation, MultiAsset)> = Some((
```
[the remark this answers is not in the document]
kianenigma: Screen Shot 2022-03-03 at 09 36 11

runtime/westend/src/weights/xcm/pallet_xcm_benchmarks_fungible.rs
```diff
@@ -1,4 +1,4 @@
-// Copyright 2017-2021 Parity Technologies (UK) Ltd.
+// Copyright 2021 Parity Technologies (UK) Ltd.
```
kianenigma, line 1:
```suggestion
// Copyright 2022 Parity Technologies (UK) Ltd.
```

xcm/pallet-xcm-benchmarks/src/fungible/benchmarking.rs
```diff
@@ -107,6 +107,25 @@ benchmarks_instance_pallet! {
 		// TODO: Check sender queue is not empty. #4426
 	}
 
+	reserve_asset_deposited {
+		let (trusted_reserve, transferable_reserve_asset) = T::TrustedReserve::get()
+			.ok_or(BenchmarkError::Skip)?;
+
+		let assets: MultiAssets = vec![ transferable_reserve_asset ].into();
```
kianenigma, line 114: just to refresh my memory on this as it was a while ago since I last looked at it: XCM benchmarks are all per-asset right? and then later on, elsewhere we sum all the weights per asset?

kianenigma (approved):
LGTM

runtime/westend/src/weights/xcm/pallet_xcm_benchmarks_fungible.rs
```diff
@@ -65,19 +64,23 @@ impl<T: frame_system::Config> WeightInfo<T> {
...
 	// Storage: Dmp DownwardMessageQueues (r:1 w:1)
 	pub(crate) fn transfer_reserve_asset() -> Weight {
-		(99_639_000 as Weight)
+		(54_395_000 as Weight)
 			.saturating_add(T::DbWeight::get().reads(7 as Weight))
 			.saturating_add(T::DbWeight::get().writes(5 as Weight))
 	}
+	// Storage: Benchmark Override (r:0 w:0)
+	pub(crate) fn reserve_asset_deposited() -> Weight {
+		(2_000_000_000_000 as Weight)
```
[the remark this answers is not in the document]
KiChjang: Actually no, because Westmint doesn't have a configured reserve location, so it can't be benched, and as such, any incoming `ReserveAssetDeposited` should have a high cost as it can't really be executed.

xcm/pallet-xcm-benchmarks/src/fungible/benchmarking.rs
KiChjang: Unsure what you mean by that, but this `transferable_reserve_asset` is provided by configuration, and so if the chain maintainer wants to benchmark different assets, they'll have to specify it via the `TrustedReserve` associated type.

runtime/westend/src/weights/xcm/pallet_xcm_benchmarks_fungible.rs
KiChjang: Looks like `file_header.txt` needs to be changed as well.

shawntabrizi (approved):
looks good

KiChjang:
bot merge

PR #4927, opened 2022-02-16
Companion for refactor election score #10834

niklasad1:
https://github.com/paritytech/substrate/pull/10834

### DIFF
--- a/utils/staking-miner/src/dry_run.rs
+++ b/utils/staking-miner/src/dry_run.rs
@@ -22,6 +22,7 @@ use crate::{
 use codec::Encode;
 use frame_support::traits::Currency;
 use jsonrpsee::rpc_params;
+use sp_npos_elections::ElectionScore;
 
 /// Forcefully create the snapshot. This can be used to compute the election at anytime.
 fn force_create_snapshot<T: EPM::Config>(ext: &mut Ext) -> Result<(), Error<T>> {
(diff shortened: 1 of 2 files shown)


utils/staking-miner/src/dry_run.rs
```diff
@@ -63,10 +64,16 @@ async fn print_info<T: EPM::Config>(
...
+
+		let score = {
+			let ElectionScore { minimal_stake, sum_stake, sum_stake_squared } = raw_solution.score;
+			[Token::from(minimal_stake), Token::from(sum_stake), Token::from(sum_stake_squared)]
+		};
+
 		log::info!(
 			target: LOG_TARGET,
 			"solution score {:?} / deposit {:?} / length {:?}",
-			&raw_solution.score.iter().map(|x| Token::from(*x)).collect::<Vec<_>>(),
```
niklasad1, line 69: @kianenigma would be nice if the `iterator API` is public :)

kianenigma: unfortunately that iter api is some implementation detail (its order matter). Let's roll with this for now.

kianenigma (approved):

cheme (approved):
I believe we don't display the field names to get more compact log  and do not change existing display.
Sounds fine to me.

PR #4929, opened 2022-02-16
[DO NOT MERGE] zer0-nine-fifteen-plus-session-info

drahnr:
Teleports back in time, cherry picks #4545 and #4547 and converts pvf pre check API calls to NOPs


drahnr:
:point_right:  abyss

PR #4931, opened 2022-02-16
[Do not merge] Pipeline jobs optimizing

alvicsam:
I divided `test-build-linux-stable` job into 3 separate jobs: `test-linux-stable`, `test-linux-stable-node-metrics` and `build-linux-stable`
Also changed stages logic. Currently, there are 4 stages (1-4) instead of test-build-deploy-publish, and all jobs are connected using [DAG](https://docs.gitlab.com/ee/ci/directed_acyclic_graph/) in a way that all of them can finish during `test-linux-stable` run.

With these changes duration of the pipeline reduced for ~10-15 min.

The only thing that should be changed in GitHub is to remove `continuous-integration/gitlab-test-build-linux-stable` from required and mark as required these ones instead:
- `continuous-integration/test-linux-stable`
- `continuous-integration/test-node-metrics`
- `continuous-integration/build-linux-stable`

Part of https://github.com/paritytech/ci_cd/issues/326


.gitlab-ci.yml
```diff
@@ -292,10 +295,8 @@ build-malus:
     - echo "polkadot-test-malus = $(cat ./artifacts/VERSION) (EXTRATAG = $(cat ./artifacts/EXTRATAG))"
     - cp -r scripts/* ./artifacts
 
-#### stage:                        build
-
 .check-dependent-project:          &check-dependent-project
-  stage:                           build
+  stage:                           test
```
TriplEight, line 299: better leave it on the next stage before we're on 14.2, but put in a artificial dependency with `artifacts: false` after something that takes ~ 15 min

.gitlab-ci.yml
```diff
@@ -317,13 +318,46 @@ check-dependent-cumulus:
     DEPENDENT_REPO: cumulus
     CARGO_UPDATE_CRATES: "sp-io polkadot-runtime-common"
 
+#### stage:                        build
```
TriplEight, line 321: just looked into an entire pipeline, this one is better be named `publish`, it doesn't build much

.gitlab-ci.yml
```diff
@@ -177,41 +185,36 @@ test-deterministic-wasm:
...
     # Enable debug assertions since we are running optimized builds for testing
     # but still want to have debug assertions.
     RUSTFLAGS: "-Cdebug-assertions=y -Dwarnings"
-  rules:
-    - if: $CI_PIPELINE_SOURCE == "schedule"
-    - if: $CI_COMMIT_REF_NAME == "master"
-    - if: $CI_COMMIT_REF_NAME =~ /^v[0-9]+\.[0-9]+.*$/              # i.e. v1.0, v2.1rc1
-    - if: $CI_COMMIT_REF_NAME =~ /^[0-9]+$/                         # PRs
   script:
-    - ./scripts/gitlab/test_linux_stable.sh
```
TriplEight, line 196: guess we don't need this file any more, right?

.gitlab-ci.yml
TriplEight: and this comment moved beneath the following job too

.gitlab-ci.yml
```diff
@@ -338,6 +372,8 @@ generate-impl-guide:
   stage:                           build
   <<:                              *test-refs
   <<:                              *docker-env
+  needs:
+    - cargo-fmt
```
TriplEight, line 376: `  artifacts: false`

.gitlab-ci.yml
```diff
@@ -348,6 +384,8 @@ build-rustdoc:
   stage:                           build
   <<:                              *docker-env
   <<:                              *test-refs
+  needs:
+    - test-deterministic-wasm
```
TriplEight, line 388: `  artifacts: false`

TriplEight (changes requested):

TriplEight (changes requested):

alvicsam:
Closing PR because there were several changes in master, so I have to start from scratch. I'll create a new PR

PR #4934, opened 2022-02-16
paras: `initialize_para_now` and `ParachainsCache`

pepyakin:
This commit adds a new primitive called `ParachainsCache` to manipulate
the `Parachains` storage entry in a more convenient way.

Then, on top of that, this commit changes the logic of
`initialize_para_now` so that it is identical to what is used for
initialization of onboarding.

### DIFF
--- a/runtime/parachains/src/paras/mod.rs
+++ b/runtime/parachains/src/paras/mod.rs
@@ -537,6 +537,8 @@ pub mod pallet {
 		StorageValue<_, Vec<ValidationCodeHash>, ValueQuery>;
 
 	/// All parachains. Ordered ascending by `ParaId`. Parathreads are not included.
+	///
+	/// Consider using the [`ParachainsCache`] type of modifying.
 	#[pallet::storage]
 	#[pallet::getter(fn parachains)]
 	pub(crate) type Parachains<T: Config> = StorageValue<_, Vec<ParaId>, ValueQuery>;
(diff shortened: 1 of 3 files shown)


pepyakin:
* master
  * **PR #4934** <a href="https://app.graphite.dev/github/pr/paritytech/polkadot/4934" target="_blank">Graphite</a>  (this PR)
    * **PR #4938** <a href="https://app.graphite.dev/github/pr/paritytech/polkadot/4938" target="_blank">Graphite</a> 

runtime/parachains/src/paras/mod.rs
```diff
@@ -1994,19 +1951,78 @@ impl<T: Config> Pallet<T> {
...
-			"Parachain not created properly"
-		);
-		Ok(())
+	fn ensure_loaded(&mut self) -> &mut Vec<ParaId> {
+		if self.parachains.is_none() {
+			self.parachains = Some(Parachains::<T>::get());
+		}
+		self.parachains
+			.as_mut()
+			.expect("ensured above that the parachains list is Some; qed")
```
ordian, line 2001: https://doc.rust-lang.org/std/option/enum.Option.html#method.get_or_insert might come in handy

runtime/parachains/src/paras/mod.rs
```diff
@@ -1994,19 +1951,78 @@ impl<T: Config> Pallet<T> {
...
+		Self { parachains: None, _config: PhantomData }
+	}
 
-		// ensure it has become a para.
-		ensure!(
-			ParaLifecycles::<T>::get(id) == Some(ParaLifecycle::Parachain),
-			"Parachain not created properly"
-		);
-		Ok(())
+	fn ensure_loaded(&mut self) -> &mut Vec<ParaId> {
```
ordian, line 1995: nit: `ensure_initialized` sounds less ambiguous to me

pepyakin:
Thanks for the suggestions, fixed!

runtime/parachains/src/paras/mod.rs
```diff
@@ -683,30 +685,11 @@ pub mod pallet {
...
-
-			parachains.sort();
-			parachains.dedup();
-
-			Parachains::<T>::put(&parachains);
-
+			let mut parachains = ParachainsCache::new();
 			for (id, genesis_args) in &self.paras {
-				let code_hash = genesis_args.validation_code.hash();
-				<Pallet<T>>::increase_code_ref(&code_hash, &genesis_args.validation_code);
```
ordian, line 701: Previously, it didn't check for `validation.code.0.is_empty()`, but I assume this is never the case for the genesis otherwise it's borked?

ordian (approved):

runtime/parachains/src/paras/mod.rs
pepyakin: Yeah, good point, I should've mentioned this explicitly in the description.

I did not care that much about this case. Ultimately, using empty wasm code will fail during candidate validation. However, this is not the only way a PVF can be broken: all PVFs with less than 4-8 bytes are broken. Many others greater than that are at least useless. I'd say vast majority of byte combination are not valid wasm files or compressed blobs. Hence there is no explicit check.

However, you question gave me a thought: we are using an empty value as a sentinel value for a code path that won't be reachable from here. So I think this particular case warrants an empty code check.

PR #4950, opened 2022-02-20
XCM v3: Support for non-fungibles

gavofyork:
Dependent on https://github.com/paritytech/substrate/pull/10896

Provides XCM/MultiAsset support for using any implementer of the `assets::nonfungible` traits including, notably, Uniques pallet.

Also introduces the `pallet_uniques::Config::CreateOrigin` type param, which allows a `ContainsPair` to govern what origins are allowed to create what asset classes (NFT collections). Something like Statemint would probably not restrict this for its main Uniques pallet instance. However it's a game changer for when we have a Uniques pallet instance whose job is to hold *foreign* NFTs in a reserve mechanism. Here we can allow the location which hosts them to *create* the NFT collection and thus determine who pays the deposits for any assets minted therein. NFTs can then be transferred in using the usual ReserveAssetTransfer mechanism, or, if the chains trust each other, then beamed in with a Teleport.

An example and demonstration of such a setup is provided via the xcm-simulator example tests.

Also introduces an approval mechanism to prevent an NFT collection being transferred to an unwilling recipient who might then get griefed through having items minted using their account for the deposit.


amitrovich:
We were just starting to work on interoperability with Uniques. You  made our life easier :)

xcm/src/v3/junctions.rs
```diff
@@ -29,7 +29,9 @@ pub(crate) const MAX_JUNCTIONS: usize = 8;
 ///
 /// Parent junctions cannot be constructed with this type. Refer to `MultiLocation` for
 /// instructions on constructing parent junctions.
-#[derive(Clone, Eq, PartialEq, Ord, PartialOrd, Encode, Decode, Debug, TypeInfo, MaxEncodedLen)]
+#[derive(
+	Copy, Clone, Eq, PartialEq, Ord, PartialOrd, Encode, Decode, Debug, TypeInfo, MaxEncodedLen,
+)]
 pub enum Junctions {
```
KiChjang, line 35: I must have missed this in the previous PRs, but since `Junctions` are now `Copy`able, a lot of methods can in fact be `const` nowadays, the most important of which is `new`. This will allow `MultiLocation`s to be constructed in a const context, and one of the more interesting things that we can do is to statically assert that every `MultiLocation` constructed by the developer is at its most simplified form, i.e. they do not contain the `Ancestry` as a prefix.

xcm/src/v3/multiasset.rs
```diff
@@ -107,6 +109,130 @@ impl From<[u8; 32]> for AssetInstance {
...
+	fn from(x: u32) -> Self {
+		Self::Index(x as u128)
+	}
+}
+
+impl From<u64> for AssetInstance {
+	fn from(x: u64) -> Self {
+		Self::Index(x as u128)
+	}
+}
```
KiChjang, line 134: Missing `From<u128>` implementation for `AssetInstance`?

xcm/xcm-builder/src/asset_conversion.rs
```diff
@@ -0,0 +1,149 @@
+// Copyright 2020 Parity Technologies (UK) Ltd.
```
KiChjang, line 1:
```suggestion
// Copyright 2022 Parity Technologies (UK) Ltd.
```

xcm/xcm-builder/src/nonfungibles_adapter.rs
```diff
@@ -0,0 +1,252 @@
+// Copyright 2020 Parity Technologies (UK) Ltd.
```
KiChjang, line 1:
```suggestion
// Copyright 2022 Parity Technologies (UK) Ltd.
```

xcm/xcm-executor/src/lib.rs
```diff
@@ -400,6 +401,11 @@ impl<Config: config::Config> XcmExecutor<Config> {
 
 	/// Process a single XCM instruction, mutating the state of the XCM virtual machine.
 	fn process_instruction(&mut self, instr: Instruction<Config::Call>) -> Result<(), XcmError> {
+		log::trace!(
+			target: "xcm::execute",
```
KiChjang, line 405:
```suggestion
			target: "xcm::process_instruction",
```
?

xcm/xcm-simulator/example/src/lib.rs
```diff
@@ -60,40 +61,68 @@ decl_test_network! {
...
 }
 
-pub fn para_account_id(id: u32) -> relay_chain::AccountId {
-	ParaId::from(id).into_account()
+pub fn parent_account_id() -> parachain::AccountId {
+	let location = (Parent,);
+	parachain::LocationToAccountId::convert(location.into()).unwrap()
+}
+
+pub fn child_account_id(para: u32) -> relay_chain::AccountId {
```
KiChjang, line 69: Is this the para account ID instead?

xcm/src/v3/multiasset.rs
KiChjang: Oh I see now, `From<u128>` is deliberately not implemented because that's reserved for fungibles.

xcm/src/v3/junctions.rs
gavofyork: Nice yeah. I wasn't planning on necessarily leaving the `Copy` trait in, but for now it's required due to `MultiLocation` being used as an `Instance` type in Uniques pallet config.

xcm/xcm-simulator/example/src/lib.rs
gavofyork: yes; a child parachain (rather than a sibling parachain, as it would be when addressed from another parachain which shared the same parent).

Issue #4921, opened 2022-02-14
Only consider fresh votes in node side dispute spam slots


eskimor:
Consequences should be minor, but they do exist as outlined [here](https://github.com/paritytech/polkadot/pull/4845#issuecomment-1039121160). We should strive for correctness as much as possible in general and this is obviously not intended behavior.

PR #4957, opened 2022-02-21
Use only fresh DisputeStatement-s for dispute spam detection  (#4921)

tdimitrov:
handle_import_statements() from  Dispute-Coordinator does spam detection
by counting unconfirmed dispute statements from each validator.

This patch adjusts the logic so that only fresh dispute statements are
counted.

Fresh dispute statement in this case means a statement which was not
loaded from overlay db.

Fixes: https://github.com/paritytech/polkadot/issues/4921

### DIFF
--- a/node/core/dispute-coordinator/src/real/initialized.rs
+++ b/node/core/dispute-coordinator/src/real/initialized.rs
@@ -831,6 +831,9 @@ impl Initialized {
 			.find(|(_, index)| controlled_indices.contains(index))
 			.is_some();
 
+		// Indexes of the validators issued 'invalid' statements. Will be used to populate spam slots.
+		let mut fresh_invalid_statement_issuers = Vec::new();
+
 		// Update candidate votes.
 		for (statement, val_index) in &statements {
 			if validators
@@ -876,6 +879,7 @@ impl Initialized {
 						continue
 					}
 
+					fresh_invalid_statement_issuers.push(*val_index);
 					votes_changed = true;
 					self.metrics.on_invalid_vote();
 				},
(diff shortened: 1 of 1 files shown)


node/core/dispute-coordinator/src/real/initialized.rs
```diff
@@ -832,7 +832,8 @@ impl Initialized {
 			.is_some();
 
 		// Update candidate votes.
-		for (statement, val_index) in &statements {
+		let mut fresh_statements = HashSet::new(); // Used to keep track if the statement is fresh or not. Needed for spam detection.
+		for (idx, (statement, val_index)) in statements.iter().enumerate() {
```
eskimor, line 836: Why don't we just simply consume statements here via `into_iter` and push the whole statement data into a `fresh_invalid_statements` vec directly, in case it was found to be fresh & invalid. This way we can avoid the indexing - which is always good as it provides room for out of bounds errors. In the loop below we can drop the check for validity then, as the vec will only contain the invalid votes to begin with.

tdimitrov: I agree, indexing is error prone and I didn't realise we care only about invalid statements.

This however means that we don't need to keep the invalid statements at all, because we only use the validator index (the actual index - `val_index`, not the index in the array). So bottom line - we can push only `val_index` in the vector.

Am I misunderstanding something?

eskimor: You are right, we only need the `ValidatorIndex` - let's just keep that.

eskimor (approved):
Nice!

node/core/dispute-coordinator/src/real/initialized.rs
```diff
@@ -832,6 +832,7 @@ impl Initialized {
 			.is_some();
 
 		// Update candidate votes.
+		let mut fresh_invalid_st_issuers = Vec::new(); // indexes of the validators issued 'invalid' statements
```
drahnr, line 835: nit: Avoid abbreviations whenever possible
nit: Move the dev comment to a separate line, two pane editor users on their laptop will thank you :)

tdimitrov: Thanks. I've fixed both.

PR #4965, opened 2022-02-22
Bump `tokio` to 1.17.0

koute:
Companion for https://github.com/paritytech/substrate/pull/10894

### DIFF
--- a/bridges/relays/client-substrate/Cargo.toml
+++ b/bridges/relays/client-substrate/Cargo.toml
@@ -14,7 +14,7 @@ jsonrpsee-ws-client = "0.3.1"
 log = "0.4.11"
 num-traits = "0.2"
 rand = "0.7"
-tokio = "1.17.0"
+tokio = "1.8"
 thiserror = "1.0.26"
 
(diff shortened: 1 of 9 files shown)


KiChjang (approved):

bridges/relays/client-substrate/Cargo.toml
```diff
@@ -14,7 +14,7 @@ jsonrpsee-ws-client = "0.3.1"
 log = "0.4.11"
 num-traits = "0.2"
 rand = "0.7"
-tokio = "1.8"
+tokio = "1.17.0"
```
bkchr, line 17:
```suggestion
tokio = "1.8"
```

These kind of unnecessary changes should be done in the bridges repo.

bkchr (approved):

bridges/relays/client-substrate/Cargo.toml
wigy-opensource-developer: If we have code here that is not needed, why do not we delete the whole folder?

PR #4972, opened 2022-02-22
Add metrics for out of view statement distribution errors

vstakhov:
This PR is intended to cover the situation when we loose some messages during the statement distribution propagation and it causes a chain of the peer reputation changes requests. The idea is to store the count of such a messages in a metric and avoid reputation change flood constrained per relay parent block.


node/network/statement-distribution/src/lib.rs
```diff
@@ -239,6 +239,10 @@ struct PeerRelayParentKnowledge {
 	/// connecting again with new peer ids, but we assume that the resulting effective bandwidth
 	/// for such an attack would be too low.
 	large_statement_count: usize,
+
+	/// We have seen a message that come out of view from this peer, so note this fact
+	/// and stop subsequent logging and peer reputation flood
```
drahnr, line 244:
```suggestion
	/// and stop subsequent logging and peer reputation flood.
```

node/network/statement-distribution/src/lib.rs
```diff
@@ -1330,15 +1347,33 @@ async fn handle_incoming_message<'a>(
 	// perform only basic checks before verifying the signature
 	// as it's more computationally heavy
 	if let Err(rep) = peer_data.check_can_receive(&relay_parent, &fingerprint, max_message_count) {
+		// This situation can happen when a peer's Seconded message was lost
+		// but we have received the Valid statement.
+		// So we check it once and then ignore repeated violation to avoid
+		// reputation change flood.
+		let mut need_report_peer = true;
+
+		if rep == COST_UNEXPECTED_STATEMENT_UNKNOWN_CANDIDATE {
```
drahnr, line 1356: nit: Not keen on using this as a discriminator, this should be the exception.

node/network/statement-distribution/src/lib.rs
```diff
@@ -1330,15 +1347,33 @@ async fn handle_incoming_message<'a>(
 	// perform only basic checks before verifying the signature
 	// as it's more computationally heavy
 	if let Err(rep) = peer_data.check_can_receive(&relay_parent, &fingerprint, max_message_count) {
+		// This situation can happen when a peer's Seconded message was lost
+		// but we have received the Valid statement.
+		// So we check it once and then ignore repeated violation to avoid
+		// reputation change flood.
+		let mut need_report_peer = true;
```
drahnr, line 1354: nit: var name needs a bit more bike shedding :)

node/network/statement-distribution/src/lib.rs
vstakhov: Yes, I was not sure it is a good idea, but `can_report` might return different outcomes, and some of them must be treated seriously (like a corrupted statement or apparent flood).

node/network/statement-distribution/src/lib.rs
```diff
@@ -1970,6 +2006,12 @@ impl Metrics {
 	) -> Option<metrics::prometheus::prometheus::HistogramTimer> {
 		self.0.as_ref().map(|metrics| metrics.network_bridge_update_v1.start_timer())
 	}
+
```
drahnr, line 2009: This should have a short description.

node/network/statement-distribution/src/lib.rs
```diff
@@ -502,6 +506,19 @@ impl PeerData {
...
+	/// Receive a notice about out of view statement and returns the value of the old flag
+	fn receive_out_of_view_statement(&mut self, relay_parent: &Hash) -> bool {
+		self.view_knowledge
+			.get_mut(relay_parent)
+			.map_or(false, |relay_parent_peer_knowledge| {
+				let old = relay_parent_peer_knowledge.seen_out_of_view;
+				if !old {
+					relay_parent_peer_knowledge.seen_out_of_view = true;
+				}
+				old
```
drahnr, line 518:
```suggestion
				let old = relay_parent_peer_knowledge.seen_out_of_view;
				relay_parent_peer_knowledge.seen_out_of_view = true;
				old
```

drahnr, line 518:
```suggestion
				let mut old = true;
				std::mem::swap(&mut old, &mut relay_parent_peer_knowledge.seen_out_of_view)
				old
```

drahnr (approved):
A few nits that should be resolved before merging, generally looks good to me :+1:

node/network/statement-distribution/src/lib.rs
vstakhov: The main issue is that it actually compares two strings in this case :(

node/network/statement-distribution/src/lib.rs
vstakhov: Well, branch predictor success (which will be successfull in all branches but the first one) is almost free on x86, whilst unconditionally set `relay_parent_peer_knowledge.seen_out_of_view = true;` emits a store instruction all the time that is not free :) But I agree that it's a completely negligible micro-optimization.

drahnr: Code simplification > micro-opt.

I think we don't really require this optimization to only log it once, as long as it's trace.

vstakhov: Then one line: `std::mem::replace(&mut relay_parent_peer_knowledge.seen_out_of_view, true)`

node/network/statement-distribution/src/lib.rs
```diff
@@ -2022,6 +2066,13 @@ impl metrics::Metrics for Metrics {
 				))?,
 				registry,
 			)?,
+			statements_out_view: prometheus::register(
+				prometheus::Counter::new(
+					"polkadot_parachain_statement_distribution_statements_out_view",
```
sandreim, line 2071: Should we also differentiate between what kind of statements which arrived out of view via some label ("kind")?  `LargeStatement`,`Seconded` or `Valid`

node/network/statement-distribution/src/lib.rs
```diff
@@ -239,6 +239,10 @@ struct PeerRelayParentKnowledge {
 	/// connecting again with new peer ids, but we assume that the resulting effective bandwidth
 	/// for such an attack would be too low.
 	large_statement_count: usize,
+
+	/// We have seen a message that come out of view from this peer, so note this fact
+	/// and stop subsequent logging and peer reputation flood.
+	seen_out_of_view: bool,
```
sandreim, line 245: This can be an integer so we can count how many times this happened. Does it add any value for debugging ?

node/network/statement-distribution/src/lib.rs
vstakhov: It is now only for tracking like `expected` out-of-view statements when we have received `Valid` but `Seconded` has been lost in transition. Probably, we might label this metric in future with different cases.

node/network/statement-distribution/src/lib.rs
vstakhov: Yes, I was thinking about adding that as `usize` but I was also not sure if it would be useful.

sandreim (approved):

vstakhov:
I have slightly reworked that addressing @sandreim comments.

sandreim (approved):
LGTM

sandreim:
bot merge

node/network/statement-distribution/src/lib.rs
```diff
@@ -1928,25 +1970,50 @@ struct MetricsInner {
...
+}
+
+impl From<UnexpectedStatementLabel> for &'static str {
+	fn from(label: UnexpectedStatementLabel) -> Self {
+		match label {
+			UnexpectedStatementLabel::Seconded => "seconded",
+			UnexpectedStatementLabel::Valid => "valid",
+			UnexpectedStatementLabel::Large => "large",
+		}
+	}
```
drahnr, line 1990: nit: Commonly implementing `std::fmt::Display` is the preferred way.

node/network/statement-distribution/src/lib.rs
```diff
@@ -1928,25 +1970,50 @@ struct MetricsInner {
...
 }
 
 /// Statement Distribution metrics.
 #[derive(Default, Clone)]
 pub struct Metrics(Option<MetricsInner>);
 
 impl Metrics {
+	/// Update statements distributed counter
```
drahnr, line 1998: nit: With content breaking the 2k loc, it might make sense to move this into a separate file `metrics.rs`

node/network/statement-distribution/src/lib.rs
drahnr: actually in this specific case, we use each variant only once, 3 constants or just string literals would be even better

node/network/statement-distribution/src/lib.rs
vstakhov: Done.

node/network/statement-distribution/src/lib.rs
vstakhov: Yes, I agree - I planned to use more variants but even 3 is redundant as we actually see only `valid` label all the time. But I have converted that to constants.

node/network/statement-distribution/src/requester.rs
```diff
@@ -161,6 +165,7 @@ pub async fn fetch(
 					);
 
 					metrics.on_received_response(false);
+					metrics.on_unexpected_statement(UNEXPECTED_STATEMENT_LARGE_LABEL);
```
drahnr, line 168: nit: we probably want to move the const to the `metrics.rs` side, and only expose `on_unexpected_statement_large()` and friends, this reduces visual complexity quite a bit. I promise this is the last thing :) - if that's resolved, feel free to merge as is :+1:

drahnr (approved):

drahnr (approved):

drahnr:
bot merge

PR #4982, opened 2022-02-24
Co #10897: Storage benchmarking

ggwpez:
- Add the `benchmark-storage` sub-command from https://github.com/paritytech/substrate/pull/10897  

Disclaimer: This **CAN CORRUPT YOUR DB** so *please* make a backup before using this on a real snapshot; [discussion](https://github.com/paritytech/substrate/pull/10897#discussion_r813904923)


shawntabrizi (approved):

cheme (approved):

cli/src/cli.rs
```diff
@@ -54,6 +54,9 @@ pub enum Subcommand {
 	#[clap(name = "benchmark", about = "Benchmark runtime pallets.")]
 	Benchmark(frame_benchmarking_cli::BenchmarkCmd),
 
+	#[clap(name = "benchmark-storage", about = "Benchmark storage speed.")]
```
NukeManDan, line 57: > Disclaimer: This CAN CORRUPT YOUR DB so please make a backup before using this on a real snapshot; https://github.com/paritytech/substrate/pull/10897#discussion_r813904923

@ggwpez should the command help not mention something to this effect to help end users not "footgun" on this?

ggwpez: Running the command as is,will also require the `--storage-version` flag, so you cannot just run it accidentally.  
The corruption can only occur for `RocksDB`, but I agree to make it a bit more explicit.  
Maybe even add a `--i-used-a-backup` flag?

PR #4990, opened 2022-02-25
[ci] Create synth-wave docker image on tag

alvicsam:
Added rule to `publish-polkadot-image` so it publishes synth-wave image on release. Also renamed it to `publish-polkadot-debug-image`

Closes https://github.com/paritytech/ci_cd/issues/328

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -74,11 +74,21 @@ default:
     - sccache -s
 
 .test-refs:                        &test-refs
-  # these jobs run always*
+  # these jobs run on master and PR branches*
   rules:
+    - if: $CI_PIPELINE_SOURCE == "web"
     - if: $CI_PIPELINE_SOURCE == "schedule"
     - if: $CI_COMMIT_REF_NAME == "master"
     - if: $CI_COMMIT_REF_NAME =~ /^[0-9]+$/                         # PRs
+
+.common-refs:                      &common-refs
+  # these jobs run always*
+  rules:
+    - if: $CI_PIPELINE_SOURCE == "web"
+    - if: $CI_PIPELINE_SOURCE == "schedule"
+    - if: $CI_COMMIT_REF_NAME == "master"
+    - if: $CI_COMMIT_REF_NAME =~ /^[0-9]+$/                         # PRs
+    - if: $CI_COMMIT_REF_NAME =~ /^v[0-9]+\.[0-9]+.*$/              # i.e. v1.0, v2.1rc1
 
 .test-pr-refs:                     &test-pr-refs
   rules:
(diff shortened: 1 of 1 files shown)


.gitlab-ci.yml
```diff
@@ -400,24 +406,18 @@ build-rustdoc:
...
+# .github/workflows/publish-docker-release.yml
+publish-polkadot-debug-image:
   stage:                           build
   <<:                              *build-push-image
+  <<:                              *common-refs
   variables:
     <<:                            *image-variables
     # scripts/dockerfiles/polkadot_injected_debug.Dockerfile
     DOCKERFILE:                    dockerfiles/polkadot_injected_debug.Dockerfile
     IMAGE_NAME:                    docker.io/paritypr/synth-wave
```
PierreBesson, line 420: Would it be possible to rename the docker image to `paritypr/polkadot-debug` ?

alvicsam: For me the naming is ok, need to ask @TriplEight and @pepoviola. 

@pepoviola do you have any dependencies in `zombienet` related to `synth-wave` image and its current name?

TriplEight: as I mentioned in the issue

> [Synth](https://fallout.fandom.com/wiki/Synth)- stands not for the musical genre, but rather for synths, synthetic humanoids that would habitate the simnet. Any suggestion to a better name? `paritytech/polkadot` will be confusing, one can find themself on a host with a `polkadot` process running without realizing if the host is prod or not.

The name was a fit for `simnet`, but not it's no longer. I'm OK with renaming the image. Maybe something like `polkadot-goule` would keep us less formal?

.gitlab-ci.yml
```diff
@@ -400,24 +406,18 @@ build-rustdoc:
...
-  rules:
-    # Don't run on releases - this is handled by the Github Action here:
-    # .github/workflows/publish-docker-release.yml
-    - if: $CI_PIPELINE_SOURCE == "pipeline"
-      when: never
```
TriplEight, line 415: this was to avoid running the job i.e. when `paritytech/scrips` builds a CI image and triggers this CI.

TriplEight (changes requested):

.gitlab-ci.yml
PierreBesson: @TriplEight  Is `polkadot-debug` not a suitable name ? This image is not only needed for the Zombienet but also for Testnets (Westend/Rococo/...) or even potentially production networks for example to enable debug symbols for easy profiling.

TriplEight: @PierreBesson  nah, just wanted to make it less boring. If you feel this would work better, let's rename it, of course.

TriplEight (approved):

PR #4992, opened 2022-02-25
Add instructions for running local Rococo

gregzaitsev:


runtime/rococo/README.md
```diff
@@ -6,6 +6,11 @@ Rococo is a testnet runtime with no stability guarantees.
 
 > TODO: figure out how to run this properly.
 
+```
+cargo run --release -- --alice --tmp --validator --chain rococo-local &
```
bkchr, line 10: Alice is already "documented " below

PR #4998, opened 2022-02-27
sudo call to establish hrmp channel

### LINKED ISSUE #4648: Root call to open / accept / establish HRMP channel
Can we have root call to open / accept / establish HRMP channel?

We do have this on para sudo wrapper but it is for testing purpose only and can't be used for Kusama/Polkadot.

https://github.com/paritytech/polkadot/blob/2f4b4178256e9f92f0416de88b31d7abc41d283d/runtime/common/src/paras_sudo_wrapper.rs#L153

Use case:

Currently, the step to open HRMP channel with a common good parachain, requires a root call to send a XCM to transact a call on the common good parachain to send a XCM to transaction a call on relay chain to init open HRMP channel.

Given a root call is already required anyway, it will be helpful to avoid the extra XCM round trip as well as the ability to avoid deposit.

stanly-johnson:
resolves #4648

### DIFF
--- a/runtime/parachains/src/hrmp.rs
+++ b/runtime/parachains/src/hrmp.rs
@@ -492,12 +492,14 @@ pub mod pallet {
 			Ok(())
 		}
 
-		/// Sudo call to establish a channel from the sender to the recipient.
+		/// Call to establish a channel from the sender to the recipient.
 		///
 		/// This is equivalent to sending an `hrmp_init_open_channel` extrinsic followed by
 		/// `hrmp_accept_open_channel`.
+		///
+		/// Origin must be Root.
 		#[pallet::weight(<T as Config>::WeightInfo::hrmp_init_open_channel() + <T as Config>::WeightInfo::hrmp_accept_open_channel())]
-		pub fn sudo_establish_hrmp_channel(
+		pub fn force_establish_hrmp_channel(
 			origin: OriginFor<T>,
 			sender: ParaId,
 			recipient: ParaId,
--- a/runtime/parachains/src/hrmp.rs
+++ b/runtime/parachains/src/hrmp.rs
@@ -492,6 +492,29 @@ pub mod pallet {
 			Ok(())
 		}
 
+		/// Sudo call to establish a channel from the sender to the recipient.
+		///
+		/// This is equivalent to sending an `hrmp_init_open_channel` extrinsic followed by
+		/// `hrmp_accept_open_channel`.
+		#[pallet::weight(<T as Config>::WeightInfo::hrmp_init_open_channel() + <T as Config>::WeightInfo::hrmp_accept_open_channel())]
+		pub fn sudo_establish_hrmp_channel(
+			origin: OriginFor<T>,
+			sender: ParaId,
+			recipient: ParaId,
+			proposed_max_capacity: u32,
+			proposed_max_message_size: u32,
+		) -> DispatchResult {
+			ensure_root(origin)?;
+			Self::init_open_channel(
+				sender,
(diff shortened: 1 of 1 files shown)


runtime/parachains/src/hrmp.rs
```diff
@@ -492,6 +492,29 @@ pub mod pallet {
 			Ok(())
 		}
 
+		/// Sudo call to establish a channel from the sender to the recipient.
+		///
+		/// This is equivalent to sending an `hrmp_init_open_channel` extrinsic followed by
+		/// `hrmp_accept_open_channel`.
+		#[pallet::weight(<T as Config>::WeightInfo::hrmp_init_open_channel() + <T as Config>::WeightInfo::hrmp_accept_open_channel())]
+		pub fn sudo_establish_hrmp_channel(
```
bkchr, line 500:
```suggestion
		/// Call to establish a channel from the sender to the recipient.
		///
		/// This is equivalent to sending an `hrmp_init_open_channel` extrinsic followed by
		/// `hrmp_accept_open_channel`.
		///
		/// Origin must be Root.
		#[pallet::weight(<T as Config>::WeightInfo::hrmp_init_open_channel() + <T as Config>::WeightInfo::hrmp_accept_open_channel())]
		pub fn force_establish_hrmp_channel(
```

runtime/parachains/src/hrmp.rs
```diff
@@ -492,6 +492,29 @@ pub mod pallet {
...
+		) -> DispatchResult {
+			ensure_root(origin)?;
+			Self::init_open_channel(
+				sender,
+				recipient,
+				proposed_max_capacity,
+				proposed_max_message_size,
+			)?;
+			Self::accept_open_channel(recipient, sender)?;
+			Ok(())
```
bkchr, line 515:
```suggestion
			Self::accept_open_channel(recipient, sender)
```

bkchr (approved):

runtime/parachains/src/hrmp.rs
```diff
@@ -492,6 +492,30 @@ pub mod pallet {
...
+		) -> DispatchResult {
+			ensure_root(origin)?;
+			Self::init_open_channel(
+				sender,
+				recipient,
+				proposed_max_capacity,
+				proposed_max_message_size,
+			)?;
+			Self::accept_open_channel(recipient, sender)
+		}
```
pepyakin, line 517: Nit: Can we slap an `Ok(())` at the end here? Just for the consistency sake with the code around

pepyakin (approved):
Thanks,

One thing that we could be done here or in a follow-up is to see if we can remove this function from `paras_sudo_wrapper` to avoid duplication.

runtime/parachains/src/hrmp.rs
bkchr: This is my fault, I exactly proposed to remove the `Ok(())` :P

joepetrowski:
Would be nice to pass the `deposit`s into `{init, accept}_open_channel` so that channels opened by `Root` don't need a deposit. It would remove the need to transfer from the Treasury, and makes sense given that governance is the spam-limiting means.

runtime/parachains/src/hrmp.rs
gilescope: I'm with Basti on this one.

hbulgarini:
@stanly-johnson , do you think that you can address [this comment](https://github.com/paritytech/polkadot/pull/4998#issuecomment-1063869068) from joe?  I think this feature will very simplify the usage for opening hrmp channels.

PR #5000, opened 2022-02-28
lower kusama staking limits

kianenigma:
This PR is part of my effort to re-configure Kusama's staking system for long term safety, and prepare it for nomination pools. 

Recall that the NPoS election process, in order to be both successful and safe, contains 4 main steps. 
1. Snapshot creation.
2. Signed submissions from staking-miner.
3. Unsigned submissions from validators, as backup. 
4. Solution verification.

Our experimentation has unraveled that, as it stands now, the bottleneck for the Kusama network (and Polkadot and Westend) is the third step, and the largest input data for which this step could be executed, in WASM, without running out of resources (weight, memory, block-length), is when there are at most around 12500 nominators in the system. Note that this number is expectedly lower than Polkadot’s 22500, because Kusama opted to allow for up to 24 nominations per nominator. Thus, we need to assume the worst case in our experimentation, namely that each nominator is indeed using all of their 24 nominations. 

More information: https://docs.google.com/document/d/1aKWbaYnzyIp7mnWlQYccrQtGLOIuVw8SD277HZuR1vw/edit?usp=sharing


shawntabrizi:
> Note that this number is expectedly lower than Polkadot’s 22500, because Kusama opted to allow for up to 24 nominations per nominator. Thus, we need to assume the worst case in our experimentation, namely that each nominator is indeed using all of their 24 nominations.

I think with this in mind, it is better to reduce this number from 24 than to reduce the voter snapshot per block.

The upside of having people nominate 24 people instead of 16 people seems way smaller than the upside of including another 10,000 people into the staking system calculation logic.

bkchr:
I'm seconding @shawntabrizi

kianenigma:
One minor problem with going down from 24 to 16 will be that we need to give sufficient heads up for people to update their nominations, which might be tricky. If the maximum is set back to 16, all of those who are voting for more than 16 will no longer be treated as a nominator, and can also be chilled via `chill_other`.

Nonetheless, I am open to explore this path as well.

emostov (approved):
If we want to go down the path of 16 targets on Kusama I think that should be done as a follow up since it's more controversial and would force any nominators with more than 16 targets to change their strategy. Conversely, this current proposed change shouldn't require any intervention by active nominators and will improves safety by ensuring OCW (or any stage in the election process) won't OOM. 

I don't see any reason to punt off this fix while the community does bikesheding on the ideal number of targets.

kianenigma:
Indeed, and looking at the numbers again, Kusama is only about halfway into filling the proposed 12_500 limit, and with the current growth rate, it would be a long time until it is reached. 

We also have PRs like https://github.com/paritytech/substrate/pull/10340 which could solve this problem in a more novel way. In the meantime though, I strongly advice we move forward with this.

shawntabrizi (approved):
We will use this as a stop gap until we lower the number of nominations per nominator.

runtime/kusama/src/lib.rs
```diff
@@ -410,11 +410,7 @@ parameter_types! {
 
 	// 1 hour session, 15 minutes unsigned phase, 8 offchain executions.
 	pub OffchainRepeat: BlockNumber = UnsignedPhase::get() / 8;
-
-	/// Whilst `UseNominatorsAndUpdateBagsList` or `UseNominatorsMap` is in use, this can still be a
-	/// very large value. Once the `BagsList` is in full motion, staking might open its door to many
-	/// more nominators, and this value should instead be what is a "safe" number (e.g. 22500).
-	pub const VoterSnapshotPerBlock: u32 = 22_500;
+	pub const VoterSnapshotPerBlock: u32 = 12_500;
```
KiChjang, line 413: But y u remove comment?

kianenigma: comment was referring to the pre-bags-list world, we are passed that now :)

kianenigma:
Will leave this open for more feedback, or pushback, but ideally it whould be merged to be part of 9190 runtime.

runtime/kusama/src/lib.rs
KiChjang: Yeah but what I mean was, it should still have a comment explaining why we have set it to be `12_500`.

kianenigma:
next release might branch off soon, so merging this.

kianenigma:
bot merge

Issue #4874, opened 2022-02-08
Malus: implement malicious node cooperation


sandreim:
This will enable a malicious node to initiate a disputes and all the other malicious validators in the test net will vote invalid for that candidate.
We can extend this further to garbage candidate suggesting and backing.

rphmeier:
How do you plan to implement this?

sandreim:
My plan is to start with enhancing `dispute valid candidate` functionality. In this case the easiest way is to intercept the candidate validation subsystem and make it always return `ValidationResult::Invalid(reason)` as a proof of concept. We don't even need to (At this moment I am having some difficulties with not being able to intercept the messages to that subsystems.) This should allow to run high scale tests where 1/3 malicious nodes that not just start disputes, but also vote invalid for any candidates they are supposed to be validated as part of backing or approval.

Next I will implement real coordination via the Zombienet backchannel. Then malus nodes triggering malicious behaviour will notify the others of the event and let them know the `candidate_hash` for each everyone should prevent finalizing.  I am currently looking at the semantics of the different `ValidationResult::Invalid` reasons to see if we also need to coordinate on these or we can just always use for example `InvalidOutputs`.

At this point we can extend the Zombienet backchannel support to the other nemesis types, such that for example in case of a garbage candidate we inform the others to actually shortcircuit validation and back it up and approve it.

I'm not 100% sure that this is the best way to go about this, and I'd appreciate other ideas and thoughts.

rphmeier:
> but also vote invalid for any candidates they are supposed to be validated as part of backing or approval.

Note that nodes just ignore candidates in backing that are invalid and don't start disputes. We haven't implemented 'remote-unavailable' disputes.

sandreim:
Created a separate issue for switching to the backchannel. https://github.com/paritytech/polkadot/issues/5378

PR #4711, opened 2022-01-13
Malus: Implement storing invalid chunks

Lldenaurois:
This PR implements a first attempt at intercepting AvailabilityStoreMessages and writing invalid chunks into the availability store for every `StoreAvailableData` message intercepted by the MessageInterceptor.

PR #5011, opened 2022-03-01
Malus: improvements in `dispute ancestor` and `suggest garbage candidate` implementation

sandreim:
First step(s), see https://github.com/paritytech/polkadot/issues/4874

- Implemented fake validation as a component that is reused in `dispute_valid_candidate.rs` and `suggest garbage candidate.rs`.
- The fake validation is configurable to report valid candidates as invalid or report malicious candidates as valid to the backing/approval subsystems.
- Implemented `suggest garbage candidate`  using the draft from https://github.com/paritytech/polkadot/pull/4711 as a starting point

These two malus nemesis implementations will be used in the parachain zombienet test suite:
- malicious nodes initiate disputes of valid candidates (`dispute_valid_candidate`)
- honest nodes initiate disputes of invalid candidates backed by malicious nodes

I attached some screenshots to show how `suggest garbage candidate` malus nodes perform in a zombienet test. The first image shows disputes for malicious candidates concluding as valid and the inclusion blocks being finalized. The next two show disputes concluding as invalid and chain reversions.  

This shows up on honest validators when they run the PVF with a malicious PoV: 
```
2022-03-30 14:02:30 panicked at 'invalid block data format.: Error', /home/sandreim/polkadot/parachain/test-parachains/undying/src/wasm_[validation.rs:30](http://validation.rs:30/):58
```
Screenshot 2022-03-30 at 17 04 47
Screenshot 2022-03-30 at 17 48 43
Screenshot 2022-03-30 at 17 59 32


node/malus/src/malus.rs
```diff
@@ -48,11 +48,24 @@ enum NemesisVariant {
...
 #[allow(missing_docs)]
 struct MalusCli {
 	#[clap(subcommand)]
 	pub variant: NemesisVariant,
+
+	#[clap(long, arg_enum, ignore_case = true)]
+	pub fake_backing_validation: Option<FakeCandidateValidation>,
+
+	#[clap(long, arg_enum, ignore_case = true)]
+	pub fake_approval_validation: Option<FakeCandidateValidation>,
```
drahnr, line 68: Shouldnttis be part of the variant?

node/malus/src/variants/common.rs
```diff
@@ -0,0 +1,133 @@
...
+use crate::{interceptor::*, shared::MALUS, FakeCandidateValidation};
+use polkadot_node_primitives::{InvalidCandidate, ValidationResult};
+use polkadot_node_subsystem::messages::CandidateValidationMessage;
+
+#[derive(Clone, Debug)]
+/// An interceptor which fakes validation result with a preconfigured result.
+/// Replaces `CandidateValidationSubsystem`.
+pub struct ReplaceValidationResult {
+	fake_backing_validation: Option<FakeCandidateValidation>,
+	fake_approval_validation: Option<FakeCandidateValidation>,
```
drahnr, line 32: Default `false` should be sufficient?

drahnr: Or defining a tristate enum

node/malus/src/malus.rs
sandreim: Hmm. I think you are right, I considered this initially but then I was thinking that this makes sense as a general configuration option. Now, I think if it's not implemented for all variants it will confuse users.

node/malus/src/variants/common.rs
sandreim: There is also the option of combining these two into a single enum:
- Disabled
- BackingInvalid
- BakingAndApprovalInvalid
- ApprovalInvalid

Naming scheme is not the best but IMO it is not ambiguous at least.

node/malus/src/variants/dispute_valid_candidates.rs
```diff
@@ -21,19 +21,20 @@
 
 #![allow(missing_docs)]
 
+use clap::{ArgEnum, Parser};
```
drahnr, line 24: CC

node/malus/src/variants/dispute_valid_candidates.rs
```diff
@@ -47,6 +48,86 @@ use sp_keystore::SyncCryptoStorePtr;
 
 use std::sync::Arc;
 
+#[derive(ArgEnum, Clone, Debug, PartialEq)]
+#[clap(rename_all = "kebab-case")]
+pub enum FakeCandidateValidation {
```
drahnr, line 53:
```suggestion
#[clap(rename_all = "kebab-case")]
#[non_exhaustive]
pub enum FakeCandidateValidation {
```

node/malus/src/variants/dispute_valid_candidates.rs
```diff
@@ -136,8 +207,8 @@ impl OverseerGen for DisputeValidCandidates {
 		let crypto_store_ptr = args.keystore.clone() as SyncCryptoStorePtr;
 		let backing_filter = ReplaceApprovalsWithDisputes;
 		let validation_filter = ReplaceValidationResult::new(
-			self.fake_backing_validation.clone(),
-			self.fake_approval_validation.clone(),
+			self.opts.fake_validation.clone(),
+			self.opts.fake_validation_error.clone(),
```
drahnr, line 211: nit: these are flat byte sized enums, make em `Copy`.

node/malus/src/variants/common.rs
```diff
@@ -352,8 +373,22 @@ where
 						pov,
 						timeout,
 						response_sender,
+						commitments_hash,
 					),
 			} => {
+				// Behave normally if the `PoV` is not known to be malicious.
+				if pov.block_data.0.as_slice() != MALICIOUS_POV {
```
drahnr, line 380: I talked to @pepoviola and we want this to be determined by the backchannel rather than the PoV itself. But that's for a follow up PR :) the initial goal is achieve without it.

drahnr (changes requested):
Looks very good, there is one small thing we should discuss and a few cleanups required. Generally speaking :+1:

node/malus/src/variants/common.rs
sandreim: Yes, the plan is to use the backchannel long term, this is just the first step

node/core/backing/src/lib.rs
```diff
@@ -469,41 +477,39 @@ async fn validate_and_make_available(
 				"Validation successful",
 			);
 
-			// If validation produces a new set of commitments, we vote the candidate as invalid.
-			if commitments.hash() != expected_commitments_hash {
```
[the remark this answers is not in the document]
sandreim: This change actually moves the check from calling subsystems right after the parachain block has been executed and we have the outputs (ValidationResult). I believe this is an improvement as the `candidate-validation-subsystem` now avoids calling runtime to validate the candidate commitments when the initial commitments differ from what we get by executing the block. Does this make sense ?

node/core/approval-voting/src/lib.rs
```diff
@@ -2277,6 +2277,7 @@ async fn launch_approval(
 					available_data.pov,
 					APPROVAL_EXECUTION_TIMEOUT,
 					val_tx,
+					candidate.commitments_hash,
```
eskimor, line 2280: With `CandidateReceipt` we already have a struct combining both the descriptor and the commitments ... why not just pass that, instead of both items separately?

node/core/backing/src/lib.rs
```diff
@@ -469,41 +477,39 @@ async fn validate_and_make_available(
...
+						target: LOG_TARGET,
+						candidate_hash = ?candidate.hash(),
+						actual_commitments = ?commitments,
+						"Erasure root doesn't match the announced by the candidate receipt",
+					);
+					Err(candidate)
+				},
 			}
 		},
+		ValidationResult::Invalid(InvalidCandidate::ComittmentsHashMismatch) => {
```
eskimor, line 504: Typo

node/core/candidate-validation/src/lib.rs
```diff
@@ -413,29 +417,45 @@ where
...
-				// If neither the assumption of the occupied core having the para included or the assumption
-				// of the occupied core timing out are valid, then the persisted_validation_data_hash in the descriptor
-				// is not based on the relay parent and is thus invalid.
-				return Ok(ValidationResult::Invalid(InvalidCandidate::BadParent))
-			},
-			AssumptionCheckOutcome::BadRequest =>
-				return Err(ValidationFailed("Assumption Check: Bad request".into())),
+		match find_validation_data(&mut new_sender, &descriptor).await? {
+			Some((validation_data, validation_code)) => (validation_data, validation_code),
+			None => return Ok(ValidationResult::Invalid(InvalidCandidate::BadParent)),
```
eskimor, line 458: Honestly, not sure this refactoring improved readability.

node/core/candidate-validation/src/tests.rs
```diff
@@ -414,6 +423,7 @@ fn candidate_validation_ok_is_ok() {
 		Arc::new(pov),
 		Duration::from_secs(0),
 		&Default::default(),
+		commitments.hash(),
```
eskimor, line 426: I am missing a dedicated test, that makes sure that the commitment hash is now actually checked. It looks like we only recycled other tests and make them fail for a different reason?

eskimor (commented):
First pass, got half way through. Remaining review coming soon.

node/core/backing/src/lib.rs
sandreim: indee, ... it's sprinkled by copy paste everywhere ...

node/core/candidate-validation/src/lib.rs
sandreim: Totally agree on this one; I aimed to expose some means of getting the validation data from Malus. I assume there are better ways to do it and this is not necessary, but couldn't find them at the time of writing.

eskimor (commented):
Looks good overall - code is clean and relatively easy to follow as well.

node/malus/src/variants/dispute_valid_candidates.rs
```diff
@@ -105,15 +138,17 @@ impl OverseerGen for DisputeValidCandidates {
...
+			spawner.clone(),
+		);
 
 		prepared_overseer_builder(args)?
-			.replace_candidate_backing(move |cb| {
-				InterceptedSubsystem::new(
-					CandidateBackingSubsystem::new(spawner, crypto_store_ptr, cb.params.metrics),
-					filter,
-				)
+			.replace_candidate_backing(move |cb| InterceptedSubsystem::new(cb, backing_filter))
```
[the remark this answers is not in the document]
sandreim: Hmm, you are right, I added the fake validation subsystem to make sure the disputes can be forcefully concluded as invalid, but never tried to use just that to initiate the disputes.

node/core/candidate-validation/src/tests.rs
sandreim: done

node/malus/src/variants/common.rs
```diff
@@ -0,0 +1,347 @@
...
+		target: MALUS,
+		para_id = ?candidate_descriptor.para_id,
+		"ValidationResult: {:?}",
+		&result
+	);
+
+	response_sender.send(result).unwrap();
+}
+
+// TODO: Only fake validation for `MALICIOUS_POV`, otherwise behave normally.
```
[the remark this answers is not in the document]
sandreim: done

node/malus/src/variants/common.rs
```diff
@@ -0,0 +1,347 @@
...
+								msg: CandidateValidationMessage::ValidateFromChainState(
+									descriptor,
+									pov,
+									timeout,
+									response_sender,
+									commitments_hash,
+								),
+							})
+						}
+						self.send_validation_response(
```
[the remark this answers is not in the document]
sandreim: implemented

node/malus/src/variants/common.rs
```diff
@@ -0,0 +1,347 @@
...
+						descriptor,
+						pov,
+						timeout,
+						response_sender,
+						commitments_hash,
+					),
+			} => {
+				match self.fake_validation {
+					FakeCandidateValidation::BackingValid |
+					FakeCandidateValidation::BackingAndApprovalValid => {
```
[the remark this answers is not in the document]
sandreim: This is because this message is sent by the backing subsystem only. There is an identical check for the other susbsytem message sent by approval voting where check ApprovalValid | BackingAndApprovalValid for the same reason.

node/malus/src/variants/dispute_valid_candidates.rs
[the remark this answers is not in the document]
sandreim: refactored.

eskimor (approved):

drahnr (approved):

PR #5021, opened 2022-03-02
Revert "req/resp: use IfDisconnected::ImmediateError (#4253)"

rphmeier:
This reverts commit c18d42d495d686979d96c1777628064ca074bc35.

### DIFF
--- a/node/network/availability-distribution/src/pov_requester/mod.rs
+++ b/node/network/availability-distribution/src/pov_requester/mod.rs
@@ -66,7 +66,10 @@ where
 
 	ctx.send_message(NetworkBridgeMessage::SendRequests(
 		vec![full_req],
-		IfDisconnected::ImmediateError,
+		// We are supposed to be connected to validators of our group via `PeerSet`,
+		// but at session boundaries that is kind of racy, in case a connection takes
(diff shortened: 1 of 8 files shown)


rphmeier:
I will present some graphs, scraped with github.com/ordian/kuddelmuddel from Kusama

There are 2 sets of 2 graphs showing the amount of block delay for backing and availability. The first set is at approximately block 11,000,000 (before 0.9.16). The second set is at approximately block 11,630,000 or around now.

## Backing and Inclusion at 11,000,000 on Kusama

11005000-backing-2023
11005000-inclusion-2023

## Backing and Inclusion at 11,630,000 on Kusama

11630000-backing-2023
11630000-inclusion-2023


We can see that both backing and availability have gotten markedly worse, not just backing. This is consistent with the theories in https://github.com/paritytech/polkadot/issues/4911 
  1. The issues are validator-side, not collator-side
  2. The issues manifested in 0.9.16

ordian (approved):
Thanks for the confirmation. It seem to indicate authority discovery doesn't work well?

rphmeier:
Quite possibly. Either that or reserved peers.

rphmeier:
stale, unwanted? definitely obsolete

PR #5025, opened 2022-03-03
Fix fast-runtime

zjb0807:
There have `prod_or_fast` and `fast-runtime` feature, so revert #2167

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7467,6 +7467,7 @@ dependencies = [
  "sp-offchain",
  "sp-runtime",
  "sp-session",
+ "sp-state-machine",
  "sp-storage",
  "sp-timestamp",
  "sp-transaction-pool",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7467,7 +7467,6 @@ dependencies = [
  "sp-offchain",
  "sp-runtime",
  "sp-session",
- "sp-state-machine",
  "sp-storage",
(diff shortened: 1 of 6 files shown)


bkchr:
I don't see any reason why this should be done. This changes the behavior in quite some way.

zjb0807:
> I don't see any reason why this should be done. This changes the behavior in quite some way.

https://github.com/paritytech/polkadot/blob/a13f0b96fa1838d5ef12e1b4f68b6ba911692281/runtime/rococo/constants/src/lib.rs#L40-L43

`prod_or_fast` was used here, but the `EpochDurationInBlocks` was modified in `chain_spec.rs`. It's going to be a little confusing

zjb0807:
I need a stable rococo-local network to test para chain, `EPOCH_DURATION` is 1min and will cause `UnexpectedEpochChange`.

bkchr:
> I need a stable rococo-local network to test para chain, `EPOCH_DURATION` is 1min and will cause `UnexpectedEpochChange`.

I never had problems with 1 minute epochs. However, your change here would also just lead to 1 minute epochs when you use fast. You clearly don't want to wait 1 hour until your Parachain is onboarded. With the current code, you can also change the epoch/session length in the chain spec.

bkchr (approved):

bkchr:
bot merge

PR #5028, opened 2022-03-03
Replace TransactionByteFee with LengthToFee

notlesh:
This is a companion PR for https://github.com/paritytech/substrate/pull/10785/.

It is effectively a no-op as it replaces the previous constant modifier `TransactionByteFee` with the `ConstantModifierFee` (an impl of `WeightToFeePolynomial` which results in a constant modifier) using the previous value for each runtime.

cumulus companion: https://github.com/paritytech/cumulus/pull/1152


runtime/test-runtime/src/lib.rs
```diff
@@ -234,9 +234,9 @@ parameter_types! {
 
 impl pallet_transaction_payment::Config for Runtime {
 	type OnChargeTransaction = CurrencyAdapter<Balances, ()>;
-	type TransactionByteFee = TransactionByteFee;
 	type OperationalFeeMultiplier = OperationalFeeMultiplier;
 	type WeightToFee = WeightToFee;
+	type LengthToFee = ConstantModifierFee<Balance, TransactionByteFee>;
```
shawntabrizi, line 239:
```suggestion
	type LengthToFee = frame_support::weights::ConstantMultiplier<Balance, TransactionByteFee>;
```

kianenigma (approved):

shawntabrizi (approved):

PR #5035, opened 2022-03-05
Implement a blocking barrier for XCMv3

### LINKED ISSUE #4813: XCM: Blocking Barrier
Sometimes it can be useful to block all but a particular type of XCM from being executed. This would allow for that by:

- Introducing a proper Error type for the `ShouldExecute::should_execute`. This type would include the `Suspended` variant meaning "does not pass this barrier yet, but it might pass in a later block" .
- Introducing a new struct `RespectSuspension` impl for `ShouldExecute` in PalletXcm which takes an `Inner: ShouldExecute` and which returns `Suspended` if a state value `SuspendFlag` is set to `true` and returns the result of its `Inner::should_execute` otherwise.

This can work with more broad queue suspensions to ensure that only acceptable messages are processed under certain circumstances.

KiChjang:
Fixes #4813.

Introduces a new barrier called `RespectSuspension` which takes an inner barrier type and a type that implements `CheckSuspension`, which in turn is a trait containing a method named `is_suspended` returning a boolean to indicate whether we should suspend processing of the incoming XCM.

To get this to work, `ShouldExecute` is also modified to return a proper error type called `RejectReason`, which is an enum containing all the possible reasons why an XCM did not pass a barrier condition.


xcm/xcm-executor/src/traits/should_execute.rs
```diff
@@ -46,21 +67,59 @@ impl ShouldExecute for Tuple {
...
+///
+/// The trait method is given the same parameters as `ShouldExecute::should_execute`, so that the
+/// implementer will have all the context necessary to determine whether or not to suspend the
+/// XCM executor.
+///
+/// Can be chained together in tuples to have multiple rounds of checks. If all of the tuple
+/// elements returns false, then execution is not suspended. Otherwise, execution is suspended
+/// if any of the tuple elements returns true.
+pub trait CheckSuspension {
+	fn is_suspended<Call>(
```
KiChjang, line 101: Maybe what should happen here is to put the `is_suspended` method inside of `ShouldExecute`, that way we don't need to have a new trait just for this method.

KiChjang:
Superseded by #6670.

PR #5043, opened 2022-03-07
[DO NOT MERGE] remove ambiguous DAG dependencies

alvicsam:
Some of the DAG dependencies in the pipeline are ambiguous and can be removed


alvicsam:
This didn't work out. If all jobs in `stage2` are finished, jobs without DAG in `stage3` are waiting jobs in `stage1` which ruins the optimisation.

Issue #5050, opened 2022-03-07
Brides Cargo is self-referencing `git` dependancies


NukeManDan:
`git = "https://github.com/paritytech/polkadot"` is found in the bridges cargo.toml and points to `master` throughout, this may lead to unexpected behavior when building releases/ branches.

bkchr:
Bridges is a git subtree that should not diverge too much from the bridges repo. There is no problem with the builds currently, so I don't see any reason why we should change this.

NukeManDan:
Is there any good reason to have a crate reference it's repo cyclically this way? Overall if moved to `path` it will mean less overhead to maintain and less to go wrong IMHO. 

In releases you will reference `master` within the release branch unless we use the _correct_ `release-v...` branch explicitly instead of `master`. If missed, this could lead to confusion and latter on builds that wiped the lock file would be pulling from the wrong source upstream instead of the code in the branch itself via `path`. It also maintains the repo and local versions separately IIUC in the artifacts & lock file- so bloat and confusion inducing.

This PR only touches the cargo files to remove that possibility. It will be up to the team to maintain logic changes between the bridges repo and this subtree in `master` only then - the cargo file is mostly static.

PR #5051, opened 2022-03-07
[WIP] migrate to internal deps for bridges

NukeManDan:
Closes https://github.com/paritytech/polkadot/issues/5050


bkchr:
This is a git subtree. We don't need this.

PR #5052, opened 2022-03-07
add weights that need to be generated for rococo

brunopgalvao:
In an effort to get Rococo up-to-date, there are several weights that need to be generated.


runtime/rococo/src/weights/mod.rs
```diff
@@ -15,9 +15,23 @@
 
 //! A list of the different weight modules for our runtime.
 
-pub mod runtime_parachains_configuration;
+pub mod frame_system;
```
KiChjang, line 18: This is 100% the wrong fix, `frame_system` is a Rust crate that is imported and can be seen in `Cargo.toml`.

brunopgalvao: Removed. Worth noting that this is in Westend as well here:
- https://github.com/paritytech/polkadot/blob/b567c8bfa760a1fc57e980494de18bf7182ae3dd/runtime/westend/src/weights/mod.rs#L18

brunopgalvao: I am using Westend as a base to mirror Rococo.

ggwpez: This is fine since it is a `pub mod`, not `pub use` so it will use the local `runtime/rococo/src/weights/frame_system.rs` for that once it exists.

ggwpez:
Doing these one-by-one with the bot is a bit painful.  
Last time I did this for Substrate I logged into a reference hardware server and run them all at once.  
Maybe you can ask for access in chat and do it all at once as well.

brunopgalvao:
Thanks @ggwpez is there a command to run all benchmarks without having to specify one by one? I am also running these locally on my machine before I kick them off here.

ggwpez:
There is [this](https://github.com/paritytech/polkadot/blob/master/scripts/run_benches_for_runtime.sh) script, not sure how good this works or how up-to-date it is.    
I personally use my own one, will DM you on elements when I find it.

brunopgalvao:
Awesome, thanks @ggwpez appreciate it!

runtime/rococo/src/weights/mod.rs
KiChjang: Oh wait, I mistook this file's purpose as trying to redefine the `frame_system` pallet, while in fact it's simply trying to provide the weights for `frame_system`. No need to remove it then.

shawntabrizi:
@brunopgalvao please dont use the benchbot in this way.

The devops team has access to many machines which can be used directly to run benchmarks.

The benchbot only has one machine, which is shared across all runtime developers, which is enough for our needs, but wont scale if being used like this.

NukeManDan:
FYI merging master into this branch once https://github.com/paritytech/polkadot/pull/5139/files lands should require a rerun of auctions/slots IIUC

brunopgalvao:
Also, should note that other than `pallet_session` the following benchmarks still need to be created/integrated for Rococo:
- `runtime_common_crowdloan`
- `xcm`

ggwpez:
>Looks like pallet-session benchmark [is coupled with](https://github.com/paritytech/substrate/blob/master/frame/session/benchmarking/src/lib.rs#L44) pallet-staking which is not used in Rococo.
>Any thoughts on how to move forward?

I think there is no proper quick fix.  

As makeshift solution you can probably copy the weights from Polkadot and create an issue to fix it later.

NukeManDan:
The migration is not needed, as the reset was based on [Polkadot v0.9.18-rc4](https://github.com/paritytech/polkadot/releases/tag/untagged-7e053670fa5b1fec9b30) where the pallet is already updated AFAIU.

Also made an [issue](https://github.com/paritytech/substrate/issues/11068) about the coupling you found, Bruno. :bow: Adding the weights from another runtime as suggested is likely a good idea, although they will be wrong IIUC because the coupled pallet isn't here, and thus may affect the benchmark. :thinking:

shawntabrizi (approved):

runtime/rococo/src/weights/pallet_session.rs
```diff
@@ -0,0 +1,62 @@
...
+//! Autogenerated weights for `pallet_session`
+//!
+//! THIS FILE WAS AUTO-GENERATED USING THE SUBSTRATE BENCHMARK CLI VERSION 4.0.0-dev
+//! DATE: 2022-03-15, STEPS: `50`, REPEAT: 20, LOW RANGE: `[]`, HIGH RANGE: `[]`
+//! EXECUTION: Some(Wasm), WASM-EXECUTION: Compiled, CHAIN: Some("polkadot-dev"), DB CACHE: 1024
+
+// Executed Command:
+// ./target/production/polkadot
+// benchmark
+// --chain=polkadot-dev
```
ggwpez, line 25: Polkadot dev as well

runtime/rococo/src/weights/pallet_collective.rs
```diff
@@ -0,0 +1,176 @@
...
+// --chain=polkadot-dev
+// --steps=50
+// --repeat=20
+// --pallet=pallet_collective
+// --extrinsic=*
+// --execution=wasm
+// --wasm-execution=compiled
+// --heap-pages=4096
+// --header=./file_header.txt
+// --output=./runtime/polkadot/src/weights/
```
ggwpez, line 34: You just overwrote a Polkadot weigh file.

runtime/rococo/src/weights/pallet_collective.rs
```diff
@@ -0,0 +1,176 @@
...
+//! Autogenerated weights for `pallet_collective`
+//!
+//! THIS FILE WAS AUTO-GENERATED USING THE SUBSTRATE BENCHMARK CLI VERSION 4.0.0-dev
+//! DATE: 2021-08-27, STEPS: `50`, REPEAT: 20, LOW RANGE: `[]`, HIGH RANGE: `[]`
+//! EXECUTION: Some(Wasm), WASM-EXECUTION: Compiled, CHAIN: Some("polkadot-dev"), DB CACHE: 128
+
+// Executed Command:
+// target/release/polkadot
+// benchmark
+// --chain=polkadot-dev
```
ggwpez, line 25: Should be `production` and `rococo-dev`.

ggwpez (changes requested):

shawntabrizi:
@ggwpez Bruno created a follow up tracker for some of those things. I dont think it is a blocker to merge what has been done so far

ggwpez (approved):

shawntabrizi:
bot merge

runtime/rococo/src/weights/pallet_collective.rs
shawntabrizi: no, he copy pasted the polkadot weight file to here

runtime/rococo/src/weights/pallet_collective.rs
brunopgalvao: This was just copied from polkadot, this needs to be replaced with proper rococo weights.

PR #5060, opened 2022-03-08
Westend sync with Rococo - `use` uniformity & Babe `Randomness`  in Rococo

NukeManDan:
There is no intended code changes here it side the `Randomness` source that AFAIK requires no explicit migration or action in the next runtime upgrade including this.

We uniformity set `use` statements. The first in a few PRs to 0.9.18 to release Rococo and update the network to more closely parallel Westend where possible.


runtime/rococo/src/lib.rs
```diff
@@ -715,19 +708,14 @@ impl paras_registrar::Config for Runtime {
 	type WeightInfo = paras_registrar::TestWeightInfo;
 }
 
-/// An insecure randomness beacon that uses the parent block hash as random material.
-///
-/// THIS SHOULD ONLY BE USED FOR TESTING PURPOSES.
-pub struct ParentHashRandomness;
```
NukeManDan, line 721: This is the exception in this PR, Randomness is now derived from Babe only, as in westend.

bkchr (approved):

bkchr:
Next time please make the pr against master and then we backport it.

coderobe:
does this really need to be backported to the current release branch?

coderobe:
we think it's best for this to wait for 0.9.19, which shouldn't be that far into the future. i'd say let's close this PR and have the one targeting master land there, then it'll make it into .19

bkchr:
Who is we and why? This doesn't change anything important.

coderobe:
i spoke to @NukeManDan and @EgorPopelyaev. My point was that we're mostly through the release process already and i would like to avoid merging in things at this stage unless they're acutely necessary, I'd been told that it's technically fine for this to wait. We are picking up a faster release pace anyway.

If you think this one PR should really make it then alright, we can get it in - I just don't want to set a precedent for squeezing in changes.

bkchr:
Okay :)

PR #5064, opened 2022-03-09
Runtime dependent weights

ggwpez:
This MR makes the `BlockExecution`, `ExtrinsicBase`, `RocksDb` and `ParityDb` Weights runtime dependent.  
Later on https://github.com/paritytech/substrate/pull/10977 will be used to generate weights for each runtime instead of using the Substrate defaults.

Changes:
- Copy the Substrate weights from `frame_support` into each `*_runtime_constants` module
- Move the parameter weight types out of `runtime_common` into each runtime with a new `impl_runtime_weights!` macro
- Make `NORMAL_DISPATCH_RATIO` public since it is mentioned in the doc of `TargetBlockFullness`. Can make it private again if that is better.
- Fix some test in rococo

I checked that the runtime did not change: https://github.com/paritytech/polkadot/pull/5064#issuecomment-1064222621, CI confirmation would still be appreciated.

cumulus companion: https://github.com/paritytech/cumulus/pull/1076


runtime/common/src/lib.rs
```diff
@@ -71,53 +64,76 @@ pub type NegativeImbalance<T> = <pallet_balances::Pallet<T> as Currency<
 pub const AVERAGE_ON_INITIALIZE_RATIO: Perbill = Perbill::from_percent(1);
 /// We allow `Normal` extrinsics to fill up the block up to 75%, the rest can be used
 /// by  Operational  extrinsics.
-const NORMAL_DISPATCH_RATIO: Perbill = Perbill::from_percent(75);
+pub const NORMAL_DISPATCH_RATIO: Perbill = Perbill::from_percent(75);
```
ggwpez, line 67: This is the only intended change.

runtime/rococo/constants/src/lib.rs
```diff
@@ -97,15 +99,16 @@ mod tests {
...
 
 	#[test]
-	// This function tests that the fee for `MaximumBlockWeight` of weight is correct
+	// This function tests that the fee for `MAXIMUM_BLOCK_WEIGHT` of weight is correct
 	fn full_block_fee_is_correct() {
 		// A full block should cost 16 DOLLARS
-		println!("Base: {}", BlockWeights::get().get(DispatchClass::Normal).base_extrinsic);
-		let x = WeightToFee::calc(&BlockWeights::get().max_block);
+		println!("Base: {}", ExtrinsicBaseWeight::get());
+		let x = WeightToFee::calc(&MAXIMUM_BLOCK_WEIGHT);
```
ggwpez, line 111: Test here looks weird so I changed it to look like in the other runtimes; [in polkadot](https://github.com/paritytech/polkadot/blob/bb02373ef87d5d1efb84a9f22c1e32a71e000fc7/runtime/polkadot/constants/src/lib.rs#L109), [kusama](https://github.com/paritytech/polkadot/blob/bb02373ef87d5d1efb84a9f22c1e32a71e000fc7/runtime/kusama/constants/src/lib.rs#L106) and [westend](https://github.com/paritytech/polkadot/blob/bb02373ef87d5d1efb84a9f22c1e32a71e000fc7/runtime/westend/constants/src/lib.rs#L108) it is all done this way.  
Could also be macro generated :shrug:

shawntabrizi:
would be good to sanity check the metadata constants have not changed with regard to these weight parameters.

ggwpez:
Waiting for this to be resolved, since cumulus *should* :crossed_fingers: already build with this version: https://github.com/paritytech/polkadot/pull/5037#issuecomment-1063385282  

> would be good to sanity check the metadata constants have not changed with regard to these weight parameters.

I will extract the metadata manually and check but CI would be much preferred.

ggwpez:
I wrote a small script to compare the metadata between branches and there is no change for `kusama-dev`, `westend-dev`, 
`polkadot-dev`, `rococo-dev`. [compare.sh.txt](https://github.com/paritytech/polkadot/files/8224973/compare.sh.txt) maybe the CI devs can reuse that somehow.

kianenigma (approved):
Didn't look at the whole diff, but sounds right to me, and much cleaner now 👍

ggwpez:
bot merge

PR #5066, opened 2022-03-09
[ci] cancel pipeline if test-linux-stable fail

alvicsam:
Added job that will cancel pipeline if `test-linux-stable` fails.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -257,6 +257,8 @@ test-linux-stable:
     # but still want to have debug assertions.
     RUSTFLAGS: "-Cdebug-assertions=y -Dwarnings"
   script:
+    - sleep 10
+    - exit 1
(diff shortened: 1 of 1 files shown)


.gitlab-ci.yml
```diff
@@ -803,3 +803,23 @@ publish-rustdoc:
     - echo "___Rustdoc was successfully published to https://paritytech.github.io/polkadot/___"
   after_script:
     - rm -rf .git/ ./*
+
+#### stage:                        .post
+
+# This job cancels the whole pipeline if test-linux-stable job fails
```
TriplEight, line 809:
```suggestion
# This job cancels the whole pipeline if any of the DAG starter jobs fail.
# In a DAG, every jobs chain is executed independently of others. The `fail_fast` principle suggests
# to fail the pipeline as soon as possible to shorten the feedback loop.
```

TriplEight (changes requested):

TriplEight (approved):

PR #5076, opened 2022-03-10
Use ss58-registry instead of sub-token

wigy-opensource-developer:


utils/staking-miner/src/token.rs
```diff
@@ -0,0 +1,37 @@
...
+// along with Polkadot.  If not, see <http://www.gnu.org/licenses/>.
+
+use ss58_registry::{Token as WrappedToken, TokenAmount};
+
+static mut TOKEN: WrappedToken = TokenRegistry::Dot.into();
+
+pub struct Token;
+
+impl Token {
+	pub fn init(token: WrappedToken) {
```
niklasad1, line 24: I would rather mark these functions as unsafe because it really depends how it's used thus the type itself can't ensure that's is single threaded or not.

utils/staking-miner/src/token.rs
```diff
@@ -0,0 +1,37 @@
...
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with Polkadot.  If not, see <http://www.gnu.org/licenses/>.
+
+use ss58_registry::{Token as WrappedToken, TokenAmount};
+
+static mut TOKEN: WrappedToken = TokenRegistry::Dot.into();
```
niklasad1, line 19: no way around the `static mut` here?

wigy-opensource-developer:
@niklasad1 I agree with everything you said. My only goal here was to make an adapter in this project that uses the new TokenAmount code from ss58-registry instead of the sub-token crate. If you want to make deeper changes in this repo, feel free to take over from here, as we discussed with @kianenigma.

utils/staking-miner/src/token.rs
wigy-opensource-developer: The rest of the code needs to be injected with a Token or needs another ambient context implementation. The environmental crate comes to my mind. But since the whole crate is single threaded and there is no risk for reentrancy between Token method calls, I was not too concerned about the implementation.

utils/staking-miner/src/token.rs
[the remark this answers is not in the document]
wigy-opensource-developer: Thought about `impl !Sync`, but there is no self argument on the methods, so would not help much.

niklasad1:
Alright, this looks good to me just make the functions unsafe on `Token` then it should be ready to go.

utils/staking-miner/src/main.rs
```diff
@@ -630,8 +630,7 @@ async fn main() {
 			sp_core::crypto::set_default_ss58_version(
 				sp_core::crypto::Ss58AddressFormatRegistry::PolkadotAccount.into(),
 			);
-			sub_tokens::dynamic::set_name("WND");
-			sub_tokens::dynamic::set_decimal_points(1_000_000_000_000);
+			Token::init(WrappedToken { name: "WND", decimals: 12 });
```
kianenigma, line 633: hmm westend is not part of this crate? strange..

kianenigma (approved):
Looks overall good to me, and happy to see sub-tokens retire. Not sure why Westend is not in there though 🤔

utils/staking-miner/src/main.rs
wigy-opensource-developer: Yeah, it is missing from the JSON in the ss58-registry

niklasad1: https://github.com/paritytech/ss58-registry/issues/85

niklasad1:
@kianenigma 

Attila explained to me that his goal is to re-export `Token` from `sp_core` to avoid duplicated dependencies of `ss58` however we could do it in a followup PR

utils/staking-miner/src/token.rs
```diff
@@ -0,0 +1,37 @@
...
+		unsafe {
+			TOKEN = token;
+		}
+	}
+
+	pub fn from(value: u128) -> TokenAmount {
+		// safety: this program will always be single threaded, thus accessing global static is
+		// safe.
+		unsafe { TOKEN.amount(value) }
+	}
```
emostov, line 36: Same as above - at the very least it will help people using some types of IDE plugins when they hover over the function.

PR #5082, opened 2022-03-10
Rococo to use BABE randomness, Westend & Rococo runtime synchronization in file format

NukeManDan:
#5060 companion, targeting `master`


bkchr (approved):

coderobe (approved):
lgtm but there are some differing changes compared to https://github.com/paritytech/polkadot/pull/5060/files - make sure all the intended changes made it over into this PR

runtime/rococo/src/lib.rs
```diff
@@ -98,7 +92,7 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
 	spec_name: create_runtime_str!("rococo"),
 	impl_name: create_runtime_str!("parity-rococo-v2.0"),
 	authoring_version: 0,
-	spec_version: 9170,
+	spec_version: 9180,
```
KiChjang, line 95: I think we shouldn't be fiddling with this value unless we're making a release, but I'm not a release engineer so I'm not sure about this.

coderobe: hi i'm releng
this one is fine, it's already 9180 in the release branch - syncing that to master doesn't hurt

KiChjang (approved):
alles gut.

KiChjang:
bot merge

PR #5085, opened 2022-03-11
MultiAssets can be indexed and constructed with capacity

Dengjianping:
1. Allow to construct `MultiAssets` with capacity.
2. `MultiAssets` can be indexed.


KiChjang:
Any changes to XCM must be directed towards the `gav-xcm-v3` branch.

In addition, can you explain what is the use case for this change?

rdovgan (approved):

bkchr:
Please recreate if still required.

PR #5093, opened 2022-03-12
Adding `Fallback` on election failure

georgesdib:
Companion to paritytech/substrate#10988

Use the newly introduced `BoundedOnchainExecution `, and `UnboundedOnchainExecution ` for Genesis.

@kianenigma @emostov


kianenigma (changes requested):
`type Fallback` in our production networks should remain `NoFallback` since we don't quite know what are some good bounds. 

As for the `GovernanceFallback` as well, since this transaction can impose its own limits, I think actually for now we should set it to `UnboundedExecution`, because we don't know any 'reasonable static bounds' that we can assign to any `BoundedExecution` type. 

The missing step to figure all of this out is to create an accurate weight function for all of this `Solver` implementations, will make an issue for it.

georgesdib:
> `type Fallback` in our production networks should remain `NoFallback` since we don't quite know what are some good bounds.
> 
> As for the `GovernanceFallback` as well, since this transaction can impose its own limits, I think actually for now we should set it to `UnboundedExecution`, because we don't know any 'reasonable static bounds' that we can assign to any `BoundedExecution` type.
> 
> The missing step to figure all of this out is to create an accurate weight function for all of this `Solver` implementations, will make an issue for it.

I have reverted the changes including for Westend.

kianenigma (approved):

runtime/test-runtime/src/lib.rs
```diff
@@ -342,11 +337,15 @@ impl pallet_staking::Config for Runtime {
...
-	type VoterList = pallet_staking::UseNominatorsAndValidatorsMap<Runtime>;
+	type ElectionProvider = frame_election_provider_support::onchain::UnboundedExecution<
+		runtime_common::elections::OnChainSeqPhragmen<Self, Staking>,
+	>;
+	type GenesisElectionProvider = frame_election_provider_support::onchain::UnboundedExecution<
+		runtime_common::elections::OnChainSeqPhragmen<Self, Staking>,
+	>;
+	// Use the nominator map to iter voter AND no-ops for all SortedListProvider hooks. The migration
+	// to bags-list is a no-op, but the storage version will be updated.
+	type VoterList = pallet_staking::UseNominatorsMap<Runtime>;
```
kianenigma, line 348:
```suggestion
	type VoterList = pallet_staking::UseNominatorsAndValidatorsMap<Runtime>;
```

kianenigma:
@georgesdib thanks for making the bot work with your fork, helps with the automated process.

kianenigma:
Must look into why and how this caused benchmarks to break. 

Alternatively, we can merge this to keep the normal build of master working.

georgesdib:
> Must look into why and how this caused benchmarks to break.
> 
> Alternatively, we can merge this to keep the normal build of master working.

You think it's caused by the changes here? I tried to go back in time and see merged PRs from 2 days ago, and they seem to fail the same thing.

kianenigma:
hope I am not breaking anything for anyone, merging this manually to keep the master working.

Issue #5104, opened 2022-03-14
Histogram metric for collator-protocol indicating collation fetch times and failures


rphmeier:
Failures can be represented as +inf or some other way. But we'd like to visualize these together to get a picture of how long collation fetching takes and how often it succeeds.

rphmeier:
(noted duplicate by @vstakhov )

PR #5108, opened 2022-03-14
Add metrics timer for the `PerRequest` structure lifetime

vstakhov:
This PR simply adds a timer to observe `PerRequest` structure lifetime in the collator protocol module (validator side). Addresses #5104.

### DIFF
--- a/node/network/collator-protocol/src/validator_side/mod.rs
+++ b/node/network/collator-protocol/src/validator_side/mod.rs
@@ -131,6 +131,13 @@ impl Metrics {
 			.as_ref()
 			.map(|metrics| metrics.collator_peer_count.set(collator_peers as u64));
 	}
+
+	/// Provide a timer for `PerRequest` structure which observes on drop.
+	fn time_collation_request_duration(
+		&self,
+	) -> Option<metrics::prometheus::prometheus::HistogramTimer> {
+		self.0.as_ref().map(|metrics| metrics.collation_request_duration.start_timer())
+	}
 }
 
 #[derive(Clone)]
@@ -139,6 +146,7 @@ struct MetricsInner {
 	process_msg: prometheus::Histogram,
 	handle_collation_request_result: prometheus::Histogram,
 	collator_peer_count: prometheus::Gauge<prometheus::U64>,
+	collation_request_duration: prometheus::Histogram,
 }
 
 impl metrics::Metrics for Metrics {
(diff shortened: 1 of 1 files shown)


node/network/collator-protocol/src/validator_side/mod.rs
```diff
@@ -181,6 +189,15 @@ impl metrics::Metrics for Metrics {
 				)?,
 				registry,
 			)?,
+			collation_request_duration: prometheus::register(
```
rphmeier, line 192: the default buckets may need some adjustment

vstakhov: The default buckets are `0.005, 0.01, 0.025, 0.05, 0.1, 0.25, 0.5, 1.0, 2.5, 5.0, 10.0` seconds - should that be adjusted to somehow more long intervals or vice versa?

rphmeier: I feel like we'd want more info within the 100ms to 1s range especially.

node/network/collator-protocol/src/validator_side/mod.rs
```diff
@@ -194,6 +211,8 @@ struct PerRequest {
 	to_requester: oneshot::Sender<(CandidateReceipt, PoV)>,
 	/// A jaeger span corresponding to the lifetime of the request.
 	span: Option<jaeger::Span>,
+	/// A metric histogram for the lifetime of the request
+	_lifetime_timer: Option<metrics::prometheus::prometheus::HistogramTimer>,
```
sandreim, line 215:
```suggestion
	maybe_span_timer: Option<metrics::prometheus::prometheus::HistogramTimer>,
```

sandreim (approved):
LGTM + one nit.

drahnr (approved):

node/network/collator-protocol/src/validator_side/mod.rs
drahnr: I agree that this would be the correct approach, it's not an unused field after all

vstakhov: Well, this will cause compile warning, as the only usage of this field is an implicit observe on drop. On the contrary jaeger `span` field is actually used in the code explicitly.

vstakhov: I can rename `_lifetime_timer` to `_maybe_span_timer` of course but I'm not sure if that will improve readability.

drahnr:
bot merge

PR #5125, opened 2022-03-15
Additional tracing in provisioner

tdimitrov:
Add debug logs for `fn send_inherent_data`.

### DIFF
--- a/node/core/provisioner/src/lib.rs
+++ b/node/core/provisioner/src/lib.rs
@@ -272,9 +272,7 @@ impl ProvisionerJob {
 			gum::debug!(
 				target: LOG_TARGET,
 				signed_bitfield_count = self.signed_bitfields.len(),
-				signed_bitfields = ?self.signed_bitfields,
 				backed_candidates_count = self.backed_candidates.len(),
-				backed_candidates = ?self.backed_candidates,
 				disputes_enabled,
 				"inherent data sent successfully"
 			);
@@ -348,10 +346,8 @@ async fn send_inherent_data(
 	gum::debug!(
 		target: LOG_TARGET,
 		disputes_enabled = disputes_enabled,
-		candidates = ?candidates,
-		availability_cores = ?availability_cores,
+		availability_cores_len = availability_cores.len(),
 		disputes_count = disputes.len(),
-		bitfields = ?bitfields,
 		bitfields_count = bitfields.len(),
 		candidates_count = candidates.len(),
 		"inherent data prepared",
@@ -385,8 +381,15 @@ fn select_availability_bitfields(
 ) -> Vec<SignedAvailabilityBitfield> {
 	let mut selected: BTreeMap<ValidatorIndex, SignedAvailabilityBitfield> = BTreeMap::new();
 
+	gum::debug!(
+		target: LOG_TARGET,
+		bitfields_count = bitfields.len(),
+		"bitfields count before selection"
+	);
+
 	'a: for bitfield in bitfields.iter().cloned() {
 		if bitfield.payload().0.len() != cores.len() {
+			gum::debug!(target: LOG_TARGET, "dropping bitfield due to length mismatch");
 			continue
 		}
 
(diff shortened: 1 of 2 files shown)


node/core/provisioner/src/lib.rs
```diff
@@ -269,6 +269,13 @@ impl ProvisionerJob {
 			self.metrics.on_inherent_data_request(Err(()));
 		} else {
 			self.metrics.on_inherent_data_request(Ok(()));
+			gum::debug!(
+				target: LOG_TARGET,
+				len_signed_bitfield = self.signed_bitfields.len(),
+				len_backed_candidates = self.backed_candidates.len(),
```
drahnr, line 275: Since this is only executed every 6s by one validator at `debug!` level, I think we can afford printing some additional data.

drahnr:
bot merge

eskimor:
The provisioner does some filtering by its own, adding debug logs here:

https://github.com/paritytech/polkadot/blob/bd3f56fd7d4f2ee9f36ac91092f33af6376f586a/node/core/provisioner/src/lib.rs#L369
https://github.com/paritytech/polkadot/blob/bd3f56fd7d4f2ee9f36ac91092f33af6376f586a/node/core/provisioner/src/lib.rs#L383

would be a good idea, I think, also here a debug log that we are not providing bitfields because the leaf is stale would be good:

https://github.com/paritytech/polkadot/blob/bd3f56fd7d4f2ee9f36ac91092f33af6376f586a/node/core/provisioner/src/lib.rs#L334

and as @rphmeier mentioned in the chat - logging the number of bitfields also before the node side filtering would make sense.

eskimor:
If we can relate before and after filtering as suggested by @rphmeier, like:

```
"selected X of Y bitfields from Z unique validators"
```

that would be even better.

node/core/provisioner/src/lib.rs
```diff
@@ -269,6 +269,15 @@ impl ProvisionerJob {
 			self.metrics.on_inherent_data_request(Err(()));
 		} else {
 			self.metrics.on_inherent_data_request(Ok(()));
+			gum::debug!(
+				target: LOG_TARGET,
+				signed_bitfield_count = self.signed_bitfields.len(),
+				signed_bitfields = ?self.signed_bitfields,
+				backed_candidates_count = self.backed_candidates.len(),
+				backed_candidates = ?self.backed_candidates,
```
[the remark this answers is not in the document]
drahnr: They are only printed once per block by the block producer, so that's a tradeoff imho is reasonable

@tdimitrov we should reduce the print of the backed candidates down to their hashes, I suggest reducing to trace-level and printing a reduced set in debug, aligned with what @eskimor suggested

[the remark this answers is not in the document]
rphmeier: > They are only printed once per block by the block producer, so that's a tradeoff imho is reasonable

The provisioner is triggered by every node on every block for some reason. That wasn't the original behavior, but somewhere during an inherent-data refactoring in Substrate in 2021 it changed.

drahnr (approved):

eskimor (changes requested):
Let's pass in the activated leaf into [`select_availability_bitfields`](https://github.com/paritytech/polkadot/blob/bd3f56fd7d4f2ee9f36ac91092f33af6376f586a/node/core/provisioner/src/lib.rs#L333), so we can add the [leaf's hash](https://github.com/paritytech/polkadot/blob/f06e87e7ee3ddeb0aecba3946ef86de8cb62bd4c/node/subsystem-types/src/lib.rs#L75)  to the logs, this should help with relating them to each other.

node/core/provisioner/src/lib.rs
```diff
@@ -361,11 +381,19 @@ async fn send_inherent_data(
 fn select_availability_bitfields(
 	cores: &[CoreState],
 	bitfields: &[SignedAvailabilityBitfield],
+	leaf_hash: &Hash,
 ) -> Vec<SignedAvailabilityBitfield> {
 	let mut selected: BTreeMap<ValidatorIndex, SignedAvailabilityBitfield> = BTreeMap::new();
 
+	gum::debug!(
+		target: LOG_TARGET,
+		bitfields_count = bitfields.len(),
```
eskimor, line 390: Can we have the leaf hash here as well?

node/core/provisioner/src/lib.rs
```diff
@@ -403,6 +451,8 @@ async fn select_candidates(
 	let mut selected_candidates =
 		Vec::with_capacity(candidates.len().min(availability_cores.len()));
 
+	gum::debug!(target: LOG_TARGET, "{} candidates before selection", candidates.len());
```
eskimor, line 454: same: leaf hash?

PR #5126, opened 2022-03-15
[ci] scripts: don't rebuild for every pallet benchmark

coderobe:
the `cargo run` in the benchmark loop was rebuilding polkadot for every pallet benchmark, which in our tests took approximately 1 hour per pallet with the production profile enabled

instead, build once and re-use the binary

### DIFF
--- a/scripts/run_benches_for_runtime.sh
+++ b/scripts/run_benches_for_runtime.sh
@@ -1,17 +1,18 @@
 #!/bin/bash
 
+set -e
+
 # Runs all benchmarks for all pallets, for a given runtime, provided by $1
 # Should be run on a reference machine to gain accurate benchmarks
 # current reference machine: https://github.com/paritytech/substrate/pull/5848
 
 runtime="$1"
-profile="$2"
-standard_args="--profile $profile --locked --features=runtime-benchmarks"
 
 echo "[+] Running all benchmarks for $runtime"
 
-# shellcheck disable=SC2086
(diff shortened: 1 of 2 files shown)


coderobe:
motivation for this change is that the `&update_weights` CI jobs started timing out as they surpassed 24h of run time. this should reduce the run time for `update_polkadot_weights` by... 24-40h

EgorPopelyaev (approved):

scripts/run_benches_for_runtime.sh
```diff
@@ -1,17 +1,18 @@
...
-profile="$2"
-standard_args="--profile $profile --locked --features=runtime-benchmarks"
 
 echo "[+] Running all benchmarks for $runtime"
 
-# shellcheck disable=SC2086
-cargo +nightly run $standard_args benchmark \
+cargo +nightly build --profile production --locked --features=runtime-benchmarks
+
+target/production/polkadot benchmark \
```
ggwpez, line 15:
```suggestion
./target/production/polkadot benchmark \
```

coderobe:
bot merge

coderobe:
oh. excellent point. i think some of them do fail

ggwpez:
> oh. excellent point. i think some of them do fail

Yea that is unfortunate. Please post them here so we can investigate and create an issue.  
Trying locally; `runtime_common::crowdloan` failed for me.

PR #5129, opened 2022-03-15
Collator protocol debug

vstakhov:
This PR is intended to add some missing debug messages to the collator protocol part.

### DIFF
--- a/node/network/collator-protocol/src/validator_side/mod.rs
+++ b/node/network/collator-protocol/src/validator_side/mod.rs
@@ -547,11 +547,11 @@ impl CollationsPerRelayParent {
 	/// the passed in `finished_one` is the currently `waiting_collation`.
 	pub fn get_next_collation_to_fetch(
 		&mut self,
-		finished_one: Option<CollatorId>,
+		finished_one: Option<&CollatorId>,
 	) -> Option<(PendingCollation, CollatorId)> {
 		// If finished one does not match waiting_collation, then we already dequeued another fetch
 		// to replace it.
-		if self.waiting_collation != finished_one {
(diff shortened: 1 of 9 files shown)


node/network/collator-protocol/src/validator_side/mod.rs
```diff
@@ -662,6 +662,14 @@ async fn fetch_collation<Context>(
 
 	if state.peer_data.get(&peer_id).map_or(false, |d| d.has_advertised(&relay_parent)) {
 		request_collation(ctx, state, relay_parent, para_id, peer_id, tx).await;
+	} else {
```
eskimor, line 665: This does also trigger if there is no peer data for that peer.

node/network/collator-protocol/src/validator_side/mod.rs
```diff
@@ -883,6 +909,12 @@ async fn process_incoming_peer_message<Context>(
 
 			let peer_data = match state.peer_data.get_mut(&origin) {
 				None => {
+					gum::debug!(
+						target: LOG_TARGET,
+						peer_id = ?origin,
+						?relay_parent,
+						"Unknown peer",
```
eskimor, line 916: maybe more descriptive: "Message received from unknown peer"

node/network/collator-protocol/src/validator_side/mod.rs
vstakhov: Yes, I think this message shoul be on the `warn!` level as it is quite unlikely.

rphmeier (approved):

rphmeier:
bot merge

PR #5133, opened 2022-03-16
make more subsystems blocking, fix blocking subsystem spawning

rphmeier:
Approval Voting, Dispute Coordinator, and Chain Selection subsystems all use the underlying DB and block on it. The first 2 of these actually do a lot of writes.  'blocking' subsystems were meant to be spawned with `spawn_blocking`, which gives them their own dedicated thread. They weren't, so I fixed that here.


drahnr (approved):
Mea culpa

sandreim (approved):

sandreim:
bot merge

rphmeier:
I wonder if we should worry about this in any case? Maybe switching to `blocking` will cause more problems...

drahnr:
I doubt it, we saw some outliers in the poll duration (10s of millis), so this would at least mitigate that. And we only spawn 3 additional threads really.

Let's do a burnin

PR #5137, opened 2022-03-16
Integrate `benchmark-overhead` command

ggwpez:
Integrate the `benchmark-overhead` command from paritytech/substrate#10977 to benchmark the per-block and per-extrinsic execution overhead.

- Add `with_signed_payload!` macro which creates `SignedPayload` for each relay-runtime
- Add `benchmark_inherent_data` which creates inherent data for all relay-runtime
- Integrate `benchmark-overhead` for each relay-runtime
- Add test for each relay- and relay-dev-runtime :sparkles: 

The only problem that I currently have here is that it prints `CurrentBlockRandomness did not provide entropy` for every extrinsic or block, I don't know.


ggwpez:
bot rebase

node/client/src/lib.rs
```diff
@@ -567,3 +569,250 @@ impl sp_blockchain::HeaderBackend<Block> for Client {
...
+					extra,
+				);
+				Ok(ext.into())
+			}
+		}
+	}
+}
+
+/// Generates inherent data for: Polkadot, Kusama, Westend and Rococo.
+pub fn provide_inherent_data(
```
bkchr, line 796: You should make it more obvious that the context is for benchmarking. You should not use this for normal block production.

node/client/src/lib.rs
```diff
@@ -567,3 +569,250 @@ impl sp_blockchain::HeaderBackend<Block> for Client {
...
 }
+
+/// Provides a `SignedPayload` for any runtime.
+///
+/// The first code block should set up all variables that are needed to create the
+/// `SignedPayload`. The second block can make use of the `SignedPayload`.
+///
+/// This is not done as a trait function since the return type depends on the runtime.
+/// This macro therefore uses the same approach as [`with_client!`].
+macro_rules! with_raw_payload {
```
bkchr, line 580: Why do we need a macro for this?

runtime/kusama/src/lib.rs
```diff
@@ -75,6 +75,8 @@ use sp_version::NativeVersion;
 use sp_version::RuntimeVersion;
 use static_assertions::const_assert;
 
+#[cfg(feature = "std")]
```
bkchr, line 78: Here and in the other runtimes you don't really need this feature.

tests/benchmark_overhead_works.rs
```diff
@@ -0,0 +1,69 @@
...
+		.arg("--weight-path")
+		.arg(base_path)
+		.args(["--warmup", "5", "--repeat", "5"])
+		.args(["--add", "100", "--mul", "1.2", "--metric", "p75"])
+		.args(["--max-ext-per-block", "5"])
+		.status()
+		.map_err(|e| format!("command failed: {:?}", e))?;
+
+	if !status.success() {
+		return Err("Command failed".into())
```
bkchr, line 62: Would be good to print stderr/stdout of the process to find out why it failed.

tests/benchmark_overhead_works.rs
```diff
@@ -0,0 +1,69 @@
...
+#[cfg(unix)]
+fn benchmark_overhead_works() {
+	for runtime in RUNTIMES {
+		assert!(benchmark_overhead(format!("{}-dev", runtime)).is_ok());
+	}
+}
+
+/// `benchmark-overhead` rejects all non-dev runtimes.
+#[test]
+#[cfg(unix)]
```
bkchr, line 36: Why does this only works on Unix?

node/client/src/lib.rs
ggwpez: The problem is that every runtime has their own `SignedExtra` and `SignedPayload`.  
These types do not share any trait, so I cannot use a trait instead.  
The `with_client` macro was done for a similar reason which did not work with a trait.  

I personally don't like it but did not see an easier solution.

tests/benchmark_overhead_works.rs
ggwpez: All the other tests in that directory have it as well, don't know why.

tests/benchmark_overhead_works.rs
ggwpez: The error output of the command is printed on failure, so it for example on an invalid argument:
```pre
error: Found argument '--unknown' which wasn't expected, or isn't valid in this context

        If you tried to supply `--unknown` as a value rather than a flag, use `-- --unknown`

USAGE:
    polkadot benchmark-overhead --chain <CHAIN_SPEC>

For more information try --help
```

node/client/src/lib.rs
```diff
@@ -567,3 +569,252 @@ impl sp_blockchain::HeaderBackend<Block> for Client {
...
+				$( $setup )*
+
+				let $extra: runtime::SignedExtra = (
+					frame_system::CheckNonZeroSender::<runtime::Runtime>::new(),
+					frame_system::CheckSpecVersion::<runtime::Runtime>::new(),
+					frame_system::CheckTxVersion::<runtime::Runtime>::new(),
+					frame_system::CheckGenesis::<runtime::Runtime>::new(),
+					frame_system::CheckMortality::<runtime::Runtime>::from(sp_runtime::generic::Era::mortal(
+						$period,
+						$current_block,
```
shawntabrizi, line 693: Looks like a lot of duplicate code that can continue to be macro-ized?

PattanaiSgulrax: Re: [paritytech/polkadot] Integrate `benchmark-overhead` command (PR #5137)

ggwpez: The `SignedExtras` are slightly different for Polkadot, since they have one more entry. But I can sim up the other three ones, which would make the macro a bit more complicated though.

shawntabrizi (approved):

node/client/src/lib.rs
ggwpez: Renamed to `benchmark_inherent_data` and added doc.

ggwpez:
I will merge this as-is and then think about a better structure for the commands and how to use them from polkadot.  
With the current state we already got the results that we were looking for, so I think its worth it.

ggwpez:
bot merge

Issue #5138, opened 2022-03-16
Benchmark Regression Tracking Issue


shawntabrizi:
In https://github.com/paritytech/polkadot/pull/5097, it was shown some benchmarks were not working.

- [x] runtime/kusama/src/weights/frame_system.rs
- [ ] runtime/kusama/src/weights/pallet_collective_council.rs
- [ ] runtime/kusama/src/weights/pallet_collective_technical_committee.rs
- [ ] runtime/kusama/src/weights/pallet_election_provider_multi_phase.rs
- [ ] runtime/kusama/src/weights/pallet_elections_phragmen.rs
- [ ] runtime/kusama/src/weights/runtime_common_crowdloan.rs
- [x] runtime/polkadot/src/weights/frame_system.rs
- [ ] runtime/polkadot/src/weights/pallet_collective_council.rs
- [ ] runtime/polkadot/src/weights/pallet_collective_technical_committee.rs
- [ ] runtime/polkadot/src/weights/pallet_election_provider_multi_phase.rs
- [ ] runtime/polkadot/src/weights/runtime_common_auctions.rs
- [ ] runtime/polkadot/src/weights/runtime_common_crowdloan.rs
- [ ] runtime/polkadot/src/weights/runtime_common_slots.rs
- [ ] runtime/polkadot/src/weights/runtime_parachains_hrmp.rs
- [x] runtime/westend/src/weights/frame_system.rs
- [ ] runtime/westend/src/weights/pallet_election_provider_multi_phase.rs
- [ ] runtime/westend/src/weights/runtime_common_crowdloan.rs

This issue will track making sure each of these works on the appropriate runtime, and work on a fix if not.

shawntabrizi:
thank you @coderobe

ggwpez:
The `runtime_parachains::ump` benchmark still reports `Invalid input: UnknownMessageIndex`.  
Do you know what that means @hirschenberger? We are currently trying to fix some regressions.  
cc @KiChjang

hirschenberger:
> The `runtime_parachains::ump` benchmark still reports `Invalid input: UnknownMessageIndex`. Do you know what that means @hirschenberger? We are currently trying to fix some regressions. cc @KiChjang

Is it in the `service_overweight` benchmark? It means that the message has not been overweight and was not enqueued to the overweight queue. That might be bc. of some bound-value changes.

So the `create_message_overweight` function may not have produced a overweight message.

Is it urgent? I can have a look on it at the weekend.

[...]

PR #5139, opened 2022-03-16
Fix Benchmark Regressions in Polkadot Parachain Auction System

shawntabrizi:
https://github.com/paritytech/polkadot/issues/5138

Most of these breaks were introduced with the Lease Period Offset, and just needed to adjust the block number that events happen on.

### DIFF
--- a/runtime/common/src/auctions.rs
+++ b/runtime/common/src/auctions.rs
@@ -1772,6 +1772,10 @@ mod benchmarking {
 
 		// Worst case scenario a new bid comes in which kicks out an existing bid for the same slot.
 		bid {
+			// If there is an offset, we need to be on that block to be able to do lease things.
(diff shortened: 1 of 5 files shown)


ggwpez:
These two still fail:  
- [ ] `runtime_parachains::ump`
- [ ] `pallet_election_provider_multi_phase`

shawntabrizi:
@ggwpez yeah, these are just the pallet I know about, so lets have this be one pr, and then look at the others in follow up

ggwpez (approved):

coderobe (approved):

ggwpez:
I think we should wait until https://github.com/paritytech/substrate/pull/11057 is merged and then pull lastest Substrate/master in here.  
Then the `runtime_parachains::ump` should be the only failing one.

shawntabrizi:
bot merge

PR #5142, opened 2022-03-16
provisioner logs minor janitor duties

drahnr:


tdimitrov (approved):

node/core/provisioner/src/lib.rs
```diff
@@ -289,8 +288,16 @@ impl ProvisionerJob {
 			ProvisionableData::Bitfield(_, signed_bitfield) =>
 				self.signed_bitfields.push(signed_bitfield),
 			ProvisionableData::BackedCandidate(backed_candidate) => {
+				let candidate_hash = backed_candidate.hash();
+				gum::debug!(
```
rphmeier, line 292: probably better as `trace!`. This is pretty heavy. We also have a similar log in candidate-backing already

drahnr: Done

PR #5151, opened 2022-03-18
fix freshness calculations in statement distribution

rphmeier:
This breaks a lot of tests, even though it's now the 'correct' behavior.

### DIFF
--- a/node/network/statement-distribution/src/lib.rs
+++ b/node/network/statement-distribution/src/lib.rs
@@ -271,7 +271,7 @@ impl PeerRelayParentKnowledge {
 
 				let was_known = self.is_known_candidate(h);
 				self.sent_candidates.insert(h.clone());
-				was_known
+				!was_known
 			},
 			CompactStatement::Valid(_) => false,
 		};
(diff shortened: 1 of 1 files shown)


node/network/statement-distribution/src/lib.rs
```diff
@@ -269,7 +269,9 @@ impl PeerRelayParentKnowledge {
 			CompactStatement::Seconded(ref h) => {
 				self.seconded_counts.entry(fingerprint.1).or_default().note_local(h.clone());
 
-				self.sent_candidates.insert(h.clone())
+				let was_known = self.is_known_candidate(h);
+				self.sent_candidates.insert(h.clone());
+				was_known
```
eskimor, line 274: Shouldn't this be !was_known ... the function is supposed to return whether the pear learned from this statement the first time, which is equivalent to it was not known by the peer.

node/network/statement-distribution/src/lib.rs
```diff
@@ -338,14 +340,14 @@ impl PeerRelayParentKnowledge {
 					return Err(COST_UNEXPECTED_STATEMENT_REMOTE)
 				}
 
-				h
+				(h, self.is_known_candidate(h))
```
eskimor, line 343: Same - it is fresh if it was not known.

eskimor:
My understanding is the missing `!` is a logic error. The actual purpose of this PR is to also take into consideration candidates in `received_candidates` instead of just in `sent_candidates` - correct?

rphmeier:
Yes, that's right. This can lead to too many messages being sent, possibly.

eskimor:
bot merge

eskimor (approved):

PR #5154, opened 2022-03-18
Some improvements in the statement-distribution histograms

vstakhov:
This PR is intended to do the following neats:

* Add more reasonable buckets for latencies
* Add granular timers for different network events instead of measuring everything in general
* Remove timers from events that actually does not involve any networking (less noise)

### DIFF
--- a/node/network/statement-distribution/src/metrics.rs
+++ b/node/network/statement-distribution/src/metrics.rs
@@ -16,6 +16,10 @@
 
 use polkadot_node_subsystem_util::metrics::{self, prometheus};
 
+/// Buckets more suitable for checking the typical latency values
+const HISTOGRAM_LATENCY_BUCKETS: &[f64] =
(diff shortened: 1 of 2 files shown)


node/network/statement-distribution/src/metrics.rs
```diff
@@ -16,14 +16,18 @@
 
 use polkadot_node_subsystem_util::metrics::{self, prometheus};
 
+/// Buckets more suitable for checking the typical latency values
+const HISTOGRAM_LATENCY_BUCKETS: &[f64] =
```
sandreim, line 20: This could be moved to `polkadot-node-metrics`

sandreim (approved):
LGTM

drahnr (approved):

eskimor (approved):

drahnr:
bot merge

